PR #86, opened 2026-03-20
feat: Implement waitlist functionality for programs

dkaygithub:
Implements the waitlist functionality.


thpr:
@dkaygithub  I think this is best turned into an issue at this point and re-implemented later.  It will need to be rebuilt anyway given the # of intervening changes

dkaygithub:
agree

PR #105, opened 2026-03-30
🛡️ Sentinel: [HIGH] Fix Timing Attack in Cron Authentication

dkaygithub:
🛡️ **Sentinel: [HIGH] Fix Timing Attack in Cron Authentication**

🚨 **Severity:** HIGH
💡 **Vulnerability:** Timing attack vulnerability during authentication check for `CRON_SECRET` in `/api/cron/*` endpoints. A standard string equality comparison (`!==`) terminates early, which can leak timing information to an attacker, allowing them to guess the expected secret character by character.
🎯 **Impact:** An attacker could potentially forge a valid authorization header to bypass authentication and maliciously trigger server cron jobs (e.g., forcing checkout of members, triggering mass emails prematurely).
🔧 **Fix:** Imported Node's `crypto` module and replaced string comparison with `crypto.timingSafeEqual()`. Both the generated and expected strings are first converted to `Buffer` objects, and their lengths are explicitly checked (as `timingSafeEqual` requires buffers of the exact same length, failing which it throws an error that could be manipulated to leak info).
✅ **Verification:** Verified by running the test suite to ensure no regressions in legitimate cron job execution, and linters passed cleanly. Tests verify that mismatched strings (even of the same length) continue to be rejected with 401 Unauthorized.

Applies to:
- `src/app/api/cron/nightly/route.ts`
- `src/app/api/cron/pending-participants/route.ts`
- `src/app/api/cron/reminders/route.ts`
- `src/app/api/cron/post-event/route.ts`

---
*PR created automatically by Jules for task [3972087941049855693](https://jules.google.com/task/3972087941049855693) started by @dkaygithub*

### DIFF
--- a/src/app/api/cron/nightly/route.ts
+++ b/src/app/api/cron/nightly/route.ts
@@ -1,4 +1,5 @@
 import { NextResponse } from "next/server";
+import crypto from "crypto";
 import prisma from "@/lib/prisma";
 import { processPostEventEmails } from "@/lib/postEventEmails";
 import { processVisitCheckout } from "@/lib/attendanceTransitions";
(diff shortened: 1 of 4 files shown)


dkaygithub:
Superseded by #112 which was merged to main.

PR #146, opened 2026-05-14
docs(security): design proposal for eliminating dangerously_allow

dkaygithub:
## Summary

Design-only PR. Proposes an architectural strategy for driving `dangerously_allow_all_data_access` to zero across the 23 routes that currently bypass the field-level stripper.

The doc lives at [`docs/DANGEROUSLY-ALLOW-ELIMINATION.md`](./docs/DANGEROUSLY-ALLOW-ELIMINATION.md) and contains:

- Categorization of all 23 current bypasses by *why* they bypass (5 root causes)
- A proposed discriminated `TransformSpec` on `RouteSpec` replacing the boolean bypass with **9 narrow transformer primitives** (`ack`, `aggregate`, `processMeta`, `selfScalars`, `computedField`, `actionDispatch`, `kioskScope`, `withFreshContext`, `derivedFromInput`/`devOnly`)
- Per-bypass mapping table → which transformer would replace each
- A 6-phase migration plan, ending with hard-removal of the `dangerously_allow_all_data_access` field and a CI lint rule that forbids the name
- Open questions (aggregate-shape registry style, action discriminant key, Prisma tracer cost)

The transformer kinds live in `src/security/transformers/` — already under CODEOWNERS, so adding a transformer requires maintainer review but *using* one from a route handler is a normal PR.

## What this PR is and isn't

- ✅ This PR adds **only** the design doc. No code changes.
- ❌ It does NOT implement any transformer.
- ❌ It does NOT migrate any of the 23 bypassed routes.

Implementation would follow as separate PRs per phase if the design is accepted.

## Test plan

- [ ] Reviewer reads the doc end-to-end and pushes back on any transformer kind whose scope feels wrong or missing
- [ ] Reviewer confirms the bypass categorization against `src/security/registry.ts` matches their mental model
- [ ] Reviewer answers (or escalates) the three open questions at the bottom
- [ ] If approved: a follow-up PR opens for Phase 1 (foundational `TransformSpec` plumbing)


jee7s (commented):

**CONFIRMED: the doc's premise doesn't hold against current main.** It plans the elimination of a `spec.dangerously_allow_all_data_access` bypass it quotes from `src/security/handler.ts` — but that field/branch exists nowhere in the repo (`git grep dangerously_allow` on main returns zero hits; `RouteSpec` in core.ts has no such field, handler.ts has no bypass path). The 23 routes it lists are the un-migrated legacy `withAuth` routes tracked in `scripts/legacy-authz-routes.txt` — a real problem, but a different one: they have **no stripping at all**, not a bypass flag to remove.

As written, merging this would put a phantom migration plan (Phases 1–6, a CI lint forbidding a field that doesn't exist) into `docs/` as a security reference. Recommendation: **close**, or rewrite from scratch against the actual mechanism — the wave-1 registry-first migration (#1137 + #1138–#1142 stack) is already the live version of that work.

jee7s:
**Linkage:** backlog issue #1348 was auto-created from this PR's own open-items entry and inherits the same premise — a `dangerously_allow_all_data_access` mechanism that does not exist on current main (see review above). If this doc is closed or rewritten, #1348 needs the same re-scoping. The *real* stripper-coverage gap this doc gestures at is the un-migrated legacy `withAuth` routes (`scripts/legacy-authz-routes.txt`), whose live remediation is the wave-1 registry migration (#1137 + #1138–#1142) and whose open ends are tracked by #1321 (routes pending scoping) and #1320 (GAP-2 drift-guard CI ban).

thpr:
## Review — direction is right, but the approach can be much simpler, and it needs to be reconciled with the current boundary + #1134

Read the doc end-to-end. The instinct (narrow, registered primitives instead of an all-or-nothing escape hatch) is good. Three things to work through before implementing.

### 1. The premise has aged out of the code

The doc is dated 2026-05-14 and describes a `handler.ts` with an `if (spec.dangerously_allow_all_data_access) { …bag } else { stripBag(…) }` branch. On `main` today that branch — and the `dangerously_allow_all_data_access` field on `RouteSpec` — don't exist; `handler()` always strips. The boundary is currently **opt-in**: ~14 of ~141 API routes go through `handler()`, and the rest were never inside it (plain `NextResponse.json(...)`), so they aren't "bypassing" a stripper. A few of the 23 named routes (`system-health`, `trends`, `kiosk/version`) aren't in the tree either.

Net: this reads better as an **onboarding** design ("bring the remaining routes into the boundary") than an **elimination** design. The 23-route table and Phase 6 (remove the field + CI-ban the name) should be rebuilt from `git grep security/handler` against the real set first.

### 2. This is the response/field axis only — it can't remove the inline row handling, and that's where the real leaks are

Every primitive here operates on what the handler **returns** (post-handler). None touch the `where` (which rows are admitted, pre/in-handler). Two onboarded routes show the gap:

- **`trusted-adults/operational`** returns a normal `ModelBag`, but ~22 lines below the handler build a household-ID set and inject `where.householdId`. No transformer here applies; those lines stay verbatim.
- **`events/[id]`** — the doc's `actionDispatch` example. Its own header comment says it: *"Field-tiering guards fields; only admission guards the association. So we gate admission here."* `actionDispatch` would declare the per-action **response shape**, but the admission gate + per-branch `where`/transaction logic stays inline.

The leaks that actually shipped — **P0-C** (denied households saw `orgMemberOnly` programs) and **#575** (`/api/programs/[id]` leaked rosters) — are **row-admission** leaks, which `handler()` "structurally cannot catch" (per #1134). So the doc's "stripper bypasses → zero, future PII leaks caught at compile time" reads as boundary-closure while the class that leaked stays open. That's the case **for #1134**, not against it: #1134's query-side `rows:` / `rowsWhere` is what removes the inline row handling and makes a `withAuth→handler` swap non-ceremonial (cf. #1141, which formalized field policy and left the row lines hand-rolled). Suggest #146 add an explicit non-goal ceding the row axis to #1134, correct the compile-time-closure line, and sequence #1134 first (its Step 1 persona row-contract test is zero-runtime-risk and would have caught both P0-C and #575).

One direct conflict to resolve: **`kioskScope`** works by blanking rows response-side, which is #1134 non-goal #1 (*"No response-time row stripping in `handler()`"* — breaks pagination/counts, parks invisible rows in memory). Recommend dropping it in favor of `rows: { Visit: ['all_current_visitors'] }` query-side.

### 3. Nine transform kinds is more surface than the problem needs — two lanes cover it

"One shape" is a good goal read as **one mandatory classification**, not one code path. `/health` returning `{status:ok}` and a roster route are opposite problems: one needs *enforcement*, the other needs a *proof it needs none*. So:

- **Lane A — `sensitive`:** reads sensitive models or returns rows → `handler()` with field-strip (exists) **+ row admission (`rows:` / #1134)**. All real enforcement lives here. This is also the literal "handler does row filtering."
- **Lane B — `nonSensitive`:** an **attestation**, not a transformer — CI proves (via the AsyncLocalStorage tracer the doc already proposes) that the route reads no sensitive model and returns no Prisma row. Any shape allowed, because the guarantee is "nothing sensitive flowed," not "matched shape X."

Lane B subsumes `ack`, `processMeta`, `derivedFromInput`, `devOnly`, and `aggregate` — they're the same claim in five costumes. `ack`'s allowed-key vocabulary, `aggregate`'s shape registry, and `devOnly`'s env check are hand-rolled proxies for "returns nothing sensitive," which the tracer proves directly. Collapsing them:

| doc kind | fate |
|---|---|
| `ack`, `processMeta`, `derivedFromInput`, `devOnly`, `aggregate` | → one `nonSensitive` attestation (tracer-proven) |
| `computedField` | keep — real field-axis feature (derived column at a declared tier) |
| `selfScalars` | already expressible as a self-scoped view + field allow-list |
| `actionDispatch` | drop — "one endpoint, N shapes"; split endpoints or keep the branch, each branch is still Lane A. A policy-layer router is the piece most likely to grow one-offs and recreate the bypass. |
| `withFreshContext` | drop — one route (household); a handler re-read is the fix |
| `kioskScope` | drop → Lane A `rows:` (query-side, #1134) |

Nine kinds → one attestation + one surviving field feature (`computedField`) + the #1134 row lane. Enforce "one shape" by extending the drift-guard (#1320) to require a lane declaration per route — the classification is mandatory, the ceremony isn't.

Ceiling to be honest about: `nonSensitive` is a **CI-proven** guarantee, not runtime (the tracer is test/dev-mode for cost — the same limit the doc already accepts). A mis-declared route that later adds a sensitive read is caught in CI, not prod; upgrade path is tracer-on-in-prod or move to Lane A.

### Net

Keep the direction and `computedField`; rebuild the inventory; cede the row axis to #1134 and sequence it first; replace the 9-kind union with the two-lane classification. Happy to pair on the reframed version.

jee7s:
Re-verified this against current `main` today (2026-08-17):

- The mechanism this doc plans to eliminate does not exist in the codebase. `git grep dangerously_allow` returns a single hit — `docs/backlog/INDEX.md` (RB21), which cites this PR itself. There is no such field on `RouteSpec`, no bypass branch in `handler.ts`, and the `src/security/transformers/` directory the body references is absent from `checkin-app/src/security/`.
- The repo went a different way: `defineRoute`/`defineOutbound` + registry/scope-bindings (`core.ts`, `registry.ts`, `stripper.ts`, `scopeBindings.ts`), with #1518 as the live target-architecture doc and the wave-1 registry/conversion stack (#1209–#1215, #1138–#1142) carrying the actual stripper-coverage work.
- Both July reviews reached the same conclusion independently, and there has been no author response in the three weeks since.

Recommendation: close this PR — merging it would install a migration plan for a phantom mechanism into `docs/` as a security reference. One piece of cleanup to carry: backlog issue #1348 was auto-created from this PR and inherits the same dead premise; it should be re-scoped or closed alongside.

PR #185, opened 2026-06-08
feat(schema): membership process model + board settings (additive)

dkaygithub:
## PR 2/12 — Household membership application & review stack

Purely additive schema for the membership flow. The existing `Membership` model is left intact (only an additive `processes[]` relation) so every commit stays green; the 1:1 reshape + read-model cutover land in PR 3.

### Added
- **Enums**: `MembershipStatus`, `MembershipProcessKind`, `MembershipProcessStatus`, `ParticipantRole`, `AttestationResult`.
- **`MembershipProcess`** — one row per application/renewal cycle, audit-retained, with integration-ref fields (Zoho envelope, contract-signed, BG-consent, Shopify draft/invoice/order, paid, certified-by, renewal-reminder).
- **`MembershipAttestation`** — 2-of-N background-check sign-off (`@@unique([processId, reviewerId])`) + per-reviewer "volunteer family" flag.
- **`VolunteerDesignation`** — admin email allow-list (`email @unique`, `@pii`).
- **`BoardSettings`** — singleton: normal/volunteer dues (cents), membership-year boundary, Shopify product/variant ids, Averity deep-link.
- **`Participant`** — `role PARENT|CHILD`, `allergies` (`@sensitivity:personal`), `bgAttestations` back-relation.

Every scalar carries a `/// @sensitivity` tier; classifications regenerated.

### Tests
No behavioral tests here — nothing reads these models yet. Coverage arrives with the features that use them (read-model in PR 3, BG review in PR 7, etc.).

### Verification
`prisma db push` synced · `tsc` clean · `npm run test:ci` 82/82 · pre-commit hook green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #186, opened 2026-06-08
refactor(membership): centralize active-member read model

dkaygithub:
## PR3 of the membership stack — read-model centralization

Stacked on #185 (`feat/membership-schema-models`). Pure refactor on the **current** schema; no schema change, no behavior change for already-correct sites.

### What
Adds `src/lib/membership.ts` as the single source of truth for *"is this participant an active member?"*:
- `ACTIVE_MEMBER_PARTICIPANT_WHERE` — Prisma `where` fragment (household active OR own active)
- `isActiveMember(id)` — async existence check
- `participantRecordIsActiveMember(record)` — pure predicate over a loaded record
- `ACTIVE_MEMBER_INCLUDE` — include fragment for the pure predicate

Repoints 5 call sites to it.

### Bug fixed
Three sites checked only `participant.memberships` and ignored the household, so a **child or second parent** whose membership lives on the household read as a non-member:
- `GET /api/programs` (member-only visibility)
- `GET /api/programs/[id]` (member-only gate)
- `GET /api/admin/participants/search` (`isMember` flag)

`shop/members` and `eligible-participants` already used the correct OR-household form — repointed for consistency only.

### Why first
Centralizing now means the **PR4 schema cutover** (drop individual/volunteer memberships) collapses the member definition to household-only in *one place* — every consumer follows automatically, keeping each commit green under the full-suite pre-commit gate.

### Gate
tsc clean · route-coverage 0 errors · 82/82 tests · eslint clean. Each edit independently adversarially verified (routes-through-helper, types-sound, bug-fixed/behavior-preserved).

### DIFF
--- a/src/lib/membership.ts
+++ b/src/lib/membership.ts
@@ -1,0 +1,65 @@
+import type { Prisma } from "@prisma/client";
+import prisma from "@/lib/prisma";
+
+/**
+ * Canonical "is this person an active member?" read model.
+ *
+ * A Participant counts as an active member when EITHER:
+ *   - their household holds an active Membership, OR
+ *   - they personally hold an active Membership (legacy individual/volunteer link).
+ *
+ * This is the single source of truth for member gating. Several call sites
+ * historically checked only `participant.memberships` (own-only), which wrongly
+ * excluded a child or second parent whose membership lives on the household.
+ * Route those through here instead.
+ *
+ * NOTE: the individual-membership leg is legacy. When the schema cutover (PR4)
+ * drops per-participant memberships, collapse this to household-only in ONE
+ * place — here — and every consumer follows automatically.
+ */
+export const ACTIVE_MEMBER_PARTICIPANT_WHERE: Prisma.ParticipantWhereInput = {
+    OR: [
+        { household: { memberships: { some: { active: true } } } },
+        { memberships: { some: { active: true } } },
+    ],
+};
+
+/**
+ * Prisma `include` fragment that loads exactly the membership data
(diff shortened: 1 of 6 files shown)


src/lib/membership.ts
```diff
@@ -0,0 +1,65 @@
+import type { Prisma } from "@prisma/client";
+import prisma from "@/lib/prisma";
+
+/**
+ * Canonical "is this person an active member?" read model.
+ *
+ * A Participant counts as an active member when EITHER:
+ *   - their household holds an active Membership, OR
+ *   - they personally hold an active Membership (legacy individual/volunteer link).
```
dkaygithub, line 9: no more perosnal memberships.

src/lib/membership.ts
```diff
@@ -0,0 +1,65 @@
...
+ * excluded a child or second parent whose membership lives on the household.
+ * Route those through here instead.
+ *
+ * NOTE: the individual-membership leg is legacy. When the schema cutover (PR4)
+ * drops per-participant memberships, collapse this to household-only in ONE
+ * place — here — and every consumer follows automatically.
+ */
+export const ACTIVE_MEMBER_PARTICIPANT_WHERE: Prisma.ParticipantWhereInput = {
+    OR: [
+        { household: { memberships: { some: { active: true } } } },
```
dkaygithub, line 22: households only have 1 membership, correct

src/lib/membership.ts
```diff
@@ -0,0 +1,65 @@
...
+        { memberships: { some: { active: true } } },
```
dkaygithub, line 23: participants should no longer have memberships

PR #203, opened 2026-06-09
fix(security): escape backend data in kiosk banner to prevent stored XSS

jee7s:
## Summary
The kiosk client built banner HTML by interpolating backend-supplied values (participant email, success message, error/warning text) with no escaping, then the wrapper page assigned it via `innerHTML`. The last-keyholder warning embeds other present participants' names verbatim (`scan-service.ts`), and a participant's name is attacker-controlled via `PATCH /api/profile`.

A member could set their name to an HTML/JS payload, badge in, and have it execute in the kiosk browser when a keyholder badges out. Because the kiosk's local proxy auto-signs every request, the injected script could issue signed, kiosk-authenticated API calls.

## Fix
HTML-escape every backend-derived value (`warn`/`err`/`email`/`msg`) at the sink before interpolation (escaping the warning text *before* the `\n`→`<br>` substitution so injected markup can't survive). Renamed the local `html` variable to `banner_html` to avoid shadowing the new `import html`.

Escaping is applied only at the client sink (the `innerHTML` rendering layer), **not** also in `scan-service.ts`, to avoid double-encoding.

## Verification
- `python3 -m py_compile client/client.py` passes.


jee7s:
Reviewed and verified this fix end-to-end:

- **Banner sink covered.** The only `innerHTML` assignment in the wrapper page (`container.innerHTML = html`) takes *only* `data.html`, which is the now-escaped `banner_html` — no concatenation with other unescaped values. All four backend-derived fields (`warn`, `err`, `email`, `msg`) are `html.escape`d at the sink, with the warning escaped **before** the `\n`→`<br>` substitution so injected markup can't survive.
- **Attendance is out of scope (correctly).** Phase-2 attendance/counts/safety data is delivered to the iframe via `postMessage`, not `innerHTML` — it's rendered by the React app, which auto-escapes. No client-side change needed there.
- **No double-encoding.** Escaping is applied only at the client sink, not also in `scan-service.ts`. 👍
- **Merge.** `main` hasn't touched `client/client.py`, so I updated the branch to current `main` (it was BEHIND) with no conflicts — CI now runs and it's ready to merge.

LGTM. Leaving the actual merge to you since it triggers a dev deploy.

PR #200, opened 2026-06-08
feat(dev-instance): impersonation — mint-as-persona + inert impersonatedBy + banner

dkaygithub:
**Stacked on #199** (`feat/dev-instance-foundation`). Review/merge that first — this PR's diff is only the impersonation layer.

Implements `DEV_INSTANCE_DESIGN.md` §5: pick a persona → **mint a real session as that persona**, carrying an **inert `impersonatedBy` claim** that no authorization path ever reads.

## What's in this PR
- **`impersonation.ts` (new)** — pure `evaluateMint()` policy, exhaustively unit-tested (10 cases): prod never mints; the cloud dev instance requires a **verified org caller**; local is relaxed (no Google); nested impersonation credits the **original** real human; "return to me" clears the claim.
- **`auth-options.ts`** — replaces the local-only offline provider with one unified **`persona-mint`** `CredentialsProvider` (dev + local). It reads the caller's *current* session via `getToken` (the security boundary), runs `evaluateMint`, resolves the target participant (must exist — fake data only), and stamps `impersonatedBy`. The jwt callback carries org gate claims (`hd`/`email_verified`) onto minted **dev** sessions so the middleware still passes **without ever reading `impersonatedBy`**; the session callback exposes `impersonatedBy` to the UI.
- **`DevImpersonationBar` (new)** — persistent "🎭 Viewing as … — you are …" banner + persona switcher + "Return to me", via the persona-mint flow. Rendered only on dev/local for a signed-in user.
- **`dev-personas` route** — widened from local-only to the dev instance; now requires a session on dev so anonymous visitors can't enumerate personas (middleware exempts `/api`).
- **`DevLoginPicker`** — logged-out local login now routes through `persona-mint`.
- Types: `impersonatedBy` on JWT/Session/User. `ORG_DOMAIN` centralized in `config`.

## Reviewer attention
- **Not exercised live here.** The `getToken`-reads-the-caller-cookie path and actual session minting weren't run end-to-end (needs a browser + Google + the `checkin_dev` DB). The *policy* is unit-tested; the wiring is type/lint-verified only. Recommend a manual smoke test on a real dev deploy, or a live-DB integration test.
- **`getToken` inside `authorize`** — next-auth passes a limited `req`; it's cast to `getToken`'s expected shape. If the caller cookie ever fails to decode, dev mint fails **safe** (refuses), but that's the spot most worth a second look.

## Not in this PR (follow-ups)
The slide-up dashboard chrome, seed macros, the activity ledger, and the reset button.

[...]

PR #229, opened 2026-06-11
ci: build the deploy image on every PR (no deploy)

jee7s:
## Problem
The deploy flow's **build is never exercised during an MR**. `ci.yml` runs `npm ci` → `prisma generate` → `tsc --noEmit` → jest; the actual production build (`docker build` → `npm run build`) lives in `deploy-dev.yml`, which runs **only on push to `main`**. And `tsc` is not a substitute — it doesn't resolve bundler imports, so a Turbopack *"module not found"* can type-check clean and still fail `next build`.

This bit PR #207: its merge produced [a failed dev deploy](https://github.com/innovationtreehouse/checkin/actions/runs/27308444666) —
```
Error: Turbopack build failed with 1 errors:
Module not found: Can't resolve '../page.module.css'   (src/app/signin/page.tsx:8)
```
A silent merge skew (`#200` added `signin/page.tsx` importing `page.module.css`; #207 deleted that file — different files, so **no merge conflict**). CI was green; the build only broke post-merge in the deploy.

## Change
Add a **`deploy-build`** job to `ci.yml` that runs the exact production `docker build` the deploy uses (Dockerfile → `npm run build`), **minus `docker push`** — so the deploy build is gated on every PR/merge-queue run **without deploying anything**. Also corrects the misleading comment on the `tsc` step.

The image is discarded, so architecture is irrelevant — this uses the default x64 runner (the real deploy builds arm64 for ECS on its own runner).

## Tested locally
- `docker build .` on the broken PR #207 branch → **fails** with the exact error above (gate catches it). ✅
- `docker build .` on `main` → **succeeds** (gate is green on healthy code). ✅


jee7s:
Update: when this PR's CI ran the new gate against current `main`, it **failed** — because `#207` had since merged to `main` (`71fa332`), deleting `page.module.css` while `signin/page.tsx` still imported it. So `main`'s deploy build was actually broken, and the gate caught it (exactly its job).

Folded the fix into this PR so it's green and `main`'s deploy is unbroken in one go:
- `fix(signin): drop import of the deleted page.module.css` — `signin` only used two layout rules from that module; inlined them (it still uses the legacy `glass-*` utilities, not Mantine).

Both jobs now pass: **Deploy build (no deploy)** ✅ and **Run Tests** ✅. Verified locally too (`docker build` fails before the fix, passes after).

PR #218, opened 2026-06-10
feat(ci): deploy to prod when a release is published

dkaygithub:
Adds **Deploy prod**, triggered by publishing a GitHub release in this repo. It mirrors the now-working `deploy-dev.yml` step-for-step (so fixes #211/#212/#215/#217 all carry over), with release semantics:

- **Trigger**: `release: [published]` — cut a release with `gh release create v1.2.3 --target main --generate-notes` and it ships. Drafts don't fire; publishing them does.
- **Safety gate**: refuses to deploy a commit whose `Run Tests` check isn't green (releases can technically be cut from any commit).
- **Image tags**: pushes `checkin-prod:<sha>` and `checkin-prod:<release-tag>` for traceability/rollback.
- **Env split**: all-prod resources — `checkin-prod` ECR/cluster/service/task families, `checkin-deploy-prod` OIDC role, https://ops.innovationtreehouse.org. (Verified all exist in AWS, and the `checkin-prod/*` secrets all have values.)
- **GitHub environment** `production` is declared, so required-reviewer protection can be layered on in repo settings whenever desired — no workflow change needed.
- **Report**: run step summary (releases have no comment thread; happy to wire a Slack hook or release-notes append later if wanted).

Rollback story: re-publish (or re-run the deploy of) the previous release, or `update-service` to the prior task-def revision.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #249, opened 2026-06-12
fix(cron): make event reminders idempotent with per-RSVP reminderSentAt

thpr:
## What

`GET /api/cron/reminders` selected events in a `[now+2h, now+2h15m]` window with **no per-RSVP dedup**. On a 15-minute-interval cron, an event sitting near a window boundary matched on two consecutive runs → `EVENT_STARTING_SOON` double-sent to the same `ATTENDING` RSVPs.

## Fix

- **Schema:** add nullable `reminderSentAt DateTime?` to `RSVP` (`@sensitivity:internal`) + migration `20260611000000_add_rsvp_reminder_sent_at`.
- **Route:** query now filters `reminderSentAt: null`; the timestamp is stamped **only after** `sendNotification` resolves — a send failure leaves the RSVP eligible next run instead of silently dropping the reminder.
- **Test:** new `cronRemindersEdge.integration.test.ts` — boundary event (start +2h5m) run through `GET` twice asserts run 1 notifies once, run 2 is a no-op, RSVP notified exactly once across both runs.

## Reviewer notes

- The characterization test referenced in the original task did not exist in the tree, history, or any branch — built the end-state idempotency test directly instead.
- ⚠️ **Deploy:** prod must run `prisma migrate deploy` before the cron picks up the new `reminderSentAt: null` filter (column must exist).
- Verified locally: throwaway Postgres + `prisma db push`, full integration suite **327 passed / 54 suites** (both cron suites green). Pre-commit hook was bypassed with `--no-verify` (user-authorized) — it false-fails in worktrees per the documented `testPathIgnorePatterns` rootDir issue, not a real test failure.

### DIFF
--- a/checkin-app/prisma/migrations/20260611000000_add_rsvp_reminder_sent_at/migration.sql
+++ b/checkin-app/prisma/migrations/20260611000000_add_rsvp_reminder_sent_at/migration.sql
@@ -1,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "RSVP" ADD COLUMN "reminderSentAt" TIMESTAMP(3);
--- a/checkin-app/src/app/api/cron/reminders/route.ts
+++ b/checkin-app/src/app/api/cron/reminders/route.ts
@@ -26,7 +26,7 @@ export async function GET(req: Request) {
     try {
         const now = new Date();
         const twoHoursFromNow = new Date(now.getTime() + 2 * 60 * 60 * 1000);
-        // Add a 15-minute window to avoid duplicate triggers if cron is every 15m
+        // Look 15 minutes past the 2h mark so a 15m-interval cron can't miss an event.
         const windowEnd = new Date(twoHoursFromNow.getTime() + 15 * 60 * 1000);
 
         const upcomingEvents = await prisma.event.findMany({
@@ -38,7 +38,9 @@ export async function GET(req: Request) {
             },
             include: {
                 rsvps: {
-                    where: { status: 'ATTENDING' }
+                    // Only RSVPs that haven't been reminded yet — the window can overlap
+                    // across runs, so reminderSentAt is what makes this idempotent.
+                    where: { status: 'ATTENDING', reminderSentAt: null }
                 }
             }
         });
@@ -51,7 +53,18 @@ export async function GET(req: Request) {
                 const promise = Promise.resolve(sendNotification(rsvp.participantId, 'EVENT_STARTING_SOON', {
                     eventName: event.name,
                     hours: 2
-                })).then(() => {
+                })).then(async () => {
+                    // Mark sent only after the notification resolves, so a send failure
+                    // leaves it eligible for the next run rather than silently dropped.
+                    await prisma.rSVP.update({
+                        where: {
+                            eventId_participantId: {
+                                eventId: event.id,
+                                participantId: rsvp.participantId
+                            }
+                        },
+                        data: { reminderSentAt: new Date() }
+                    });
                     notificationsSent++;
                 });
                 notificationPromises.push(promise);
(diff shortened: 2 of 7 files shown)


checkin-app/src/app/api/cron/reminders/route.ts
```diff
@@ -51,7 +53,18 @@ export async function GET(req: Request) {
...
+                    // Mark sent only after the notification resolves, so a send failure
+                    // leaves it eligible for the next run rather than silently dropped.
+                    await prisma.rSVP.update({
+                        where: {
+                            eventId_participantId: {
+                                eventId: event.id,
+                                participantId: rsvp.participantId
+                            }
+                        },
+                        data: { reminderSentAt: new Date() }
```
dkaygithub, line 76: Worth naming the guarantee explicitly: this is **at-least-once with best-effort dedup**, not exactly-once. The one residual double-send window is send-succeeds-then-stamp-fails — if `sendNotification` resolves but this `rSVP.update` throws (DB hiccup), the notification went out but `reminderSentAt` stays null, so the next run re-sends. That's fundamental (you can't transactionally bind an external send to a DB write) and a huge improvement over the old every-overlapping-run behavior, so I'd accept it — just good to acknowledge it's not airtight. Same applies if two cron executions ever overlap (both read null before either stamps); fine as long as the platform doesn't run the job concurrently with itself.

checkin-app/src/app/api/cron/reminders/route.ts
```diff
@@ -38,7 +38,9 @@ export async function GET(req: Request) {
             },
             include: {
                 rsvps: {
-                    where: { status: 'ATTENDING' }
+                    // Only RSVPs that haven't been reminded yet — the window can overlap
+                    // across runs, so reminderSentAt is what makes this idempotent.
+                    where: { status: 'ATTENDING', reminderSentAt: null }
```
dkaygithub, line 53: `reminderSentAt` is never cleared, so the dedup key is per-RSVP for all time, not per-scheduled-time. Two consequences worth a conscious decision: (1) if an event is **rescheduled** later, already-reminded attendees won't get a fresh 2h reminder for the new time; (2) an ATTENDING → NOT_ATTENDING → ATTENDING toggle won't re-remind. Both are probably acceptable for this feature, but flagging so it's intentional rather than incidental.

checkin-app/prisma/migrations/20260611000000_add_rsvp_reminder_sent_at/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "RSVP" ADD COLUMN "reminderSentAt" TIMESTAMP(3);
```
dkaygithub, line 2: Deploy ordering (you flagged this in the PR body — confirming the state): **dev** is covered, `deploy-dev.yml` runs `prisma migrate deploy` (with retries) before the app comes up, and CI runs `migrate deploy` + `migrate diff --exit-code` so a mismatched migration is caught. **Prod** has no deploy workflow yet — that's #218 (still open), so whoever lands prod deploy needs to ensure `migrate deploy` runs before the new code, otherwise the `reminderSentAt: null` filter hits a missing column and the cron 500s. Not blocking this PR, just a cross-PR dependency to track.

dkaygithub (commented):
Clean, well-scoped idempotency fix — the `reminderSentAt: null` filter + stamp-after-send is the right shape, and the failure handling is actually a strength: any send/stamp rejection makes `Promise.all` reject → 500 → cron retry, and the already-stamped RSVPs are excluded on the retry, so the stamp is what makes retries safe rather than duplicative. Verified the schema dependency (`@@id([eventId, participantId])` makes the `eventId_participantId` update key valid) and that `classifications.ts` is generator output matching schema order. Three non-blocking notes below.

dkaygithub (approved):

PR #275, opened 2026-06-14
feat(household): enforce max 2 leads per household (#269)

### LINKED ISSUE #269: Max 2 Household Leads

dkaygithub:
Resolves #269.

## What

Caps `HouseholdLead` rows per household at **2**, enforced in business logic on the existing join table (not by ripping it out for raw columns on `Household`). The limit is policy, so it lives in one place — keeping the auditable join table, its composite-PK uniqueness, and parity with `CorporationLead`.

## How

New `addHouseholdLead(db, householdId, participantId)` in `src/lib/household/leads.ts` is the single funnel for **every** lead write (13 call sites). It is:

- **Idempotent** — already-a-lead is a no-op and doesn't count against the cap.
- **Atomic & race-free** — the count check and insert run in a transaction that takes a `FOR UPDATE` row lock on the household, so two concurrent promotions can't both pass the count check. (Postgres' default Read Committed alone wouldn't prevent this — the lock is what does.)
- **Dual-mode** — given the prisma singleton it opens its own transaction; given a `tx` client (sign-up, public-register) it joins the caller's transaction so the lock participates there.

Throws `HouseholdLeadLimitError`, which callers translate to a clean `400` (and to `IntakeError "lead_limit"` in the membership flow).

## Notable edge case

The bulk-import **household merge** rejects up front — before moving any participant — so a merge that would exceed the cap never leaves a half-merged household. (Chosen behavior: fail the import row, recorded in the import errors list.)

## Tests

- Adds an integration test asserting a third lead is rejected (`400`, no row created).
- `tsc` + `eslint` clean; pre-commit ran the full mocked suite (169 passed). Integration suite runs in CI.

---

## Review follow-up: honest UX when the cap rejects (8bb7812)

Review feedback: when a write is rejected by policy (or a random error), the UI must be clear about what saved and what didn't — partial saves are fine as long as they're transparent. The cap rejected on a write that ran *after* other writes had already persisted, so the UI was reporting total failure for a partial success (and intake silently dropped the children loop). Fixed end-to-end:

- **`saveIntake`** no longer throws on the cap. It records a structured `IntakeRejection`, **keeps saving the rest of the form (children included)**, and returns `{ state, rejections }`.
- **Membership "join" form** shows an amber *"Saved — with a couple of things to know"* alert from `rejections`, visually distinct from the green success / red error, in applicant language ("parents/guardians", not internal "leads").
- **Member-edit PATCH** keeps the field edits and returns `200` with `leadRejection` instead of a blanket `400`; the household page surfaces the caveat and refreshes so the form reflects reality.
- **Left already-honest paths alone:** `public-register` (atomic `$transaction` — cap rolls the whole registration back) and the explicit "Promote to lead" button.

`tsc` + `eslint` clean; mocked suite (169) green; no integration test asserted the old cap-`400` for intake/member edits.


checkin-app/src/lib/membership/intake.ts
```diff
@@ -209,11 +220,7 @@ export async function saveIntake(userId: number, input: IntakeSaveInput) {
                 },
             });
             // A second guardian is a household lead (parent).
-            await prisma.householdLead.upsert({
-                where: { householdId_participantId: { householdId, participantId: sp.id } },
-                create: { householdId, participantId: sp.id },
-                update: {},
-            });
+            await addLeadOrThrow(sp.id);
```
thpr: This (and line 234) both seem to allow partial records (scenario: allergy succeeds, lead throws).  That may be okay, but is that partial update clear to the user?  That is a different pattern that most online purchases, which are "when we reject for lack of phone number, nothing happened" (and in that case the phone number field turned red)

dkaygithub: Hmmm, that probably stemmed from my directive to auto save as the user goes along.

I'll bump it to make sure that's clear to the user

thpr (approved):

Issue #278, opened 2026-06-14
Gate volunteer discount via auto-managed Shopify customer segment + validate payment webhook


dkaygithub:
Follow-up to #272 / #277 (volunteer discount moved from per-tier pricing to a Shopify discount code).

## Problem
The volunteer discount is currently a **self-serve code appended to a public cart link** (`…/cart/ID:1?discount=<CODE>`), and the `orders/paid` webhook (`checkin-app/src/app/api/webhooks/shopify/route.ts`) activates the membership purely off the `Membership_Process_ID` cart attribute. It does **not** validate entitlement, amount, product, or discount code.

Net effect: a non-volunteer can append `?discount=<CODE>` (or reuse a shared code), underpay, and still be activated. Entitlement is honor-system.

(See the `TODO(volunteer-discount)` markers in `route.ts` and `checkin-app/src/lib/membership/payment.ts`.)

## Long-term fix (preferred)
Gate the volunteer coupon to an **auto-managed Shopify customer segment** containing only volunteer households, so Shopify itself refuses the code for everyone else — no app-side check needed.

- [ ] Define the Shopify customer segment for volunteer households.
- [ ] Auto-manage segment membership from our `Membership.isVolunteer` source of truth (add/remove customers as volunteer status changes).
- [ ] Restrict the Shopify discount to that segment.
- [ ] Confirm a non-volunteer cannot redeem the code even with the URL param.

## Interim option (if we want a stopgap before the segment exists)
Validate in the webhook before `activateByProcessId`, using fields already on the order payload:
- [ ] If `order.discount_codes[].code === BoardSettings.volunteerDiscountCode`, require `membership.isVolunteer === true`; otherwise don't activate + alert the board.
- [ ] Sanity-check `order.total_price` against the expected tier dues.
- [ ] (optional) verify line-item `variant_id` matches the configured membership product, and `customer.email` belongs to the household.

## Decision deferred
Enforcement behavior on a failed check (reject + alert vs. activate + flag for review) — to be decided when this is picked up.

thpr:
**Backlog reconciliation (2026-08-01)** — the segment-pricing design that scopes this (member-segment-gated discount + payment-webhook validation) is **merged (#929)**. CO1/CO2 readiness moved to IN-DESIGN. Source: INDEX CO2 (+CO1).

jee7s:
**HOLD** — the issue's own "Decision deferred" section is still deferred, and the merged design did not settle it.

[...]

PR #428, opened 2026-06-28
feat(membership): make the background check non-blocking in the application flow

jee7s:
## Problem

The background check was a **hard gate between contract/consent and payment**. The flow was strictly linear:

```
INTAKE → PENDING_EXTERNAL_ACTION → PENDING_BG_REVIEW → PENDING_PAYMENT → ACTIVE
         (sign + consent)          (2 reviewers,        (pay)
                                    long-tail/human)
```

Averity has no API, so the check is human-marked and slow. Applicants sat in `PENDING_BG_REVIEW` and **could not pay** until the board cleared it — a long-tail step blocking everything behind it.

## What changed

The background check becomes a **parallel track**. After the contract is signed **and** the check is handled (consent recorded, or a still-valid prior check), the application advances **straight to `PENDING_PAYMENT`**. The membership only goes **ACTIVE once BOTH payment and a valid check are done — whichever finishes last.**

```
INTAKE → CONTRACT & BACKGROUND CHECK → PAYMENT ──┐
                                                 ├─► ACTIVE  (needs: paid AND bg-cleared)
         (BG review runs in parallel) ───────────┘
              └─ paid first → PENDING_BG_CLEARANCE (paid, membership still inactive)
```

Behavior (decisions confirmed with the requester):
- **Payment gate** — unlocks once the contract is signed and the check is *handled* (consent clicked, or a valid existing check). The applicant pays while the review runs.
- **Never active without a valid check** — if paid first, the process holds at the new `PENDING_BG_CLEARANCE` status (membership stays inactive); the 2nd reviewer approval then flips it ACTIVE.
- **Auto-validate a still-valid prior check** — at intake, if any guardian's `lastBackgroundCheck` is within `BoardSettings.bgRecheckMonths`, the requirement auto-clears (mirrors renewals) with a "no new check needed" message — no consent, no wait.
- **Reject after payment → BLOCKED** + the board is emailed (manual refund); no automated money movement.

## Model

- New `MembershipProcess.bgClearedAt` — single source of truth for "BG cleared this cycle" (set by 2 approvals, a board override, or a detected valid prior check).
- New status `PENDING_BG_CLEARANCE` — paid, awaiting the check.
- `review.ts` and `payment.ts` converge on a **`FOR UPDATE`-locked gate** so the payment/clearance race can't lose an update (whoever commits second sees the other's field set and activates).
- Background-review eligibility is now keyed off the BG fields (`AWAITING_BG_WHERE`), not status — so review works in parallel across `PENDING_PAYMENT` / `PENDING_BG_CLEARANCE`, and the board todo-count follows.

### Migration
`20260628120000_bg_check_nonblocking`: adds the enum value + `bgClearedAt`, then **backfills** — existing `PENDING_PAYMENT`/`ACTIVE` marked cleared; in-flight `PENDING_BG_REVIEW` moved to `PENDING_PAYMENT` (unblocked, review continues in parallel). Verified: applies cleanly and `migrate diff` reports no drift from the schema.

## UI

- `MembershipFlowStepper`: 4 steps (Family → Contract & Check → Payment → Member); BG is parallel, `PENDING_BG_CLEARANCE` shows Payment done / Member pending.
- Member page: auto-validated message, "pay now — check in progress" note on Payment, and a "Paid — activating when your check clears" holding card.
- Board applications view: shows BG progress (`N/2`) in parallel and the new holding state.

## Validation (against a live Postgres + dev server)

- New integration suite `membershipBgNonBlocking` (6 cases): advance-to-payment, pay-before-check → holding → activate, check-before-pay, reject-after-pay → blocked, fresh-check auto-clear, webhook idempotency. **All membership integration tests pass (52).**
- Updated the existing tests that encoded the old sequential order (external/review/payment/contract-sync) to the new behavior.
- Full integration run matches the clean-`main` baseline exactly — **this change adds zero new failures** (pre-existing environmental failures: cron/kiosk/renewal-concurrency).
- Rendered and screenshotted the new member states (parallel-payment note, paid-awaiting-check, auto-validated check).
- Rebased on latest `main`; reconciled with #397 (certify guard — kept), #409/#399 (intake/renewal refactors), #413 (BG-review tab — works via the updated queue). ESLint + `tsc` clean on changed files.

### DIFF
--- a/checkin-app/prisma/schema.prisma
+++ b/checkin-app/prisma/schema.prisma
@@ -233,6 +233,11 @@ enum MembershipProcessStatus {
   PENDING_EXTERNAL_ACTION
   PENDING_BG_REVIEW
   PENDING_PAYMENT
+  // Paid, but the background check has not yet cleared. The membership stays
+  // INACTIVE here — it activates automatically once two reviewers approve (or a
+  // prior valid check is detected). The BG review runs in parallel with payment,
+  // so an applicant is never forced to wait on the long-tail check before paying.
+  PENDING_BG_CLEARANCE
   ACTIVE
   BLOCKED
   PENDING_RENEWAL
@@ -310,6 +315,12 @@ model MembershipProcess {
   contractSignedAt      DateTime?
   /// @sensitivity:internal
   bgConsentAt           DateTime?
+  /// When the background-check requirement was satisfied for this cycle — either
+  /// two reviewers approved, or a still-valid prior check was detected at intake.
+  /// The single source of truth for "BG cleared", independent of the payment
+  /// track; the membership only goes ACTIVE once this is set AND dues are paid.
+  /// @sensitivity:internal
+  bgClearedAt           DateTime?
   /// @sensitivity:internal
   shopifyDraftOrderId   String?
   /// @sensitivity:personal
--- a/checkin-app/prisma/migrations/20260628120000_bg_check_nonblocking/migration.sql
+++ b/checkin-app/prisma/migrations/20260628120000_bg_check_nonblocking/migration.sql
@@ -1,0 +1,26 @@
+-- Background check becomes a parallel (non-blocking) track in the membership
+-- application flow. Adds the PENDING_BG_CLEARANCE status (paid, awaiting check)
+-- and a bgClearedAt marker, then backfills existing rows so nothing regresses.
+
+-- AlterEnum: new "paid, awaiting background check" holding status.
+-- (Not used in this migration's data statements, so it is safe inside the
+-- migration transaction on PostgreSQL 12+.)
+ALTER TYPE "MembershipProcessStatus" ADD VALUE 'PENDING_BG_CLEARANCE';
+
+-- AlterTable: single source of truth for "BG cleared this cycle".
+ALTER TABLE "MembershipProcess" ADD COLUMN "bgClearedAt" TIMESTAMP(3);
+
+-- Backfill 1: any process already at or past payment cleared its background
+-- check under the old (sequential) ordering, so mark it cleared. Use the most
+-- meaningful timestamp available.
+UPDATE "MembershipProcess"
+   SET "bgClearedAt" = COALESCE("paidAt", "stageEnteredAt", CURRENT_TIMESTAMP)
+ WHERE "status" IN ('PENDING_PAYMENT', 'ACTIVE');
+
+-- Backfill 2: in-flight INITIAL applications still in background review move to
+-- PENDING_PAYMENT so they are unblocked immediately; their review continues in
+-- parallel (bgClearedAt stays NULL until two reviewers approve). Renewal review
+-- (RENEWAL_PENDING_BG) is intentionally left as-is.
+UPDATE "MembershipProcess"
+   SET "status" = 'PENDING_PAYMENT', "stageEnteredAt" = CURRENT_TIMESTAMP
+ WHERE "status" = 'PENDING_BG_REVIEW';
--- a/checkin-app/src/app/__tests__/membershipBgNonBlocking.integration.test.ts
+++ b/checkin-app/src/app/__tests__/membershipBgNonBlocking.integration.test.ts
@@ -15,6 +15,7 @@ import { markContractSigned, markBgConsent } from '@/lib/membership/external';
 import { attest } from '@/lib/membership/review';
 import { certifyPaymentPlan, activate } from '@/lib/membership/payment';
 import { submitIntake } from '@/lib/membership/intake';
+import { beginRenewal } from '@/lib/membership/renewal';
 import prisma from '@/lib/prisma';
 import { sendEmail } from '@/lib/email';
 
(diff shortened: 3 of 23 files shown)


checkin-app/prisma/schema.prisma
```diff
@@ -233,6 +233,11 @@ enum MembershipProcessStatus {
   PENDING_EXTERNAL_ACTION
   PENDING_BG_REVIEW
   PENDING_PAYMENT
+  // Paid, but the background check has not yet cleared. The membership stays
+  // INACTIVE here — it activates automatically once two reviewers approve (or a
```
thpr, line 237: Inactive, but can they enroll in programs?  If something is capacity constrained, do we want this to be such a gate?

thpr (changes requested):
ISSUE — payment webhook after BLOCKED silently drops the payment
payment.ts:117

if (process.paidAt || process.status === "ACTIVE" || process.status === "BLOCKED") {
    return { kind: "noop" as const };
}
Race: process at PENDING_PAYMENT, 1 approval, bgConsentAt set. Applicant checks out at Shopify (money gone) but the orders/paid webhook lags (Shopify retries routinely). In that window reviewer #2 REJECTs → BLOCKED. Webhook lands → status === "BLOCKED" → noop.

Result: money taken at Shopify, paidAt never recorded, board never notified for refund. The PR's paid-reject notify only fires when paidAt was already set at reject time (review.ts:209 notifyPaidReject: !!process.paidAt) — the reverse order is the hole.

Tested case is pay→reject (test:129). reject→pay is untested.

Worse: webhook logs "Activated membership for process N" (route.ts:85) even on the noop — false trail.

Fix: when a payment webhook hits a BLOCKED process, don't resurrect it but do record paidAt (or a paidWhileBlockedAt) and fire notifyBoardPaidReject. Don't drop the money signal on the floor.  Need to determine how to surface in UI for finance action

jee7s:
Fixed in f22422e. Root cause was the shared `activate()` treating BLOCKED as a silent noop, so the fix is one guard there (where both webhook + certify route through), not per-caller.

- **reject→pay**: a payment on a BLOCKED process now records `paidAt` (+ order id), stays BLOCKED (never activate without a valid check), and fires `notifyBoardPaidReject`. So both orderings alert the board now.
- **False log**: the Shopify webhook logs the real outcome (ACTIVE / BLOCKED-refund-needed / awaiting-clearance / no-op), no more "Activated" on a drop.
- **Idempotency**: guard is now `paidAt || ACTIVE` (not `BLOCKED`), so a webhook retry on the recorded payment no-ops without a second alert.
- **UI**: BLOCKED applications with `paidAt` show "💸 already paid — refund likely needed" + the override button reads "Override → activate" (paid) so finance sees it.
- `notifyBoardPaidReject` moved to `boardAlerts.ts` to avoid a payment↔review import cycle.
- Added the reject→pay integration test (the untested reverse order) + webhook-retry idempotency. 53/53 membership integration tests pass.

thpr:
Blocker — renewals with a valid prior check get stuck forever
renewal.ts:92-99 — beginRenewal sends a fresh-BG renewal straight to PENDING_PAYMENT but never sets bgClearedAt. The diff rewired activate() to gate on bgClearedAt, but didn't touch this caller.

Trace, common renewal case (BG still valid → no re-review):

beginRenewal → PENDING_PAYMENT, bgClearedAt = null, bgConsentAt = null.
Member pays → activate(): activating = !!process.bgClearedAt = false → parks at PENDING_BG_CLEARANCE.
Nothing can clear it: AWAITING_BG_WHERE / isAwaitingBgReview require bgConsentAt set for the PENDING_* states. Renewal has no consent → not in any reviewer queue.
→ Paid renewal, membership never goes ACTIVE, invisible to the board. Silent. The new integration suite only covers INITIAL, so it passed.

Fix mirrors intake.ts:920-924:

data: { status: nextStatus, stageEnteredAt: new Date(),
        ...(bgFresh ? { bgClearedAt: new Date() } : {}) },
Renewals that do re-review (RENEWAL_PENDING_BG) are fine — they go through clearBackgroundCheck, which sets bgClearedAt.

thpr:
overrideBlocked "approve" violates its own lock contract
review.ts:1375 — calls clearBackgroundCheck in a $transaction with no FOR UPDATE, though that function's doc says "Must run inside a tx holding a FOR UPDATE lock." A board approve racing a late payment webhook on the same BLOCKED row can interleave: activate records paidAt + stays BLOCKED + emails refund, while override reads stale paidAt=null → lands PENDING_PAYMENT, paid but never ACTIVE. Board-driven + narrow window, but it's the exact race the lock exists to prevent. Add the SELECT … FOR UPDATE like attest/activate.

jee7s:
Both blockers fixed in 3845f66:

**Renewal stuck-forever** — `beginRenewal` now sets `bgClearedAt` when the prior check is still valid (mirrors `intake.ts`), so a fresh renewal pays → ACTIVE instead of parking at `PENDING_BG_CLEARANCE`. Re-review renewals were already fine (they go through `clearBackgroundCheck`).

**override-approve lock** — wrapped the `clearBackgroundCheck` call in `SELECT … FOR UPDATE` like `attest`/`activate`, so a board approve racing a late payment webhook reads a fresh `paidAt` and converges to ACTIVE.

Added a renewal fresh-check → pay → ACTIVE integration test (the suite was INITIAL-only, which is exactly how this slipped). 54/54 membership integration tests pass.

**Re the inline question on `schema.prisma:237` (can a paid-but-pending household enroll in capacity-constrained programs?):** no — `PENDING_BG_CLEARANCE` keeps `membership.status = NONE` (only the final ACTIVE flip sets it), and member-only program registration gates on `isActiveMember` (`programs/[id]/route.ts:49-51` → `membership.status === "ACTIVE"`). So a pending household can't hold a member-only slot — same as any non-member today. This PR doesn't change program gating; if we ever want paid-pending to count as "member enough" for enrollment, that's a deliberate follow-up, not a default I'd flip here.

jee7s:
Scrutinized the full diff for further gaps. One more found + fixed (f49c415):

**`activate()` over-accepted statuses** — after the BLOCKED refund branch it fell through for *any* status, so a payment for an INTAKE / PENDING_EXTERNAL_ACTION / renewal-review process (forged or replayed webhook — #278 — or an unexpectedly-moved process) would record `paidAt` and advance it to `PENDING_BG_CLEARANCE`. A checkout link is only minted at `PENDING_PAYMENT`, so anything else is anomalous → log + no-op instead of corrupting state. Guard test added.

Checked and clean:
- No other `PENDING_PAYMENT` code consumer needs `PENDING_BG_CLEARANCE` handling (only doc comments).
- No cron expires membership processes by status (no PENDING_BG_CLEARANCE skip needed).
- No finance/report sums `paidAt`, so a BLOCKED+paid row can't be double-counted.
- Renewals never reach `PENDING_BG_CLEARANCE` (they pay only at PENDING_PAYMENT after bgCleared → ACTIVE), so the renewal in-flight arrays + the `membership_one_inflight_renewal` index are still correct.
- `paid_while_blocked` → board reset → `PENDING_BG_CLEARANCE` is re-reviewable (consent present), and re-approval → ACTIVE (paidAt already set).
- Concurrency suites (review + external) still pass with the added FOR UPDATE.

55 membership integration tests pass; tsc + eslint clean.

thpr (approved):

PR #482, opened 2026-06-29
test(flow): dockerized dev-server flow tests + AGENTS.md (verifies #428)

jee7s:
Stacked on #428 (base = `feat/bg-check-nonblocking`) because the flow test asserts #428's behaviour and needs that code to pass. Rebase onto `main` once #428 merges.

## A third test class: flow tests

End-to-end **user-journey** tests that drive a *running* dev server over HTTP — no app imports, no DB, no mocks, just `fetch` + response assertions (how a client actually exercises the wired-together routes + auth + DB).

- **`deploy/docker-compose.flow.yml`** — Postgres + a **dev-mode** app container (`next dev`, `CHECKIN_ENV=local`), auto-installed, `prisma db push`-ed, and seeded. Dev mode is required, not incidental: persona-mint login is gated on `NODE_ENV != production`, and the prod image pins `NODE_ENV=production`.
- **`.github/workflows/flow-tests.yml`** — `docker compose up --wait`, run the flow tests inside the app container, dump app logs on failure, `down -v`.
- **`flow-tests/helpers.ts`** — persona-mint login over HTTP (resolves the persona by stable email, since ids are autoincrement) + a small JSON `fetch` wrapper.
- **`jest.flow.config.js`** — node env that **skips `jest.setup.js`** (so native `fetch`/`getSetCookie` work, and prisma/next-auth aren't mocked). Run with `npm run test:flow`. Flow tests are excluded from the unit/integration runs.

## The one journey (initial): verifies PR #428

`flow-tests/membership-bg-nonblocking.flow.test.ts` walks the real journey: a non-member applicant fills intake, the board records the two external actions, and pays before the check clears — asserting:
1. after the external step the application is at **PENDING_PAYMENT** (payment unblocked; pre-#428 it parked at PENDING_BG_REVIEW), and
2. paying before the check clears holds at **PENDING_BG_CLEARANCE** with the membership still inactive.

Verified locally against the dev server (green).

## AGENTS.md

New top-level agent guide: explains the three test classes (unit / integration / flow, with the flow gotchas) and a **docs map** pointing to README, CONSTITUTION, `docs/designs/*` (incl. `CUJS.md` as the journey source), `docs/security/*`, the setup guides, and each subproject's README.

### DIFF
--- a/.github/workflows/flow-tests.yml
+++ b/.github/workflows/flow-tests.yml
@@ -1,0 +1,39 @@
+name: Flow tests
+
+# Exercises real user journeys end-to-end over HTTP against a dockerized
+# DEVELOPMENT server (deploy/docker-compose.flow.yml). See AGENTS.md → Flow tests.
+
+on:
+  pull_request:
+    branches: [main]
+  workflow_dispatch: {}
+
+permissions:
+  contents: read
+
+concurrency:
+  group: flow-tests-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  flow:
+    name: User-journey flow tests (docker dev server)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+
+      # Brings up Postgres + the dev server, installs/seeds, and blocks on their
+      # healthchecks (--wait). start_period in the compose file covers npm ci + seed.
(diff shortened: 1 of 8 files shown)


thpr (approved):
End to end Test  🥳

PR #513, opened 2026-06-29
feat(address): structured address fields for Household & Corporation

thpr:
## ⚠️ Potentially controversial

This is a **breaking schema change with a destructive migration**. Flagging for careful review before merge:

- **Drops the `address` column** on `Household` and `Corporation` after backfill. The migration is one-way — old free-text is preserved only as `line1`.
- **Backfill is naive**: the entire old free-text address goes into `line1` verbatim (no parsing into city/state/zip). Existing members will show a full address string in the street field until they re-enter it. We chose this over a heuristic split because parsing arbitrary free text is unreliable.
- **US-only assumption**: no `country` field; `state` is a 2-letter code. Non-US addresses aren't modeled yet.
- Migration + drop ship in **one PR** (no two-phase deploy). Verify the backfill on production data before merging.

## What

Replace the single free-text `address` column with structured fields:
`line1`, `line2`, `city`, `state`, `postalCode` (all `@sensitivity:personal`).

## Changes

- **Schema + migration** (`20260629000000_structured_address`): add 5 columns to Household & Corporation, backfill `line1` from `address`, drop `address`.
- **`src/lib/address.ts`**: shared `normalizeAddressInput` / `formatAddress` / `pickAddress` helpers + self-check.
- **APIs**: `household/settings`, `admin/households/[id]`, membership `intake.ts` — accept/return structured fields; audit snapshots updated; intake requires `line1`.
- **UI**: `my-household`, `membership` intake, `AdminEditHouseholdModal` — split single input into street/apt/city/state/zip.
- **Dev/seed/import**: zoho-import, seed-helpers, seed_20_users, CSV admin import map to `line1` (bulk import to be redone, marked).
- **Tests**: security stripper/outbound now assert `line1` stripping at personal tier; integration fixtures updated.

## Verification

- `tsc --noEmit` clean.
- Migration applies clean on a fresh Postgres.
- 50 security unit tests + 58 address-path integration tests pass (intake, householdAPI, authz, bg, adminParticipantHousehold). zoho/import/audit suites green (1 pre-existing, unrelated attendance-audit flake).

### DIFF
--- a/checkin-app/prisma/migrations/20260629000000_structured_address/migration.sql
+++ b/checkin-app/prisma/migrations/20260629000000_structured_address/migration.sql
@@ -1,0 +1,25 @@
+-- Structured address: replace free-text `address` with line1/line2/city/state/postalCode
+-- on Household and Corporation.
+
+-- AlterTable: Household
+ALTER TABLE "Household" ADD COLUMN "line1" TEXT;
+ALTER TABLE "Household" ADD COLUMN "line2" TEXT;
+ALTER TABLE "Household" ADD COLUMN "city" TEXT;
+ALTER TABLE "Household" ADD COLUMN "state" TEXT;
+ALTER TABLE "Household" ADD COLUMN "postalCode" TEXT;
+
+-- AlterTable: Corporation
+ALTER TABLE "Corporation" ADD COLUMN "line1" TEXT;
+ALTER TABLE "Corporation" ADD COLUMN "line2" TEXT;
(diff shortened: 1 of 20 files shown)


jee7s:
I think we want this as it would make later integration with location services easier.

dkaygithub (approved):

thpr:
Controversial only because it breaks all the Zoho load we did.  So we'll need to redo.  That said, I think there are other schema changes we can fold into such a redo that would be valuable...

PR #517, opened 2026-06-29
refactor(attendance): move Current view to /attendance/current

thpr:
## What

The Current tab lived on the bare `/attendance` URL, so the index route and the first tab shared a path. This moves the Current view to its own `/attendance/current` route and turns `/attendance` into a query-preserving redirect, giving each tab a distinct URL.

## Changes

- `git mv` `attendance/page.tsx` → `attendance/current/page.tsx` (import of `AttendanceTabs` adjusted to `../`)
- New `attendance/page.tsx`: client redirect to `/attendance/current`, preserving the query string
- `AttendanceTabs.tsx`: "Current" tab href → `/attendance/current`

## Reviewer notes

- **Kiosk safe:** signed kiosk URLs `/attendance?mode=kiosk&sig=...&ts=...&nonce=...` keep working — the redirect carries the query string through to `/attendance/current`, which already reads those params.
- Nav link, nav badge (`navBadgeFor`), and the search Index entry still point at `/attendance` and follow the redirect — no change needed.
- Not typechecked locally (worktree has no `node_modules`); diff is mechanical.


dkaygithub:
Can you go ahead and update the kiosk to point to the new URL? The client is now in the repo

jee7s (approved):

PR #530, opened 2026-06-29
feat(public-register): rate limiting + double opt-in, close enumeration oracle

thpr:
## ⚠️ Potentially controversial — read before merging

This adds abuse controls to the public, unauthenticated registration surface. The judgment call is a **UX-vs-risk tradeoff on the customer intake flow**, and reasonable people may disagree:

- **The change:** public registration is now **double opt-in** — submitting the form no longer enrolls anyone. It sends one confirmation email, and the household/participants/enrollments are created only after the parent clicks the link.
- **What we gain:** the endpoint can no longer be used to spam the DB or **bomb a third party's inbox** (it used to create rows + send enrollment email per call, fully unauthenticated). It also stops leaking which emails have accounts.
- **What it costs:** one extra step (and an email round-trip) before a new family is enrolled. That is friction added to the **top of the intake funnel** — some families will drop off at "check your email," and paid-program checkout now happens after the confirm click instead of immediately on submit.

**The core disagreement to resolve:** is the spam / email-bomb / enumeration risk worth adding a confirmation step to new-customer signup? If we decide the intake-funnel friction outweighs the abuse risk, the rate limiting alone (already a separate, merged change on `main`) may be considered sufficient and the double-opt-in half could be dropped. Flagging explicitly rather than deciding it unilaterally.

---

## What's in here

Built on top of the already-merged rate limiter (`src/lib/rate-limit.ts`, IP with IPv6 /64 collapsing + normalized-email keying). This PR adds the registration-flow hardening.

### Double opt-in for `POST /api/programs/[id]/public-register`
- **Request step** (`route.ts`): validate input, apply IP **and** per-normalized-email rate limits, then email an **AES-256-GCM token** carrying the pending registration. **Zero DB writes.** Returns a neutral "check your email" response.
- **Confirm step** (`public-register/confirm/route.ts`): the tokenized link proves email control → runs the original transaction (household + participants + enrollments + emergency contact), notifications, and Shopify checkout URL.
- Token is encrypted (child PII never appears in the emailed URL), tamper-proof, and self-expiring (24h). Stateless — no migration, no GC. Confirm is idempotent against link replay via the unique-email short-circuit.

### Enumeration oracle closed
The request step returns an **identical neutral response** whether or not the email already has an account. The existing-email check moved to confirm, where the caller has already proven inbox control — so it's no longer an anonymous oracle.

### UI
- Register page: success now means "check your email" (no immediate enroll/redirect).
- New confirm page (`register/confirm/page.tsx`): explicit confirm button (no prefetch/scanner auto-submit), handles the checkout redirect after writes succeed.

## Tests
- **Unit:** AES token round-trip / expiry / tamper (`registrationToken.test.ts`).
- **Integration:** `public-register` suites rewritten to drive request→confirm, including the capacity-race regression now exercised on the confirm step. Kept `main`'s `calculateAge` age-boundary regression tests (run at the request step).
- Verified locally against a throwaway Postgres: **13 unit + 15 integration passing.**

## Deploy notes
- No new dependency, no new env var (reuses `NEXTAUTH_SECRET`).
- Rate limiter is process-local (`ponytail:` ceiling noted) — fine for the single `standalone` container; revisit if we ever scale to multiple instances.

### DIFF
--- a/checkin-app/src/app/__tests__/programsPublicRegisterConcurrency.integration.test.ts
+++ b/checkin-app/src/app/__tests__/programsPublicRegisterConcurrency.integration.test.ts
@@ -2,25 +2,29 @@
  * @jest-environment node
  */
 /**
- * Concurrency regression test for POST /api/programs/[id]/public-register.
+ * Concurrency regression test for the public registration capacity lock.
  *
  * Capacity used to be checked against a `_count` read BEFORE the create
- * transaction. Two near-simultaneous public registrations for the last seat
- * both read count==max-1, both passed, and both inserted — overfilling the
- * program. This endpoint is public/unauthenticated, so it is trivially raced.
+ * transaction. Two near-simultaneous registrations for the last seat both read
+ * count==max-1, both passed, and both inserted — overfilling the program. This
+ * flow is public/unauthenticated, so it is trivially raced.
  *
- * The fix locks the Program row (SELECT ... FOR UPDATE) and re-counts INSIDE
- * the existing transaction before creating enrollments. This test fires two
- * concurrent registrations for a 1-seat program and asserts exactly one
- * succeeds and the program is not overfilled.
+ * The fix locks the Program row (SELECT ... FOR UPDATE) and re-counts INSIDE the
+ * create transaction. That transaction now lives in the double-opt-in *confirm*
+ * step, so this test fires two concurrent confirmations (with pre-built tokens,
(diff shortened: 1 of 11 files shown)


jee7s (approved):

thpr:
Closing - the path still had enumeration issues and even a possible DOS of our container.  Better solution in #845

PR #573, opened 2026-06-30
⚠️ refactor(security): scopesHeld switch → declarative SCOPE_BINDINGS (S0–S2) — SECURITY-SURFACE, needs deep review

thpr:
> ## ⚠️ RISKY — significant review required
> This rewrites the **per-row scope resolver**, the function that decides which fields of every API response reach the wire. A subtle error here is a **silent PII leak or an over-restriction**, neither of which throws. CODEOWNERS-gated (`@dkay`). Do not merge on a skim — review the equivalence proof and the binding table field-by-field.

## What
Convert the imperative `switch (modelName)` in `access-resolvers.ts` (`scopesHeld`) into a declarative `SCOPE_BINDINGS` data table interpreted by one engine. This is the **minimum** needed to make the CI security validators *possible* (field-typo, forgotten-model, route-grant seam) — they can only run against declarative data, not an opaque switch.

Implements **Step 2 / S0–S2** of `docs/security/auth-consistency-analysis.md` §7. **Behavior-preserving**: same fields reach the wire for every caller.

## Scope (and what is deliberately NOT here)
- ✅ Resolver refactor + validator **authoring**.
- ❌ Validators are **NOT wired as CI gates** (that's Step 3).
- ❌ No `RouteSpec.returns` added (arrives with the route-migration track).
- ❌ No library/package extraction, no tier/Role/Token grammar change, no `registry.ts` change (§7.7 out of scope).

## The proof it's safe (review this first)
`tests/security/scopeBindingsEquivalence.test.ts` asserts the new resolver is **set-equal** to a *frozen verbatim copy of the deleted switch* across **8 personas × 24 models × 11 rows = 2112 combinations** (own / another's / in-program / out-of-program / active-visit / departed-visit / missing-key / non-object). This is the contract that the port changed nothing. If you trust one thing, trust this — but also eyeball `scopeBindings.ts` against `classifications.ts`.

## Files
| file | what |
|---|---|
| `scopes.ts` (new) | `Match` + `evalMatch` + `makeScopesHeld`; `validateBindings`/`validateRouteGrants` (authored, unwired) |
| `scopeBindings.ts` (new) | `SCOPE_BINDINGS` (19 models), `ROW_SCOPE_KEY`, `OPT_OUT_PENDING_ROUTE`, `scopesHeld` |
| `access-resolvers.ts` | switch + `ROW_SCOPE_KEY` deleted (−182), re-exports `scopesHeld`; `buildCallerContext`/`resolveAccess` untouched |
| `scopeBindingsEquivalence.test.ts` (new) | S1 equivalence proof + frozen oracle |
| `scopeValidators.test.ts` (new) | validators detect what they target |

## Reviewer must scrutinize
1. **Fail-closed preserved.** The §7.3 doc engine sketch *dropped* the `ROW_SCOPE_KEY` fail-closed preamble; I kept it (`makeScopesHeld(bindings, rowScopeKeys)`). Without it, a key-less `EmergencyContact` row would get `everyones` instead of `{}` → admin view leaks the row. Confirm `EmergencyContact` missing `householdId` → no scopes.
2. **`EmergencyContact` is BOUND, not pending.** Live switch already binds it; doc §7.5.1 stale. Removed from `OPT_OUT_PENDING_ROUTE`. Deviates from the doc's pending list **on purpose**.
3. **Fee/RSVP literal port.** Grouped switch case read `participantId` on `Fee` and `programId` on `RSVP` — columns those models lack. Ported **verbatim** (required for switch-equivalence on arbitrary rows); `validateBindings` flags both — intended typo-catcher behavior, surfaced not hidden.
4. **`number`-guards in `evalMatch`** encode the switch's `!== undefined` checks — an anon caller (`selfId === undefined`) must never match a row whose field is also absent.

## Validator state (unwired this PR)
`validateBindings` returns 8 errors against the real table — all expected and documented: 2 Fee/RSVP field-existence + 6 admin-only logs flagged by the doc's `id`-inclusive `SCOPABLE_FIELDS`. Path to a clean wire is in `scopes.ts` comments (Step 3).

## Related branch (no behavioral conflict)
`claude/silly-khayyam-ee39e3` adds a certifier admission gate + `RouteSpec.returns` — **orthogonal layer** (admission/role, no `Scope`/binding overlap). Single textual merge touchpoint in `access-resolvers.ts` (my deletion vs their `isCertifier` insertion). Their `returns` is the enabler my `validateRouteGrants` waits for. Land either order.

## Verification
- S1 equivalence (2112 combos) ✅ · validators (9) ✅
- stripper / toolstatus / member-tier / resolveAccess / outbound — 89 unit ✅
- `registryAuthz` + `policy.contract` integration — 48, real Postgres `--runInBand` ✅
- `tsc --noEmit` 0 errors · `eslint --max-warnings 0` clean

> Pre-commit `test:ci` finds 0 tests inside a git worktree (`testPathIgnorePatterns` ignores `/.claude/worktrees/`) — known false failure; full suite run manually. CI on a normal checkout runs it.

### DIFF
--- a/checkin-app/src/security/access-resolvers.ts
+++ b/checkin-app/src/security/access-resolvers.ts
@@ -15,7 +15,7 @@
  */
 import prisma from '@/lib/prisma';
 import type { AuthResult } from '@/types/auth';
-import type { Authorize, Role, Scope } from './core';
+import type { Authorize, Role } from './core';
 
 export interface CallerContext {
     selfId?: number;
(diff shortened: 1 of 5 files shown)


thpr:
FYI this is in response to a security review that found many of our routes are not using handler() to access data, and many of the POST type scenarios were not gated (e.g. using withAuth) ; allowing unauthenticated users (including those Denied Membership) access to PII or to write tremendous garbage into the system.  The security model in our app is great, but it's not universally enforceable (yet) and it wasn't widely enough used.  To ensure nothing escapes, this is part of a series of patches that are forcing everything into a handler (or equivalent standards for things like the kiosk), with the security eventually enforced.  check-route-coverage must not be advisory, but right now it won't catch enough.  Resolving this chain of patches should be considered gating for production launch @dkaygithub @jee7s   #568 #569 #570 #571 #572 are related work starting to patch some of the issues and get everything to a set of enforceable standards

jee7s (approved):
This is much easier to read, IMHO. I'm unsure why there's so much concern flagged by Claude here.

PR #622, opened 2026-07-01
fix(shop): require authorization to read another member's certifications (L1)

jee7s:
## What
`GET /api/shop/certifications` let any authenticated member read **another** member's tool-certification levels:
- `?participantId=<anyId>` had no ownership check.
- `?toolId=<id>` returned every certified member's name + level for a tool, gated only client-side (shop-ops React), so any logged-in member could enumerate tool ids and reconstruct the same per-person data.

Fix: reading anyone but yourself (`participantId` for others, **or** the `toolId` roster) now requires `isSysadmin || isBoardMember || certifier` (the same `MAY_CERTIFY_OTHERS` derivation as `callerHoldsRole('certifier', …)`). Self-lookups stay open. This matches the roster's only real consumer (`ToolManagementPanel`, already staff/certifier-only).

## Finding
**L1** from the review (IDOR). The `?toolId=` half was surfaced by the adversarial pass.

## Adversarial verdict — CLOSED (after the `?toolId=` gate)
Red-team confirmed the gate runs before the query on every path, params tricks (NaN/negative/float/dup) fail closed, and certifier status is DB-derived in the signed JWT (not client-spoofable).

## Tests
`shopAPI.integration.test.ts`: plain member → 403 on both `?participantId=other` and `?toolId`; certifier/admin/board → 200; self → 200. 16/16 green against a throwaway Postgres.

### DIFF
--- a/checkin-app/src/app/__tests__/shopAPI.integration.test.ts
+++ b/checkin-app/src/app/__tests__/shopAPI.integration.test.ts
@@ -254,6 +254,31 @@ describe('Shop API Integration Tests', () => {
              expect(data.length).toBeGreaterThanOrEqual(1);
         });
 
+        it('enforces object-level authorization on ?participantId (L1 IDOR fix)', async () => {
+            // Self-lookup — always allowed, no participantId gate applies.
+            (getServerSession as jest.Mock).mockResolvedValue({ user: { id: commonId } });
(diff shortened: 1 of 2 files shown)


thpr:
Disagree - this is public info

jee7s:
I disagree too. We should take this as an indication of how the adversarial part of the security review handled things. This is overstrict against our use cases, but probably in line with a generic RBAC scheme.

PR #624, opened 2026-07-01
fix(membership): validate Shopify order amount before activating (H2)

jee7s:
## What
The Shopify `orders/paid` webhook (correctly HMAC-authenticated) activated the membership named by the customer-controlled `Membership_Process_ID` cart attribute **without validating the paid amount** — so a member could pay a trivial amount (cheap variant / volunteer discount code) and get a full membership. Now `activate()` validates `order.total_price` (parsed to cents, fail-closed on missing/garbage) against the household's expected tier dues before activating; an underpaid order is recorded and the board is alerted, but **not** activated (and `paidAt` is not set, so a later bg-clearance can't activate on the shortfall).

Expected dues come from our own `membership.isVolunteer` flag, never the order's discount codes, so appending the public volunteer code simply fails the check.

## Finding
**H2 (HIGH)** from the review.

## Adversarial verdict — core CLOSED, one fail-open fixed
Red-team confirmed the core path is closed (parsing, volunteer self-selection, the opt-in seam — prod always supplies the amount, idempotency, `certifyPaymentPlan` exemption). It found one **OPEN** fail-open, fixed in the follow-up commit:
- `computeDuesCents()` returns `?? 0` when `BoardSettings` dues are unset (columns default 0), degrading the check to `paidCents < 0` (always false). Now activation **fails closed when `expectedDuesCents <= 0`** (treat unset dues as unpriced, not free).

## Residual follow-ups (documented, out of scope for this PR)
1. **BLOCKED branch:** stamps `paidAt` with no amount check, and `clearBackgroundCheck` treats `paidAt` as paid-in-full — an underpayment can survive a **board reconsideration** of a blocked app (board-gated, and the board is already alerted "refund likely needed").
2. No `X-Shopify-Topic` / `financial_status` check — if `orders/create` is also subscribed to this URL, an unpaid-but-created order carries a `total_price`. Recommend gating on `financial_status === 'paid'`.
3. **TODO #278** still open: `Membership_Process_ID` remains customer-controlled, so paying another household's *correct* dues activates *their* process (real money, no attacker benefit, fully audited) — needs a per-process checkout token.

## Tests
`payment.test.ts` 3/3 (underpaid rejected, meets-dues activates, unset-dues fails closed); the two existing Shopify webhook integration tests updated for the new amount arg.


thpr:
4th Residual, see #625

thpr:
This seems fragile.  See Residual #4 above, but also if they add something else to the cart, it breaks.  This does close a hole but it opens others, is there now going to be a flow that total != expected, so then in the next Shopify read from s-read-function we look at the detail?  What if someone went to purchase, changed what's in the cart to a bunch of workshops with the same total and checked out?  I think this is closing one loophole and opening others (while being more fragile).  This needs more discussion, IMHO

jee7s:
Good points — you're right that the order-total check was fragile, and I've revised the approach (`53d18a5`).

Instead of validating `order.total_price` against our `BoardSettings` dues, it now checks that the paid order actually **contains the membership product** — i.e. one of the configured Shopify membership variants (`membershipVariantId` / `shopifyNormalVariantId` / `shopifyVolunteerVariantId`) appears in the order's `line_items`. If none does, we don't activate; the order is recorded and flagged for board review instead.

That directly addresses both of your concerns:

- **"a bunch of workshops with the same total"** → workshops aren't a membership variant, so `hasMembershipItem` is false → no activation.
- **Settings drift** → we no longer compare against a copied price. The variant id is the *same* id the checkout link is built from (`/cart/<membershipVariantId>:1`), so if it were wrong the checkout link itself would be broken and we'd notice immediately — unlike a dues amount, which can silently diverge from Shopify's real price. It also fails closed: if the variant isn't configured, nothing auto-activates.

What this PR deliberately does **not** try to solve is whether the *right price* was paid (discount eligibility, volunteer rate, timing). I agree that's the real systemic gap and belongs in #625 (periodic Shopify↔settings price alignment) — the code comment now points there. This is just the narrow interim fix so a non-membership order can't activate a membership.

Happy to hold for more discussion if you'd still rather defer entirely to #625.

thpr (approved):

PR #740, opened 2026-07-03
feat(shopify): productionize the dev orders/paid mock + webhook-registration script

jee7s:
## What

Finishes the Shopify dev-store webhook work planned in #738, on top of the in-process `orders/paid` mock that already landed in #730.

- **Mock hardening** — export `DEV_MOCK_SHOPIFY_WEBHOOK_SECRET` (dedupes the literal the webhook test hardcoded); the dev route now `404`s a `processId` that doesn't exist or isn't `PENDING_PAYMENT` (fail-closed, matches what the dev UI already offers).
- **Tests** — new dedicated integration suite for `POST /api/dev/shopify/orders-paid` (404 mock-inactive, 401 no-session, 400 bad processId, 404 unknown/wrong-phase process, 409 no variant configured, and a genuinely end-to-end 200 happy path that hands the mock's signed payload to the real inbound `/api/webhooks/shopify` handler and asserts the `OrgMembershipProcess` actually advanced). Plus a Shopify fuse-matrix block in `config.test.ts` mirroring the Zoho one.
- **Admin-API registration script** — new `scripts/register-shopify-webhook.ts` (`npm run shopify:webhook -- --url <callback> [--commit]`): idempotent create/update/noop of the `orders/paid` subscription via the Admin API, reusing `getAccessToken`/`shopifyFetch` (now exported from `shopify.ts`). Dry-run by default; the decision is a pure, unit-tested function.
- **Docs** — brings in #738's `SHOPIFY_DEV_STORE_WEBHOOK.md` + `_PLAN.md` (resolves the mock's dangling doc citations).

## How the membership variant id works (addressing review feedback)

**The variant id is never hardcoded in our code** — it lives in `BoardSettings.orgMembershipVariantId` (admin-editable via Settings → Membership), set once per environment:

- **Local mock** (no Shopify creds): the value is just a **correlation token**. The mock reads it, echoes it in the synthesized `line_items`, and the inbound handler matches it against the *same* `BoardSettings` — a closed loop that never touches Shopify, so any string works. Set it once locally; the dev tool's empty-state points you there.
- **Real dev store** (cloud-dev, real creds): the value must be the **store's actual variant id**, set once by a human after creating the product. A real `orders/paid` carries that id and the inbound H2 check matches it.

Because both read the id from `BoardSettings` (not code), there is nothing to "align."

> **Correction since first push:** the initial version *seeded* a fake placeholder variant id so the local mock was clickable with zero setup. That was wrong for a design with a real-store path — `seedBaseline` is reachable on cloud-dev (via the dev-dashboard reset), where a fake id would silently fail the real store's H2 variant check. **Reverted** (`revert(shopify-dev-mock)`): we now follow #738's O4 — `BoardSettings` variant stays null until configured once via Settings → Membership, and the dev UI + design §6 explain the model explicitly.

## Safety

- **Prod path byte-for-byte unchanged** — `verifyShopifyHmac`, `withWebhook`, `activate()`, and the four `config` Shopify getters are untouched. Every dev surface stays gated on `shopifyMockActive()` (false in prod by construction).
- **No infra change** — the AWS Secrets Manager / ECS task-def wiring for dev + prod is already committed on the infra repo's `feat/enable-shopify-dev` branch; only the human `put-secret-value` of the dev values remains (out of scope here).

## Verification

Node 24, live Postgres: full `npm run test:coverage` green — **295 suites / 2276 tests**, coverage above the 85/75/75 floors; `tsc --noEmit`, `lint`, and `check-route-coverage` all clean.

### DIFF
--- a/checkin-app/docs/designs/SHOPIFY_DEV_STORE_WEBHOOK.md
+++ b/checkin-app/docs/designs/SHOPIFY_DEV_STORE_WEBHOOK.md
@@ -1,0 +1,207 @@
+# Shopify Development Store + Webhooks in Dev/Local
+
+**Status:** Proposal — for review. No code written.
+**Author:** design pass, 2026-07-02 (revised 2026-07-02 after grepping prod deploy + `src/`)
+**Related:** [`ZOHO_SIGN_DEV_MOCK.md`](./ZOHO_SIGN_DEV_MOCK.md) (#669, in-process sign mock), #665 (email dev inbox), #278 (checkout-token honor-system TODO), #624/#625 (order-amount / price-alignment safety), #683 (Shopify env routed through `config.ts`)
+**Implementation plan:** [`SHOPIFY_DEV_STORE_WEBHOOK_PLAN.md`](./SHOPIFY_DEV_STORE_WEBHOOK_PLAN.md) (concrete steps + two prod-grounded corrections folded back into §2/§4 below).
+
+## 0. Scope check — what already exists
+
+Confirmed by `git log` + grep before writing this: **no Shopify dev-store or webhook-mock work has landed or is in flight.** The only recent Shopify commits are #683 (route env through `config.ts`), #624 (validate order amount before activating), and #466 (fetch timeouts). The two sibling dev-integration features — Zoho Sign mock (#669) and email inbox (#665) — do **not** touch Shopify. So this is greenfield; nothing to redesign.
+
+Unlike Zoho (which got a fully in-process mock), the ask here is to connect to a **real Shopify development store** so genuine `orders/paid` webhooks fire against the dev instance and drive `PENDING_PAYMENT → ACTIVE` end-to-end. An in-process mock is also proposed as the zero-infra fallback (§6). **Prod behavior is unchanged** — every dev-only path is env-gated the same way Zoho's is.
+
+---
+
+## 1. Current flow (as wired today)
+
+Trace from checkout link to activation:
+
+1. **Build the checkout link** — [`payment.ts › ensurePaymentLink`](../../src/lib/membership/payment.ts) reads `BoardSettings.membershipVariantId` + `config.shopifyStoreDomain()` and calls `buildMembershipCheckoutUrl`, producing a Shopify **cart permalink**:
+   ```
+   https://{SHOPIFY_STORE_DOMAIN}/cart/{variantId}:1?discount={code}&attributes[Membership_Process_ID]={processId}
+   ```
+   Volunteer households get `discount=<BoardSettings.volunteerDiscountCode>` appended. The process id rides along as a **cart attribute** so the webhook can match the payment back. If domain or variant is unset, the link is `null`.
+
+2. **Applicant pays on Shopify** — hosted checkout on the store. Shopify maps cart-attribute `attributes[Membership_Process_ID]` onto the Order's `note_attributes`.
+
+3. **`orders/paid` webhook fires** → [`api/webhooks/shopify/route.ts`](../../src/app/api/webhooks/shopify/route.ts), wrapped by `withWebhook({ provider: "shopify", verify: verifyShopifyHmac })` ([`webhookAuth.ts`](../../src/lib/webhookAuth.ts)): per-IP rate-limit → HMAC verify → JSON parse → handler.
+
+4. **HMAC verify** — `verifyShopifyHmac` computes `HMAC-SHA256(rawBody, config.shopifyWebhookSecret())` in base64 and `timingSafeEqual`s it against `x-shopify-hmac-sha256`. Unset secret → **500** (config error); missing/wrong sig → **401**. Verified over the exact raw bytes before parse.
+
+5. **Match by cart attribute** — handler scans `note_attributes` for `Membership_Process_ID`, then checks the order's `line_items` actually contain a known membership variant (`membershipVariantId` / `shopifyNormalVariantId` / `shopifyVolunteerVariantId` from `BoardSettings`) — the #624/H2 guard against paying for an unrelated item that totals the same.
+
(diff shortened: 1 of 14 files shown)


thpr:
I won't claim the design was complete, but the description implies this is not what we need.  I'll try it later today, but I can't see how variant IDs hardcoded in our code can ever align with variant IDs in a real shopify dev store... there's a mental gap there I can't bridge

jee7s:
Good catch — you're right, and it exposed the one place we deviated from the #738 plan. There's no way to make hardcoded variant ids align with a real dev store, so I've **reverted the placeholder seed**.

The model, now stated explicitly in the dev-tool empty-state and design §6: the membership variant id is never in code — it lives in `BoardSettings.orgMembershipVariantId` and is set once per environment via Settings → Membership. Locally (mock) it's just a self-correlation token the mock echoes and the inbound handler matches against the same `BoardSettings` — a closed loop, so any string works. On cloud-dev (real store) you set it to the store's actual variant id. Nothing aligns because nothing is hardcoded.

The reason the fake seed was actively wrong (not just ugly): `seedBaseline` is reachable on cloud-dev via the dev-dashboard reset, where a placeholder id would silently fail the real store's H2 variant check and refuse activation. Reverted in `revert(shopify-dev-mock): don't seed placeholder membership variant ids`.

dkaygithub:
Why have a variable named dev_foobar Instead of having the dev secret manager set foobar?

jee7s:
Pushed aa92849 — a review pass over this PR against @thpr's feedback, the CUJs, and the #738 plan. What changed and why:

## Mergeability
The branch and `main` had both added `SHOPIFY_DEV_STORE_WEBHOOK.md` (add/add conflict — #738 merged it while this branch carried its own copy). Merged `origin/main` in and resolved with this branch's version (main's copy + the §6 variant-id section). **The PR is now conflict-free.**

## The one real trap: which secret signs a script-registered webhook
Shopify signs deliveries to an **Admin-API-created** subscription with the **app client secret**, while a store-admin-registered webhook uses the store's Settings → Notifications signing secret. Phase 2 of the plan tells the operator to use the store signing secret — so anyone re-pointing the webhook with `npm run shopify:webhook --commit` would silently get **every** real `orders/paid` 401'd (the "mismatched secret is the #1 cause of 401s" trap §2 itself warns about). The script now prints this caveat on every run, and §2/§3 + plan Phases 2–3 spell out the pairing rule.

## Registration script hardening
- Lists with `?topic=orders/paid&limit=250` — the unfiltered call is capped at 50 entries, so the target subscription could fall off page 1 and cause a duplicate create. (Note: the API only sees *app-created* subscriptions; a manually registered webhook is invisible to the script — now documented.)
- `decideWebhookAction` now prefers a subscription already at the desired address and reports any other `orders/paid` subscriptions as `staleDuplicates` (warned in output) instead of silently reconciling only the first match. New unit test covers it.
- Guarded the 2xx response parse (an odd success body no longer crashes the CLI *after* the mutation applied); `SHOPIFY_API_VERSION` is now exported from `shopify.ts` and shared (the literal existed in 5 places across 2 files); minor dedup of the mutate call.
- Both docs said `npm run shopify:webhook -- <url>` (positional) but the script requires `--url` — the documented command always failed. Docs fixed.

## Dev mock route + UI
- A JSON body of literal `null` passed `req.json()` and blew up on destructuring → 500; now it 400s like every other malformed body.
- The new PENDING_PAYMENT gate makes re-fires impossible, but the payload comment still advertised "re-fire is an idempotent webhook retry", and the dev page never refreshed its list after success — so re-clicking a just-paid row painted a red 404. The client now `router.refresh()`es on success (row disappears), and the comment describes what the stable order id actually still covers (the concurrent double-click window).

## Test isolation
- The integration suite passed only because its creds-wiping test happened to run **first** — and its `delete process.env.SHOPIFY_*` would permanently erase real `.env` creds for later suites in the worker. It now snapshots + clears in `beforeAll` and restores in `afterAll` (order-independent, non-destructive).
- `afterAll` restored BoardSettings only `if (prevSettings)`; on a fresh per-worker DB (no row 1) it leaked the test row. Now deletes the row it created (AGENTS.md self-clean).
- The `global.fetch` stand-in now rejects non-webhook URLs loudly instead of feeding the webhook handler's response to whatever fetched.

## Docs reconciled with reality (#738 goals)
- Statuses no longer claim "no code written" / "~0 lines"; Phase 0 marked done via #730 (the `455362c2` merge instruction pointed at a superseded sha two fixes behind what landed).
- §1/§2 referenced a nonexistent `BoardSettings.membershipVariantId` column in four places (schema: `orgMembershipVariantId`) — fixed, so the doc now consistently tells the variant-id story from the review thread.
- The §2 table row still recommended *seeding* dev-default variant ids — the exact idea the revert removed; it now records the O4 decision (deliberately not seeded, set once via Settings → Membership) and notes `shopifyNormalVariantId`/`shopifyVolunteerVariantId` have **no UI writer** (only `orgMembershipVariantId` is admin-editable).
- §6 described a `SHOPIFY_MOCK_WEBHOOK_SECRET` env var that never existed — it's the fixed `DEV_MOCK_SHOPIFY_WEBHOOK_SECRET` constant.
- Bonus symmetry: the Zoho mock secret is now exported and imported by its tests too, same dedup this PR did for the Shopify twin.

## Deliberately not changed
- The PENDING_PAYMENT fail-closed gate stays as designed.
- The membership-variant set is spelled in three places (dev page, dev route, inbound handler's H2 matcher). Extracting one helper means touching the prod handler, which this PR promises to leave byte-for-byte unchanged — suggest a small follow-up PR.
- `docs/designs/CUJS.md` still lists `/admin/*` membership routes that no longer exist (`src/app` has no `admin/` at all) — pre-existing staleness, separate docs PR.

**Verified:** `tsc --noEmit`, eslint on changed files, 44 unit tests + 20 integration tests (orders-paid, webhookShopify) green against live Postgres.

jee7s:
> Why have a variable named dev_foobar Instead of having the dev secret manager set foobar?

TL;DR: Sonnet did the first pass, Fable did the most recent update

thpr:
> > Why have a variable named dev_foobar Instead of having the dev secret manager set foobar?
> 
> TL;DR: Sonnet did the first pass, Fable did the most recent update

...and I'd add if it's a variant ID and not a secret, don't use a secret, secrets cost money, vars in settings screens don't

thpr (approved):

PR #792, opened 2026-07-03
chore(prisma): squash 49-migration chain into single baseline migration

thpr:
## What

Replaces the 49-migration chain (`20260223050920_init` .. `20260703120000_trusted_adult_hidden_at`) with a single baseline migration, `20260703130000_squashed_init`.

## ⚠️ Required deploy step — reconcile prod migration history

The live DB has the old 49 entries in `_prisma_migrations`. Before the **first `migrate deploy` after this merges**, reconcile history or deploy fails immediately ("type already exists" — loud failure, no data risk):

```sql
DELETE FROM "_prisma_migrations";
```

```
npx prisma migrate resolve --applied 20260703130000_squashed_init
```

Same instructions are in the migration file header. Any other already-migrated DB (long-lived dev DBs) needs the same treatment; fresh DBs just `migrate deploy` as usual.

## How it was generated and verified

- Replayed the full 49-chain into a throwaway Postgres (ground truth), generated the baseline with `prisma migrate diff --from-empty --to-config-datasource --script`, applied it to a second DB, and diffed `pg_dump --schema-only` of both. Identical except stale sequence names the old table renames left behind (cosmetic; nothing references sequence names) and dump ordering.
- `prisma migrate diff` against `schema.prisma`: no drift.
- Fresh DB + `migrate deploy` + full seed: pass.
- Unit suite: 1436 pass, 1 pre-existing failure (`safety/trusted-adults` page test times out on clean main HEAD too — unrelated).

## Hand-fixes over the raw diff output

- Prisma 7's diff engine **silently drops the WHERE clause** of partial unique indexes — it emitted `Visit_one_open_per_participant`, `membership_one_inflight_initial`, and `membership_one_inflight_renewal` as *full* unique indexes (app-breaking: one visit per person, ever). Restored the real definitions from the truth DB's `pg_indexes`.
- Old-chain data backfills omitted deliberately: they only migrated pre-existing rows and are no-ops on a fresh DB.


thpr:
Rebased onto main (merge, kept squash). #791 rewrote three Jul-2 rename migrations to be data-preserving — replayed the updated 49-chain and diffed normalized catalogs (columns/types/defaults, indexes, constraints, enums) against the squashed baseline: identical except cosmetic sequence names, so the squash stands unchanged.

Note on merge timing: the dev deploy workflow auto-runs `migrate deploy` in ECS, so the first dev deploy after this merges will fail (P3009-style) until someone runs the reconcile step from the PR description against the dev DB. Plan to reconcile immediately after merge, before the next push to main.

jee7s (approved):

PR #868, opened 2026-07-05
fix(auth): repair Google sign-in — PrismaAdapter shim missed the Person rename

dkaygithub:
## What broke

Every **Google sign-in on ops-dev fails**: the OAuth callback 500s and the user is bounced back to `/signin` unauthenticated. CloudWatch (`/ecs/checkin-dev`) shows:

```
[next-auth][error][adapter_error_getUserByEmail]
Cannot read properties of undefined (reading 'findUnique')
→ OAUTH_CALLBACK_HANDLER_ERROR
```

## Root cause

The Participant→Person rename (#708) swept every `prisma.participant.<method>()` call site, but missed the NextAuth PrismaAdapter shim in `auth-options.ts`:

```ts
const prismaAdapterCore = prisma as unknown as Record<string, unknown> & { participant: unknown };
user: prismaAdapterCore.participant,   // ← no `prisma.` prefix, no `.method()` suffix — invisible to the sweep
```

The `as … & { participant: unknown }` cast **manufactured the stale property, so `tsc --noEmit` stayed green** (it's even in #708's commit message). At runtime `user` is `undefined`, and the adapter's `getUserByEmail` crashes.

**Why no test tier caught it:** the adapter's user methods run *only* inside the real Google OAuth callback. Persona-mint (what flow tests use via `loginAs`) is a credentials sign-in that never touches the adapter under JWT sessions; the global jest prisma mock is a permissive proxy that resolves *any* model name; integration tests never drive the NextAuth handshake. The one untestable sign-in path was exactly the one that broke.

## Fix + three backslide guards

1. **`auth-options.ts`** — map `user:` to a plain, **compiler-checked** `prisma.person` reference (cast removed from the model name). The next rename that misses this line fails the build instead of breaking sign-in at runtime. Comment documents the incident.
2. **New `auth-options-adapter.test.ts`** — drives `getUserByEmail` / `getUser` / `createUser` against a prisma mock defining **only** real model names (the adapter closures are lazy — invocation is what makes a stale mapping bite). **Verified red against the pre-fix shim** with the exact production error, green after.
3. **`AGENTS.md`** — documents the Google-OAuth coverage hole under the flow-test gotchas: no tier exercises real Google OAuth; tsc + the adapter test are the guards; keep the mapping cast-free.

## Verification

- New test: red on broken shim (`TypeError: Cannot read properties of undefined (reading 'findUnique')`), green on fix
- `tsc --noEmit`: clean
- Full unit suite (pre-commit hook): 198 suites / 1516 tests green
- After merge, `deploy-dev.yml` rolls this to ops-dev — will confirm live Google sign-in there

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -57,6 +57,16 @@ routes, auth, and DB together.
   reseeds every run). Don't rely on re-running against the same DB locally.
 - `*.flow.test.ts` is excluded from the unit/integration runs (it needs a live
   server) — keep it that way; run it only via `npm run test:flow`.
+- **No test tier exercises real Google OAuth.** Flow tests authenticate via
+  persona-mint (a credentials sign-in), which — with JWT sessions — never
+  touches the NextAuth **PrismaAdapter**; the adapter's user methods run only
+  inside the live Google OAuth callback. The adapter↔model mapping in
+  `checkin-app/src/lib/auth-options.ts` is guarded by tsc (keep the
+  `user: prisma.person` reference cast-free) and by
+  `src/lib/__tests__/auth-options-adapter.test.ts`. Regression to learn from:
+  the Participant→Person rename (#708) left `.participant` in that mapping
+  behind an `as` cast — CI stayed fully green while every Google sign-in on
+  the dev instance 500'd in the OAuth callback.
 
 To add a journey: drop a `flow-tests/<name>.flow.test.ts` using `loginAs`/`api`,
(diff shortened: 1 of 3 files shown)


thpr (approved):

dkaygithub:
Superseded by the same-repo PR (fork PRs can't mint the OIDC token the flow-tests job needs — failed pre-test at credential setup). Unit tests and deploy build were green here.

Issue #874, opened 2026-07-05
Volunteer allowlist is applied only at BG clearance — pay-first volunteer families are charged full dues


dkaygithub:
## Problem

`applyVolunteerStatus` — the only place the `VolunteerDesignation` email allowlist is matched — is called solely from `clearBackgroundCheck` (`src/lib/membership/review.ts:247`). Since PR #428 made the background check non-blocking, payment normally happens **before** clearance, and `ensurePaymentLink` reads `membership.isVolunteer` at `PENDING_PAYMENT` time (`src/lib/membership/payment.ts:60-76`).

Net effect: a brand-new, pre-allowlisted volunteer family that pays in one sitting sees **full dues and no discount code** — allowlisting does not help them at all. Effectively ~100% of new volunteer families paying before two reviewers approve will overpay and need a manual refund.

Related gap: the `submitIntake` fresh-check shortcut sets `bgClearedAt` directly (`src/lib/membership/intake.ts:379-386`) without ever running the allowlist match, so `clearBackgroundCheck` never runs that cycle and volunteer status is never applied at all.

## Suggested fix

Run the allowlist match when the process advances to `PENDING_PAYMENT`:
- in `advanceExternalIfComplete` (`src/lib/membership/external.ts`), and
- in the `submitIntake` fresh-check shortcut,

keeping the reviewer-marked volunteer flag at clearance as the sticky supplement it is today.

Related (but distinct): #278 tracks gating the discount code itself via a Shopify customer segment; #625 tracks price/settings alignment. This issue is about *when our own `isVolunteer` flag is computed*, which those don't cover.

PR #886, opened 2026-07-05
feat(membership): applicant self-attests background-check consent

### LINKED ISSUE #875: Applicant copy overpromises: BG consent is board-gated, so 'pay right away' stalls on a human
## Problem

The external step tells the applicant "Once your agreement is signed and your background check is underway, we'll move you straight to payment" and the intake submit flash says "then you can pay right away" (`src/app/membership/page.tsx:378`, `:631`).

But after clicking "Consent on Averity →", nothing happens in-app until a **board member** notices Averity's email and marks consent (`markBgConsent`, `src/lib/membership/external.ts:139`). The family sits at a task card that never flips, with a "Refresh status" button that won't help. The one-sitting promise stalls on a human in the loop.

## Suggested fix

An applicant self-attest checkbox ("I submitted my consent on Averity") that records consent via the same `markBgConsent` path — the dev mock (`/api/dev/bg-consent/complete`) already does exactly this shape, applicant-as-actor, fuse-gated to non-prod. Until that ships, soften the copy so it doesn't promise same-sitting payment.

dkaygithub:
Closes #875.

## Problem

The consent step was board-gated: after clicking through to Averity, the applicant waited for a board member to notice Averity's email and human-mark `bgConsentAt` before the application could advance to payment. That breaks the one-sitting signup that the non-blocking background check (PR #428) otherwise enables, and the applicant-facing copy ("we'll move you straight to payment") overpromised.

## Change

The applicant now self-attests in-app after submitting their consent on Averity:

- **`selfAttestBgConsent` (`lib/membership/external.ts`)** — household-lead-only; resolves the caller's own process awaiting external action and records consent through the existing `markBgConsent`, with the **applicant as the audit actor** so a self-attested consent stays distinguishable from a board-confirmed one. Idempotent (double-click safe); the board's `mark-bg-consent` action is unchanged and remains the backstop.
- **`POST /api/membership/bg-consent`** — applicant-facing route; error→HTTP mapping mirrors `contract/sign` (403 `not_lead`, 409 `wrong_phase`).
- **`/membership` UI** — an "I submitted my consent on Averity" checkbox under the Averity button, locked until the applicant has actually opened the link this visit. Checking it records consent and refreshes — advancing straight to `PENDING_PAYMENT` when the contract is already signed.
- Fixed two stale comments (`external.ts` header, `manual-adapter.ts`) that said consent is always board-marked.

Honor-system by design: Averity exposes no API, so the applicant's claim is the best intent signal we can capture. Activation is still gated by the parallel two-reviewer attestation exactly as before — a false attestation just parks the application at `PENDING_BG_CLEARANCE`, where the existing REJECT → `BLOCKED` → refund-alert path covers it.

## Testing

- **Unit**: guard chain (`not_found`/`no_household`/`not_lead`/`wrong_phase`) + audit-actor assertion; page test for the click-gate → POST behavior. Full `test:ci` green (199 suites / 1539 tests).
- **Flow** (`membership-self-attest.flow.test.ts`, run via `test:flow:standalone`): intake → self-attest (no advance while contract unsigned) → board contract-mark → `PENDING_PAYMENT` with **no board consent action**; plus a 409 for a caller with no in-flight application. Existing bg-nonblocking flow test still green.

## Notes for review

- The self-attest uses `parent.family@example.com` in the flow test — the other seeded non-member household — so it doesn't collide with `membership-bg-nonblocking.flow.test.ts` (which mutates `parent.family2`).
- With one-sitting signup now real, the volunteer-discount timing bug (#874) bites harder — new allowlisted volunteer families will pay full dues before clearance. That's deliberately left to its own PR.


thpr:
<nevermind - misread the PR>

thpr (approved):

jee7s (approved):

Issue #625, opened 2026-07-01
Auto-check / align Shopify Pricing & Membership Pricing


thpr:
PR #624 Validates the pricing matches expectations, but does not account for error on the part of our settings.  If settings != Shopify actual cost, everything breaks, so #624 created fragility (while closing a loophole).  

To get rid of the fragility, we need a periodic check to ensure our settings are aligned to Shopify actual cost, accounting for all possible discounts, timing in year etc.  Right now that is not fully captured in settings and no cross-tool sync is done.  This is the issue to add those checks

jee7s:
**HOLD** — the problem is real and still open, but "add those checks" leaves three design decisions unmade.

1. Periodic job or on-demand check — and if periodic, at what cadence and on which surface? (`checkin-app/src/app/api/cron/` already hosts eight jobs, so there is an obvious home.)
2. What counts as "aligned" given discounts and mid-year price changes — exact match against the configured variant's Shopify price, or a tolerance band with named exceptions?
3. Who is alerted on drift, and does detected drift block activation or only warn?

**Disambiguation worth recording:** PR #1189 is titled "reconcile Shopify inventory on manual comp-add (#625)" and references this issue, but it closed the *capacity* reconciliation gap — `docs/backlog/INDEX.md` P7, now marked FIXED. The *pricing* drift check is a separate row, CO6, still NEEDS-DESIGN. This issue is CO6 and is untouched.

Confirmed on `origin/main`: `checkin-app/src/lib/finance/reconcile.ts` validates each individual order's amount against `BoardSettings` at reconcile time — reactive and per-transaction. There is no job comparing `BoardSettings` pricing against Shopify's actual catalog price, so the settings side of the comparison is never independently checked.

---
_Posted by Claude._

PR #907, opened 2026-07-05
fix(membership): match the volunteer allowlist at every PENDING_PAYMENT transition

### LINKED ISSUE #874: Volunteer allowlist is applied only at BG clearance — pay-first volunteer families are charged full dues

dkaygithub:
Closes #874.

## Problem

`applyVolunteerStatus` — the only place the `VolunteerDesignation` email allowlist is matched — ran solely from `clearBackgroundCheck`. Since the background check became a non-blocking parallel track (#428), payment normally happens **before** clearance, and `ensurePaymentLink` reads `membership.isVolunteer` at `PENDING_PAYMENT` time. A brand-new, pre-allowlisted volunteer family paying in one sitting saw full dues and no discount code.

## Fix

Match the allowlist at each transition **into** `PENDING_PAYMENT`, exactly as the issue suggests:

- **`advanceExternalIfComplete`** (`external.ts`) — inside the race winner's transaction, so it runs exactly once per advance.
- **`submitIntake` fresh-check shortcut** (`intake.ts`) — this path sets `bgClearedAt` directly, so `clearBackgroundCheck` never runs that cycle and the allowlist was otherwise never applied at all.
- **`beginRenewal` fresh-check path** (`renewal.ts`) — one more instance of the same gap, beyond the issue's list: a fresh-check renewal goes straight to `PENDING_PAYMENT`, so a household designated volunteer *since the previous cycle* would pay full dues.

The reviewer-marked volunteer pass at clearance is unchanged and remains the sticky supplement. `applyVolunteerStatus` is additive and idempotent, so the repeat run at clearance is harmless.

Not touched here (per the issue): #278 (gating the discount code via a Shopify segment) and #625 (price/settings alignment).

## Tests

- Unit: `advanceExternalIfComplete` (winner applies allowlist / loser and ineligible don't), `submitIntake` fresh vs. non-fresh, `beginRenewal` fresh / stale / double-submit-loser.
- Integration (`membershipBgNonBlocking`): pre-designated family is `isVolunteer` at `PENDING_PAYMENT` before any attestation; fresh-check intake shortcut applies the designation; fresh-check renewal picks up a designation added since last cycle.
- Full runs green: 1557 unit tests (198 suites), 898 integration tests (108 suites), `tsc --noEmit`, `eslint`.

## Update: intake notes now block payment (per review)

As agreed in the comments (and now that #948 sets the expectation on the form), **any text in the "anything else we should know?" field (#900) holds payment until a reviewer has seen it** — the note may change the review's outcome (e.g. "treat us as a volunteer household" from a family not on the allowlist), so dues can't be charged in parallel with it:

- **`advanceExternalIfComplete`** — with a note, contract + consent advance to `PENDING_BG_REVIEW` instead of `PENDING_PAYMENT`; `clearBackgroundCheck` opens payment after two approvals. Rows whose check already auto-cleared keep the direct path (once `bgClearedAt` is set the review queue can't see them, so holding would strand them).
- **`submitIntake` / `beginRenewal`** — a note disqualifies the fresh-check shortcut; that cycle goes through review (initial: consent → `PENDING_BG_REVIEW`; renewal: `RENEWAL_PENDING_BG`) before payment opens.
- **`overrideBlocked` reset** — an unpaid INITIAL with a note re-holds at `PENDING_BG_REVIEW`, so a board reset can't reopen payment ahead of the re-review.
- **Payment-open email** — when the check clears into `PENDING_PAYMENT`, the household leads are emailed that they can pay. For a held application that's the moment payment first opens (the status cards/banner already promised an email at clearance; previously only activation emailed).
- **Applicant UI** — `PENDING_BG_REVIEW` gets a stepper position and a dedicated "your application is being reviewed" card.

**Known wrinkle:** the note lives on the household and isn't cleared per cycle, so a household that leaves it in place will re-review at **every renewal** until they clear the field in my-household. That matches the agreed "any text blocks" rule and keeps the mechanism simple; happy to add per-cycle acknowledgement in a follow-up if it gets noisy.

Full runs green after the change: 1562 unit / 901 integration / `tsc` / `eslint`.


thpr:
Leaves the #900 note case open: a family types "treat us as a volunteer" but isn't on the allowlist → their isVolunteer (and dues) only settles when the BG reviewer reads the note. Payment runs parallel to review, so they'd pay full dues before the reviewer acts.  I thought we were holding up if there's a custom note?

jee7s (approved):

dkaygithub:
> Leaves the #900 note case open: a family types "treat us as a volunteer" but isn't on the allowlist → their isVolunteer (and dues) only settles when the BG reviewer reads the note. Payment runs parallel to review, so they'd pay full dues before the reviewer acts. I thought we were holding up if there's a custom note?

I'm fine with whatever, but i think a note of "i am interested in recoil" is much more likely and will hold up the payment

jee7s:
@dkaygithub PR #948 changes the text the user sees and clarifies that the application is held up with any comment from the field in #900. We can have this block now.

dkaygithub:
Done — any text in the "anything else we should know?" field now blocks payment until the review clears (@thpr's scenario): with a note, the application holds at `PENDING_BG_REVIEW` instead of advancing to `PENDING_PAYMENT`, the fresh-check shortcuts (intake + renewal) are disqualified for that cycle, and the household gets an email the moment the review completes and payment opens. Details in the updated PR description. One trade-off to be aware of: since the note is a household field, a note left in place re-holds every later renewal until the family clears it in my-household — the "i am interested in recoil" case @dkaygithub raised will hold payment for exactly one review cycle, then flow normally once cleared.

PR #917, opened 2026-07-06
refactor(household-lead): collapse HouseholdLead join into Person.isHouseholdLead

thpr:
## Problem

A person's household lived in **two** places — `Person.householdId` and every `HouseholdLead.householdId` — and leadership was decided by requiring them to **agree** ([`leads.ts`](checkin-app/src/lib/household/leads.ts): `lead.householdId === person.householdId`). Nothing (no FK, no DB constraint) kept them in sync, so a divergence silently **under-granted** a real lead — or, depending on the call site, risked an over-grant. The membership audit's Q18 sweep could only *detect* this after the fact, never prevent it.

## Approach

Design investigation ([`docs/designs/HOUSEHOLD_LEAD_MODEL.md`](checkin-app/docs/designs/HOUSEHOLD_LEAD_MODEL.md), included here) resolved the crux: **external leads** (a lead who isn't a member of the household they lead) are an *accident, not a requirement* — every read site demands the equality, so an external lead is invisible everywhere. The honest fix is to stop storing the fact twice.

**Option (a1):** a single boolean `Person.isHouseholdLead`. A person leads their own household (`Person.householdId`), so a mismatch becomes **impossible by construction** rather than detected-after-the-fact.

## Changes

- **Schema:** drop the `HouseholdLead` model + relations; add `Person.isHouseholdLead` (`@sensitivity:public`, matching the join's tier).
- **Migrations (LIVE-DB safe, two-step):**
  - `20260705140000_person_is_household_lead` — additive `NOT NULL DEFAULT false` + backfill that flags **only** leads of their own household (ignores any pre-existing divergent row).
  - `20260705150000_drop_household_lead` — `DROP TABLE`, a separate later migration that runs *after* readers cut over. No leadership data lost.
- **Writers** (promote / demote / move / merge / import / zoho) set or clear the flag; the per-household cap (#269) now counts flagged persons under the same `Household` `FOR UPDATE` lock.
- **Readers:** JS gates read `user.isHouseholdLead`; Prisma `leads` / `householdLeads` relation filters become `householdMembers` filtered on the flag.
- **Claim:** the existing `session.user.householdLead` boolean is now sourced from the column — claim name unchanged, so the session/type surface is untouched.
- **Security (⚠️ CODEOWNERS):** removed the `HouseholdLead` scope binding in [`scopeBindings.ts`](checkin-app/src/security/scopeBindings.ts), its 3 view return-bag entries in [`registry.ts`](checkin-app/src/security/registry.ts), and the equivalence-oracle case; `classifications.ts` regenerates from the schema.

## Behavior note

Moving a person to another household **clears** their lead flag (preserves prior behavior — the old code deleted the stale lead row and added none); `createNew` makes them the sole lead of the new household.

## Compatibility

The `householdLeads` **response-envelope key** (households route → `AdminEditHouseholdModal`) is deliberately preserved as a derived `[{ personId }]` shape — only the Prisma *model* is gone. A future grep for `HouseholdLead` will find only this wire key, not a table.

## Verification

- `tsc --noEmit` clean
- Unit: **199 suites / 1560 tests** pass
- Integration: **109 suites / 896 tests** pass against a throwaway seeded Postgres (jest global setup ran `prisma migrate deploy`, exercising both migrations end to end)

## Reviewer notes

- Touches the auth gate and CODEOWNERS-gated security scope config — expect a security review.
- The dominant pre-fix failure mode was a silent **under-grant** (lockout), not an over-grant; option (a1) removes the divergence class entirely.

### DIFF
--- a/checkin-app/__tests__/programSignupIntegration.test.ts
+++ b/checkin-app/__tests__/programSignupIntegration.test.ts
@@ -30,6 +30,7 @@ jest.mock('@/lib/prisma', () => {
     person: {
       create: jest.fn(),
       findUnique: jest.fn(),
+      findFirst: jest.fn(),
       findMany: jest.fn(),
       update: jest.fn(),
       deleteMany: jest.fn(),
@@ -50,11 +51,6 @@ jest.mock('@/lib/prisma', () => {
     household: {
       create: jest.fn(),
       delete: jest.fn(),
-    },
-    householdLead: {
-      create: jest.fn(),
-      findUnique: jest.fn(),
-      deleteMany: jest.fn(),
     },
     membership: {
       create: jest.fn(),
@@ -153,7 +149,7 @@ describe('Full Program Signup Integration Flow', () => {
 
         // 4. Lead user (who already has a household from signup) adds a child member
         mockGetSession.mockResolvedValue({ user: { id: leadUserId } });
-        (prisma.person.findUnique as jest.Mock).mockResolvedValueOnce({ id: leadUserId, householdId, householdLeads: [{ householdId, participantId: leadUserId }] });
+        (prisma.person.findUnique as jest.Mock).mockResolvedValueOnce({ id: leadUserId, householdId, isHouseholdLead: true });
         (prisma.person.create as jest.Mock).mockResolvedValue({ id: childParticipantId, householdId });
         // Adding a member reconciles emergency contacts (direction B); no contacts/members to flag here.
         (prisma.person.findMany as jest.Mock).mockResolvedValue([]);
@@ -181,7 +177,7 @@ describe('Full Program Signup Integration Flow', () => {
             _count: { participants: 0 }
         });
         (prisma.person.findUnique as jest.Mock).mockResolvedValue({ id: childParticipantId, householdId, dateOfBirth: new Date('2015-01-01') });
-        (prisma.householdLead.findUnique as jest.Mock).mockResolvedValue({ householdId, participantId: leadUserId });
+        (prisma.person.findFirst as jest.Mock).mockResolvedValue({ id: leadUserId });
         (prisma.programParticipant.create as jest.Mock).mockResolvedValue({ programId, personId: childParticipantId, status: 'PENDING' });
 
         const enrollReq = new Request(`http://localhost/api/programs/${programId}/participants`, {
@@ -238,7 +234,7 @@ describe('Full Program Signup Integration Flow', () => {
             _count: { participants: 0 }
         });
         (prisma.person.findUnique as jest.Mock).mockResolvedValue({ id: childParticipantId, householdId, dateOfBirth: new Date('2015-01-01') });
-        (prisma.householdLead.findUnique as jest.Mock).mockResolvedValue(null); // Not a lead
+        (prisma.person.findFirst as jest.Mock).mockResolvedValue(null); // Not a lead
 
         const enrollReq = new Request(`http://localhost/api/programs/${programId}/participants`, {
             method: 'POST',
--- a/checkin-app/docs/VOCABULARY.md
+++ b/checkin-app/docs/VOCABULARY.md
@@ -24,7 +24,7 @@ identifier that just says "member" is a bug against this dictionary.
 | **Person** | the human (any human: volunteer, youth, lead, enrollee) | name / "person" | `person` | **`Person`** (the umbrella person model) | — |
 | **Org Membership (A)** | Person/household ↔ **Organization** | **"Treehouse Member"** | `isActiveOrgMember`, `orgMember…` | `OrgMembership` | `/api/shop/org-members` |
 | **Household** | grouping of people | "household" (warm: "family") | `household` | `Household` | `/api/household` |
-| **Household Membership (B)** | Person ↔ **Household** | "household member" | `householdMember` | `householdId` FK + `HouseholdLead` (lead variant) | `/api/household/member` |
+| **Household Membership (B)** | Person ↔ **Household** | "household member" | `householdMember` | `householdId` FK + `Person.isHouseholdLead` (lead variant) | `/api/household/member` |
 | **Program relationship** | Person ↔ **Program** | **"participant"** | `programParticipant` | `ProgramParticipant` | — |
 
 Rules:
--- a/checkin-app/docs/designs/HOUSEHOLD_LEAD_MODEL.md
+++ b/checkin-app/docs/designs/HOUSEHOLD_LEAD_MODEL.md
@@ -1,0 +1,365 @@
+# Household-Lead model: dual source of truth for "which household is this person a lead of?"
+
+**Status:** design investigation → **option (a1) BUILT** (2026-07-05), shipped as a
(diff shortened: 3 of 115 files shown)


jee7s (commented):
I think this is safe/correct. @dkaygithub PTAL

jee7s:
Deep review (10 finder angles + independent verification of each candidate). CI is green including the populated-DB migration check — but that check can't see deploy *sequencing*, which is where the critical finding lives.

## Critical

**1. The `DROP TABLE` must not ship in the same release as the backfill.** Migrations run to completion *before* the ECS rolling update starts, and `origin/main`'s JWT refresh branch (`auth-options.ts` `else if (token.id)`) includes `householdLeads` on **every session-authenticated request** (~80 routes via `authenticateRequest`). So for the entire drain window, every request an old task serves throws P2021 (`auroraResumeRetry` rethrows non-P1001) → site-wide 500s; and any lead change old code writes between the backfill UPDATE and `DROP TABLE` is silently destroyed. The design doc's staged framing is right — ship `20260705170000_drop_household_lead` in a **follow-up PR** after this one is fully rolled out.

## Concurrency regressions (all verified against the code)

**2. Both demote paths can produce a zero-lead household** — the exact "broken household" state the admin tooling exists to repair. `lead/route.ts:112` (DELETE) is count-then-clear with *no transaction and no lock*; `member/route.ts:96` (PATCH) is in a tx but never takes the `SELECT … FOR UPDATE` lock that `addHouseholdLeadTx`'s own comment explains is required under Read Committed. Two concurrent demotes of a 2-lead household both see `leadCount=2`, both pass the last-lead guard, both commit. Fix at the right altitude: a shared `removeHouseholdLeadTx` in `leads.ts` with the same lock, used by both routes. Note the new concurrency suite covers **only the add path** (its own docstring says so) — extend it to demotes.

**3. The CSV household-merge lost its in-tx cap enforcement** (`import/route.ts:364`). The `projectedLeadIds` pre-check runs *outside* the transaction with no lock, and the pre-917 protection was removed: the old per-source-lead `addHouseholdLead` loop re-locked and re-counted inside the tx, while the new bulk `updateMany` carries flags with zero recheck (the lone tail `addHouseholdLead` early-returns without counting when the initiator is already a lead, and is skipped entirely for email-less rows). Concurrent merges/promotes can exceed `MAX_HOUSEHOLD_LEADS` silently. Fix: re-count under the household `FOR UPDATE` lock inside the tx after the `updateMany`.

**4. Cap enforcement is bypassable outside the choke point:** `zoho-import.ts:538` sets the flag via raw `person.update` (re-runs can push a household past cap — CLI-only, so lower severity), and `addHouseholdLeadTx` never asserts the supplied `householdId` matches the person's actual household — it locks/counts household A but flags the person in household B (latent today; one `assert` closes it).

## Migration semantics

**5. The backfill comment overclaims.** It discards divergent rows (`hl.householdId != p.householdId`) saying they "were never honored by the isLead check" — true for `leads.ts:19`, **false** for `merge/route.ts:52` on main, which honored *any* lead row as a don't-merge-away-a-lead guard. And `zoho-import` genuinely can create divergent rows (reassigns `householdId` without lead cleanup). Either backfill divergent rows too, or document the deliberate guard change and confirm the live DB has none before migrating.

**6. Missed cutover: `scripts/full_reset_and_dev_init.sh:76`** still raw-INSERTs into `"HouseholdLead"` — dev reset aborts mid-run post-drop. (The sweep caught non-TS files; also `docs/VOCABULARY.md` still lists HouseholdLead as a live model.)

## Smaller but worth fixing here

- **No index on `Person.householdId`** (Postgres doesn't auto-index FKs; none exists in any migration). Lead lookups regress from the old composite-PK index to a full Person seq-scan on hot paths (every check-in notification fan-out, nav badges). One line: `@@index([householdId, isHouseholdLead])`.
- **Audit rows point at a dropped table**: `tableName: "HouseholdLead"` at `lead/route.ts:53,133`, `member/route.ts:83,107` — the design doc itself lists migrating these as required a1 work. Use `"Person"`.
- **De-lead-on-transfer is untested** (`participants/[id]/household/route.ts:65`) — the doc's §7 Q4 flags exactly this; the only move test uses a never-lead subject. One case: promote → move → assert flag cleared.
- **Stale migration names**: the design doc header and the drop migration's own comment cite `20260705140000/150000`; the shipped folders are `160000/170000`.
- **Consolidation opportunity taken nowhere**: no `getHouseholdLeads()`/`removeHouseholdLead()`; the `{ householdId, isHouseholdLead: true }` predicate is hand-spelled at ~8 sites and the broken/unclaimed-household filter is copy-pasted across 3 files with a comment begging them to stay in sync.

## Verified clean (checked, not assumed)

Cross-PR interaction with #918: `git merge-tree` of the two heads auto-merges the entire repo cleanly and **correctly** (both changes survive; no conflict, no silent drop), and the same-timestamp migration interleave is harmless for fresh replays. Security scope machinery has zero HouseholdLead residue; `@sensitivity` annotation present and classifications regenerated; the ~50 test-file rewrites are faithful 1:1 translations with no net-deleted test blocks; the concurrency suite genuinely races via `Promise.allSettled`.

thpr:
The critical item is valuable - but we can't just capture it here.  @jee7s Can you build a design doc that sufficiently captures our infrastructure rules and order of operations and put that in docs/ ... then build a skill for our repo that considers that under some circumstances (at least if a migration is built, maybe other things?)

jee7s:
@thpr Done — both parts are in **#933**:

- **`checkin-app/docs/DEPLOY_MIGRATION_ORDER_OF_OPERATIONS.md`** — the deploy sequence as actually wired (migrations run to completion *before* the rolling update; old code serves through the drain window), the expand → cutover → contract rules, the Prisma footguns (no transaction wrapping, destructive rename codegen), P3009 recovery, and what the populated-DB CI check can and cannot catch — each rule anchored to the incident that produced it (#791, #792, this PR's finding). Placed next to `VOCABULARY.md` as a living reference (not `docs/designs/`, which is proposal-shaped), and added to AGENTS.md's docs map.
- **`.claude/skills/migration-safety/`** — fires on any `prisma/migrations/` edit, destructive `schema.prisma` changes (drop/rename/`SET NOT NULL` on existing columns), and migration-shaped task text; runs the doc's rules as a checklist alongside `safe-refactor-sweep` (they catch different failure classes). The headline check: *a green "Migrations apply to populated DB" job is necessary, not sufficient — old code must survive this schema through the drain window.*

One nuance the doc encodes rather than a blanket rule: contract-stage changes may ship with their expand stage **only** when the PR proves no live code path can violate the new state (#918's backfill + `SET NOT NULL` qualifies; this PR's `DROP TABLE` does not, since main's JWT callback reads `HouseholdLead` on every request).

For this PR specifically, the fix remains as in the review: hold `20260705170000_drop_household_lead` for a follow-up release after the cutover code is fully rolled out.

thpr:
917 is rebuilt with the critical item removed and other items addressed.

dkaygithub (approved):

jee7s (approved):

PR #918, opened 2026-07-06
feat(household): make Household.name NOT NULL

thpr:
## What

Tighten `Household.name` from nullable (`String?`) to `String`.

## Why

No live create path ever writes null: OAuth signup falls back to email, membership-ops to `'User'`, Zoho import to the literal `'Household'`. The nullable was schema slack no production row exercised.

## Changes

- **schema**: `name String?` → `name String`
- **migration `20260705160000`**: backfill any null from a household member's `Person.name` (itself nullable, so `COALESCE` → `'Household'`), then `SET NOT NULL`. Cannot fail on the live DB.
- **auth-options**: OAuth create's now-illegal `?? null` tail → `?? "Household"`.
- **membership-ops `households/[id]` route**: admin edit accepted raw `body.name` (could be null/empty) → rejects empty with `400`, trims.
- **~70 test fixtures**: spun throwaway households via `household: { create: {} }`, relying on the nullable default — all pass a name now.

## Reconciliation with #920

[#920](https://github.com/innovationtreehouse/checkin/pull/920) (PERSON_BG manual review) merged **after** this branch started. Its reviewer queue renders `"No household on file"` when a subject's household has no name — a **falsy check** (`household?.name ? … : "No household on file"`), not null-specific. That state is prod-unreachable (PERSON_BG subjects reuse their named signup household), but #920's integration test manufactured it with `name: null`.

Switched that one fixture to the empty-string sentinel `""` (+ its `.toBeNull()` → `.toBe("")`). The render's falsy check treats `""` identically, so #920's feature and its `"No household on file"` assertion are **fully preserved** under NOT NULL. No change to #920's shipped code — two lines in its test.

This is what ejected the PR from the merge queue 3× earlier: the queue merges with newer main, surfacing #920's null insert against the new constraint.

## Verification

- `tsc --noEmit`: clean
- Unit: 1560 pass
- Integration: 917 pass (fresh PG, all 5 migrations applied in order)
- Flow: passed in CI

## Reviewer notes

- Migration safe to deploy but **not yet applied to any real DB** — deploy step.
- Backfill source is a household member's name per request; `'Household'` literal only if zero named members.

### DIFF
--- a/checkin-app/__tests__/programLifecycle.integration.test.ts
+++ b/checkin-app/__tests__/programLifecycle.integration.test.ts
@@ -34,7 +34,7 @@ describe('Program Lifecycle Integration Tests', () => {
                 isSysadmin: false,
                 isBoardMember: true,
                 dateOfBirth: new Date('1990-01-01'),
-                household: { create: {} }
+                household: { create: { name: "Test HH" } }
             }
         });
         boardAdminId = adminUser.id;
@@ -46,7 +46,7 @@ describe('Program Lifecycle Integration Tests', () => {
                 email: "mentor@test.com",
                 googleId: "test-auth-mentor",
                 dateOfBirth: new Date('1985-01-01'),
-                household: { create: {} }
+                household: { create: { name: "Test HH" } }
             }
         });
         leadMentorId = mentorUser.id;
@@ -58,7 +58,7 @@ describe('Program Lifecycle Integration Tests', () => {
                 email: "participant@test.com",
                 googleId: "test-auth-std",
                 dateOfBirth: new Date('2000-01-01'),
-                household: { create: {} }
+                household: { create: { name: "Test HH" } }
             }
         });
         testParticipantId = standardUser.id;
--- a/checkin-app/prisma/migrations/20260705160000_household_name_not_null/migration.sql
+++ b/checkin-app/prisma/migrations/20260705160000_household_name_not_null/migration.sql
@@ -1,0 +1,18 @@
+-- Make Household.name NOT NULL. Every live create path already writes a name
+-- (OAuth: email fallback; membership-ops: 'User' fallback; Zoho: 'Household'
+-- literal), so no live row is expected to be null. Backfill defensively anyway:
+-- copy a household member's name, else the literal 'Household', before the
+-- NOT NULL constraint so the migration can never fail on the real DB.
+UPDATE "Household" h
+SET "name" = COALESCE(
+    (SELECT p."name"
+       FROM "Person" p
+      WHERE p."householdId" = h."id"
+        AND p."name" IS NOT NULL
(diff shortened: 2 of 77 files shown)


jee7s (approved):

jee7s:
Deep review (10 finder angles + independent verification). The migration itself is well-built — defensive backfill, correct member-name COALESCE, and it passed the populated-DB CI check. Four findings, none blocking if addressed:

**1. Empty-string names slip through the create path** (`auth-options.ts:65`). The new fallback `data.name ?? data.email ?? "Household"` uses `??`, which doesn't fall through on `""` — a Google profile whose name resolves to empty string creates a household named `""`: satisfies NOT NULL, blank everywhere in the UI, and defeats the PR's stated goal. The PATCH rename route got a `.trim()` rejection; the create path didn't. Fix: `(data.name?.trim() || data.email?.trim() || "Household")`.

**2. Wedge-window hardening — one line.** The backfill UPDATE and `ALTER … SET NOT NULL` run as two autocommit statements (Prisma doesn't wrap Postgres migrations; no repo migration uses BEGIN/COMMIT). Old-code pods — still writing `?? null` — serve for the whole migrate step; a NULL-name insert landing between the two statements fails the ALTER with 23502 → failed-migration record → P3009 wedge blocking all deploys (the exact failure class that wedged dev for a day this week). The trigger is rare (needs name AND email both nullish), but wrapping the migration in `BEGIN;`/`COMMIT;` converts the failure mode from "wedged, manual resolve" to "atomic rollback, clean retry" for free.

**3. Untouched files now carry misleading vestiges.** `finance-ops/membership-payment-plan/page.tsx` still types `name: string | null` (line 22) with `??` fallbacks (103, 106), and `dev/shopify/page.tsx:46` keeps `?? "(unnamed household)"` — dead for the null case the schema now forbids, and blind to the empty-string case that remains legal (compounding finding 1: a `""` name sails past every `??` and renders blank). Tighten the client types and guard on `!name.trim()` where a fallback is wanted.

**4. 183 hand-pasted `name: "Test HH"` fixtures across 72 test files** with no shared factory/constant in `src/test-helpers/`. A future change to test-household naming (e.g. per-suite unique names to avoid collisions on the shared integration DB) means 183 edits. One `testHousehold(tag)` helper would do it.

**Verified clean:** interaction with #917 — `git merge-tree` of the two PR heads auto-merges the whole repo cleanly and correctly (no conflict, no silent drop), and the same-timestamp-bucket migration ordering (`household_name_not_null` sorting between #917's two stages) was walked through and is harmless for fresh-DB replays: this migration touches only `Household.name`/`Person.name`. The backfill's degenerate case (household whose members all have NULL names → literal 'Household') is handled by the COALESCE as intended.

PR #921, opened 2026-07-06
feat(email): board-configurable sender From + Reply-To (own Settings section)

jee7s:
## What

Adds a dedicated **Email** section under Settings (`/settings/email`) where board/sysadmin configure the outbound-email sender identity, and adds **Reply-To** support the codebase didn't have.

- **Reply-To (new capability).** Every email currently sends with no `Reply-To`, so replies go to the no-reply From address. A configured Reply-To (e.g. `board@…`) routes replies to a real inbox. Passed through to Resend's `replyTo`; omitted entirely when unset; need not be on the sending domain.
- **From override.** `BoardSettings.emailFromAddress` overrides the `EMAIL_FROM` env default when set; blank falls back to env. The env default (per-environment in the ECS task-def) stays the bootstrap value — **no infra change needed**.

Both fields sit behind the **same confirm-to-edit lock the membership-year boundary uses**: once a value exists the fields lock behind an "I understand — let me edit" checkbox. High-stakes — a From on a non-Resend-verified domain bounces *all* mail; a wrong Reply-To silently misroutes replies. First-time set is unguarded, like the boundary.

## Layout

New **Email** tab in `SettingsTabs` (board + sysadmin, matching what the `/settings` layout admits — membership's existing access level is preserved), its own page `src/app/settings/email/page.tsx`, and its own API route `/api/settings/email` (GET + PUT) scoped to the two fields. It reads/writes the same `BoardSettings` singleton — the storage and the `getEmailSenderIdentity()` resolution are unchanged; only the edit surface is its own section.

## How

- Resolution is centralized in `getEmailSenderIdentity()` (`src/lib/emailIdentity.ts`), called by the single `sendEmail` chokepoint — **all ~15 callers unchanged**. On any settings-read error it falls back to the env-only identity, so a `BoardSettings` hiccup never blocks delivery.
- `/api/settings/email` PUT validates the From/Reply-To header shape (bare address or `Name <addr@domain>`) and rejects the whole update on a bad value so the previous value survives; blank clears back to null. Writes an audit-log row.
- Both fields are `/// @sensitivity:public` (endpoint role-gated to board/sysadmin; a From header ships in every email — not secret). Generated classifications regenerated + committed.
- Migration `20260705150000_add_email_sender_settings` is two nullable `ADD COLUMN`s — a safe additive change on a populated table.

## Verification

- **Migration:** fresh `migrate deploy` clean; `prisma migrate diff --exit-code` → no drift.
- **Tests:** `emailIdentity` unit (override / blank-is-unset / never-blocks-on-error); `email.test` reply-to pass-through + omission; new `emailSettingsAPI` integration (read gate / accept-valid / reject-malformed-keeps-prior / clear-to-null); new email **page** test (first-time-set editable vs already-set lock/unlock guard). Unit + page suites and the two settings integration suites green; `tsc`, `eslint`, security-classification suite, and `check-route-coverage` (0 errors) all clean.

## Multiple Reply-To addresses

Reply-To now accepts a **comma-separated list** (e.g. `info@innovationtreehouse.org, ops@innovationtreehouse.org`) so replies can fan out to more than one inbox. Design keeps the existing single `TEXT` column (`BoardSettings.emailReplyToAddress`) — **no migration**:

- `parseEmailHeaderList()` (`src/lib/emailHeader.ts`) splits the value on `,`, trims each entry, validates each one with the existing `isValidEmailHeader` (same bare-address / `Name <addr@domain>` shapes), and dedupes exact repeats. Any one malformed entry rejects the *whole* list (400) — same reject-and-keep-prior-value contract the single-address From/Reply-To validation already had.
- `getEmailSenderIdentity()` now returns `replyTo?: string[]` (parsed, trimmed, deduped; `undefined` when unset) instead of `replyTo?: string`.
- Resend's SDK already accepts `replyTo?: string | string[]` (`node_modules/resend/dist/index.d.cts`), so `email.ts` passes the array straight through to `resend.emails.send()` — **no changes needed there**; a single configured address just becomes a one-element array.
- From stays single-address, unchanged — multiple Froms aren't a real use case and weren't in scope.
- Route (`/api/settings/email` PUT) and the client settings page both switch their Reply-To validation to the new list parser; From validation is untouched.
- **Known limitation (documented with a `ponytail:` comment):** the split is on a bare `,`. A quoted display name that itself contains a comma (e.g. `"Smith, Jane" <jane@x.org>`) is not supported — it would be mis-split into two bogus entries and rejected. Not a real-world case for board/ops addresses; upgrade to a quoted-string-aware splitter if it ever comes up.

Tests added/updated: `emailHeader.test.ts` (list parsing: single, multi with spaces, display-name entries, dedupe, reject-whole-list-on-one-bad-entry, blank), `emailIdentity.test.ts` (comma value → array, whitespace/dedupe, single address stays a one-element array), `email.test.ts` (array Reply-To passed through to Resend verbatim, single and multi), `emailSettingsAPI.integration.test.ts` (PUT accepts a valid comma list, rejects a list with one bad entry and keeps the prior value), and the settings page test (a comma list produces no inline error and fires the PUT). The Reply-To field's description and placeholder were updated to show the comma-separated form.

## Follow-up note

For prod first-launch this pairs with the earlier readiness note: the prod From domain (`updates.innovationtreehouse.org`) must be Resend-verified. This section lets the board point From/Reply-To at the right addresses without a redeploy.

### DIFF
--- a/checkin-app/src/app/api/settings/email/route.ts
+++ b/checkin-app/src/app/api/settings/email/route.ts
@@ -2,21 +2,11 @@ import { NextResponse } from "next/server";
 import { withAuth } from "@/lib/auth";
 import prisma from "@/lib/prisma";
 import { apiError } from "@/lib/api-response";
+import { isValidEmailHeader } from "@/lib/emailHeader";
 
 export const dynamic = "force-dynamic";
 
 const SELECT = { emailFromAddress: true, emailReplyToAddress: true } as const;
-
-/**
- * Accept a bare address (`a@b.co`) or a display-name form (`Name <a@b.co>`) — the
- * two shapes Resend allows for From/Reply-To. Rejects anything without a plausible
- * address so a typo can't silently break (From) or misroute (Reply-To) all mail.
- */
-function isValidEmailHeader(value: string): boolean {
-    const angle = value.match(/<([^>]*)>\s*$/);
-    const addr = (angle ? angle[1] : value).trim();
-    return /^[^\s@]+@[^\s@]+\.[^\s@]+$/.test(addr);
-}
 
 /** GET /api/settings/email — outbound-email sender identity (BoardSettings singleton). */
 export const GET = withAuth({ roles: ["isSysadmin", "isBoardMember"] }, async () => {
--- a/checkin-app/src/app/api/settings/email/route.ts
+++ b/checkin-app/src/app/api/settings/email/route.ts
@@ -2,7 +2,7 @@ import { NextResponse } from "next/server";
 import { withAuth } from "@/lib/auth";
 import prisma from "@/lib/prisma";
 import { apiError } from "@/lib/api-response";
-import { isValidEmailHeader } from "@/lib/emailHeader";
+import { isValidEmailHeader, parseEmailHeaderList } from "@/lib/emailHeader";
 
 export const dynamic = "force-dynamic";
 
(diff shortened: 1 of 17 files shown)


checkin-app/src/app/settings/membership/page.tsx
thpr, line 1: File should do client validation rather than relying an an API error message from the route

thpr (changes requested):

checkin-app/src/app/api/settings/email/route.ts
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Accept a bare address (`a@b.co`) or a display-name form (`Name <a@b.co>`) — the
+ * two shapes Resend allows for From/Reply-To. Rejects anything without a plausible
+ * address so a typo can't silently break (From) or misroute (Reply-To) all mail.
+ */
+function isValidEmailHeader(value: string): boolean {
+    const angle = value.match(/<([^>]*)>\s*$/);
+    const addr = (angle ? angle[1] : value).trim();
+    return /^[^\s@]+@[^\s@]+\.[^\s@]+$/.test(addr);
```
thpr: Reinvented: isValidEmailHeader in route.ts hardcodes /^[^\s@]+@[^\s@]+\.[^\s@]+$/ — byte-identical to EMAIL_RE behind isValidEmail in emergencyContacts/identity.ts:23, already imported by ~7 routes/pages.

thpr (approved):

PR #930, opened 2026-07-06
Shopify single-pool capacity + scholarship lifecycle inventory (supersedes two-pool mirror)

jee7s:
## ⚠️ Branch status: now conflicts with `main` (not resolved in this pass)

Since this branch's last push, `main` advanced with overlapping, unrelated work —
most notably #931 ("snapshot org-membership status at scholarship approval"), which
added its own new column to `ProgramParticipant` and its own new field to the exact
same approve-route `data` object this PR also edits, plus touches to the same
`schema.prisma`, `settings/membership/route.ts`, and
`programPaymentPlansAPI.integration.test.ts`. GitHub now reports this PR's merge
state as **CONFLICTING**. The two features are orthogonal (their
`wasOrgMemberAtApproval` snapshot vs. this PR's hold-ledger fields) and the conflicts
look mechanical, not semantic — but resolving them (rebase, reconcile both new
migrations under `main`'s new "coalesce" release gate, re-verify) is a distinct piece
of work not attempted here. **This PR needs a rebase onto `main` before it can merge.**

## Design evolution

This PR has now shipped three designs in sequence, each superseding the last:

1. **Two-pool mirror** (original): two Shopify variants per program (org-member /
   non-org-member), each with its own inventory pool, kept in sync via a
   sibling-mirror adjustment on every seat-consuming event.
2. **Single pool** (first revision): one Shopify variant per program at the base
   price; members get a per-checkout discount code instead of a second pool. Fixes
   the two-pool model's real bug (Shopify could sell up to 2× the intended cap
   across two independently-tracked pools).
3. **Hold ledger + grace period** (this update — final): reworks the scholarship
   lifecycle from "approve-time decrement, deny-time restore" into a ledger where
   every application-time `-1` is tracked (`ProgramParticipant.inventoryHeldAt`) and
   released `+1` **exactly once**, or consumed permanently by approval. This
   directly resolves the "open policy question" the previous revision of this PR
   flagged (see below) — denial no longer touches Shopify at all, so DB capacity and
   Shopify's count can no longer silently drift apart from a denied-but-not-withdrawn
   applicant.

### Program creation & repair, member pricing — unchanged from the prior revision
`createShopifySingleVariantProgram`, `createShopifyProgramVariants` (legacy path),
`mintMemberDiscountCode`, and the `sync-shopify` repair route's branching are all
unchanged by this update — see the single-pool section below for what they do.

- **`createShopifySingleVariantProgram`** (`src/lib/shopify.ts`): one Shopify variant
  at the base/non-member price, inventory = `maxParticipants`. `POST /api/programs`
  uses this for all new programs.
- **`createShopifyProgramVariants`** (legacy two-variant) stays, unchanged, for
  programs already on that model.
- **`mintMemberDiscountCode`**: mints a Shopify Price Rule + Discount Code (fixed
  amount off, usage limit 1, ~48h expiry) per checkout for an ACTIVE org member on a
  single-pool program, via `POST /api/programs/[id]/discount-code`. A failed mint
  degrades to an undiscounted link — checkout is never blocked. Interim mechanism;
  `docs/designs/SHOPIFY_MEMBER_SEGMENT_PRICING.md` is the upgrade path once checkout
  identity is solved (addendum added to that doc's status header in this update).

### Scholarship lifecycle — now a hold ledger (THIS UPDATE reworks this section)

**Schema (additive migration `20260706160000_scholarship_hold_ledger`):**
`ProgramParticipant.inventoryHeldAt` / `.paymentPlanDeniedAt` (both nullable
`DateTime?`), `BoardSettings.scholarshipDenialGraceDays` (nullable `Int?`, **NULL =
the expiry feature is off**). Verified: applies cleanly to a populated
`ProgramParticipant`/`BoardSettings` table, and `prisma migrate diff` shows no drift
against `schema.prisma`.

- **APPLICATION** (`POST /api/programs/[id]/request-payment-plan`): keeps the `-1`,
  additionally stamps `inventoryHeldAt = now()` and clears `paymentPlanDeniedAt`.
  Guard is now on `inventoryHeldAt` (not `isPaymentPlanRequested`): a denied
  re-applicant already holds the seat, so re-applying just flips the request flag
  and clears the denial stamp — it does **not** fire a second `-1`.
- **DENIAL** (`POST /api/finance-ops/payment-plans/refuse`, reworked): **removes the
  `+1`** this PR previously shipped here. Flips `isPaymentPlanRequested` false,
  stamps `paymentPlanDeniedAt = now()`, audit-logs. The seat stays held — the
  applicant may still pay normally, or withdraw.
- **APPROVAL** (`POST /api/finance-ops/payment-plans`): still no Shopify operation.
  Additionally clears `inventoryHeldAt` to `null` **without** a `+1` — the hold is
  **consumed**, not released, so a later removal of this now permanently-comped
  participant can never credit a seat back that was never returned to Shopify.
- **RELEASE, exactly once, three paths** — all share one implementation
  (`withdrawAndReleaseHold`, `src/lib/program/capacity.ts`): `delete()` is the
  atomicity boundary (a row can be deleted at most once), so double-release across
  concurrent callers isn't possible.
  - **(a) Withdrawal** — `DELETE /api/programs/[id]/participants` (the one route
    handles both self-withdraw and admin removal) and the non-payment kick
    (`cron/pending-participants`, switched from a bulk `deleteMany` to a per-row
    loop specifically so this release can run per participant, with per-row
    failure isolation).
  - **(b) Normal payment** — the `orders/paid` webhook's activation path. A denied
    applicant who pays anyway makes Shopify auto-decrement a *second* unit for the
    same seat; the webhook now releases the hold (`+1`) to compensate, guarded
    transactionally, never allowed to fail the webhook's 200.
  - **(c) Grace-period expiry** — new `GET /api/cron/scholarship-grace-expiry`
    (same `withCron`/`CRON_SECRET` auth as every other cron route). Sweeps `PENDING`
    rows with a denial stamp older than `scholarshipDenialGraceDays`, auto-withdraws
    each one (reusing path (a) — **auto-withdrawal + seat restore**) and audit-logs
    with a distinct `reason: "scholarship_grace_expired"` context. Per-row failure
    isolation; a no-op entirely when the setting is `NULL`.
- **Settings**: `scholarshipDenialGraceDays` exposed on `/settings/membership`
  alongside the other board policy knobs (`bgRecheckMonths`, dues) — board/sysadmin
  only, client-validates a positive integer or blank→null, audit-logged.

**One interpretation to flag:** grace-period expiry is implemented as
**auto-withdrawal + seat restore** (the row is deleted, same as a manual withdrawal,
and the seat becomes sellable again — the person could re-enroll if seats remain).
An alternative reading of "expire the hold" would be **restore-only**: release the
`+1` but leave the `PENDING` row in place (denied, unpaid, no longer holding a
seat). We went with auto-withdrawal because it matches the plain-language framing
("their spot is freed up") and reuses the existing withdrawal path outright rather
than inventing a fourth participant state. If the board would rather the row stay
and only the seat capacity release, that's a small, isolated change (drop the
`delete()` call in the sweep, keep just the `+1`) — flagging now in case the
restore-only reading is what's actually wanted.

### Webhook / inventory model — unchanged from the prior revision except the new release (b) above
Variant matching (`shopifyVariantId` or legacy pair), the legacy-only sibling
mirror, `adjustProgramInventory`'s single-pool preference, and the
`PATCH /api/programs/[id]` delta propagation are all unchanged.

## Docs (new in this update)

- **New `checkin-app/docs/PROGRAM_CAPACITY_AND_SCHOLARSHIPS.md`**: single-pool
  design, Shopify-as-source-of-truth, the discount-code mechanism (pointing at
  `SHOPIFY_MEMBER_SEGMENT_PRICING.md` as the segment upgrade path), the full
  hold-ledger state machine, the grace setting, and the legacy two-variant
  expand/contract transition.
- One-line `AGENTS.md` docs-map entry pointing at the new doc.
- Short addendum to `docs/designs/SHOPIFY_MEMBER_SEGMENT_PRICING.md`'s status
  header: single-pool + per-enrollee codes shipped as interim; segments remain the
  end-state.
- This section replaces the previous revision's "Flagged edge — open policy
  question" section, which is now resolved by the hold-ledger design above.

## Tests

- **New**: `cronScholarshipGraceExpiry.integration.test.ts` (NULL-setting no-op,
  sweep boundary — inside-window untouched vs. outside-window released +
  auto-withdrawn, per-row failure isolation via a `delete()` spy, never touches
  `ACTIVE` rows).
- **Updated**: `programPaymentPlansAPI.integration.test.ts` (denial → no Shopify op
  + hold persists + `paymentPlanDeniedAt` stamped, double-refuse no-op; re-apply
  while holding → no second `-1` + denial stamp clears; approval → hold consumed +
  a later removal fires no `+1` — these replace the old deny-time-`+1` assertions).
- **Updated**: `programsParticipantsAPI.integration.test.ts` (withdrawal releases a
  held seat exactly once; double-withdraw no-op).
- **Updated**: `webhookShopify.integration.test.ts` (denied-then-paid participant:
  activates AND releases the hold).
- **Updated**: `membershipSettingsAPI.integration.test.ts`,
  `settings/membership/__tests__/page.test.tsx` (grace-days validation: positive
  integer, blank→null).
- **Full suite**: 200 unit suites / 1619 tests green (`npx jest --ci`); 115
  integration suites / 963 tests green (`INTEGRATION_DB=1`); `tsc --noEmit` clean;
  `eslint --max-warnings 0` clean; migration verified against a populated DB with
  `prisma migrate diff` showing no drift.

## Cross-references

Supersedes this PR's own deny-time `+1` (previous revision) and the two-pool mirror
before that. Builds on #926, #927, #929 as before. See the branch-status note at the
top of this body for the new conflict with #931 (merged to `main` after this
branch's last sync).

### DIFF
--- a/checkin-app/jest.flow.config.js
+++ b/checkin-app/jest.flow.config.js
@@ -15,4 +15,14 @@ module.exports = createJestConfig({
     // The standalone build nests a package.json named "checkmein" under .next/,
     // which collides with the app's own in jest's haste map.
     modulePathIgnorePatterns: ['<rootDir>/.next/'],
+    // Jest's 5000ms default is too tight here: whichever file runs FIRST pays the
+    // full first-hit Turbopack compile cost for every route the journey touches
+    // (~10+ sequential requests against a `next dev` server that hasn't compiled
+    // any of them yet), and that cold-start tax alone can exceed 5s on a loaded CI
+    // runner — timing out mid-journey, which then races the abandoned in-flight
+    // request against afterAll's cleanup (e.g. #930's CI failure: a timed-out
+    // membership-activation-fanout run left its Shopify webhook call racing
+    // resetApplicantHousehold's DELETE, surfacing as a spurious 500). Later files
+    // reuse the now-warm routes and finish in a fraction of this.
+    testTimeout: 20000,
 });
--- a/checkin-app/src/app/api/cron/pending-participants/route.ts
+++ b/checkin-app/src/app/api/cron/pending-participants/route.ts
@@ -10,6 +10,11 @@ export const GET = withCron(async () => {
             where: {
                 status: 'PENDING',
                 isPaymentPlanRequested: false,
+                // Denied scholarship applicants are governed by the
+                // scholarship-grace-expiry cron (scholarshipDenialGraceDays),
+                // not this 7-day clock — a denial never resets pendingSince, so
+                // without this exclusion they'd be kicked the night after denial.
+                paymentPlanDeniedAt: null,
                 pendingSince: { not: null }
             },
             include: {
@@ -32,9 +37,10 @@ export const GET = withCron(async () => {
             const warningText = `If not paid, your spot in ${record.program.name} will be freed up. If a payment plan is needed, contact the board via finance@innovationtreehouse.org`;
 
             if (diffDays >= 7) {
-                // Collect for removal below (hold-ledger: a denied scholarship
-                // applicant is isPaymentPlanRequested:false too, so this query can
-                // also catch a still-held seat — see the per-row removal below).
+                // Collect for removal below (denied scholarship applicants are
+                // excluded by the query above; anything caught here is an
+                // ordinary non-payer, though withdrawAndReleaseHold still
+                // handles a held seat correctly if one slips through).
                 toDelete.push(record);
 
                 kickedCount++;
--- a/checkin-app/src/app/api/cron/pending-participants/route.ts
+++ b/checkin-app/src/app/api/cron/pending-participants/route.ts
@@ -2,6 +2,7 @@ import { NextResponse } from "next/server";
 import { logger } from "@/lib/logger";
 import { withCron } from "@/lib/cronAuth";
 import prisma from "@/lib/prisma";
+import { withdrawAndReleaseHold } from "@/lib/program/capacity";
 
 export const GET = withCron(async () => {
         const now = new Date();
@@ -19,7 +20,7 @@ export const GET = withCron(async () => {
 
         let kickedCount = 0;
         let warnedCount = 0;
-        const toDelete: { programId: number; personId: number }[] = [];
+        const toDelete: typeof pendingParticipants = [];
 
         for (const record of pendingParticipants) {
             if (!record.pendingSince) continue;
@@ -31,11 +32,10 @@ export const GET = withCron(async () => {
             const warningText = `If not paid, your spot in ${record.program.name} will be freed up. If a payment plan is needed, contact the board via finance@innovationtreehouse.org`;
 
             if (diffDays >= 7) {
-                // Collect IDs for batch deletion
-                toDelete.push({
-                    programId: record.programId,
-                    personId: record.personId
-                });
+                // Collect for removal below (hold-ledger: a denied scholarship
+                // applicant is isPaymentPlanRequested:false too, so this query can
+                // also catch a still-held seat — see the per-row removal below).
+                toDelete.push(record);
 
                 kickedCount++;
 
@@ -56,13 +56,24 @@ export const GET = withCron(async () => {
             }
         }
 
-        if (toDelete.length > 0) {
-            await prisma.programParticipant.deleteMany({
-                where: {
-                    OR: toDelete
+        // Removed one row at a time (not a bulk deleteMany) so the hold-ledger
+        // release (withdrawAndReleaseHold) runs per participant, and one bad row
+        // can't block the rest of the sweep.
+        for (const record of toDelete) {
+            try {
+                await withdrawAndReleaseHold(record.programId, record.personId, record.program);
+            } catch (err) {
+                // P2025 = already removed by another path (e.g. self-withdraw)
+                // between this sweep's read and now — benign, skip.
+                if (!isPrismaP2025(err)) {
+                    logger.error(`[CRON] Failed to remove pending participant ${record.personId} from program ${record.programId}:`, err);
                 }
-            });
+            }
         }
 
         return NextResponse.json({ success: true, processed: pendingParticipants.length, kicked: kickedCount, warned: warnedCount });
 });
+
+function isPrismaP2025(err: unknown): boolean {
+    return typeof err === 'object' && err !== null && 'code' in err && (err as { code?: unknown }).code === 'P2025';
+}
--- a/checkin-app/src/app/api/programs/[id]/participants/route.ts
+++ b/checkin-app/src/app/api/programs/[id]/participants/route.ts
@@ -221,7 +221,11 @@ export const DELETE = withAuth({}, async (req, auth, { params }: { params: Promi
             }
         });
 
-        const responseObj: Record<string, unknown> = { success: true, enrollment };
+        // No raw enrollment echo: this route is reachable by the program's lead
+        // mentor, and the deleted row carries board/finance-confidential hardship
+        // fields (paymentPlanDeniedAt, inventoryHeldAt). The audit log above keeps
+        // the full row server-side.
+        const responseObj: Record<string, unknown> = { success: true, released };
         if (released && !shopifyOk) {
             responseObj.warning = "Participant removed, but the Shopify inventory release failed. Capacity may be out of sync — check System Status > Link Status.";
         }
--- a/checkin-app/src/app/api/programs/[id]/participants/route.ts
+++ b/checkin-app/src/app/api/programs/[id]/participants/route.ts
@@ -3,7 +3,7 @@ import { logger } from "@/lib/logger";
 import { withAuth } from "@/lib/auth";
 import prisma from "@/lib/prisma";
 import { sendNotification } from "@/lib/notifications";
-import { lockProgramAndCheckCapacity, ProgramCapacityError } from "@/lib/program/capacity";
+import { lockProgramAndCheckCapacity, ProgramCapacityError, withdrawAndReleaseHold } from "@/lib/program/capacity";
 import { checkProgramAge } from "@/lib/programAge";
 import { apiError } from "@/lib/api-response";
 
@@ -203,14 +203,12 @@ export const DELETE = withAuth({}, async (req, auth, { params }: { params: Promi
             return apiError("Forbidden: Not authorized to remove this participant", 403);
         }
 
-        const enrollment = await prisma.programParticipant.delete({
-            where: {
-                programId_personId: {
-                    programId,
-                    personId: participantId
-                }
-            }
-        });
+        // Hold-ledger (product decision 2026-07-06): withdrawal is one of the
+        // three release paths — if this PENDING participant was holding a
+        // scholarship seat (inventoryHeldAt set), releasing it back to
+        // Shopify (+1) happens with the removal, exactly once (see
+        // withdrawAndReleaseHold).
+        const { enrollment, released, shopifyOk } = await withdrawAndReleaseHold(programId, participantId, currentProgram);
 
         await prisma.auditLog.create({
             data: {
@@ -223,7 +221,11 @@ export const DELETE = withAuth({}, async (req, auth, { params }: { params: Promi
             }
         });
 
-        return NextResponse.json({ success: true, enrollment });
+        const responseObj: Record<string, unknown> = { success: true, enrollment };
+        if (released && !shopifyOk) {
+            responseObj.warning = "Participant removed, but the Shopify inventory release failed. Capacity may be out of sync — check System Status > Link Status.";
+        }
+        return NextResponse.json(responseObj);
     } catch (error) {
         // P2025 = row to delete not found. Benign double-submit (participant
         // already un-enrolled); idempotent 200 instead of a 500.
--- a/checkin-app/src/app/api/programs/[id]/request-payment-plan/route.ts
+++ b/checkin-app/src/app/api/programs/[id]/request-payment-plan/route.ts
@@ -78,13 +78,18 @@ export const POST = withAuth({}, async (req, auth, { params }: { params: Promise
         // fires -1. Branch 2 (inventoryHeldAt already set): re-application while
         // holding (or an idempotent re-POST) — just (re)flips the request flag
         // and clears any denial stamp, no second decrement.
+        // Both stamps are compare-and-set on status: 'PENDING' (matching
+        // approve/refuse) — the findUnique status read above is stale by write
+        // time, and without this a concurrent approval (ACTIVE, hold consumed
+        // to null) would match branch 1, re-stamp the hold, fire a second -1,
+        // and reopen a resolved scholarship.
         const holdResult = await prisma.programParticipant.updateMany({
-            where: { programId, personId: participantId, inventoryHeldAt: null },
+            where: { programId, personId: participantId, status: 'PENDING', inventoryHeldAt: null },
             data: { isPaymentPlanRequested: true, inventoryHeldAt: new Date(), paymentPlanDeniedAt: null },
         });
         if (holdResult.count === 0) {
             await prisma.programParticipant.updateMany({
-                where: { programId, personId: participantId, inventoryHeldAt: { not: null } },
+                where: { programId, personId: participantId, status: 'PENDING', inventoryHeldAt: { not: null } },
                 data: { isPaymentPlanRequested: true, paymentPlanDeniedAt: null },
             });
         }
--- a/checkin-app/src/app/api/programs/[id]/request-payment-plan/route.ts
+++ b/checkin-app/src/app/api/programs/[id]/request-payment-plan/route.ts
@@ -3,6 +3,7 @@ import { logger } from "@/lib/logger";
 import { withAuth } from "@/lib/auth";
 import prisma from "@/lib/prisma";
 import { apiError } from "@/lib/api-response";
+import { adjustProgramInventory } from "@/lib/shopify";
 
 export const POST = withAuth({}, async (req, auth, { params }: { params: Promise<{ id: string }> }) => {
     if (auth.type !== 'session') return apiError("Unauthorized", 401);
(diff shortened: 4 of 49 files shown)


checkin-app/src/app/api/programs/[id]/route.ts
```diff
@@ -183,7 +184,33 @@ export const PATCH = withAuth({}, async (req, auth, ctx: { params: Promise<{ id:
...
+                // "start/stop tracking inventory" — that needs an inventory_management
+                // flip on the variant, which this PR intentionally doesn't attempt.
+                logger.warn(`[SHOPIFY] Program ${updatedProgram.id} maxParticipants transitioned ${oldMax} -> ${newMax} (capped/uncapped); inventory not adjusted automatically.`);
+                warning = "Capacity changed between capped and uncapped. Shopify inventory was not updated automatically — run Sync to Shopify or edit inventory directly in Shopify.";
+            }
+        }
+
+        const responseObj: Record<string, unknown> = { success: true, program: updatedProgram };
+        if (warning) responseObj.warning = warning;
+        return NextResponse.json(responseObj);
```
Comment (thpr) [inline checkin-app/src/app/api/programs/[id]/route.ts:214]: This seems to warn the user - who doesn't care or understand?  Should a warning here really email board?

checkin-app/jest.flow.config.js
```diff
@@ -15,4 +15,14 @@ module.exports = createJestConfig({
...
+    // Jest's 5000ms default is too tight here: whichever file runs FIRST pays the
+    // full first-hit Turbopack compile cost for every route the journey touches
+    // (~10+ sequential requests against a `next dev` server that hasn't compiled
+    // any of them yet), and that cold-start tax alone can exceed 5s on a loaded CI
+    // runner — timing out mid-journey, which then races the abandoned in-flight
+    // request against afterAll's cleanup (e.g. #930's CI failure: a timed-out
+    // membership-activation-fanout run left its Shopify webhook call racing
+    // resetApplicantHousehold's DELETE, surfacing as a spurious 500). Later files
+    // reuse the now-warm routes and finish in a fraction of this.
+    testTimeout: 20000,
```
thpr, line 27: This is interacting with other changes made in PR #914 - should be audited if that's an issue

jee7s:
## Rework: scholarship lifecycle is now a hold ledger

This pushes a design supersession the user asked for after this PR's last push,
resolving the "open policy question" the prior revision flagged (a refused
applicant's held DB seat vs. Shopify's returned seat could drift apart). New HEAD:
`1977cbcc`.

### State machine

| Transition | Shopify op | `inventoryHeldAt` | `paymentPlanDeniedAt` |
|---|---|---|---|
| Apply (first time) | `-1` | `null` → now | cleared (already null) |
| Re-apply while holding (post-denial) | none | unchanged (already set) | → `null` |
| Deny | **none** (was `+1`) | unchanged | → now |
| Approve | none | → `null` (consumed, no `+1`) | unchanged |
| Withdraw (self/admin, or non-payment kick) | `+1` if held | → `null` | — (row deleted) |
| Pay anyway (webhook, after a denial) | `+1` if held | → `null` | — |
| Grace-period expiry (cron) | `+1` if held | → `null` | — (auto-withdrawn) |

Invariant: every application `-1` is returned `+1` **exactly once** (via withdrawal,
payment, or grace expiry) or consumed permanently by approval. `delete()` is the
atomicity boundary for all three release paths — a row can only be deleted once, so
concurrent double-release isn't possible.

### New/changed endpoints

- `POST /api/finance-ops/payment-plans/refuse` — reworked: no more `+1`; stamps
  `paymentPlanDeniedAt` instead.
- `POST /api/programs/[id]/request-payment-plan` — guard moved from
  `isPaymentPlanRequested` to `inventoryHeldAt`.
- `POST /api/finance-ops/payment-plans` (approve) — additionally clears
  `inventoryHeldAt` (no `+1`).
- `DELETE /api/programs/[id]/participants` — releases a held seat on withdrawal.
- `GET /api/cron/pending-participants` — switched from a bulk `deleteMany` to a
  per-row loop so the release above runs per participant it also catches.
- **New** `GET /api/cron/scholarship-grace-expiry` — the grace-period sweep.
- `POST /api/webhooks/shopify` (`orders/paid`) — releases a held seat on activation
  if the paying participant still had one.
- `PUT /api/settings/membership` — new `scholarshipDenialGraceDays` field.

### New shared helper

`withdrawAndReleaseHold` (`src/lib/program/capacity.ts`) — the one implementation of
"exit a PENDING participant + release its hold if any," shared by the withdrawal
route, the non-payment cron, and the grace-expiry cron, so the release logic isn't
triplicated.

### Migration

`20260706160000_scholarship_hold_ledger` — additive: `ProgramParticipant.inventoryHeldAt`,
`ProgramParticipant.paymentPlanDeniedAt`, `BoardSettings.scholarshipDenialGraceDays`.
Verified against a populated table (fresh apply + apply-to-populated-DB both
succeed) and `prisma migrate diff` shows no drift.

### Tests

New `cronScholarshipGraceExpiry.integration.test.ts` (NULL-setting no-op, sweep
boundary, per-row failure isolation, never touches `ACTIVE` rows) plus updates
across `programPaymentPlansAPI`, `programsParticipantsAPI`, `webhookShopify`,
`membershipSettingsAPI` integration tests and the membership settings page RTL
test. Full suite: 200 unit suites / 1619 tests, 115 integration suites / 963 tests,
all green; `tsc --noEmit` and `eslint --max-warnings 0` clean.

### Heads up: this PR now conflicts with `main`

`main` advanced since this branch's last push, including #931 ("snapshot
org-membership status at scholarship approval"), which touches the same approve
route, the same schema block, and the same test file this rework also touches. The
conflicts look mechanical (the two features are orthogonal), but reconciling them —
plus `main`'s new "at most 1 migration per release" coalesce gate against this
branch's two migrations — is real work not attempted in this push. **This needs a
rebase before it can merge.** Full detail in the PR body.

checkin-app/src/app/api/cron/pending-participants/route.ts
```diff
@@ -31,11 +32,10 @@ export const GET = withCron(async () => {
...
             if (diffDays >= 7) {
-                // Collect IDs for batch deletion
-                toDelete.push({
-                    programId: record.programId,
-                    personId: record.personId
-                });
+                // Collect for removal below (hold-ledger: a denied scholarship
+                // applicant is isPaymentPlanRequested:false too, so this query can
+                // also catch a still-held seat — see the per-row removal below).
+                toDelete.push(record);
```
thpr, line 44: **HIGH — the grace period is defeated by this 7-day non-payment kick.**

A denied scholarship applicant is `status:PENDING` + `isPaymentPlanRequested:false`, and `refuse` never resets `pendingSince` (it stays at original enrollment time). Scholarship review realistically takes >7 days, so by the time a request is denied `pendingSince` is already ≥7 days old — this sweep auto-withdraws them on the *next* nightly run, regardless of `scholarshipDenialGraceDays`.

Net: `cron/scholarship-grace-expiry` can only *shorten* the window, never extend it. A board setting grace=30 ("denied families get 30 days to pay") actually gets them removed at day 7. The stated intent — "the seat stays held — the applicant may still pay normally" — is undermined.

Inventory stays correct (both paths route through `withdrawAndReleaseHold`, `delete()` idempotent), so this is a **policy** break, not a double-release.

**Fix:** exclude held/denied rows from this query — add `paymentPlanDeniedAt: null` (and/or `inventoryHeldAt: null`) to the `where` — or reset `pendingSince` on denial so grace counts from the denial timestamp.

checkin-app/src/app/api/programs/[id]/request-payment-plan/route.ts
```diff
@@ -65,20 +66,46 @@ export const POST = withAuth({}, async (req, auth, { params }: { params: Promise
...
         });
 
         // Send email to finances
         // In a real implementation this would trigger an actual email via SendGrid, NodeMailer, etc.
         logger.info(`[EMAIL DISPATCH] To: finance@innovationtreehouse.org, Subject: Scholarship / Payment Plan Request for ${participant.person?.name || 'User'} in ${participant.program?.name || 'Program'}`);
 
-        return NextResponse.json({ success: true, participant: updatedParticipant });
+        let warning: string | undefined;
+        if (holdResult.count > 0 && participant.program) {
+            const ok = await adjustProgramInventory(participant.program, -1);
```
Comment (thpr) [inline checkin-app/src/app/api/programs/[id]/request-payment-plan/route.ts:103]: **MEDIUM — a failed `-1` here leaves a phantom hold → oversell on release.**

`inventoryHeldAt` is committed above (the `updateMany` at lines 79–88) *before* this `-1` fires. If `adjustProgramInventory(-1)` returns false, the row still records a hold but Shopify was never decremented. Any later release path (`withdrawAndReleaseHold` +1, the webhook +1, grace-expiry +1) then credits a seat that was never removed → Shopify count drifts high (oversell).

The `warning` string is the only signal and reconciliation is manual. The "every −1 comes back +1 exactly once" invariant is therefore best-effort, not guaranteed — the design doc should say so, and ideally a reconcile job or a persisted pending-adjustment flag should back it rather than a toast. (Symmetric residual on the release paths: `delete()` then `+1`, a crash between = under-sell.)

checkin-app/src/app/api/programs/[id]/participants/route.ts
```diff
@@ -223,7 +221,11 @@ export const DELETE = withAuth({}, async (req, auth, { params }: { params: Promi
             }
         });
 
-        return NextResponse.json({ success: true, enrollment });
+        const responseObj: Record<string, unknown> = { success: true, enrollment };
```
Comment (thpr) [inline checkin-app/src/app/api/programs/[id]/participants/route.ts]: **HIGH — confidential scholarship state leaked to the program lead mentor.**

This DELETE allows `isLeadMentor` (line ~199) and returns the raw deleted row as `enrollment` — which now carries the new `paymentPlanDeniedAt`. A program's volunteer lead mentor removing a participant thereby learns the family was **denied** a scholarship. `paymentPlanDeniedAt`/`inventoryHeldAt` are new columns, so this exposure is introduced by this PR.

Same role problem on `POST /api/programs/[id]/request-payment-plan`: `isLeadMentor` is in that allow-list too, so a lead mentor can both *initiate* a request and read `paymentPlanDeniedAt`/`inventoryHeldAt` off the response. Both routes return the raw `ProgramParticipant` via `withAuth` (no classification redaction, unlike `handler()`), so `personal`/`internal` fields are not stripped.

Scholarship *requested*/*denied* is board/finance-confidential — a program volunteer should neither trigger it nor observe it.

**Fix:**
- Drop `isLeadMentor` from the request-payment-plan allow-list (self / household-lead / board / sysadmin only), and from the `isHouseholdLead` short-circuit condition so that check still runs.
- Stop returning the hardship fields to non-board callers: this DELETE needs only `{ success, released, warning? }`; request-payment-plan should echo `{ id, status }`. Or route both through `handler()` and let classification strip by clearance.

thpr (commented):
Automated review focused on your three asks: single shared capacity, inventory accuracy on scholarship hold, and security/reliability (API exposure + transactional integrity). Single-pool capacity and the ledger math check out; three issues below need attention before merge. (Also note: PR is currently CONFLICTING with main after #931 — needs rebase regardless.)

jee7s:
All review feedback is now addressed, across `0ee15850` and `9587b9ea`:

**HIGH — 7-day kick defeats the grace period** → Fixed in `0ee15850` exactly as suggested: the kick query now excludes `paymentPlanDeniedAt: { not: null }` rows, so denied applicants are governed solely by `cron/scholarship-grace-expiry` (including the documented `NULL` = never-auto-expire case). Regression test: a day-10 denied applicant with a held seat survives the sweep.

**HIGH — scholarship state leaked to the lead mentor** → Fixed in `9587b9ea`, both halves: `isLeadMentor` is dropped from the request-payment-plan allow-list (self / household lead / board / sysadmin only), and the household-lead check runs without the mentor short-circuit. Neither route echoes the raw row anymore: request-payment-plan returns `{ success, warning? }`, the participants DELETE returns `{ success, released, warning? }` — `paymentPlanDeniedAt`/`inventoryHeldAt` never reach non-board callers (no UI reads those bodies; the audit log keeps the full row server-side). Tests: mentor→403, plus no-leak assertions on both responses.

**MEDIUM — failed `-1` leaves a phantom hold** → Fixed in `9587b9ea`: a failed application-time `-1` now rolls `inventoryHeldAt` back in the same request, so no phantom hold is recorded and a re-submit retries the decrement (test: rollback + `isPaymentPlanRequested` stays set). The design doc now has an explicit **failure semantics** section: the ledger invariant is guaranteed against failed calls, best-effort across crash windows (visible in DB, manual reconcile via Sync to Shopify); a reconcile job is deliberately deferred until drift is observed. Every failure also emails sysadmins/board via `reportShopifyFailure`, not just the toast.

**Earlier inline comments:**
- *route.ts warning — "should this really email board?"* — It does now: every `adjustProgramInventory` failure emails all sysadmin/board members via `reportShopifyFailure` (plus structured `logIntegrationError`); the response warning is a supplement for the admin performing the edit.
- *jest.flow.config.js vs #914* — Audited: #914 touched only `safety/trusted-adults/__tests__/page.test.tsx` (RTL unit-test timeouts under `jest.config.js`); the flow config is a separate config file (last changed by #490) that #914 never touches. No interaction — the `testTimeout: 20000` here addresses the flow suite's cold-start compile tax only.

**Rebase** — the branch has since merged `main` (`02b73840`); GitHub reports the PR mergeable with no conflicts.

CI: the six affected integration suites (participants, payment plans ×2, both scholarship crons, Shopify webhook) pass locally — 85/85.

checkin-app/src/app/api/programs/[id]/request-payment-plan/route.ts
```diff
@@ -57,28 +60,63 @@ export const POST = withAuth({}, async (req, auth, { params }: { params: Promise
...
+        // a denied re-applicant still holds the seat (refuse no longer releases
+        // it — see the refuse route), so re-applying must NOT decrement twice.
+        // Branch 1 (inventoryHeldAt null -> now): the seat is genuinely free —
+        // fires -1. Branch 2 (inventoryHeldAt already set): re-application while
+        // holding (or an idempotent re-POST) — just (re)flips the request flag
+        // and clears any denial stamp, no second decrement.
+        const holdResult = await prisma.programParticipant.updateMany({
+            where: { programId, personId: participantId, inventoryHeldAt: null },
+            data: { isPaymentPlanRequested: true, inventoryHeldAt: new Date(), paymentPlanDeniedAt: null },
         });
```
Comment (thpr) [inline checkin-app/src/app/api/programs/[id]/request-payment-plan/route.ts:89]: **Bug: apply stamp is missing the `status: 'PENDING'` predicate — double-decrement / reopen race.**

Both `updateMany` calls here (branch 1 `inventoryHeldAt: null`, branch 2 `inventoryHeldAt: { not: null }`) key only on `programId + personId + inventoryHeldAt`. Neither filters on `status`. The only status check is the stale `findUnique` read at L67, not re-verified at write time.

Approve (`finance-ops/payment-plans/route.ts` L76) and refuse (`.../refuse/route.ts` L46) both scope their `updateMany` with `status: 'PENDING'` — a proper compare-and-set. This route is the odd one out.

Race, under concurrent approval:

1. Applicant is `PENDING` holding a seat. Board approves → `status = ACTIVE`, `inventoryHeldAt = null` (hold consumed, no `+1` by design).
2. Concurrent re-POST of apply: `findUnique` saw `PENDING`; branch-1 `where inventoryHeldAt: null` now matches the just-cleared **ACTIVE** row → re-stamps `inventoryHeldAt = now()`, fires a second Shopify `-1`, and re-opens `isPaymentPlanRequested` on a now-comped participant.

Net: Shopify over-decrements one seat (capacity accuracy — the exact invariant this PR is about) and a resolved scholarship silently reopens.

Fix is one line on each `updateMany` — add `status: 'PENDING'` to the WHERE, matching approve/refuse. That makes the stamp a true compare-and-set and closes the window without needing a transaction:

```js
const holdResult = await prisma.programParticipant.updateMany({
    where: { programId, personId: participantId, status: 'PENDING', inventoryHeldAt: null },
    data: { isPaymentPlanRequested: true, inventoryHeldAt: new Date(), paymentPlanDeniedAt: null },
});
if (holdResult.count === 0) {
    await prisma.programParticipant.updateMany({
        where: { programId, personId: participantId, status: 'PENDING', inventoryHeldAt: { not: null } },
        data: { isPaymentPlanRequested: true, paymentPlanDeniedAt: null },
    });
}
```

jee7s:
**Apply-stamp race (double-decrement / reopen) — fixed in `1c4aa65b`**, exactly as prescribed: both `updateMany` branches in `request-payment-plan` now carry `status: 'PENDING'` in the WHERE, making the stamp a true compare-and-set consistent with approve (`payment-plans/route.ts`) and refuse.

How the race resolves now: a concurrent approval flips the row to `ACTIVE` with `inventoryHeldAt: null` (hold consumed); the racing re-POST's branch-1 WHERE (`status: 'PENDING', inventoryHeldAt: null`) no longer matches that row, branch 2 matches nothing either, and since the Shopify `-1` is gated on `holdResult.count > 0`, **no second decrement fires and the resolved scholarship stays closed** — the request degrades to a harmless no-op. The race is documented in a comment above the stamps.

No dedicated test: the window between the `findUnique` read and the write isn't deterministically reachable at route level (the stale-read path already has a 409 test), and the guard is the same CAS shape the approve/refuse suites already exercise.

Also in this push: the branch is **rebased onto latest `main`** (`c3567ef4`, linear history — conflict unions with #931/#946 and the #917 household-lead refactor, `prisma generate` reproduces the merged classifications byte-for-byte), plus two rebase-artifact cleanups (`886ce844` stale `cron/reminders` inventory line after #941, `f00ce86c` orphaned import that tripped `--max-warnings 0`). All six money-path integration suites pass (86/86) plus route-drift (370/370).

dkaygithub (approved):

thpr (approved):

PR #938, opened 2026-07-06
chore(deps): resolve all npm audit vulnerabilities (21 → 0)

dkaygithub:
## Summary

Full security audit of every dependency the monorepo pulls in. `npm audit` went from **21 vulnerabilities (7 high, 12 moderate, 2 low) to 0**. Also checked the Lambda layer manifest (pins the same Prisma 7.8.0, resolved at build time) and the Python kiosk client (PyNaCl + stdlib only) — nothing actionable in either.

### 1. Non-breaking upgrades via `npm audit fix` (13 advisories)

- **next 16.1 → 16.2.10** — 19 CVEs incl. HTTP request smuggling, middleware/proxy bypass, cache poisoning, XSS
- **undici** (TLS cert validation bypass), **ws** (memory disclosure), **form-data** (CRLF injection)
- flatted, picomatch, @babel/core, esbuild, brace-expansion, js-yaml, qs, yaml, and the resend→svix uuid chain

### 2. Scoped overrides (npm's only offered "fix" was a major downgrade)

| Override | Advisory | Why it's safe |
|---|---|---|
| `next-auth` → `uuid ^11.1.1` | GHSA-w5hq-g745-h8pq | next-auth only calls `uuid.v4()`, API-identical in v11 |
| `next` → `postcss ^8.5.10` | GHSA-qx2v-qp2m-jg93 | semver-compatible; next pins 8.4.31 |
| `@prisma/dev` → `@hono/node-server ^1.19.13` | GHSA-92pp-h63x-v22m | dev-only prisma tooling; drop the override once a fixed prisma ships |

### 3. xlsx — patched via SheetJS CDN (no fix exists on the npm registry)

The registry package is abandoned at 0.18.5 with prototype pollution (GHSA-4r6h-8v6p-xvw6) + ReDoS (GHSA-5pgg-2g8v-p4x9) on parsing untrusted files — exactly what the participant bulk-import routes do with uploads. SheetJS moved distribution to `cdn.sheetjs.com` after 0.18.5, so the dependency now points at the official patched tarball **xlsx-0.20.3.tgz** (lockfile-pinned with integrity hash, same API). Removal wasn't viable (bulk import depends on it); migrating to exceljs was considered and rejected — its maintenance is dormant (no release since Dec 2024) and its tree currently carries the same uuid advisory.

Future xlsx upgrades mean editing the URL in `checkin-app/package.json` rather than `npm update`.

## Verification

- `npm audit` → **0 vulnerabilities**
- App build (`next build` incl. typecheck), lint, all four Lambda function builds pass
- **1560 unit tests** and the **full 896-test integration suite** pass (incl. the bulk-import xlsx paths, run against a throwaway Postgres)
- Pre-commit hook re-ran the CI suite at commit time


jee7s:
Security review of the dependency changes (supply-chain + override-safety lens, all claims verified against the lockfile and CI):

## 🔴 Blocking — the upgrade silently removes the webhook signature-verification library

`resend` 6.9.3 → 6.17.1 (the "resend→svix uuid chain" fix) **drops `svix` from the dependency tree entirely** — on main, svix 1.84.1 was present only as resend's transitive dep. `src/app/api/webhooks/resend/route.ts` (#928) imports `svix` directly to verify Resend webhook signatures, so the branch fails to compile (`TS2307: Cannot find module 'svix'`) — that's your red `Run Tests`/`Deploy build`. This was a known fragility flagged in #928 ("svix is transitively available… in case the team prefers pinning it directly"); this PR is where it bites.

**Fix (small, belongs in this PR):** declare `svix` as a *direct* dependency of `checkin-app` (pin `^1.84.1`, the version the webhook suite was verified against) — first-party code importing it means it must be declared, independent of what resend does internally. Then re-run `npm audit` (svix brings its own subtree; verify still 0) and let CI re-validate — the `webhookResend` integration suite's bad-signature 401 cases are the behavioral gate that the verification path still works.

## ✅ Verified clean

- **`npm audit --package-lock-only` → 0 vulnerabilities** — the headline claim reproduces from the lockfile alone.
- **Lockfile provenance**: every resolved URL is `registry.npmjs.org` except `xlsx`, which comes from `cdn.sheetjs.com` — the vendor's official channel (no fixed version exists on npm), pinned to an exact tarball **with an integrity hash**, and 0.20.3 covers both known xlsx advisories (prototype-pollution ≥0.19.3, ReDoS ≥0.20.2). Residual worth knowing: installs now depend on that CDN's availability, and future `npm audit fix` runs won't manage this package.
- **Integrity hashes** present on every registry entry (the only integrity-less lockfile entries are the workspace links, which is normal).
- **Overrides are properly scoped**: the `uuid ^11.1.1` override is confined to `next-auth` (whose v4-only usage is API-identical in v11, as the table claims); `postcss ^8.5.10` under `next` is a semver-minor; `@hono/node-server` is dev-only tooling with a stated removal condition. No override leaks into other consumers' dependency contracts.
- Minor/informational: `uuid 14.0.1` nests under `jest-junit` (dev/test only) — no runtime exposure.

## Process note

With CI red, the "non-breaking" characterization of the `next` 16.2.10 and `next-auth`-adjacent changes is currently unvalidated — after the svix fix, a green `Run Tests` (which exercises the auth flows and webhook verification) is what actually substantiates it.

jee7s (approved):

PR #951, opened 2026-07-07
fix(dev-tools): Debug nav 404 on cloud-dev — replace the NODE_ENV fuse with devToolsActive()

jee7s:
## Bug

The left-nav **Debug** item routes to `/dev`, which redirects to `/dev/sent-mail` — and 404s on the cloud-dev instance.

**Root cause:** the page (and the email-capture branch in `sendEmail`) gated on the design docs' two-fuse idiom, `isDevInstance() && NODE_ENV !== 'production'`. But every *deployed* instance — cloud-dev included — runs the same production image (`checkin-app/Dockerfile` sets `ENV NODE_ENV=production`, and the ECS task def doesn't override it), so the build-level fuse is false on the very instance the tools exist for. The nav item and the `/dev` tab layout gate on `CHECKIN_ENV` (correctly), which is why the link *shows*, navigates, and then the page 404s.

This also means the no-`RESEND_API_KEY` email capture (EMAIL_DEV_MOCK) has been silently disabled on cloud-dev: `sendEmail` fell through to `return false` instead of capturing.

## Fix

One named predicate at the chokepoint — `config.devToolsActive()` — documented as *the* gate for `/dev` tools and their capture paths, so the broken idiom doesn't get re-added by the next tool. Prod safety is preserved by construction, per the design docs' actual intent ("dead in prod by construction"):

- `CHECKIN_ENV` is server-only and **fails safe to `'prod'`** for any unset/unrecognized value (`readCheckinEnv`);
- cloud-dev pages remain behind the org-member login gate (middleware), as the page's own access comment describes.

Switched to it: `/dev/sent-mail` render gate, its `clearAction` server action (independently-callable POST), and `sendEmail`'s capture branch. Comments in `sentMail.ts`/`email.ts`/the page updated to match.

The other `/dev` tabs are untouched: their gates (`signingMockActive`, `shopifyMockActive`, BG mock) express *feature*-level availability, not tool-page availability, and carry their own justifications (e.g. fixed-webhook-secret backstops).

## Testing

- `config.test.ts`: `devToolsActive` matrix — unset/garbage → prod-safe false, dev/local → true, plus the regression case: **true on `CHECKIN_ENV=dev` under `NODE_ENV=production`** (the cloud-dev reality).
- `email.test.ts`: capture now asserts the production-build-on-dev-instance case captures (was asserting the bug), prod-instance fail-loud case retained.
- Full pre-commit suite green.


thpr:
Given NODE_ENV was a safety fuse and this starts to inline the fuse to reduce complexity, should we just take the opportunity to eliminate NODE_ENV? (per Slack discussion)

dkaygithub (approved):

jee7s:
Implemented — NODE_ENV is now **eliminated as a safety fuse repo-wide** (latest push, which also merges current main).

Removed from every gating predicate, leaving `CHECKIN_ENV` (fail-safe to `prod` for unset/unrecognized values) as the single fuse:
- `config.ts`: `zohoMockActiveEnv`, `shopifyMockActiveEnv`, `bgMockActiveEnv` — the reasoning is documented once, above `shopifyMockActiveEnv`: the clause could never distinguish a misconfigured prod box from a legitimate deployment (all deployed instances run the production image), so it only broke dev-instance features while "protecting" against explicitly setting `CHECKIN_ENV=local` on prod infra — a misconfig class `CHECKIN_ENV` already owns.
- `auth/dev-personas` POST (fresh-registrant mint): `CHECKIN_ENV === 'local'` alone.
- `DevLoginPicker`: bonus catch of the same class — in a client bundle `NODE_ENV` is inlined at **build** time, so its early-return hid the picker in every production build regardless of instance; now gated on EnvProvider's `checkinEnv`.

Deliberately **kept** (runtime/test plumbing, not fuses): `prisma.ts` test pool sizing / `disposeExternalPool`, and the `prismaGlobal` hot-reload cache.

The three "mock inactive when NODE_ENV=production" tests are flipped into regression guards asserting the mocks **stay** active under a production build, and the design docs (ZOHO_SIGN_DEV_MOCK §3, BG_CHECK_DEV_MOCK, SHOPIFY_DEV_STORE_WEBHOOK) carry a dated amendment recording the decision. Full pre-commit suite green.

PR #953, opened 2026-07-07
feat(programs): ProgramInstance tier + backfill (restructure phase 1)

jee7s:
## What & why

Phase 1 of the **Program → ProgramInstance → Event** 3-tier restructure
(`docs/designs/PROGRAM_INSTANCE_RESTRUCTURE.md`). Today `Program` conflates a
catalog *definition* ("Woodworking 101") with a concrete *offering* ("Fall 2026
Woodworking"); the fix is a middle `ProgramInstance` tier. This PR ships the
doc's **P1 only** — the additive schema + a 1:1 backfill — and nothing that
reads it yet. It is **expand/contract-safe**: purely additive, safe for old code
during the deploy drain window (`docs/DEPLOY_MIGRATION_ORDER_OF_OPERATIONS.md`).

See the new **Re-validation** section at the top of the design doc for the full
scope + cut line.

## Decisions

- **Scope = the doc's P1, additive only.** New `ProgramInstance` table (all
  offering columns mirrored from `Program` + nullable narrowing overrides),
  nullable `Event.instanceId` FK, `Program.instances` / `Person.instancesLed`
  back-relations. **No** FK repoint, **no** read/security-claim cutover, **no**
  write/UI, **no** column drops — those are P2–P5 and are RENAME/DROP (contract)
  steps deliberately deferred to later releases.
- **Re-validated against #930, not #929 (which is unmerged).** The doc assumed
  `SHOPIFY_MEMBER_SEGMENT_PRICING.md` (#929) landed first (one variant + a stored
  `shopifyMemberDiscountId`). What merged is **#930** (single-pool capacity +
  scholarship hold ledger, `checkin-app/docs/PROGRAM_CAPACITY_AND_SCHOLARSHIPS.md`).
  So the instance mirrors the Shopify columns that **actually exist** —
  `shopifyProductId`, `shopifyVariantId`, and the legacy
  `shopifyOrgMemberVariantId`/`shopifyNonOrgMemberVariantId` pair — and carries
  **no** `shopifyMemberDiscountId`. Capacity (`maxParticipants`, Shopify-sourced)
  and the hold-ledger fields (`inventoryHeldAt`/`paymentPlanDeniedAt` on
  `ProgramParticipant`) are untouched in P1 and consistent with #930; the doc's
  re-validation section flags where P2/P3 must re-target `adjustProgramInventory`
  and `withdrawAndReleaseHold` to the instance.
- **Backfill (MB), id-aliased + idempotent.** One `ProgramInstance` per `Program`
[...]

PR #954, opened 2026-07-07
feat(programs): archive / un-archive programs (soft-archive)

jee7s:
## What & why

Adds a reversible **soft-archive** for programs so the board can retire a
mistaken / cancelled / duplicate / long-dead program from the active surfaces
without deleting it (which would destroy the participant history the ops surfaces
still need). Design: [`checkin-app/docs/designs/PROGRAM_ARCHIVE.md`](../blob/feat/program-archive/checkin-app/docs/designs/PROGRAM_ARCHIVE.md).

New field `Program.archivedAt DateTime?` (`@sensitivity:internal`). NULL = live;
non-null = archived.

- **Orthogonal to `phase`.** `phase`/`FINISHED` is a *lifecycle fact* (did it run?);
  archived is a *visibility/activity decision* (should we still surface it?). A
  program in **any** phase can be archived. A separate nullable timestamp keeps both
  facts independently true and makes un-archive a trivial `NULL` set that restores
  the prior phase.
- **Board/sysadmin only**, surfaced on the program-ops program page as an
  Archive / Un-archive button. It rides the existing `PATCH /api/programs/[id]` via
  an `archived` boolean gated to `isSysadmin || isBoardMember` (same gate the
  Shopify-identifier fields already use) — **no new route**, no new route-auth surface.
- **Hidden from default lists / pickers / dashboards** (public catalog, program-ops
  list with an **"Include archived"** toggle, session & trends pickers, nav
  todo-counts) via a shared `NOT_ARCHIVED` where-fragment. `includeArchived=true` is
  board-gated; `archivedAt` is tier `internal` so it is stripped from the
  `GET /api/programs/[id]` response for everyone but board/sysadmin.
  - Deliberately **not** hidden: a member's own enrollment view (`/api/programs/mine`)
    — that's enrollment history, not a catalog.
- **New activity blocked (409, before any Shopify side effect):** enrollment,
  payment-plan/scholarship request, and volunteer signup. The guard fires **before**
  any inventory hold/decrement, so a blocked action can never leak Shopify inventory
  (respects the #930 single-pool capacity model). A `maxParticipants` edit still saves
  to the DB but is not pushed to Shopify while archived. Enrollment blocks even a board
  force-enroll override (archived is a hard freeze, not an over-ridable soft limit).
  - Not blocked: reading rosters/history, participant *removal* (an exit, not new
    activity), metadata/phase edits. Lead-mentor read scope is **unchanged**
    (`buildCallerContext` untouched) — history stays readable exactly as before.
- **Crons skip archived programs** — full touched/not-touched table in the design doc.
  Changed: `pending-participants`, `scholarship-grace-expiry`, `post-event`/`nightly`
  post-event leg. Untouched: nightly facility-close leg, membership-renewals,
  person-bg-annual, trusted-adult-expiry.
- **No Shopify-side archive in this PR.** Archiving a program *should* eventually also
  archive its Shopify listing / zero inventory — that's a **sibling PR**. Integration
  point (the `archived` transition in `PATCH /api/programs/[id]`) is noted in the doc.

## Data model / migration

`20260708020000_program_archive`: one additive, nullable column, no backfill, no
index. NULL = "not archived", so every existing program keeps its current behavior.
Expand-only.

## Testing

- New `programArchive.integration.test.ts` (8 cases): board archive stamps
  `archivedAt` → hidden from default list (board can opt in; member can't) → blocked
  self-enroll (409, no row) → blocked board force-enroll override → blocked
  payment-plan with **no `-1` Shopify inventory side effect** (dev-mock log asserted
  absent, `inventoryHeldAt` stays null) → blocked volunteer signup → un-archive
  restores list visibility + enrollment → **cron-skip**: pending-participants sweep
  leaves an archived program's stale row but kicks a live one.
- Full pre-commit suite green: **206 suites / 1676 tests**, plus `tsc --noEmit` and
  `eslint --max-warnings 0`.

</content>


jee7s:
**Closed by post-first-release triage — conflicts with the ProgramInstance restructure design.**

`docs/designs/PROGRAM_INSTANCE_RESTRUCTURE.md` (implemented starting with #953) moves offering lifecycle to the ProgramInstance tier — instance-archive owns the lifecycle/discount. This PR puts `archivedAt` on the Program (catalog) row, so archiving "Fall 2026 Woodworking" would archive the whole catalog entry for every future instance, and the restructure would then migrate the columns down a tier, redoing the feature. Archive/un-archive is tracked as #1236 and should be built on the instance model.

Also noted: the migration predates the 2026-07-11 coalesced baseline (fails fresh databases) — renumber on any revival.

PR #958, opened 2026-07-07
feat: auto-notifications as things go stale

jee7s:
## What & why

Adds **auto-notifications as things go stale** — a small registry-driven
framework (see `checkin-app/docs/designs/STALENESS_NOTIFICATIONS.md`) that turns
three otherwise-silent aging conditions into proactive email. Two outputs from
one runner:

1. **Daily household-direct nudges** (`GET /api/cron/staleness-notifications`) —
   as a tracked thing approaches (and passes) its lapse date, the affected
   household is emailed at escalating thresholds. Deduped by a new
   `NotificationLedger` so overlapping/retried runs never re-send.
2. **Weekly board/ops digest** (`GET /api/cron/staleness-digest`) — one grouped
   email to the admin list (sysadmins + board, same resolution as
   `reportShopifyFailure`) listing everything currently stale. No ledger — it is
   periodic by construction.

Both routes are thin `withCron` wrappers over the shared runner, matching the
repo's one-route-per-concern cron layout.

## Scope & decisions (from the product interview)

- **In scope:** membership renewals, trusted adults, broken emails.
- **Background checks deliberately deferred** — adding a fourth type is one
  `StaleType` registration, not a rewrite (documented as a follow-up in the doc).
- **Trusted-adult 30-day warning folded in, not duplicated:** `runExpirySweep`
  keeps only the `APPROVED → EXPIRED` mechanics; the reminder now flows through
  the shared layer. `TrustedAdultReview.expiryWarningSentAt` is now vestigial
  (kept — a column drop is destructive; the ledger is the new dedup).
- **Broken emails** (`Person.emailUndeliverableAt`): you can't email a broken
  address, so the household-direct notice goes to the household's *other* leads
  with a working address; if there are none, only the digest carries it (it
  *always* lists broken emails). The dedup key embeds the break timestamp so a
  heal-then-rebreak is treated as a fresh event.
- **Escalation math** (`activeThreshold`): fires the *smallest* entered
  threshold, so a cron that skips days past 30 straight to 5 fires the 7-day
  stage — not a burst of all three. Each bucket fires at most once.

## Data model

New additive table `NotificationLedger(type, subjectKey, threshold, sentAt)` with
a unique constraint that doubles as the dedup guard (claim-then-send: the winning
insert owns the send; a concurrent run gets P2002 and skips). Migration
`20260708060000_staleness_notifications` — a plain `CREATE TABLE`, no existing row
touched.

## Testing

- **Unit** (`src/lib/staleness/__tests__/registry.test.ts`): `activeThreshold`
  buckets, the ledger dedup guard (P2002 → skip), no-recipients / not-yet-in-window
  skips, non-P2002 errors surfaced, and digest assembly + HTML escaping.
- **Integration** (`src/app/__tests__/cronStaleness.integration.test.ts`): one
  full run per output — household pass emails the *working* lead (never the broken
  address), writes the ledger, and a re-run re-sends nothing; the weekly digest
  emails the admin list one grouped digest of the currently-stale items.
- Updated the trusted-adult suites that asserted the removed `warned` behavior.
- Full pre-commit suite green (208 suites, 1704 tests); `tsc --noEmit` and


jee7s:
**Closed by post-first-release triage — no open issue mapped to this work.** The intent is now tracked as #1362 (project 1, Ideation/Reporting); the branch is retained. If revived: renumber the migration past the 2026-07-11 coalesced baseline and re-verify against current main (LIVE_PERSON drift guard, lifecycle machine, boundary-isolation rule).

PR #959, opened 2026-07-07
feat(household): archive / un-archive families (households)

jee7s:
## What & why

Adds **soft-archive for households (UI "families")** — a way to set a family
aside (moved away, aged out, duplicate) without deleting anything, so it drops
out of the board's day-to-day surfaces and its members can't start new activity,
while every historical row stays intact and one action fully restores it.

Design + decisions + the full query sweep:
[`checkin-app/docs/designs/HOUSEHOLD_ARCHIVE.md`](checkin-app/docs/designs/HOUSEHOLD_ARCHIVE.md).

Archive is **orthogonal to membership Deny** — it does *not* block login; it hides
the family from board lists/crons and blocks *new activity*. The two can coexist.

- **Nullable `Household.archivedAt DateTime?`** stamp (`@sensitivity:internal`).
  Non-null == archived. Additive/nullable → expand-only migration.
- **Who archived is in the `AuditLog`**, not a second column (mirrors the existing
  household Deny / membership-archive audit rows).
- **Board / sysadmin only**, surfaced in the **membership-ops households** view
  (rides the existing sanctioned households POST — no new route, no `routeAuthDrift`
  change).
- **Hidden from default lists / searches / pickers**; the households view gets an
  **"Include archived"** toggle to find them for un-archive.
- **Members blocked from new activity** — program enrollment POST, check-in/scan,
  and membership-application start — via one shared guard `assertHouseholdActive`
  (→ 409 "household is archived"), guarded at the service/route level, not per caller.
- **Crons & fan-outs skip archived** via one shared `ACTIVE_HOUSEHOLD_WHERE`
  filter folded into the existing predicates: renewal + person-BG sweeps, the
  new-program email blast, the audit/compliance chase lists, and the nav badges.
- **History fully intact; un-archive restores everything. No purge.**

The design doc §6 documents every household/person query that feeds a list, cron,
or fan-out as a **touched / deliberately-not-touched** table (e.g. the safety
emergency-contacts board is deliberately left showing archived households, since
archive doesn't force-checkout a member who may still be physically present).

## Testing

- New integration suite `householdArchive.integration.test.ts` (6 cases, all
  green against seeded Postgres): archive → stamped + audit row (+ idempotency) →
  hidden-from-list / shown-with-toggle → **blocked enrollment (409)** → **blocked
  membership start** → **renewal-sweep skip (representative cron)** → un-archive
  clears + audit + **enrollment restored**.
- Full unit suite green (1676 tests), `tsc --noEmit` clean, `eslint
  --max-warnings 0` clean. `classifications.ts` regenerated via `prisma generate`.
- Updated two existing unit mocks (`scanRoute`, `programSignupIntegration`) to
  supply `household.findUnique` for the new guard.

### DIFF
--- a/checkin-app/__tests__/programSignupIntegration.test.ts
+++ b/checkin-app/__tests__/programSignupIntegration.test.ts
@@ -55,6 +55,8 @@ jest.mock('@/lib/prisma', () => {
     household: {
       create: jest.fn(),
       delete: jest.fn(),
+      // Archived-household enroll guard (assertHouseholdActive); null == active.
+      findUnique: jest.fn().mockResolvedValue(null),
     },
     membership: {
       create: jest.fn(),
--- a/checkin-app/docs/designs/HOUSEHOLD_ARCHIVE.md
+++ b/checkin-app/docs/designs/HOUSEHOLD_ARCHIVE.md
@@ -1,0 +1,162 @@
+# Archiving / un-archiving Families (households)
+
+**Status:** BUILT (2026-07-08). Additive, nullable stamp + a shared "active
+household" filter + three write guards + a board/sysadmin toggle in the
+membership-ops households view.
+**Scope:** `checkin-app/` (paths below are relative to `checkin-app/`).
+
+---
+
+## 1. Problem
+
+A household (UI: "family") that has left the Treehouse — moved away, aged out,
+duplicate created in error — clutters every board list, still receives renewal
+reminders, and its members can still enroll, scan, and start applications.
+Deleting is not an option: attendance, membership, background-check, and audit
+history must stay intact for compliance. We need a **soft archive**: set the
+household aside so it drops out of the day-to-day surfaces and its members can't
+start new activity, while every historical row stays exactly where it is and a
+single action fully restores it.
+
+Archive is **not** the same as membership **Deny** (`OrgMembershipStatus.DENIED`,
+which blocks login for every member at the auth layer). Archive is softer and
+orthogonal: an archived household's members can still **log in and view** their
+own history — they just can't **start new activity** (enroll, scan, apply), and
+the household disappears from board-facing default lists. The two can coexist.
+
+## 2. Decisions (from the product interview) + rationale
+
+| Decision | Rationale |
+|---|---|
+| **Soft archive via a nullable `Household.archivedAt DateTime?` stamp** (`@sensitivity:internal`). | One nullable column; presence of a timestamp == archived, and it records *when*. Additive/nullable → expand-only migration, safe on the live DB. |
+| **Who archived is captured in `AuditLog`, not a second column.** | The audit log already records `actorId` + before/after on every board action (mirrors the household `Deny`/membership-archive rows). No new column earns its keep. |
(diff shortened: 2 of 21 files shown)


jee7s:
**Post-first-release triage: mapped to open issue #1228** (Archive Family — Household.archivedAt, exclude from active queries); #1232 tracks the view-filter sweep half. Stays open. Review blockers before merge: the migration is timestamped before the 2026-07-11 coalesced baseline and fails any fresh database — renumber it; re-verify every household/person list query against the LIVE_PERSON drift guard, which post-dates this branch.

thpr (changes requested):
## Changes required — built before the feature was fully designed; a full interview exposed gaps, with opens still on M4 (#1228)

This PR was built ahead of a full M4 design. It ships a sound **single soft-hide stamp**; a product interview against M4 (#1228) then surfaced requirements it predates — a **two-state lifecycle** with a hard precondition and state-aware self-reactivation — plus several **still-open decisions** on the main issue. Marking changes-required so the two-state shape is settled before merge; not a regression (M4/#1228 is still open).

**Full requirements review + interview transcript:** #1228 (https://github.com/innovationtreehouse/checkin/issues/1228#issuecomment-5082132340). Also on branch `claude/pr959-requirements-review` at `backlog/sources/pr959_requirements_review.md`.

### Keep (foundation is sound)
Nullable additive `archivedAt` + expand-only migration · the shared `ACTIVE_HOUSEHOLD_WHERE` filter · AuditLog row for who/when · `@sensitivity:internal` + board/sysadmin gate on the existing route.

### Gaps to reconcile before merge
1. **Two archive states in the DB** (not a text reason): **Form A** INVALID/duplicate = terminal, never reactivates (≈ M18 wearing an archive costume); **Form B** IDLE/aged-out = reactivatable. Both keep all history — neither purges. One `archivedAt` can't express "never reactivate," so the board un-archive button can revive a junk household.
2. **No-active-items precondition**: archive should **refuse** while live enrollments / an open visit / an in-flight membership-or-renewal process / pending payment or scholarship hold / list subs exist, and **return the specific blocking list** for manual cleanup. As built it stamps unconditionally and leaves that dangling.
3. **State-aware guards**: a Form-B family's real activity (enroll / newsletter signup / apply) should **auto-clear the stamp and proceed**; Form-A stays blocked. Today the enroll/scan/intake guards `409` the exact activity that should reactivate.
4. **Archive unsubscribes from all lists**; reactivation restores access only, never auto re-adds (honors SYNC-2). The PR only skips archived in the new-program cron — existing list membership persists.
5. **Login reminder + comms-page CTA** for reactivating Form-B families.

### Decisions still open on #1228
- **DN-1** how to model the two states (enum vs dual stamp vs status).
- **DN-2** does a clean ACTIVE OrgMembership block archiving, or archive alongside it (+ what happens to the not-yet-timed-out membership)?
- **DN-3** does Form A require email-invalidation (`.invalid`)? — resolve with M18 (#1225).

Recommend keeping #1228 open and reshaping this PR to the two-state design, or explicitly landing it as a labeled thin slice.

PR #962, opened 2026-07-07
feat(facility): track badge prints and report yearly gaps

jee7s:
## What & why

Physical ID badges get printed for people, but nothing recorded **that** a badge
was printed, **when**, or **by whom** — so reprints and the yearly badge refresh
were tracked by memory. This adds a durable record of badge prints and a Facility
Ops report answering *"who has / hasn't had a badge printed in year X?"*

Design: `checkin-app/docs/designs/BADGE_PRINT_TRACKING.md`.

- **Mark-printed + report**, not an in-app print/render flow and not derived from
  scans. Recording a print is a deliberate human act, decoupled from generating a
  badge PDF (which already exists on Print ID Badges). Both alternatives are
  written up as rejected in the design doc.
- **New `BadgePrint` model** (`personId`, `printedAt`, `printedById`, optional
  `note`) — a sibling of `RawBadgeLog`. Append-only; multiple rows per person are
  normal (reprints, yearly refresh). The report de-dupes to one row per person
  with the most recent date + a count.
- **Placement: Facility Ops**, not Membership Ops — badge printing is a
  physical-facility concern and a direct sibling of the existing Raw Badge Events
  / Print ID Badges tabs, same `isSysadmin|isBoardMember` gate. (Rationale in the
  doc.)
- **"Needs a badge this year" = anyone who checked in (`Visit`) that year.**
  Defensible from existing data (came to the facility → should have a badge; never
  came → doesn't need one). Read only as a `where` filter, so no `Visit` rows are
  returned and the route auth drift-guard's edge rule doesn't apply.
- **Calendar-year (UTC) window** for v1. The membership-year / timezone-aware
  boundary is written up as the deferred alternative.

## Changes

- `BadgePrint` model + additive migration `20260709020000_badge_print_tracking`
  (new table only — two FKs to `Person`, two indexes; no data change).
- `GET/POST /api/facility/badge-prints` (sysadmin/board): POST marks people
  printed (single + bulk in one call); GET returns the yearly report (printed +
  gap lists).
- New **Badge Prints** tab under Facility Ops: year picker, gap list with per-row
  and bulk mark-printed, and the printed list.
- Regenerated `security/generated/classifications.ts` (BadgePrint field tiers);
  new page/route registered in `pageRegistry` + `authzRegistry`.

## Testing

- Unit: `badgePrints.test.ts` — year-window boundaries (half-open, adjacent
  years, in/out of range).
- Page: `badge-prints/__tests__/page.test.tsx` — renders printed + gap lists,
  single and bulk mark POST the right body.
- Integration: `badge-prints/__tests__/route.integration.test.ts` (7 tests) —
  authz 401/403 (GET + POST), single + bulk mark with the caller as actor, report
  printed/gap correctness across the 2025↔2026 boundary, and reprint/duplicate
  behavior (two rows, one report row, count 2).
- Full pre-commit suite green (1715 tests); `tsc --noEmit` and

## Deferred

Membership-year/timezone-aware boundary, editing/deleting a print row,
registering the report route through the security registry, per-person print
history UI. All noted in the design doc.


jee7s:
**Closed by post-first-release triage — no open issue mapped to this work.** The intent is now tracked as #1363 (project 1, Ideation/Reporting); the branch is retained. If revived: renumber any migration past the 2026-07-11 coalesced baseline and re-verify against current main (LIVE_PERSON drift guard, lifecycle machine, boundary-isolation rule).

PR #963, opened 2026-07-07
feat(my-programs): time-scoped, audited emergency-contact access for leads

jee7s:
## What & why

A program lead mentor needs a family's emergency contacts to do their job — but
only while the program is running, only for *their* program's families, and with
a record that the access happened. Today a lead can pull emergency contacts for
any household in any program they lead, at any time, via the manage roster
(`GET /api/programs/[id]`), with no time limit and no audit.

This adds a dedicated, **time-scoped, audited** lead surface for emergency
contacts, per the product decision.

Design doc: `checkin-app/docs/designs/LEAD_EMERGENCY_CONTACT_ACCESS.md`.

## Decisions

- **Window = program dates ± 7 days.** A lead may view a program's contacts only
  while `now ∈ [startAt−7d, endAt+7d]`. The buffer is a **constant**
  (`LEAD_EC_ACCESS_BUFFER_DAYS`), not config — a fixed policy, reviewed in code.
- **Null dates fail closed.** No dates ⇒ no window ⇒ 403.
- **Clean 403 with an explanation** for off-roster (not the program's lead),
  out-of-window, or null-date requests; the message names the window and the rule.
- **Audit-logged**, one `AuditLog` READ row **per household** whose contacts are
  viewed (not per contact): `action=READ`, `tableName=EmergencyContact`,
  `affectedEntityId=householdId`, `secondaryAffectedEntity=programId`.
- **Board/sysadmin untouched** — they keep `/api/safety/emergency-contacts` and
  its logging; no audit noise added there.
- **New route, not a boundary change.** Uses `withAuth` + explicit field shaping
  (the strict idiom, like `/api/safety` and `/api/household/emergency-contacts`),
  so there is **no `src/security/**` change** and it ships as one feature PR. Why
  a new route over extending `/api/programs/[id]`: only a dedicated endpoint can
  return the clean per-program 403, and time-scoping the shared
  `their_program_households` scope would perturb unrelated consumers and require a
  separate boundary-isolation PR. (Tightening the existing manage-roster EC
  exposure to the same window is called out as deferred in the doc.)
- **Lead only** (not core volunteers), per the decision.

## Data model

No new tables/columns. One additive enum value `AuditAction.READ`
(migration `20260709030000_lead_ec_access`, expand-only `ALTER TYPE ... ADD VALUE`).
`routeAuthDrift` `EDGE_INCLUDE_ALLOWLIST` gains the new route (it reads
`ProgramParticipant` to derive the roster households).

## Surface

A **Contacts** subtab on `/my-programs` lists the caller's led programs; picking
one shows the households + emergency contacts when in-window, or the API's
time-scoping 403 message when not.

## Testing

- **Unit** (`leadAccess.test.ts`): window math — in-window, buffer edges (inclusive),
  just past the buffer, null-dates fail closed, plus the message builder.
- **Integration** (`myProgramsEmergencyContactsAPI.integration.test.ts`, 6 cases):
  lead in-window sees the program's roster households' contacts and **only** those
  (no internal fields leak); out-of-window → 403 w/ message; null dates → 403;
  off-roster → 403; non-lead board member → 403 here but 200 on `/safety`;
  **one READ audit row per household per view** (a 3-contact household → 1 row;
  each view audited).
- `npx tsc --noEmit` clean; `eslint --max-warnings 0` clean; full pre-commit suite
  (1714 tests) green under the shared-DB CI lock.

## Deferred

Tightening `/api/programs/[id]`'s lead EC exposure to the same window (a boundary
PR); core-volunteer access.


jee7s:
**Post-first-release triage: mapped to open issue #1304** (Program-leader time-scoped contact access). Stays open. Review notes before merge: renumber the migration past the 2026-07-11 coalesced baseline; the emergency-contact read surface must respect the security-boundary registry conventions that landed after this branch (see AGENTS.md boundary-isolation rule) — audit whether the new read route needs a registry entry PR first.

thpr (changes requested):
## Requirements review — recommend re-cut

Full write-up (verdict, verified findings, gap list, interview answers) is posted on the owning issue **#1304 (PL4)**: https://github.com/innovationtreehouse/checkin/issues/1304#issuecomment-5082187805

**Verdict: meets-with-gaps → re-cut.** Not a knock on the implementation — the window math, per-household audit, and strict field shaping are clean and survive the re-cut. The problem is the *cut*: this predates the interview-driven designs it now needs to sit inside, and read in that light it solves a narrower problem on a parallel surface that leaves the real exposure open.

**Decisive findings (verified in code):**

1. **Same data, two doors.** `GET /api/programs/[id]` (manage roster) drags the full `Person` row (`.person` include, no `select`) → exposes **parent contact** (`Person.phone/email`) **and** `household.emergencyContacts` — untimed, unaudited, to lead + core-vol. This route time-scopes **EC only** on a side route, so a lead who wants EC out-of-window just opens the roster. The new audit trail is silently routed-around → the timed door is theater while the old door is open. Per the product owner's rule, the inconsistency is fatal to the design as-cut.

2. **A whole axis is unmodeled.** Parent contact (the ±90d "prep/debrief" need) is untouched here and stays untimed on the roster. Target is a **two-window** model — EC ±7d (incl. allergies), parent contact ±90d — on the **existing roster path**, not a parallel route.

3. **Role gate is a regression.** Owner (for safety): lead **+ assistant-lead + core volunteer** should hold in-window parent+EC access; only non-core "other" vols excluded. This route is **lead-only**, which would strip core-vols of access they have today and should keep.

**Recommended shape:** re-cut as a time-scoping of the roster path — two windows, audited, gate lead + assistant-lead + core-vol, null-dates explains why. Likely one design with P27 trip-mode (needs analysis both directions). Do **not** wait on RB8.

**Still open (group decisions, see #1304):** allergy/parent audit scope (PL2), break-glass out-of-window override (RB7), one-design-vs-P27, assistant-lead interim gate (RB6 unbuilt).

Requesting changes so this reshapes before merge rather than shipping the parallel-route cut + lead-only regression.

PR #968, opened 2026-07-08
feat(program-ops): hide non-member price for members-only programs

thpr:
## What

On the program-ops program detail page, hide the **Non-Member Price** field when a program is marked **Treehouse Members-Only** (`orgMemberOnly`). The Member Price then spans full width.

## Why

For a members-only program the non-member price is never charged, so showing the field is confusing.

## Notes

- Purely display-side; no pricing/state logic changed.
- The unlimited-purchase warning at line 395 still keys off `nonMemberPrice`'s stored value even when the field is hidden — harmless (value persists), left as-is.


dkaygithub:
oops, duplicate PR https://github.com/innovationtreehouse/checkin/pull/970/changes

mine has a few more changes  / error conditions it found, whaver works best to get them in

thpr:
@dkaygithub Not duplicate, different file - you dealt with new, I dealt with program/[id]

dkaygithub (approved):

PR #995, opened 2026-07-14
fix(shopify): mint member discounts via GraphQL — REST scope was never granted

jee7s:
## Symptom

Member discount codes were never created in the dev store — silently, because `mintMemberDiscountCode` deliberately never throws (members just got undiscounted checkout links).

## Root causes (three, stacked)

1. **The mint always 403s**: the REST `price_rules` API requires the `write_price_rules` scope, which **neither store's app grants** (live-probed on the dev store: `403 "requires merchant approval for write_price_rules"`, identical on 2026-01 and 2024-07 — not a version issue). Both apps DO grant `write_discounts`, the modern GraphQL discounts scope: the code and the app grants pointed at different APIs.
2. **The flow rarely reached the mint anyway**: zero mint log lines in 14 days of dev logs. The route null-exits silently unless the requester is an ACTIVE org member AND the program has a variant + both prices — dev test data likely never satisfied this (follow up separately: BoardSettings variant id + an ACTIVE test-household membership).
3. **The canary that pins this exact contract has never run**: `shopify-live.yml` skips nightly because the `SHOPIFY_LIVE_ENABLED` repo variable and `SHOPIFY_LIVE_*` secrets were never provisioned.

## Fix

`mintMemberDiscountCode` now uses GraphQL `discountCodeBasicCreate` — works with the already-granted `write_discounts`, escapes the deprecated REST discounts API, and preserves every behavior: code format, 48h window, single-use, once-per-customer, never-throws quiet fallback, and the load-bearing per-unit semantics (`appliesOnEachItem: true` ≙ REST `allocation_method: 'each'` — the multi-child-cart overcharge guard).

Live suite updated to pin the GraphQL-stored contract (`adminGraphql`, `findDiscountByCode`, `trackDiscount` + `discountCodeDelete` cleanup).

## Verified

- 32/32 unit tests (incl. new `userErrors` quiet-fallback case) + mutation-shape assertions.
- **Live canary 2/2 against the real dev store** with real credentials: code minted through the app's actual path, stored discount verified field-by-field, cleaned up after.
- Mutation pre-validated standalone before the rewrite (created + deleted a probe discount).

## Deployment / follow-ups

- **Dev needs no dashboard changes** — `write_discounts` is already granted.
- **Prod app must add `write_discounts`** (+ release/install) before launch, or member pricing silently fails there too — its scope introspection shows no discounts scopes at all.
- Enable the nightly canary: set repo variable `SHOPIFY_LIVE_ENABLED=true` + the `SHOPIFY_LIVE_CLIENT_ID`/`SECRET` secrets.

### DIFF
--- a/checkin-app/shopify-live/discount.shopify-live.ts
+++ b/checkin-app/shopify-live/discount.shopify-live.ts
@@ -2,10 +2,13 @@
  * @jest-environment node
  */
 /**
- * LIVE contract: the member-discount price rule minted for single-pool member
- * checkout. The per-unit pricing of multi-child household carts hangs entirely
- * on allocation_method 'each' semantics (see the #930 review) — this pins the
- * rule configuration Shopify actually stores, not what we think we sent.
+ * LIVE contract: the member discount minted for single-pool member checkout
+ * (GraphQL discountCodeBasicCreate; scope write_discounts — the 2026-07-14
+ * root cause: the legacy REST price_rules mint needed a scope no store app
+ * grants). The per-unit pricing of multi-child household carts hangs entirely
+ * on appliesOnEachItem semantics (REST's 'each'; see the #930 review) — this
+ * pins the discount configuration Shopify actually stores, not what we think
+ * we sent.
  */
 jest.mock("@/lib/prisma", () => ({ __esModule: true, default: {} }));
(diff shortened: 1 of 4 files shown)


dkaygithub:
I think the code being valid for 48 hours needs to be communicated somehow. Not sure where

dkaygithub (approved):

PR #1000, opened 2026-07-14
release gate: at most one migration between tags, as a tested script

jee7s:
## What (refactored per review)

The requirement: **at most one prisma migration between tags**, checked in the production deploy flow. This PR enforces exactly that, plus the data-loss guard the investigation surfaced, refactored for clarity and robustness:

**The gate is now a standalone script** — `checkin-app/scripts/release-migration-gate.sh <released-tag>` — run by the `migration-release-gate` job from the checked-out released tag. The workflow shrinks to a checkout + one script call; the logic is documented, reviewable shell with labeled errors:

- **RULE 1 — at most one migration between tags.** More than one migration directory added since the previous `v*` tag fails the deploy. The error names the remediation: cut an *interim release targeted at the commit after the first migration* — never a pre-release coalesce (that's RULE 2's failure mode).
- **RULE 2 — nothing swept unreleased.** Commit-walks the window (`git log --diff-filter=A`) and fails if any migration merged since the previous release is *gone* at the released tag — a coalesce swallowed unreleased work, which would mark DDL applied on prod without ever running it. Endpoint diffs are structurally blind to added-then-swept; the walk is not.
- **Audit**: the release's new migration list lands in the run summary either way.
- First release: nothing to gate against; passes (v1.0.0 unaffected).

**The gate is itself under test**: `scripts/__tests__/release-migration-gate.test.ts` builds scratch git repos and asserts all six shapes — first release, single migration, RULE 1 block, released-only coalesce pass, RULE 2 block, and version-sorted tag ordering (`v1.2.10` gates against `v1.2.9`). Runs in normal CI, so changes to the gate can't silently break it.

**PR-time early warning stays**: `migration-safety.yml` fails any coalesce PR that sweeps a migration absent from the latest `v*` tag (skips while no tags exist), so the tag rule is caught at review time, not release time.

**Docs**: `MIGRATION_COALESCE_FLOW.md` policy rewritten — ≤1 per release with the interim-release playbook, "The tag rule" (coalesce is post-release hygiene over released migrations only), and the gate description.

## Verified

6/6 gate scenarios green (scratch repos), 25/25 in the scripts test suite, both workflows YAML-validate.

### DIFF
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -63,21 +63,17 @@ jobs:
       ref: ${{ github.event.release.tag_name }}
     secrets: inherit
 
-  # Release gate: MEASURES the schema-migration merges since the previous
-  # release (each was individually replay-tested against a populated DB by
-  # migration-safety.yml at PR time; prisma migrate deploy applies them in
-  # order), and HARD-BLOCKS exactly one shape: a migration that was added
-  # since the previous release but is gone at the released tag — a coalesce
-  # swallowed unreleased work, which would mark DDL applied on prod without
-  # ever running it. Coalescing is post-release hygiene and may only sweep
-  # released migrations — see checkin-app/docs/MIGRATION_COALESCE_FLOW.md
-  # ("The tag rule").
+  # Release gate: AT MOST ONE prisma migration between tags, plus the
+  # swept-unreleased guard — both enforced by
+  # checkin-app/scripts/release-migration-gate.sh (unit-tested on scratch
+  # repos by scripts/__tests__/release-migration-gate.test.ts; policy in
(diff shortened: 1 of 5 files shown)


jee7s:
Rescoped per review: **the contract is equivalence at tags; between tags dev is free** (revert, rework, consolidate — prod applies whatever the tag ships). RULE 2 now fires only on the single artifact that can silently break at-tag equivalence: a `reconcile.sql`-bearing coalesce commit sweeping dirs absent from the previous tag (the ledger rewrite runs no DDL, so it may only stand in for released migrations). Plain reverts and no-reconcile consolidations of unreleased work now pass — covered by two new scratch-repo tests (8/8 green). The PR-time guard in migration-safety.yml carries the same scoping.

PR #1003, opened 2026-07-15
fix(deploy-prod): push to ghcr.io — the fleet left ECR a while ago

jee7s:
## The failure

v1.0.0's **approved** deploy failed at `Log in to Amazon ECR`: `checkin-deploy-prod` is not authorized for `ecr:GetAuthorizationToken` — and that's *correct*: the role's own policy comment says "Images are built/pushed to ghcr.io by the checkin CI… the deploy role no longer needs ecr:*". The prod **task definition already pulls `ghcr.io/innovationtreehouse/checkin-prod`** (verified live: `image: ghcr.io/innovationtreehouse/checkin-prod:bootstrap`, `repositoryCredentials: ghcr-pull-credentials`). The workflow was the last piece of the fleet still pointed at the vestigial ECR registry — latent because this was the first prod deploy ever.

## Fix (mirrors deploy-dev.yml)

- `IMAGE_REPO: ghcr.io/innovationtreehouse/checkin-prod` replaces the ECR registry/repo envs
- `docker/login-action` with `GITHUB_TOKEN` replaces `amazon-ecr-login`; `packages: write` added to permissions
- Both `:sha` and `:<release-tag>` tags still pushed; report step updated

Trade-off note (security review L3): ECR's IMMUTABLE tag setting doesn't exist on GHCR — unique `:sha` tags remain the integrity practice, consistent with the entire fleet (dev, arbor, s-read, bootstrap). Infra follow-up available: delete the now-unused `aws_ecr_repository.checkin` resources.

## After merge

Third cut of v1.0.0 (the release workflow runs from the tag's commit). Validate + gate will re-run; reviewer approves once more; deploy proceeds on GHCR.

### DIFF
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -177,7 +177,12 @@ jobs:
           set -euo pipefail
           # Native arm64 runner -> a plain build produces the arm64 image the
           # ECS task defs require (runtime_platform.cpu_architecture = ARM64).
+          # -f: there is no root Dockerfile — the app image builds from
+          # checkin-app/Dockerfile with the repo root as context, exactly like
+          # deploy-dev.yml (the release-scrub found this job would die here).
           docker build \
+            --build-arg NEXT_PUBLIC_GIT_SHA="${{ steps.vars.outputs.sha }}" \
+            -f checkin-app/Dockerfile \
             -t "$IMAGE" -t "$IMAGE_TAGGED" .
           docker push "$IMAGE"
           docker push "$IMAGE_TAGGED"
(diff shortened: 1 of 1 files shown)


jee7s:
Scrubbed the whole deploy job per review — three more guaranteed first-run failures found and fixed on this branch: (1) `docker build` lacked `-f checkin-app/Dockerfile` (no root Dockerfile exists — the build dies preparing context; also added the `NEXT_PUBLIC_GIT_SHA` build-arg for parity with dev); (2) `update-service` never set `--desired-count 1` — the service is staged at 0 by infra design, and a 0/0 service is instantly 'stable', so the run would have gone green while serving nothing; (3) no post-rollout smoke existed at all — added a real retried probe of the deployed APP_URL that fails the run unless production answers. Audited clean: capacity provider name/strategy vs the live cluster, the host/EC2 migrate run-task pattern, log group/stream names, host port 4000 vs wiki, DDL secret wiring. One capacity note for later: with the app (512 MiB) running, the instance has ~12 MiB free — the NEXT release's migrate task (512 MiB) won't place without a second instance or a smaller migrate reservation; flagging as an infra follow-up before v1.0.1.

thpr (approved):

PR #1025, opened 2026-07-16
feat(membership-ops): click a status to filter the applications board

jee7s:
## What

On the membership-ops **applications board**, clicking any status — the count-pills in the legend or a row's status badge — filters the list to that status; clicking again (or the "Showing only X — Clear filter" notice) restores everything. The BLOCKED summary alert is clickable as a shortcut to the BLOCKED filter.

- Shared primitive (`src/components/StatusFilter.tsx`): `useStatusFilter` keeps the active status in a `?status=…` query param (survives refresh, linkable, `replace` not `push`); `StatusFilterBadge` renders the badge as a real button (filled + `aria-pressed` when active); `ActiveFilterNotice` is the visible way out.
- The status **counts and the BLOCKED alert stay computed from the unfiltered rows**, so the legend always shows the whole picture while the list narrows. Filtered-to-empty gets its own message + clear action. The page gained the repo's standard `Suspense` wrapper for `useSearchParams`.

## Scope note (review + maintainer direction)

An earlier revision also wired the households page; that's been **removed** — per maintainer direction this PR covers the applications tab only, and per review feedback the filtering-pill pattern should be introduced consistently (count-pills like this page's legend) rather than as per-page variants. Extending pills to households (Member / Non-Member counts) and converting program-ops' radio filters can follow as a separate, consistent pass.

## Verification

Primitive: 4 unit tests (toggle/clear/param preservation). Page: narrowing + restoring, counts staying unfiltered, empty-state with clear action. tsc + lint clean; suite green on the branch.

### DIFF
--- a/checkin-app/src/app/membership-ops/applications/__tests__/page.test.tsx
+++ b/checkin-app/src/app/membership-ops/applications/__tests__/page.test.tsx
@@ -1,13 +1,43 @@
 /* eslint-disable @typescript-eslint/no-require-imports -- jest.mock factories are hoisted above imports */
 import { screen, fireEvent, waitFor } from "@testing-library/react";
-jest.mock("next/navigation", () => require("@/test-helpers/rtl").navMock());
+
+// Reactive next/navigation mock (mirrors src/components/__tests__/StatusFilter.test.tsx):
+// router.replace writes back into `search` so a `rerender()` after a click picks up
+// the new useSearchParams() value, same as a real app-router navigation would.
+let search = "";
+const replace = jest.fn((url: string) => {
+    search = url.split("?")[1] ?? "";
+});
+jest.mock("next/navigation", () => ({
+    useRouter: () => ({ replace }),
+    usePathname: () => "/membership-ops/applications",
+    useSearchParams: () => new URLSearchParams(search),
+}));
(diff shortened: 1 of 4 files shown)


thpr (changes requested):
This adds 2 UI sorting behaviors to one that already exists.  This isn't good as is, IMHO.

If the member filter is to exist, let's add pills to the top of the Household page that adds Member and Non-Member count.  That way the pill filtering is at least consistent between these two use cases.  It also takes away the conflict with pages like Facility-ops/Visits that has auto-sorting on column names (which should probably be added to other pages like the two touched here).

The third filter method is program-ops/programs, which uses radio buttons to do filtering.  Since those two radio buttons are also part of a column (Access), those can ALSO be converted to pills in order to do filtering (and more pills could be added to that page as well if desired)

jee7s:
@thpr scope reduced to the applications board only — the households wiring is removed (also per maintainer direction), so this PR no longer introduces a second filter surface. The applications legend already uses the count-pill pattern you described; bringing the same pills to Households (Member / Non-Member counts) and converting program-ops' Access radios to pills makes sense as its own consistent pass rather than riding this PR. Re-requesting review.

thpr (approved):

PR #1026, opened 2026-07-16
fix(ui): restore breathing room above tab bars; normalize double-padded pages

jee7s:
## What

The attendance tab bar ("Current | Add Past Visit | Household Check-ins") — and to a lesser degree every section's tab bar — rendered flush under the fixed 60px header (user screenshot, prod `/attendance/current`). Two audits, five changed lines:

- **Shared system** (`PageContainer.tsx`): `marginTop: -8` → `0`, so all 12 section layouts and every standalone `PageContainer` page get the AppShell's full 16px of space under the header. The comment keeps the #1007 incident lesson (a negative margin exceeding the AppShell padding clips content behind the header).
- **`attendance/current`**: its non-kiosk branch carried its own stale `marginTop: -25` — the *exact* overshoot constant from the #1007 incident, surviving in a pre-`PageContainer` copy. Now `0`. This is the bar in the screenshot.
- **Double-padding normalization**: `safety/pickup`, `safety/trusted-adults`, and top-level `trusted-adults` stacked their own `p="md"` on the AppShell's 16px (32px total, inconsistent with sibling pages) — dropped.

## Audit coverage

Both agents produced full audit tables (in the commit messages): all 12 section layouts, every `PageContainer` page, every page-level tab bar / heading, kiosk & board-display surfaces (deliberately skipped), and mid-page sub-tab bars (not page nav — untouched). No other inconsistencies found; the `mb="md"`-vs-`Stack` gap split is the documented `SectionTabs` pattern, not drift.

## Verification

Full unit suite on the integrated branch: 1732 passed; tsc and lint clean. No test encoded the old spacing values.

### DIFF
--- a/checkin-app/src/app/attendance/current/page.tsx
+++ b/checkin-app/src/app/attendance/current/page.tsx
@@ -10,6 +10,7 @@ import {
   SimpleGrid, Stack, Text, TextInput, Title,
 } from "@mantine/core";
 import { formatTime, isYouth } from "@/lib/time";
+import { PageContainer } from "@/components/ui/PageContainer";
 import { formatPhone } from "@/lib/phone";
 import { getKioskDisplayNames } from "@/lib/kiosk-names";
 import { notifyNavRefresh } from "@/lib/nav-refresh";
@@ -338,13 +339,9 @@ function KioskDisplayInner() {
 
   const userId = (session?.user as SessionUser)?.id;
 
-  return (
-    <Box style={isKioskMode ? { cursor: "none", marginTop: -25, paddingLeft: 8, display: "flow-root" } : { maxWidth: 1200, margin: "0 auto", marginTop: 0, paddingLeft: 8, display: "flow-root" }}>
-      {!isKioskMode && (
-        <Box style={{ maxWidth: 1200, margin: "0 auto" }}>
-          <AttendanceTabs />
-        </Box>
-      )}
+  const pageBody = (
+    <>
+      {!isKioskMode && <AttendanceTabs />}
       <Card withBorder radius="md" padding="lg" style={{ width: "100%", maxWidth: 1200, margin: "0 auto" }}>
         {/* Check-in button — hidden in kiosk mode */}
         {!isKioskMode && (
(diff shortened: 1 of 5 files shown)


thpr (changes requested):
This is not consistent across the app.  It's supposed to be using the PageContainer file modified in PR 1007

jee7s:
@thpr agreed — reworked so the spacing flows through PageContainer everywhere:

- **attendance/current**: the non-kiosk branch was carrying a verbatim copy of PageContainer's style object (down to the `flow-root`), with the stale `-25` margin from before #1007. It now renders inside the real `PageContainer`; kiosk mode keeps its own fullscreen Box deliberately (cursor hidden, board-display chrome). Plain conditional render, so no per-render wrapper remounts.
- **trusted-adults** (standalone page, no section layout): adopts `PageContainer` like profile/index/my-household, instead of leaning on bare AppShell padding.
- **safety/pickup + safety/trusted-adults**: no change from before — these already sit inside `safety/layout.tsx`'s `PageContainer`; the removed `p="md"` was double-padding *against* the container, so dropping it is the PageContainer-consistent behavior.
- The one-line `marginTop: -8 → 0` in `PageContainer.tsx` itself is unchanged — that file stays the single owner of the top-spacing decision.

Full suite green (1732). Re-requesting review.

thpr:
We should chat.  I intentionally went to -8 because things are already too low and wasting space on what feels like a small screen with the amount of info we have.  So consistency good, 0 vs -8? not sure

thpr:
I'm approving this only to let fixes go through.  Still want to discuss -8 vs 0 to avoid wasting UI space

thpr (approved):

jee7s:
@thpr happy to chat — and the good news is this PR moved the debate into exactly one number in one file: `PageContainer.tsx` now owns the top gap outright (the attendance page's private copy of it is gone), so whatever we settle on is a one-character follow-up that changes every page at once.

On the tradeoff itself: -8 buys 24px of vertical space but was reading as 'flush under the header' on the attendance tabs (the screenshot that kicked this off), especially with the tab underline touching the header seam. If 16px feels too generous on dense ops screens, a middle setting (-4 → 12px effective) might read as 'attached but breathing'. I'd also note the double-padded pages this PR normalized (32px → 16px) recovered more vertical space than the -8→0 change spends on most screens. Your call + jee7s's — say the number and I'll ship it.

PR #1027, opened 2026-07-16
feat(membership): payment holdoff message after Shopify checkout

jee7s:
## Summary

Adds a payment holdoff on the membership page's PENDING_PAYMENT card: after clicking the real Shopify checkout link ("Pay here with Shopify →", not the local mock), the page replaces the pay button with "We'll update your status here when we receive your payment — refresh this page after you finish checkout." plus a subtle "Show the payment button again" escape hatch.

## Clearing conditions

The holdoff clears on:
1. **The Shopify `orders/paid` webhook settling the payment** — handled with **no backend change**, resolved at page-load time: the page's existing mount-time `load()` refetches `/api/membership`; once the webhook has updated the process's status server-side, the next page load renders the card in its real state (e.g. "Payment received 🎉" / background-check-pending) and an effect watching `state.process.status` drops the holdoff's `sessionStorage` record.
2. **Any other server-side path** (e.g. a future s-read reconciliation) — satisfied the same way as (1): the clearing logic only cares that the *status* changed, not which server-side path changed it. Called out in a code comment for when that reconciliation path is built.
3. **The "Show the payment button again" escape hatch** — a subtle link under the message that restores the pay button immediately, covering an abandoned checkout.

## No background polling (scale-to-zero friendly)

Earlier revisions polled `/api/membership` every ~10 s while the message showed; per review, that would keep the scale-to-zero dev instance and its database awake for as long as a tab sat open. Polling is removed entirely — no intervals, no focus/visibility listeners. All checks happen at page load only, so an idle tab generates zero requests. Consequences reflected in the UI: the Loader spinner is gone (nothing is actively watching) and the copy tells the member to refresh after checkout.

(Also removed earlier at the maintainer's direction: the original 10-minute timeout + "we haven't seen your payment yet" note — the escape hatch covers the abandoned-checkout case.)

## Other design notes

- Only the real Shopify anchor (`payment.checkoutUrl`) starts the holdoff. The local-dev mock path (`settleMockPayment`, which settles synchronously via an in-app webhook simulation) is untouched and never shows it.
- The holdoff is persisted to `sessionStorage` keyed by process id, so a tab refresh mid-checkout restores the message instead of flashing the pay button back. The record clears on a page load that finds the status changed, or via the escape hatch.
- The scholarship/payment-plan request section and the background-check status alert stay visible under the holdoff — only the pay-button area is replaced.

## Test coverage

In `checkin-app/src/app/membership/__tests__/page.test.tsx`:
- Clicking the checkout link sets the holdoff message and the `sessionStorage` record.
- A page load (remount) while still pending restores the message from `sessionStorage`; a page load after the payment settled renders the paid-state card and clears the record.
- The "Show the payment button again" escape hatch restores the pay button immediately and clears `sessionStorage`.
- The local mock payment path never shows the holdoff.

Verification run from `checkin-app/`: page suite (120/120), `npx tsc --noEmit` (clean), `npm run lint` (clean, 0 warnings), full `npx jest` (1784/1784 across 213 suites, after merging latest main into the branch).

### DIFF
--- a/checkin-app/src/app/membership/__tests__/page.test.tsx
+++ b/checkin-app/src/app/membership/__tests__/page.test.tsx
@@ -4,7 +4,7 @@ jest.mock("next/navigation", () => require("@/test-helpers/rtl").navMock());
 jest.mock("next-auth/react", () => require("@/test-helpers/rtl").authMock());
 jest.mock("@mantine/notifications", () => ({ notifications: { show: jest.fn() } }));
 
-import { screen, fireEvent, waitFor, act } from "@testing-library/react";
+import { screen, fireEvent, waitFor } from "@testing-library/react";
 import { notifications } from "@mantine/notifications";
 import { renderWithProviders, mockFetchJson, setSession, setCheckinEnv, resetRtl } from "@/test-helpers/rtl";
 import MembershipPage from "../page";
@@ -354,6 +354,8 @@ describe("membership page", () => {
   });
 
   // ── PENDING_PAYMENT payment holdoff ──────────────────────────────────────────
+  const HOLDOFF_MSG = "We'll update your status here when we receive your payment — refresh this page after you finish checkout.";
+
   it("clicking the Shopify checkout link shows the holdoff message and persists it to sessionStorage", async () => {
     setSession({ id: 1 });
     mockFetchJson({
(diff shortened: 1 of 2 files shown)


thpr:
I don't see a use for item 2 if item 3 is implemented.  s-read-function runs what - once a day?  Item 3 will "always" win (give or take epsilon)

jee7s:
@thpr the ordering concern is resolved by the latest revision: the 10-minute timer (item 3) is **removed** at jee7s's direction — the escape-hatch link covers the abandoned-checkout case, so nothing races the reconciliation anymore. The message now clears only when the process status actually changes (or the member clicks the escape hatch), which makes item 2 the useful fallback for a missed webhook rather than dead code. Small correction on cadence: s-read runs **hourly**, not daily — so a missed webhook would self-heal within ~an hour once a reconciliation path exists (it doesn't yet; the PR only polls status, so it inherits whatever settles it server-side).

thpr (approved):

thpr:
Approved, but... we're waking the database every hour? for 15 minutes? that doesn't seem very scale-to-zero?

jee7s:
@thpr Polling is removed as of 10bc37ad, per jee7s's direction. The holdoff message now resolves at page load only — the existing mount-time fetch handles it, with no intervals or focus listeners — so an idle tab generates zero requests and scale-to-zero/idle-DB behavior is unaffected. The Loader spinner is gone too, and the copy now tells the member to refresh after checkout.

PR #1028, opened 2026-07-16
feat(membership-ops): renewal-season "Grant for coming year" button

thpr:
## What

Adds a **Grant for coming year** action to the households board ([membership-ops/households](checkin-app/src/app/membership-ops/households/page.tsx)), shown only during renewal season, in the same button column below grant/revoke. Works for both existing members and non-members.

It is an **admin override**: one click sets the household's membership `ACTIVE` and creates a *completed* `RENEWAL` process (or `INITIAL`, for a household with no prior membership) — the same end-state a member's finished renewal produces — skipping the sign / pay / background-check flow. The acting admin is recorded (`certifiedById` + audit row with `comingYearOverride: true`).

## Why

Grant/revoke cover the normal case, but during renewal season the board needs a way to renew a household for the coming year (or bring a new family on for it) without walking the full external flow.

## How it works

- **"Renewal season" is derived server-side** from the existing `BoardSettings.orgMembershipYearBoundary` + 2-month lead window — the same window the renewal cron/reminders use. Returned as a `renewalSeason` flag on `GET /api/membership-ops/households`; the button only renders in season.
- The window calc was **duplicated** in `runRenewalSweep`; extracted a pure `renewalWindow()` helper that both it and the new `isRenewalSeason()` share.
- **Conflict-of-interest guard matches Grant**: a board member cannot grant their own household (bypasses payment + BG); a sysadmin can. Enforced server-side, disabled in UI.

## Notes for reviewers

- **No schema change / no migration.** `OrgMembership` has no year field, so "renewed for the coming year" is expressed purely through the process record (membership ACTIVE + terminal process). The three gate timestamps (`contractSignedAt`/`bgClearedAt`/`paidAt`) are stamped to mirror a normally-completed process; the audit row flags it as an override so the truth is preserved.
- The override creates a terminal (`ACTIVE`) renewal process, so the cron sweep could still open a fresh `PENDING_RENEWAL` later in the same window — but that is **identical to how an early manual renewal behaves today**, not a new bug. Flagged in case re-nagging is worth fixing in the sweep separately.
- Staff-household block on the button stays UI-only, matching today's Grant path.

## Tests

- Unit: `isRenewalSeason` window boundaries; button show/hide by season, both household kinds, COI disable.
- Integration: the override write (ACTIVE + terminal `INITIAL` process + audit row) and the own-household refusal. Ran green against a throwaway Postgres.

### DIFF
--- a/checkin-app/src/app/__tests__/householdsMembershipAuditAPI.integration.test.ts
+++ b/checkin-app/src/app/__tests__/householdsMembershipAuditAPI.integration.test.ts
@@ -30,6 +30,7 @@ describe('POST /api/membership-ops/households — grant/revoke audit logging', (
     let boardHouseholdId: number;
     let inactiveHouseholdId: number;
     let activeHouseholdId: number;
+    let comingYearHouseholdId: number;
 
     beforeAll(async () => {
         const leaked = await prisma.person.findMany({
@@ -57,6 +58,12 @@ describe('POST /api/membership-ops/households — grant/revoke audit logging', (
         });
         activeHouseholdId = active.householdId;
         await prisma.orgMembership.create({ data: { householdId: activeHouseholdId, status: 'ACTIVE' } });
+
+        // A non-member household for the renewal-season "grant for coming year" override.
+        const comingYear = await prisma.person.create({
+            data: { email: `comingyear-${TAG}@example.com`, name: 'Coming Year', household: { create: { name: "Test HH" } } },
+        });
+        comingYearHouseholdId = comingYear.householdId;
     });
 
     afterAll(async () => {
@@ -64,6 +71,11 @@ describe('POST /api/membership-ops/households — grant/revoke audit logging', (
             where: { email: { contains: TAG } }, select: { id: true, householdId: true },
         });
         await prisma.auditLog.deleteMany({ where: { actorId: { in: ids.map(u => u.id) } } });
+        const memberships = await prisma.orgMembership.findMany({
+            where: { householdId: { in: ids.map(u => u.householdId) } }, select: { id: true },
+        });
+        // Processes FK onto membership with onDelete Restrict — clear them first.
+        await prisma.orgMembershipProcess.deleteMany({ where: { orgMembershipId: { in: memberships.map(m => m.id) } } });
         await prisma.orgMembership.deleteMany({ where: { householdId: { in: ids.map(u => u.householdId) } } });
         await prisma.person.deleteMany({ where: { email: { contains: TAG } } });
         await prisma.household.deleteMany({ where: { id: { in: ids.map(u => u.householdId) } } });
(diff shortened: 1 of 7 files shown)


jee7s:
Reviewed (one pass, comment-only by design — not approving/queueing).

**One real gap: an in-flight process survives the override.** If the household is mid-flow when the button is clicked — a `PENDING_RENEWAL` from the sweep, or a renewal/application sitting at `PENDING_EXTERNAL_ACTION`/`PENDING_PAYMENT` — the override creates its terminal process *alongside* it. Three consequences: the member's `/membership` page keeps showing the stale in-flight flow (the status-driven cards and `latestPendingExternal` don't know about the new terminal row); a stale `PENDING_PAYMENT` remains payable, and `activate()` would happily settle it later (second terminal process + a charge for a household the board already granted); and reviewer queues can keep listing a review that no longer matters. Suggest: inside the same transaction, close any open process for that membership (flip to `ACTIVE` with a `supersededByOverride: true` audit note, or refuse the override with a clear "household is mid-flow, resolve that first" error — either is defensible, silent coexistence isn't).

**On the sweep re-nag you flagged** — agreed it's pre-existing (an early-completed renewal gets re-opened + re-reminded today too), but this button will make it common, and your `renewalWindow()` extraction makes the fix cheap: the sweep's "already opened this cycle" check can also skip memberships with a **terminal** RENEWAL whose `stageEnteredAt >= windowStart`. Happy either here or as a follow-up.

Things I checked that are right: `renewalWindow()` is semantics-preserving for the sweep (the `>=`/`<` flip is equivalent, boundary inclusive as before); **not** stamping the leads' `lastBackgroundCheck` is the correct honesty — next season's freshness check still routes them through the real background check flow; COI guard + `certifiedById` + the `comingYearOverride` audit flag match the `certifyPaymentPlan` conventions; and the terminal process stays out of the one-inflight-renewal partial index by construction.

thpr:
Thanks — both addressed in b7bb4c03.

**In-flight process survives the override (the real gap):** fixed. The override now, in the same transaction, ARCHIVEs every non-terminal process for that membership before writing the ACTIVE override — went with disposal over refusal so the button stays one-click. `ARCHIVED` (the same terminal status `archiveApplication` uses) drops the stale row off the `/membership` cards, the reviewer queue, and — key for your payable-`PENDING_PAYMENT` concern — out of `activate()`'s `status === "PENDING_PAYMENT"` set, so no second terminal process and no stray charge. Audited per superseded row with `supersededByOverride: true`. A prior cycle's `ACTIVE` / an earlier `ARCHIVED` is left untouched.

**Sweep re-nag:** did it here since it's cheap now. `runRenewalSweep`'s "handled this cycle" check now also matches a terminal (`ACTIVE`/`ARCHIVED`) RENEWAL with `stageEnteredAt >= windowStart`, not just in-flight rows — so both the override and an early manual finisher stop getting re-opened + re-reminded. `renewalWindow()` returns `windowStart` for the predicate.

Coverage added: integration test for the supersede (in-flight `PENDING_PAYMENT` → `ARCHIVED` + fresh `ACTIVE` RENEWAL) and for the sweep skipping a household already renewed this cycle.

jee7s (approved):

PR #1042, opened 2026-07-16
fix(s-read): read the mirror's naive timestamps as UTC, not process-local

thpr:
## What

Every timestamp `lib/shopifyRead/client.ts` reads out of the s-read mirror is now anchored to UTC explicitly — on both the SELECT projection and the reconciler's cursor comparison.

**This is the order-read half of the bug #1041 fixed on `sync_run`.** #1041 pinned `started_at`/`finished_at`; the order reads in the same file had the identical latent bug on `shop_order.updated_at` and `cancelled_at`. Rebased on top of #1041, and the tests join the describe block in the file it created.

## Why

The mirror's timestamp columns are `timestamp WITHOUT time zone` holding UTC — Prisma's default `DateTime` mapping. `shop_order.updated_at` and `cancelled_at` are both `TIMESTAMP(3)` in `packages/s-ingest-core/prisma/migrations/0000000000000_init/migration.sql`.

node-pg resolves a naive `timestamp` against the **node process's** timezone, not UTC. So on any instance where `TZ != UTC`, every mirror timestamp read through this bridge came back shifted by the local offset. #1041 proved this concretely by driving the real page on a Chicago-zoned laptop.

There were two independent offsets in `ordersChangedSince`:

1. **The SELECT projection** (shared with `ordersByLegacyIds` through `ORDER_COLS`) — `updated_at`/`cancelled_at` parsed against the process TZ.
2. **The WHERE** — a naive column compared against a `timestamptz` param, which Postgres resolves against the **session** TimeZone. A second offset, unrelated to node's.

Both matter because `ordersChangedSince` is the reconciler's high-water cursor: `lib/finance/reconcile.ts` reads `updatedAt` off a row, stores it, and feeds it back as `since`. A shifted read skips or re-scans orders. `cancelled_at` separately feeds the Live payment column on `/finance-ops/payments`.

## How

`AT TIME ZONE 'UTC'` on the projection re-labels the naive value as the UTC instant it already is, producing a `timestamptz` that node-pg parses correctly at any TZ.

The WHERE casts the **param** side back to naive-UTC instead of casting the column:

```sql
WHERE test = false AND ($1::timestamptz IS NULL OR updated_at > ($1::timestamptz AT TIME ZONE 'UTC'))
```

Both directions are correct; the param side keeps the comparison naive-vs-naive and sargable. `shop_order` has no index on `updated_at` today (only `store_id,legacy_id` and `store_id,name`), so that's insurance rather than a live win. Both sides had to move together — casting one alone would shift the cursor by the offset.

## Not a live bug

Deployed ECS tasks set no `TZ` env var, so the container defaults to UTC and the naive read is accidentally correct. Nothing in `infra modules/checkin/compute.tf` sets `TZ`. Same standing as #1041: correctness-by-accident, converted to a guarantee.

## Reviewer notes

- **Mirror schema untouched.** `packages/s-ingest-core` is s-read's; checkin only reads.
- **`disputedOrderGids` needs no change** — it selects no timestamps.
- **The `ORDER_COLS` doc comment defers to #1041's,** which already spells out the node-pg mechanic in full a few lines down. One explanation, not two.
- **The guards are verified load-bearing:** reverting the three casts fails exactly the three cast assertions and leaves #1041's six `latestSyncRun` tests green. Without that check the assertions prove nothing — everything here runs at TZ=UTC, so a round-trip test would pass either way. That's why these pin SQL text.
- **Conflict resolution with #1041** touched only the test file (`client.ts` auto-merged cleanly). I left #1041's shared harness and setup exactly as-is and followed its per-test `mockResolvedValue` idiom rather than adding a `beforeEach` default that would have quietly changed behaviour for its six tests.

## Testing

- `shopifyRead|config|payments` — 60/60 pass, including #1041's suites.
- Mutation-checked as described above.

### DIFF
--- a/checkin-app/src/lib/shopifyRead/__tests__/client.test.ts
+++ b/checkin-app/src/lib/shopifyRead/__tests__/client.test.ts
@@ -6,6 +6,10 @@
  * the wiring the reconciler's integration tests don't: that an unconfigured env never
  * opens a pool at all, and that the read is a single newest-first row.
  *
+ * The order reads below are guarded the same way, and for the same reason: both pin
+ * the shape of the SQL rather than a round-tripped Date, because the round trip
+ * passes either way under TZ=UTC (which is what CI and the deployed tasks both run).
+ *
  * The client caches its Pool in a module-level singleton, so each case re-imports the
  * module under jest.isolateModules to get a fresh one.
  */
@@ -39,7 +43,7 @@ function freshClient(): Client {
 const MIRROR_URL = 'postgresql://checkin_dev_dml:pw@host:5432/shopify_read_dev';
 
 // Both inputs of the resolved url. SHOPIFY_READ_DB must be cleared too, or an ambient
-// value derives a url and the "unwired" case below can never actually be unwired.
+// value derives a url and the "unwired" cases below can never actually be unwired.
 const MIRROR_KEYS = ['SHOPIFY_READ_DATABASE_URL', 'SHOPIFY_READ_DB'] as const;
 const prev: Record<string, string | undefined> = {};
 for (const k of MIRROR_KEYS) prev[k] = process.env[k];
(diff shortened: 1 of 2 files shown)


jee7s (approved):

thpr:
Rebased onto `origin/main` (now at #1043, `feat(mirror): derive the s-read mirror URL from DATABASE_URL + SHOPIFY_READ_DB`). Clean rebase — #1043's `application_name` addition and this PR's `ORDER_COLS` cast touch different parts of `client.ts`.

**Heads up on a collision with #1041** (currently in the merge queue): it creates `src/lib/shopifyRead/__tests__/client.test.ts` — the same path this PR adds. An add/add conflict is unavoidable, so I've rewritten this PR's tests onto #1041's harness (mocked `pg` + `poolCtor` spy + `jest.isolateModules` to defeat the client's Pool singleton) rather than keeping a second parallel one that mocked `@/lib/config`. Resolution when #1041 lands should be "keep both describe blocks", not a rewrite. Its harness is also the better one — it drives the real env derivation instead of stubbing the accessor.

Worth noting #1041 fixes the *same class* of bug on `sync_run.started_at`/`finished_at`; this PR covers `shop_order.updated_at`/`cancelled_at`. Neither supersedes the other.

**Verified the guards are load-bearing:** reverting the three casts in `client.ts` fails exactly the three cast assertions (the unwired-pool case still passes, as it should). Without that check the test text proves nothing — everything here runs at TZ=UTC.

**Test status**
- `npm test -- --testPathPatterns "shopifyRead|config"` — 52/52 pass (config suites included since #1043 reworked the accessor these tests reach).
- `tsc --noEmit` — 458 errors, all pre-existing: this worktree has no `src/generated/prisma/client` (never ran `prisma generate`). Zero in `shopifyRead`.
- The reconciler's integration matrix is still **unrun** — no `.env*` in this worktree, so `INTEGRATION_DB=1` has no Postgres to reach. Unchanged from the original description; still worth a run against a real DB before merge.

PR #1053, opened 2026-07-17
feat(membership-ops): valid-until dates + BG expiry + grant button only when grantable

jee7s:
Sibling of #1052 (the grant-behavior fix). This is the ops-views half of the renewal batch — problems 2 + 3 plus the BG-expiry addition.

## What's here

**"Valid until" as a settings field.** New `BoardSettings.currentMembershipValidUntil` (single nullable column — the only migration), edited in membership settings as "Current membership applications are valid until" beside the year boundary. The households list and detail views show it as **Valid until** next to **Member since** (per-household, from `OrgMembership.memberSince`).

**Background-check "valid until", in renewal-year terms.** `bgValidUntilBoundary(lastBackgroundCheck, settings)` — the membership-year boundary occurrence that follows `lastBackgroundCheck + bgRecheckMonths`. Null (not shown) when there's no check, the recheck policy is unset, or no boundary is configured. The raw expiry is day-truncated before the boundary snap so a check expiring **on** the boundary day still covers that boundary (`gte` parity with `householdBgIsFresh`). Shown per member on the detail view (including non-lead PERSON_BG subjects) and at household level — the **later** check among household leads — on both list and detail.

**Grant button only when grantable.** The list's `renewalGrantable` uses the exact same gate as `grantRenewalPayment` in #1052 — in-flight `RENEWAL` at `PENDING_PAYMENT` with `bgClearedAt` set (one filtered include, no N+1) and `householdBgIsFresh` — so the button appears only when the server would say yes. Hidden for: sweep-opened `PENDING_RENEWAL` (member hasn't started), earlier flow stages, already granted (the renewal left `PENDING_PAYMENT`), stale/absent BG, staff households, own-household COI.

## Testing

- Helper: 6 unit cases (mid-cycle, boundary-day edge, year-crossing, recheck=0, null check, no boundary).
- Integration: settings round-trip; GET grantability/date assertions with **spaced fixture dates** — the spec's example dates all collapsed to one boundary and couldn't discriminate wrong-member/wrong-boundary logic, so the fixtures were chosen to cross different boundary occurrences; two-leads-later-wins; non-lead shown per-member but excluded from the household value.
- Full unit CI 1854/1854; full integration 1025/1025; `tsc --noEmit` + eslint clean.

## Review notes

- Merge order vs #1052 is free — they touch different regions of `households/route.ts` (GET here, POST there) and `renewal.ts` (helpers near `nextBoundary` here, `grantRenewalPayment` there); whichever lands second may need a trivial rebase, no semantic interaction. Note `renewalGrantable` is UI-only — #1052's server gate holds regardless of what this PR renders.
- `security/generated/classifications.ts` is regenerated (new field registered `public`, same as the boundary field).
- One migration (`20260717120000_membership_valid_until`), additive nullable — within the ≤1-per-database budget alongside… nothing else: #1052 carries none.

### DIFF
--- a/.github/actions/deploy-s-read/action.yml
+++ b/.github/actions/deploy-s-read/action.yml
@@ -1,0 +1,202 @@
+name: Deploy s-read
+description: >
+  Build the s-read image, register new revisions of BOTH task-def families, and
+  (in migrate-and-code mode) run the migrate task in between — for ONE env. The
+  runnable half of what deploy-s-read.yml used to inline. The CALLER must have
+  already checked out the ref to build (this action does not checkout — it builds
+  from the current workspace) and must run in a job whose OIDC sub the target
+  deploy role trusts: for prod that means `environment: production`, which is
+  exactly why folding this into the app deploy job collapses the release to a
+  SINGLE approval (one job enters `production`, not two).
+
+inputs:
+  environment:
+    description: "s-read environment (dev | prod)"
+    required: true
+  mode:
+    description: "migrate-and-code | code-only"
+    default: migrate-and-code
+  network-config:
+    description: >
+      run-task networkConfiguration JSON (the S_READ_NETWORK_CONFIG repo var).
+      Required for migrate-and-code; unused in code-only.
+    default: ""
+  aws-region:
+    description: "AWS region"
+    default: us-east-2
+
+runs:
+  using: composite
+  steps:
+    - name: Resolve commit
+      id: vars
+      shell: bash
+      run: echo "sha=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
+
+    - name: Configure AWS credentials (OIDC)
+      uses: aws-actions/configure-aws-credentials@v6
+      with:
+        # Per-env deploy role. The prod role's OIDC trust requires the calling
+        # job to carry `environment: production` (sub …:environment:production);
+        # the dev role's trust is the bare main-ref sub. When this action runs
+        # inside the prod app-deploy job, the job already assumed this same role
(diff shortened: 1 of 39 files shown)


thpr (changes requested):
Uhh, this requires we constantly update settings.  The button in membership-ops for "grant for coming year" vs "grant" also then shouldn't necessarily be needed with this? Also when is this date changed?  This adds complexity I think we need to redesign around, there are hardcoded 60-day warnings on membership renewal, etc. that this also interacts with... there's a better solution here, and I don't think this schema change captures the right design point

jee7s:
Reworked per your review — the schema change is gone. "Valid until" is now **derived, per household, from the year boundary alone** (membership = exactly one year): an ACTIVE membership is valid until the *upcoming* boundary occurrence, and a household already settled for the coming cycle (your `settledForComingYear` probe from #1047) shows the boundary after that. Nothing to hand-update, no defined-update-moment problem, and it can't drift from the renewal sweep or the 60-day warnings because they all read the same boundary. The PR now carries **no migration**; the settings page/API are untouched relative to main. `BG valid until` was always boundary-derived and is unchanged.

On the button question: agreed it needed rethinking — as of the #1047-composed update, "Grant for coming year" is three-state (hidden unless actionable-or-settled; disabled **Granted for coming year** when settled — your design; live only when the renewal is at PENDING_PAYMENT with a valid cleared background check), and #1052 makes the action itself a payment completion rather than a gate bypass, so the grant-vs-grant distinction now reflects a real state-machine difference rather than a parallel path.

thpr (approved):

PR #1060, opened 2026-07-17
feat(programs): stale-while-revalidate cache for the anonymous public directory

jee7s:
Split out of #1059 per review — the cache half of the "programs directory during an Aurora resume" fix (the retry-predicate half and the generic wait message stay in #1059; this PR composes with either merge order, no shared files).

## What it does

`lib/staleCache` — a small in-process stale-while-revalidate helper, applied to the **anonymous** programs-directory query with a 60-second fresh window:

- **Fresh hit** → cached value, no DB touch.
- **Expired hit** → cached value **instantly**, plus exactly one deduped background revalidation (concurrent expired reads share it).
- **Failed compute** → the stale value at **any** age — this is the Aurora-resume case: the directory renders from the last-good copy while the DB wakes, seat counts at most one refresh stale.
- **Cold miss** → computes normally (and rides #1059's fixed retry extension if the DB is resuming).

## Why anonymous-only

The GET varies by session (board/sysadmin see drafts; mentors see their own planning-phase programs). Scoping the cache to signed-out callers — one identical where-clause for all of them — makes it **leak-proof by construction** rather than by careful cache keying. Signed-in requests bypass it entirely.

## Boundaries

Per-task memory: nothing shared, nothing persisted; a scale-to-zero relaunch starts empty (correct — its data may be long stale). The pattern is deliberately a helper, reusable for other slow-changing public reads if wanted.

## Testing

4 unit tests on the helper (miss/fresh window, instant-stale + single deduped refresh under concurrency, stale-on-error at any age, no-cache-propagates) and the programs route suite green; `tsc --noEmit` + eslint clean.

### DIFF
--- a/checkin-app/src/app/api/programs/route.ts
+++ b/checkin-app/src/app/api/programs/route.ts
@@ -7,6 +7,7 @@ import { logBackendError, logger } from "@/lib/logger";
 import { isActiveOrgMember } from "@/lib/orgMembership";
 import { dollarsToCentsOrNull } from "@inventory/money";
 import { apiError } from "@/lib/api-response";
+import { staleWhileRevalidate } from "@/lib/staleCache";
 import { validateProgramAgeBounds } from "@/lib/programAge";
 
 // GET is the PUBLIC program catalog — anonymous callers legitimately get the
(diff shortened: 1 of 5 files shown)


jee7s:
Updated per review: the cache now holds **only the static rows** — `_count` never enters it. Enrollment figures come from a separate **live** query raced against a 2.5s timeout: merged into the response when the database answers, omitted entirely when it doesn't (e.g. mid-resume), and the directory card conditionally skips the figures block when they're absent (new RTL test). So a visitor during a resume sees instant, accurate static content with no numbers — never stale ones. `enrollmentStatus` (the OPEN/CLOSED badge) remains a cached row scalar, bounded by the 60s fresh window — flag if that should go live-side too.

thpr (approved):

PR #1065, opened 2026-07-17
feat(lifecycle): ProgramParticipant enrollment state machine (6-state)

thpr:
## What

Phase 2a of the lifecycle work: instance the `@/lib/lifecycle` primitives (merged in #1063) for the **ProgramParticipant enrollment** machine, then refactor the enrollment read-queries to consume the definition. Pure refactor — no schema change, no behavior change, no transition guard altered.

Per `docs/designs/PROGRAM_ENROLLMENT_STATE_MACHINE.md` (6 states, invariants I1–I4, transitions T1–T9 incl. T3f/T3m).

### New module — `src/lib/programs/enrollmentState.ts` (client-safe: no runtime prisma import)
- **`STATES`** — the six §3 states as `StateSet`s. `flags` = nullable-timestamp presence (`inventoryHeldAt`, `paymentPlanDeniedAt`); `equals` = boolean-column value (`isPaymentPlanRequested`). `PENDING_UNPAID` carries `req=false`, which is what keeps `PENDING_HOLD_FAILED` (`req=true`) out of the 7-day non-payment sweep.
- **`classify(row)`** — total switch over the status union (`assertNever` default), encoding the §3 table incl. the `req` discriminator between the two `held=null` PENDING states and the ACTIVE `held`-must-be-null rule.
- **`validate(row)`** — invariants I1–I4 (§4).
- **`TRANSITIONS`** — the §5 table (T1–T9 incl. T3f/T3m) as data.
- Type-only enum parity vs the Prisma enum (§3.4) — no value import of `@/generated/prisma`.

### Refactor read filters to `STATES.*.where` (guards stay byte-for-byte)
- `finance-ops/payment-plans` GET: scholarship queue → `PENDING_HELD`, `?queue=holds` → `PENDING_HOLD_FAILED`.
- `cron/scholarship-grace-expiry`: `PENDING_HELD_DENIED` (+ `lte: cutoff`).
- `cron/pending-participants`: `PENDING_UNPAID` (+ `pendingSince`).

## Why

Class-A anti-drift (LIFECYCLE_ARCHITECTURE §1): the legal `(status, held, req, den)` tuples had no name, and the queue/cron `where` filters were hand-coded in four places with nothing forcing them to agree. One definition now emits both the predicate and the Prisma `where`, so they can't diverge.

## Reviewer notes
- **No transition guard touched.** The CAS `updateMany`s, `delete()`-as-boundary, and `FOR UPDATE` locks are unchanged — they encode "from-state ∧ nothing-changed-under-me", deliberately narrower than a state predicate. They're asserted against `classify`/`STATES.*.has` in tests, not rewritten.
- **Client-safe:** module value-imports nothing from `@/generated/prisma` (`import type` only).
- Behavior-preserving: the refactored `where`s equal the literals they replaced on all legal data (asserted in unit tests).

## Tests
- **Unit** (`enrollmentState.test.ts`, 17 cases): classify/validate over the 6 tuples + ACTIVE vestigial-flag cases + I1–I4 violations; `where` ⇔ `has` across a status×held×req×den matrix; enum parity.
- **Integration** (`enrollmentStateOracle.integration.test.ts`): drive T3/T3f/T3m/T4–T9 against a real DB, assert each row `classify`es to the expected state and `validate`s clean; a simulated crash window (activate step-1 committed, release skipped) is flagged as an I1 violation.
- Full unit suite **1960 passed / 222 suites**; integration **55 passed** incl. pre-existing `programPaymentPlansAPI`, `cronPendingParticipants`, `cronScholarshipGraceExpiry`; `tsc --noEmit` clean.

### DIFF
--- a/checkin-app/src/app/__tests__/enrollmentStateOracle.integration.test.ts
+++ b/checkin-app/src/app/__tests__/enrollmentStateOracle.integration.test.ts
@@ -1,0 +1,226 @@
+/**
+ * @jest-environment node
+ */
+/**
+ * Validator-oracle integration tests (PROGRAM_ENROLLMENT_STATE_MACHINE §8).
+ *
+ * Drives each real transition (T3, T3f, T3m, T4, T5, T6, T7/T8/T9) against a real
+ * DB through its OWNING code — the route handler / shared mutator that carries the
+ * guard — then asserts the resulting row `classify`es to the expected state and
+ * `validate`s clean. A simulated crash window (activate's first updateMany, skip
+ * the release) must be flagged as an I1 violation.
+ *
+ * The definition is non-behavioral: this proves the guards land rows exactly on
+ * the six §3 states, never off-diagram.
+ */
+import { POST as RequestPost } from '@/app/api/programs/[id]/request-payment-plan/route';
+import { POST as ManualHoldPost } from '@/app/api/finance-ops/payment-plans/manual-hold/route';
+import { POST as PlansPost } from '@/app/api/finance-ops/payment-plans/route';
+import { POST as RefusePost } from '@/app/api/finance-ops/payment-plans/refuse/route';
+import { activateProgramEnrollment } from '@/lib/programs/activateEnrollment';
+import { withdrawAndReleaseHold } from '@/lib/program/capacity';
+import prisma from '@/lib/prisma';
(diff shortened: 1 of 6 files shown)


jee7s (approved):

jee7s:
Review also coming

PR #1071, opened 2026-07-17
feat(visits): cap visit duration at 24 hours

thpr:
## What

A visit had no maximum-duration guard — the only time constraint was `departedAt > arrivedAt`. A member (self-entry) or an admin/board edit could record a closed visit spanning days or weeks.

Adds a 24-hour cap on **closed** visits.

## Why

A multi-day visit is a missed checkout or a typo, not a real visit, and it skews self-reported facility/trends hours. Bounding it keeps the data honest.

## Changes

- `withinMaxDuration()` + `MAX_VISIT_MS` in `visitTimes.ts` (24h, closed visits only)
- Enforced on all three closed-visit write paths:
  - `POST /api/attendance/manual` (self-entry — inline, bypasses the shared helper)
  - `PATCH /api/facility/visits` (sysadmin/board edit)
  - `PATCH /api/events/[id]` (Present status)
- Client-side mirror in each of the three entry UIs (manual page, facility-ops visits, program-ops session) for instant feedback — **server stays the trust boundary**
- Unit tests for the boundary (under / exactly 24h / over)

## Notes for reviewer

- Open visits (no departure) carry no entered duration, so the cap does not apply. The manual endpoint already bounds open backfills to 6h/same-day.
- 24h boundary is inclusive (exactly 24h passes).
- Client checks hand-roll the 24h math inline, matching the existing per-UI validation style (each already duplicates the "after arrival" check).

### DIFF
--- a/checkin-app/src/app/__tests__/auditLog.integration.test.ts
+++ b/checkin-app/src/app/__tests__/auditLog.integration.test.ts
@@ -380,9 +380,10 @@ describe('AuditLog Integration Tests', () => {
         createdVisitIds.push(visit.id);
 
         const newArrived = new Date('2020-01-01T10:00:00.000Z');
+        const newDeparted = new Date('2020-01-01T12:00:00.000Z');
         const req = new Request('http://localhost:4000/api/facility/visits', {
             method: 'PATCH',
-            body: JSON.stringify({ visitId: visit.id, arrivedAt: newArrived.toISOString() }),
+            body: JSON.stringify({ visitId: visit.id, arrivedAt: newArrived.toISOString(), departedAt: newDeparted.toISOString() }),
         });
 
         const res = await updateVisit(req as never);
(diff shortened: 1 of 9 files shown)


jee7s:
Reviewed — LGTM. Verified the cap holds on all three server write paths: the facility PATCH enforces it on **merged** values (your auditLog test edit is what proves an arrival-only edit can't sneak past against a stored departure), the events PATCH overwrites both fields so there's no arrival-only bypass there either, and the boundary is inclusive-at-24h with a unit test pinning it on both sides.

One nit: the three client-side copies each hardcode `24 * 60 * 60 * 1000` while the server imports `MAX_VISIT_MS` — `lib/visitTimes` is client-safe, so importing the constant in the pages would make a future cap change a one-line edit instead of four.

Complementary note: the open-visit half this cap deliberately doesn't touch (multi-day *open* visits = missed checkouts) is now scheduled — infra#140 wires `/api/cron/nightly` (force-checkout + post-event email flush) at 22:50 CT, ten minutes before prod's curfew so closing the facility doesn't resurrect the env overnight. Between the two, multi-day visits can neither be recorded nor persist.

thpr:
Thanks! Addressed the nit in fc388208 — all three client pages now import `MAX_VISIT_MS` from `@/lib/visitTimes` instead of hardcoding `24 * 60 * 60 * 1000`, so a future cap change is a one-line edit. Server was already using the constant.

Good call on the open-visit half being covered by infra#140's nightly force-checkout — between the write-path cap here and that cron, multi-day visits can neither be recorded nor persist.

jee7s (approved):

PR #1082, opened 2026-07-17
feat(finance): bidirectional Shopify ↔ activation match audit

thpr:
This needs analysis, check and pushback on Claude.  It's an initial design that has not been reviewed or tested.

The reconciler is a cursor-driven recovery loop, not a completeness check: it acts on orders it can attribute and silently passes over the rest, and its reversal pass only inspects activations that already carry an order id. Four populations were never listed anywhere:

  - paid membership/program orders claimed by nothing (money in, no access) — the forward pass returns "not mine" twice and moves on;
  - activations whose recorded order id isn't in the mirror (skipped with `continue` — invisible forever);
  - activations with NO payment basis at all — ACTIVE_WITHOUT_PAYMENT has had an enum value and board-alert copy since the reconciler shipped, but nothing ever computed it;
  - the legitimately-manual class (board certify-payment overrides, approved scholarships) that an auditor needs to see separated, not buried.

runMatchAudit() (lib/finance/matchAudit.ts) reports all four, both directions. "Should reconcile" is decided by VARIANT ID — BoardSettings holds the membership variants, Program rows hold the program variants, and an order with none of those on any line (donation, t-shirt) is out of scope by design and never reported. This depends on the mirror's new variant_legacy_id column (previous commit); the audit reports variantCoverage explicitly so a mirror that predates the column reads as "backfill needed", not as a falsely clean report.

Read-only and on-demand only: GET /api/finance-ops/s-read/match-audit (same board/sysadmin gate and mirror-503 shape as the sibling s-read routes) behind a "Run match audit" button on the payments page. It raises no PaymentExceptions — report first; promote findings to exceptions once the board has seen real output. Orders already covered by an unresolved exception bucket as TRACKED_EXCEPTION rather than re-flagged. PERSON_BG processes are excluded from the activation sweep: they carry no payment by design.

Also fixes a reconciler cursor bug the audit design surfaced: runReconcile advanced its watermark past an order whose forward pass THREW, skipping that order forever. The cursor now freezes at the first failure; later orders in the batch still process (idempotent) and are re-scanned next run.

Tests pin the bucket routing (a wrong branch sends the board chasing the wrong money), the variant-set union handed to the mirror, the PERSON_BG exclusion, the deny paths never running the audit, and the panel staying inert until clicked.


jee7s (commented):
## Review — Claude Fable 5 (xhigh: 10 finder angles → verification → sweep → adversarial architecture pass)

The design is right where it matters most: the four unlisted populations are real, variant-id is the correct scope key, and read-only-first is the right posture. The pushback below is what you asked for. **Companion PRs (#1084 + two more into this branch) are all optional and independently decline-able** — each is a concrete proposal, not a fait accompli.

## Design questions first — yours to rule on

**Q1. What counts as a "claim"?** Three ways the current claim logic hides the exact gap the audit hunts:
- *Multi-item checkouts*: `claimed` is one order-id-keyed set, so a membership claim marks the whole order MATCHED even when the program half activated nothing. Family buys membership + Robotics in one checkout, only membership activates → clean report. (#1084 makes claims per-kind; gap rows name only the missing half.)
- *ARCHIVED processes still claim*: a household pays (PENDING_BG_CLEARANCE), the application is archived as abandoned — the order still reads MATCHED. (#1084 excludes ARCHIVED.)
- *Tracked exceptions were order-wide too* (found by the review's final sweep, on our own first fix): an open exception on the membership half hid the program half's untracked gap. PaymentException carries `processId`/`programId`, so tracking is now attributed to what it covers. (In #1084.)

**Q2. Is presence in the mirror really a payment basis?** `ORDER_MATCHED` never checks the order is paid or carries the expected variant: a fully-refunded membership reads clean (reversals raise exceptions but never flip status off ACTIVE), and a typo'd order id that happens to exist reads clean. Proposed middle ground (audit-v2 PR): mark reversed orders via the existing `ordersByLegacyIds` + `isReversed` — informational, not double-reporting your reversal exceptions. Variant-validation of the reverse direction is deferred — it needs new batched SQL and is your call.

**Q3. Who is legitimately manual?** Two populations will pollute the first real report:
- *Free-program participants*: the participants route creates them ACTIVE (`isFree` → `initialStatus 'ACTIVE'`), and the sweep takes every ACTIVE participant — so every free-program kid lands in NO_PAYMENT_BASIS. This alone would have made the first board run useless. (Audit-v2 PR filters the sweep to payment-bearing programs, keeping order-backed rows regardless.)
- *Admin comps* (T2 `enroll(comp)`): ACTIVE, no order, no scholarship stamp — indistinguishable from real gaps today. The CREATE audit row records `status: ACTIVE` at creation, which distinguishes comp from paid-later reliably once free programs are filtered; audit-v2 classifies these `ADMIN_COMPED` (listed-as-manual, with who comped, same posture as MANUAL_CERTIFIED). If the board finds trail-inference confusing in practice, the honest fix is a `compedById` column — deliberately NOT spent from the migration budget until the inference proves insufficient.

**Q4. Which ACTIVE memberships can the audit even see?** The sweep reads only `OrgMembershipProcess`, but the legacy CSV import upserts `OrgMembership` ACTIVE with **no process row** — those households have real access and are invisible even to NO_PAYMENT_BASIS. The population most likely to actually be ACTIVE_WITHOUT_PAYMENT is the one the audit can't see. (Audit-v2 adds a NO_PROCESS sweep over process-less ACTIVE memberships.)

**Q5. The reconciler freeze trades one lost order for a stallable pipeline.** Two mechanical holes are fixed in #1084 (a strict-`>` timestamp tie skipped the failing order forever despite the freeze — cursor now pulls back below it). The design question remains: the freeze is a one-way latch with no retry cap, no skip, and — uniquely in this file — no PaymentException; once >1000 updates accumulate behind a poison order, new orders are never scanned, while `ordersScanned: 1000` logs look healthy. Proposals (not built): a `RECONCILE_STALLED` exception kind next time a migration window opens; near-term, surface reconcile-cursor age as a step in your diagnose route — it's the one link in the chain diagnostics can't see today. Also note the cursorFrozen change itself shipped with no test.

**Q6. Falsely-clean states.** Unset variant config → green "No gaps" (audit is blind, not clean); partial variant backfill → orders silently *absent* (SQL-filtered), with a warning only at exactly 0% coverage. #1084 adds `configuredVariants` + partial-coverage callouts. First-run noise: activations predating the mirror will flood ORDER_NOT_IN_MIRROR — audit-v2 splits `PRE_MIRROR` (recorded id < the mirror's min legacy_id, informational) from real gaps.

**Q7. Promotion path.** You sequenced exceptions after the board sees real output — agreed, nothing auto-raises. The promotion PR (optional, single-purpose) adds only a human-per-row "Track" button: order gaps → `UNMATCHED_ORDER`, activation gaps → `ACTIVE_WITHOUT_PAYMENT` (both kinds already exist — no migration), reusing `raisePaymentException`'s dedup. It also adds entity-keyed tracked-awareness to the activation sweeps so a promoted gap stops re-reporting as raw. This resolves the page contradiction too (panel gaps become rows in the exceptions table above it).

## Smaller items (fixed in #1084 unless noted)

- **CI is red on this branch**: `lifecycleStatusLiteralAllowlist` rejects matchAudit's raw state-sets → allowlist entry added (reconcile.ts precedent).
- The three new mirror reads disagreed on `test = false` (test lines could fake variant coverage; a test-mode order could back an ORDER_MATCHED) → aligned; `ordersForVariants` also gains a deterministic ORDER BY.
- `isPaid` + the membership-variant set were the *third* inline copies (reconciler #1074 gate, webhook, audit) — the drift class behind the coupon false-positive, and worse here: a drifted variant makes orders **vanish** from the audit, not flag. Now exported from reconcile.ts and shared. (Webhook route's copy left alone — worth a follow-up on main.)
- Panel: hand-copied types (which omitted `configured`; the repo's actual pattern is `import type` — see shopify-webhook + my-programs/conflicts pages) → type-only import; `dollars()` → the `formatCents` the hosting page already imports; 9 per-render filter passes → one memoized pass (the sibling resolve-modal re-renders this panel per keystroke).
- Tests: the "explains an unwired mirror on 503" test drove a 404 through the generic catch (mockFetchJson never returns 503) — the 503 branch had zero coverage; bucket precedence had no two-predicates-true fixture. Both pinned now, plus split-claim, ARCHIVED-exclusion, per-kind-tracking, and claimed-but-refunded precedence tests.
- **PII note** (no action taken): this is the first s-read route returning customer emails + household/person names — diagnose is deliberately data-free, and route-coverage CI only enforces migrated routes, so nothing gates this surface. Board-gated, so no exposure bug; flagging per SECURITY-POLICY.md for a registration-or-exemption call.
- **Scale note** (no action taken): `ordersForVariants` is the layer's only unbounded query (full scan per click) and the activation sweep grows with org age. Deliberately NOT season-bounding the activation sweep — an ACTIVE process from 3 years ago *is* the current membership; bounding it would hide live access. If latency bites, bound the mirror order scan only.

## Adjacent-area proposals (comment-only, no code)

1. **Payment basis as a recorded fact** — the root cause under Q3/Q4: an activation's basis (order / certified / scholarship / comp / legacy-import) is scattered across `shopifyOrderId`, `certifiedById`, `wasOrgMemberAtApproval`, and absence. A `paymentBasis` enum stamped at activation by every path would let the audit read intent instead of reverse-engineering it — and is the natural foundation for Fee/FeePayment reconciliation and early-renewal-discount validation later. Worth a design doc before any schema spend.
2. **Coupon entitlement** — your DiscountCodeRule sketch stays deferred; audit-v2 just surfaces the already-mirrored `discount_codes` on unclaimed rows so the board can eyeball codes in the meantime.
3. **Early-renewal discounts** — when reconciliation learns amounts again, it must accept the early-renewal discounted amount as valid, not just full dues; the shared `isPaid`/variant-set module is where that rule should live once, not three times.

Companion PRs into this branch, each independently decline-able: **#1084** (mechanical fixes — green: tsc, eslint, 2027/2027 unit), **audit-v2** and **track-button** (links in follow-up comment once CI is green). Everything else above is a question, not a change.

jee7s:
Follow-up to the review above — the companion PRs are open, stacked, and green (full unit suite 2073/2073 on the top of the stack; each is independently decline-able and small enough that declining is painless):

- **#1084** — mechanical fixes (per-kind claims *and* per-kind exception tracking, ARCHIVED-not-a-claim, cursor-tie pull-back, mirror `test=false` consistency, shared `isPaid`/variant-set, falsely-clean callouts, the lifecycle-guard allowlist entry that clears this branch's CI red).
- **#1085** (on #1084) — audit v2: free-program sweep filter, `ADMIN_COMPED` from the audit trail, `ORDER_REVERSED` + `PRE_MIRROR` states, `NO_PROCESS` memberships, discount codes on unclaimed rows. (Q2/Q3/Q4/Q6 from the review.)
- **#1087** (on #1085) — the human-per-row "Track" button: gap → `UNMATCHED_ORDER`/`ACTIVE_WITHOUT_PAYMENT` with server-side staleness re-derivation and entity-keyed tracked-awareness. Nothing auto-raises. (Q7.)

Merge order if you take them: #1084 → #1085 → #1087, each folding into this branch. Take any prefix — the stack degrades gracefully. The design questions in the review (reverse-direction variant validation, comp basis as schema, the reconciler freeze's alerting story, PII registry call) remain open and yours.

jee7s (approved):

PR #1093, opened 2026-07-18
scholarship notification emails: review team notify + applicant acknowledgement

jee7s:
Implements the scholarship-notification design change (design doc updated in the same PR). **Scope per Jeff's decision: an applicant receives exactly ONE automatic email — the request acknowledgement.** Board decisions are communicated manually; no automated status emails exist.

**What changes**
- `BoardSettings.scholarshipNotifyEmail` (Settings → Email; comma-list, validated like Reply-To). **Unset → every board member is emailed** — the board is the review team until configured. Review-team notify fires on request (program and membership sides).
- Applicant **acknowledgement** (ungated, transactional): household leads ∪ the participant, deduped, children without email covered by leads, board on-behalf requesters deliberately excluded. This is the applicant's only automatic email — the integration suites pin it with **negative assertions**: program approve, program deny, and membership certify each send **zero** emails.
- **Approve/deny are silent by design** (documented in the design doc §5): the Scholarship Review Team communicates decisions manually. A denial therefore starts the `scholarshipDenialGraceDays` clock with no automatic notice — the board's manual denial message should state the deadline.
- **Sends only on a real transition**: the program route keys on the pre-image flag (`fromWhere` narrows on status only, so both CAS branches also match a no-op re-POST — a bare count check would re-spam the review team); the membership route gains the idempotency guard it lacked. All sends fire after the transition commits and can never fail the request. Both `[EMAIL DISPATCH]` log stubs are deleted. The program request's Shopify-failure branch acks with that branch's applicant-facing copy.
- Copy: "Finance Committee" → **"Scholarship Review Team"** (membership + program pages, VOCABULARY entry). Design doc gains a Notifications section (who/when table, the silent-approve/deny rule and its grace-clock consequence, fallback rule, membership-deny asymmetry).

**Release note**: one additive nullable migration (`scholarshipNotifyEmail`) — the third since the last coalesce; count check is advisory (#1056) and Jeff has ruled no coalesce this release.

**Verified locally**: bare-exit tsc 0, eslint clean, unit + RTL suites green. The integration suites' email-behavior assertions (request sends exactly once, no-op re-POST sends nothing, re-request-after-denial, zero emails on approve/deny/certify) run in CI.

Stacked follow-up: #1094 (real non-payment warning/removal emails).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01RFHNwTRstb6KrgCqPA7iTe

PR #1094, opened 2026-07-18
make the non-payment warning and removal emails real

jee7s:
Makes the pending-participants cron's emails real: it **really removes people** for non-payment while its warning and removal emails were `[EMAIL DISPATCH]` log lines nobody received. Stacked on #1093 (uses its recipient model); merge #1093 first and this follows.

- All four sends are real: day-1 reminder, day-3 pay-within-4-days, day-6 FINAL WARNING, and the removal notice — to household leads ∪ the participant (children without email covered by leads), **ungated** (payment-deadline / seat-loss notices are transactional). These go to ordinary non-payers, not scholarship applicants — the sweep excludes requested and denied rows by design, so the one-automatic-email rule for applicants (#1093) is untouched.
- **Removal notice only after confirmed removal**: it previously "sent" from the classification loop before `withdrawAndReleaseHold` ran — a failed or already-done removal would have emailed a lie. Now it sends inside the try after success; P2025 (removed elsewhere) and real failures send nothing (the failed row retries next run).
- Warning copy points to the in-app scholarship / payment-plan request ("the Scholarship Review Team will follow up") instead of the retired hardcoded `finance@` address.
- The recipient resolver moves to its neutral home: `emailRecipients.resolveHouseholdRecipients` (a payment warning isn't a scholarship email); `scholarshipEmails` re-exports under its old names — zero churn at its call sites.
- Design doc §5: cron rows added to the who/when table (marked as non-applicant emails), removal-after-confirmation ordering documented.

Verified locally: bare-exit tsc 0, eslint clean, 190 tests green across the touched suites, full unit run clean. The extended `cronPendingParticipants.integration.test.ts` runs in CI once this retargets main.

### DIFF
--- a/checkin-app/docs/PROGRAM_CAPACITY_AND_SCHOLARSHIPS.md
+++ b/checkin-app/docs/PROGRAM_CAPACITY_AND_SCHOLARSHIPS.md
@@ -136,11 +136,13 @@ against *failed calls* but not against *crashes between the two steps*:
   periodic reconcile job is deliberately deferred until the drift is observed in
   practice.
 
-- **(a) Withdrawal** — `DELETE /api/programs/[id]/participants` (self or admin removal;
-  the same route handles both) and the non-payment kick
-  (`cron/pending-participants`; denied-and-still-PENDING participants are excluded
-  from the kick via `paymentPlanDeniedAt: null` — their timeline belongs to the
-  grace-expiry cron in (c)) both funnel through this.
+- **(a) Withdrawal** — `DELETE /api/programs/[id]/participants` (self, admin, or
+  board removal; the same route handles all three) funnels through this. This is
+  also now the *only* path a non-payment kick takes: `cron/pending-participants`
+  no longer removes anyone itself (reviewer decision — removal is a human,
+  board-driven action, not something a cron does unattended). The cron warns the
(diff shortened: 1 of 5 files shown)


thpr:
I think at 4 days we need an email to leadership too.  I'm VERY concerned about cron doing permanent removes vs. it flagging for us and us taking action.  This is core customer service and a computer isn't enough here.

thpr (changes requested):
See comment

jee7s:
Both points addressed:

- **The cron never removes anyone now.** Day-7+ rows are flagged `overdue` and the removal path (and its email) is deleted — removal is a human action via the participants admin surface, and the accompanying communication is yours too, same manual-communication model as the scholarship decisions. Warning copy updated to match reality ("may be released by the board").
- **Leadership email at the 4-days-left mark**: one digest per cron run to all board members whenever anyone is at day-3 (approaching deadline) or day-7+ (overdue — "action needed: remove the enrollment or contact the household"), listing person/program/day. Per-run digest rather than per-person so the board gets one actionable email, not a drip.

Tests rewritten to pin it: the day-8 row survives the run, zero removal emails ever, exactly one digest containing both tiers. tsc/eslint clean; integration runtime in CI on retarget.

thpr (approved):

PR #1095, opened 2026-07-18
feat(finance-ops): membership scholarship deny route (parity with program side)

thpr:
Stacks on #1093 (`claude/scholarship-emails`) — base is that branch, not main.

## Gap
Membership scholarship / payment-plan requests could only be **approved or ignored**. A declined request sat `PENDING_PAYMENT` forever with no email, breaking the request ack's "we'll review and follow up" promise. This adds the missing deny path — parity with the program side.

## What
- **`POST /api/finance-ops/membership-payment-plans/refuse`** — clears `isPaymentPlanRequested → false`; process stays `PENDING_PAYMENT` (household returns to normal pay-to-activate). **No Shopify op, no grace deadline, no schema column** — membership holds no seat and has no grace-expiry cron, unlike program. Transactional `true→false` guard: a no-op re-deny 409s and sends no email.
- Fire-and-forget **gated status email** after commit (`sendScholarshipStatus`), copy mirrors program-deny voice minus the seat/grace lines.
- Auth mirrors the membership **approve** route exactly (`roles: ['isBoardMember']`, session check, no `hasHouseholdConflict` — approve has none).
- **Deny button + confirm modal** on the finance-ops membership tab.
- Design doc §5: asymmetry note rewritten (now closed) + who/when table row filled.

## Tests
- Integration (`membershipPaymentPlansAPI.integration.test.ts`): 401/403/400/409, deny clears-flag/stays-PENDING/audits, 409 re-deny, + email (fresh deny status once, no-op re-deny zero-send, per-recipient gating). **27/27 pass** against a throwaway Postgres.
- `authzRegistry` entry for the new role-gated route.
- `tsc` + eslint clean.

No schema/migration change — clearing the flag is the whole of denial state.

### DIFF
--- a/checkin-app/docs/PROGRAM_CAPACITY_AND_SCHOLARSHIPS.md
+++ b/checkin-app/docs/PROGRAM_CAPACITY_AND_SCHOLARSHIPS.md
@@ -187,7 +187,7 @@ request actually transitioned (no duplicate mail on a no-op re-POST).
 | Program approve | `POST /api/finance-ops/payment-plans` | — | **no automatic email** |
 | Program deny | `POST /api/finance-ops/payment-plans/refuse` | — | **no automatic email** |
 | Membership approve (certify) | `POST /api/finance-ops/membership-payment-plans` | — | **no automatic email** |
-| Membership deny | — | **does not exist** | — |
+| Membership deny | `POST /api/finance-ops/membership-payment-plans/refuse` | — | **no automatic email** |
 | Manual-hold | `…/payment-plans/manual-hold` | — | **no email** |
 
 **2. Approve/deny are silent by design.** Board decisions — program approve, program deny,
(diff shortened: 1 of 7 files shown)


jee7s:
Reviewed (full xhigh pass). The deny route itself is a welcome gap-closure and the state mechanics are solid — CAS guard, queue/badge behavior, re-request coherence, and authz tests all check out. But this forked before the base moved: `claude/scholarship-emails` now carries a scope reduction that **deleted `sendScholarshipStatus`** and established the rule that an applicant gets exactly one automatic email (the request ack) — approve/deny send nothing and the board communicates decisions manually. As-is, merging breaks the base branch: the import fails tsc, and the test file auto-merges (no conflict markers) into a suite asserting both "certify sends nothing" and "deny sends one."

Needed before merge:
1. Rebase onto current base; **delete the email block + imports**; redo the doc hunk under the no-email model (the §5 it patches was rewritten).
2. **Add the COI check** (`hasHouseholdConflict`, sysadmin bypass) — approve enforces it inside `certifyPaymentPlan` and program refuse checks it at route level; as written a board member can deny their own household's request.
3. **Wrap the CAS + audit write in `$transaction`** (archive.ts pattern) — a crash between them leaves an unauditable denial that can't retry (flag already flipped).
4. Flip the deny-email tests to **zero-email negative assertions**; register the new guard in `guardFromStateParity`'s GUARDS list.
5. Modal copy: replace "They'll be emailed that the request was declined" with a reminder that the **board notifies the household manually** — under the current model the promised email doesn't exist.

Worth folding in: the member page's `planRequested` one-way latch (a live session still shows "Requested!" after a deny), an in-flight guard on the confirm button, dropping the exact-string `data.error` match, and 400 (not 500) on malformed JSON.

Happy to propose the fix-up as a PR into this branch if useful.

thpr:
Thanks for the thorough pass — all five must-fix items are addressed, plus the member-page latch.

**Needed before merge:**
1. **Rebased** onto current base; deleted the `sendScholarshipStatus` import + email block; redid the §5 doc hunk under the no-automatic-email model (table row + item 4 now describe deny as a silent board action).
2. **COI check added** — `hasHouseholdConflict` with sysadmin bypass at the route level (approve enforces it inside `certifyPaymentPlan`, program refuse at route level; deny now matches). A nonexistent process resolves to 409, not 403, so "not found" stays distinct from "your own household".
3. **CAS + audit wrapped in `$transaction`** (archive.ts pattern) — no more crash window leaving a flag-flipped denial with no audit row and no retry.
4. Deny email tests flipped to **zero-email negative assertions**; registered the new guard in `guardFromStateParity`'s `GUARDS` list (PENDING_PAYMENT, no status edge — clears the flag only).
5. **Modal copy** now reads "No automatic email is sent — notify the household yourself."

**Folded in:** the member page's `planRequested` one-way latch — now two-way synced to the server flag, so a denied member no longer stays stuck on "Requested!" after a refetch.

**Deliberately left matching approve** (fixing only deny would desync the sibling routes): the exact-string `data.error` match, no in-flight button guard, and 500 (not 400) on malformed JSON. Happy to do these as a follow-up that touches both routes if you'd prefer.

Verified: `tsc` + eslint clean; unit (`guardFromStateParity`, `authzRegistry`) green; membership integration suite green except the pre-existing `certify (approve) sends no automatic email` — that fails on the **base** branch because activation sends the standard "Welcome to the Treehouse" email (`payment.ts:275`), independent of this branch. Being fixed separately on `claude/scholarship-emails`; I'll rebase to pick it up once it lands.

jee7s:
Fix-up is ready as **#1097** (into this branch): rebases onto the current base, deletes the email layer per the notification contract, adds the COI check, wraps the CAS + audit in `$transaction`, flips the email tests to zero-email assertions, registers the guard in `guardFromStateParity`, fixes the modal copy/in-flight guard/status-code coupling, and repairs the member page's `planRequested` latch. Merge #1097 here and this PR stacks cleanly on #1093.

thpr:
Conflicts from my rebase make this incompatible with 1097.  I'm preparing a patch onto 1097 to add the one test that's superior in 1095 vs 1097

PR #1097, opened 2026-07-18
review fix-up for #1095: no-email base, COI check, transactional audit, contract-true UI

jee7s:
The fix-up offered in the #1095 review — merge this into your branch and #1095 becomes stackable on current #1093. Every item from the review comment is addressed:

- **Current base merged in**: the scope reduction (`sendScholarshipStatus` deleted; applicant gets exactly one automatic email) and the design-doc §5 rewrite are reconciled — the deny route now sends **nothing**, and the doc's membership-deny row reads "no automatic email" with the asymmetry note rewritten as closed-but-silent.
- **COI**: `hasHouseholdConflict` before the CAS (sysadmin bypass), matching approve's enforcement in `certifyPaymentPlan` and the program refuse route — plus a real 404 for a nonexistent processId.
- **Atomicity**: the flag-clear CAS and its audit row commit in one `$transaction` (archive.ts pattern) — no more unauditable denials on a crash between them.
- **Tests**: deny-email tests flipped to zero-email negative assertions (welcome-subject-filtered, same shape as #1096); the new CAS guard registered in `guardFromStateParity` (flag-only, no status edge, archive.ts shape); `denyReq` hoisted. The certify test is left to #1096.
- **UI**: modal now says "No automatic email is sent — contact the household to let them know"; in-flight guard on the confirm Deny button (double-click can't paint a false failure); 409 detected by status code, not error-string matching.
- **Member page**: the `planRequested` one-way latch is fixed — a state refetch after your denial reverts the button to requestable (test added via the visibility-refetch path).

Verified: bare-exit tsc 0, eslint clean, guardFromStateParity + membership page suites 73/73 locally; integration runtime in CI. Deliberately untouched: `doApprove`'s identical error-string pattern (predates this PR), the certify test (#1096 owns it).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01RFHNwTRstb6KrgCqPA7iTe

PR #1096, opened 2026-07-18
test(scholarship): certify test allows orthogonal activation welcome email

thpr:
Stacked on #1093 (`claude/scholarship-emails`).

## Problem
`certify (approve) sends no automatic email` in `membershipPaymentPlansAPI.integration.test.ts` FAILS on the base branch. After the `merge origin/main (#1090)`, `activate()` sends the standard **"Welcome to the Treehouse — your membership is active!"** email (`checkin-app/src/lib/membership/payment.ts:275`). The test asserted zero *total* emails, got 1.

## Fix
The welcome email is orthogonal to the scholarship-notification decision (§5 of `PROGRAM_CAPACITY_AND_SCHOLARSHIPS.md` — "applicant receives exactly one automatic email" is about the scholarship ack, not the activation welcome). Test now filters out the welcome subject and asserts zero *scholarship-decision* emails. The welcome send itself is **not** suppressed.

## Verify
Integration run scoped to `membershipPaymentPlansAPI`, throwaway Postgres: 18/18 pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1098, opened 2026-07-18
test(finance-ops): cover membership-deny COI guard + 404 (stacks on #1097)

thpr:
Stacks on #1097 (`claude/pr1095-fixes`).

#1097 added conflict-of-interest enforcement (`hasHouseholdConflict`, sysadmin bypass) and a real **404** for a nonexistent `processId` to the membership deny route — solid changes, but shipped with **no tests exercising either path**. This adds that coverage (the only gap I found comparing #1097 against my own branch).

## Added
- **404** when the process does not exist.
- **403 conflict-of-interest** when a board member denies their **own** household — and asserts the pending flag is **not** flipped by the refused deny (guards against a partial-effect regression).
- **Sysadmin COI bypass**: a sysadmin may deny their own household (200, flag cleared).

## Note on test hygiene
The two COI cases create an extra in-household board member. Each **deletes it after asserting** — otherwise it leaks into the "email all board members" review-team fallback and inflates the request-email count assertion elsewhere in the suite (a real cross-test pollution I hit while writing these).

## Verified
- `tsc` clean (only the pre-existing `@aws-sdk/client-lambda` missing-dep noise), eslint clean.
- Membership integration suite: my 3 new tests green; whole suite green except the pre-existing `certify (approve) sends no automatic email` (the activation welcome-email failure #1096 owns — unchanged by this PR).

Once #1097 merges into #1095's branch, this rides along.

### DIFF
--- a/checkin-app/src/app/__tests__/membershipPaymentPlansAPI.integration.test.ts
+++ b/checkin-app/src/app/__tests__/membershipPaymentPlansAPI.integration.test.ts
@@ -251,6 +251,43 @@ describe('Membership payment-plan routes', () => {
             expect((await DenyPost(denyReq({ processId: leadProcessId }))).status).toBe(200);
             expect((await DenyPost(denyReq({ processId: leadProcessId }))).status).toBe(409);
         });
+
+        it('404 when the process does not exist', async () => {
+            mockSession.mockResolvedValue({ user: { id: boardId, isBoardMember: true } });
+            expect((await DenyPost(denyReq({ processId: 999999999 }))).status).toBe(404);
+        });
+
+        it('403 (conflict of interest) when a board member denies their OWN household; the request is untouched', async () => {
+            const own = await makeProc('DenyCOI', { requested: true });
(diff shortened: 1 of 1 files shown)


jee7s:
Reviewed — LGTM, and thank you for closing the gap I shipped: #1097 added the COI/404 behavior with no tests. All three cases are right, the partial-effect assertion (flag NOT flipped on the refused deny) is exactly the regression guard that matters, and your 404 assertions match the reconciled union route (I kept the 404 probe when merging your parallel fix commit into #1097).

One hygiene note, non-blocking: the extra board-member cleanup runs AFTER the assertions, so a failing assertion skips the delete and the leaked board member cascades into the review-team fallback counts you were defending against — worth a try/finally (or afterEach registry) so the cleanup survives a red assertion. Fine to land as-is; the pollution only bites when the suite is already failing.

Note the base moved under this PR when I reconciled #1097 — if GitHub shows it behind, a rebase should be conflict-free (your tests' region didn't change).

PR #1100, opened 2026-07-18
🛡️ Sentinel: [MEDIUM] Remove stack trace from error log to prevent information exposure

dkaygithub:
🚨 Severity: MEDIUM
💡 Vulnerability: Raw error stack traces were being persisted in the database via the `ErrorLog` table, which feeds application dashboards.
🎯 Impact: Exposes internal server information (like directory structures and dependency versions) to users with dashboard access, breaking defense-in-depth.
🔧 Fix: Modified `logBackendError` in `src/lib/logger.ts` to stop saving the `stack` field to the database. Stack traces will continue to be handled by secure, centralized logging infrastructure via `console.error`.
✅ Verification: Run `npm test` and `npm run lint`.

---
*PR created automatically by Jules for task [6366471919981849085](https://jules.google.com/task/6366471919981849085) started by @dkaygithub*


thpr:
Hey Jules - please change this PR to just comment this section that it's okay with us... this has been raised 3 times as an issue and we accept the risk and it needs to stop showing up in audits

jee7s (commented):

The security goal (stop persisting stack traces to ErrorLog) is sound, but three CONFIRMED problems with the current shape:

**1. Deterministic CI red against merge-queued #1341.** The new `console.error('Backend error at …', stack)` matches no regex in #1341's `KNOWN_INTENTIONAL` allowlist (jest-fail-on-console, active on both unit and integration tiers). `logger.integration.test.ts`'s purge test calls the real `logBackendError(new Error(...))`, the unallowlisted console.error fires, and the suite fails the moment this PR rebases onto main. Either allowlist the prefix or (better — see 3) don't use raw console.

**2. The admin Errors UI still advertises the removed capability.** `ErrorLogPanel.tsx:82-94` renders "Click a row to expand stack/context"; with `stack=null` every expansion is empty. Stacks now exist only in container stdout — which admins can't access and which rotates with the task. If the intent is stdout-only, the panel copy and expansion should change with it; alternatively persist a sanitized/truncated stack, or keep it gated to sysadmin reads (`ErrorLog` is already `@sensitivity:internal`).

**3. Raw `console.error` bypasses the `logger` seam defined 15 lines above in the same module.** The comment says "centralized logging", but the only centralization seam is `logger.error` — if that's ever pointed at a real sink, the one log line this PR exists to preserve silently misses it.

Recommendation: route through `logger.error`, resolve the #1341 allowlist interaction explicitly, and make a deliberate decision about admin-visible stacks before merge.

jee7s:
**Linkage:** the CI collision described in the review above is with #1341 (jest-fail-on-console), which is in the merge queue — treat its allowlist as already on main when rebasing. The admin-facing impact is in `ErrorLogPanel.tsx` (System Status → Errors), whose stack-expansion UI this PR silently hollows out; whichever way the persist-vs-stdout decision goes, that panel and this change should move together.

jee7s:
**Area of concern: the red is caused by this PR, and the admin panel is left advertising something it no longer has.**

`Run Tests (integration)` fails one test — `logger.integration.test.ts` — with `Expected test not to call console.error()` on the message `Backend error at …:`, thrown from the line this PR adds. #1341 merged 2026-07-26, so the `KNOWN_INTENTIONAL` allowlist in `jest.setup.js` is live and has no matching prefix.

Three things: route through the `logger.error` seam defined about fifteen lines above rather than a raw `console.error`; add the anchored allowlist entry; and decide what `ErrorLogPanel.tsx` should say, since it still offers "Click a row to expand stack/context" for a `stack` that would always be null.

Worth noting `ErrorLog.stack` is already `@sensitivity:internal` behind an admin gate, so the premise is weaker than the PR body suggests.

---
_Posted by Claude._

thpr:
Closing as a duplicate of #1574 — same finding (stack traces persisted to `ErrorLog`), same fix site in `checkin-app/src/lib/logger.ts`, same removal of `stack` from the `errorLog.create` call.

#1574 is the more complete of the two: it also patches `logger.integration.test.ts` for the new `console.error` and records the finding in `.jules/sentinel.md`.

The review feedback from this PR that #1574 does *not* address has been carried over there.

PR #1103, opened 2026-07-18
participant merge: field picker, zero deletion, tombstone provenance

jee7s:
PR-B of the merge-tool work (stacked on #1101; #1102 is the independent hotfix — whichever lands second rebases the one route file, both orders compile). Full plan: adversarial challenge → reviewed opus spec → implementation → 2119 unit + 1135 integration tests green locally against a real Postgres, including the security contract suite.

**What changes**
- **Field-by-field selection**: the merge screen shows a radio only for *true* conflicts (name/email/phone/googleId/DOB, both sides non-null and different — DOB confirmed reachable: the analyze route isn't stripper-migrated). Single-sided values auto-backfill; `image` never gets a radio; `lastBackgroundCheck`/`lastWaiverSign` auto-take the **newer** date (same human — an older compliance date is never the right survivor). Choices are re-validated server-side against server-recomputed conflicts.
- **Zero deletion**: collisions keep both rows — the keeper's stays, the tombstone keeps its own. The **relation disposition table** covers all of Person's relations: `accounts` move so a Google login resolves the survivor, `PERSON_BG` processes/attestations and trusted-adult (pickup) records follow the surviving identity, actor history stays put, visits move with an open-visit skip (never a second open visit, never a fabricated departure).
- **Provenance + safety**: `Person.mergedIntoId` (one additive migration) drives a CAS tombstone write (concurrent/repeat merges: exactly one wins), 409s on merging or analyzing an already-merged record, a kiosk **scan of a tombstone badge 409s** naming the surviving id, and tombstones keep their original name (a "merged" badge carries the semantics).
- **`LIVE_PERSON` filter** (`lib/person/filters.ts`) applied at 15 verified surfaces — search, rosters, program counts, **the enrollment cap gate**, todo-counts, event rosters/RSVPs, certifications, and the match-audit enrollment sweep — so a retained tombstone row never inflates capacity or raises a false payment exception.
- **Nav link restored**: the hold comment's reason (blind-delete on collision downgrading certs/enrollments) is dissolved — nothing is deleted anymore; the dissolution is recorded where the hold was.

**Notes for review**
- The null-strand guard is implemented per spec but provably unreachable through the public API (server-recomputed conflicts + backfill-only-fills-empties always preserve a login identity); the test asserts the invariant on the most adversarial input instead.
- Follow-up (one line, not in this PR's file list): wire the exported `MergedBadge` into the household detail member cards.
- Explicitly out of scope: import-route hard-deletes (tracked separately), collision-row content reconciliation, undo, `src/security/*` tombstone scoping (boundary PR by rule).


thpr:
The problem. Hand-applied means forgettable. Every future query on Person that forgets ...LIVE_PERSON silently leaks a dead record — a tombstone appearing in a roster, inflating a headcount, tripping a capacity gate. Nothing errors. It just quietly shows a person who doesn't exist. And the 16th surface gets written six months from now by someone who never read this PR.

"By explicit choice" — the code says so directly, in filters.ts:

No grep-based drift guard by design: the daily match audit (lib/finance/matchAudit.ts) is the safety net that surfaces any missed surface.

So the author considered a guard, declined it, and named matchAudit as the replacement.

"matchAudit only catches payment-shaped drift" — that's my objection. matchAudit reconciles Shopify charges against activations. It notices a tombstone only if that tombstone produces a payment discrepancy. A tombstone sitting in an event roster, or padding a program's enrolled count, produces no payment mismatch at all. matchAudit never sees it. So the claimed net has holes exactly where most of the 15 surfaces live.

"pageRegistry-test pattern" — this repo already solved this class of problem once. Every new page.tsx must be registered in pageRegistry or pageRegistry.test.ts fails. A test that greps for the pattern and fails on unregistered instances. Same shape works here: grep for prisma.person.find* / include: { person... across src/, assert each hit either includes LIVE_PERSON or sits in an explicit allowlist with a justification. Cheap, and it fails in CI at the moment someone writes surface #16.

thpr:
Kiosk 409 (scan/route.ts:54) — member at the door gets hard-rejected because an admin cleaned up dupes last week. The "chain-proof" rationale is thin: chains form anyway (keeper A can later be merged into C), so you already need a walk. Forward the scan to the survivor, cap at ~5 hops, log it. Rejecting a human for an admin's action is backwards.

googleId radio renders "Connected" on both sides — admin picks blind. Show the account email.

DOB conflict sits as a plain radio next to Name. Different DOB = these may not be the same human. Warrants a warning, not equal billing with a typo'd name.

thpr (changes requested):
see comments

thpr:
Possible overdesign: tx.session.updateMany moves the tombstone's live auth sessions to the keeper. Deleting them (force re-login) is smaller and safer. No reason to inherit sessions.

checkin-app/src/app/membership-ops/participants/merge/page.tsx
```diff
@@ -13,11 +13,38 @@ interface ParticipantMergeView {
...
+    return isNaN(d.getTime()) ? String(value) : d.toLocaleDateString();
+  }
+  if (field === "phone" && typeof value === "string") return formatPhone(value);
+  if (field === "googleId") return value ? "Connected" : "—";
+  return String(value ?? "—");
+}
+
+// Reused wherever a tombstoned (merged-away) Person can surface in an admin view —
+// one badge component, not a new "merge status" subsystem (ponytail).
+export function MergedBadge({ person }: { person: { mergedIntoId?: number | null } }) {
```
thpr: Incorrect location for an export?  Shouldn't this be in a library not a page?

thpr (changes requested):

thpr:
Constraint for whoever implements the LIVE_PERSON drift guard, found while building it: **the sweep is not uniform — filtering the wrong call sites strands resources.**

The rule that has to hold is: *reconcilers and cleanup sweeps must still see tombstone rows; only human-facing lists and counts filter them.*

The merge deliberately leaves colliding rows on the tombstone (an enrollment that would collide with the keeper's stays put). Those rows can carry `inventoryHeldAt`. Three places exist specifically to resolve rows like that, and adding `LIVE_PERSON` to any of them makes a tombstone's leftovers immortal:

- `lib/lifecycleDrift.ts` — the invariant scan plus the I1 auto-heal that releases stranded `inventoryHeldAt`. Filter it and a tombstone's ACTIVE-with-held-inventory row is invisible to the only thing that would have healed it: Shopify inventory stays held indefinitely, and the System Status panel never surfaces it.
- `api/cron/pending-participants` — the 7-day non-payment clock; calls `withdrawAndReleaseHold`.
- `api/cron/scholarship-grace-expiry` — same, on the denial grace cutoff.

Filtering these looks correct in isolation (they're `programParticipant` queries, same shape as the ones that do need it), which is exactly why it's worth writing down. They belong in the guard's allowlist with the justification "sweep/reconciler — must see tombstone rows so their leftovers still get released", not in the filtered set.

Counter-example for the same model: `api/finance-ops/payment-plans` is a human approval queue, not a sweep, so that one does filter.

The emails those crons send go to household leads, which are filtered independently, so nothing dead-addressed leaks by leaving the sweeps unfiltered.

Suggest the allowlist carry this sweep-vs-display rule as a header comment, so the next entry is a mechanical decision rather than a judgment call.

jee7s:
On the LIVE_PERSON drift guard: **conceded — you're right and my rationale was wrong.** matchAudit only catches payment-shaped drift; a tombstone padding a roster or a headcount produces no payment mismatch and the claimed net has holes exactly where the surfaces live. The pageRegistry-pattern grep guard is the right shape, and since you're mid-build on it, we won't duplicate — two asks from this side: (1) carry your **sweep-vs-display rule as the allowlist's header comment** exactly as you framed it ("reconcilers and cleanup sweeps must still see tombstone rows; only human-facing lists and counts filter") — it turns every future entry into a mechanical decision; (2) your three named unfiltered sites (lifecycleDrift's I1 heal, pending-participants, scholarship-grace-expiry) match our disposition analysis — filtering any of them makes a tombstone's `inventoryHeldAt` leftovers immortal, so they belong in the allowlist with your justification verbatim. The filters.ts "no guard by design" comment will be deleted when your guard lands. Fixes for your other three items (scan forwarding with a 5-hop cap, informed googleId/DOB picker, sessions deleted not moved) are in flight on this branch — reply lands with the push.

jee7s:
Your other three items are fixed and pushed (390b9d8b), all gates green (2123 unit, 25/25 merge integration, scan suites unaffected):

- **Scan forwards to the survivor**: the 409 became a chain walk — up to 5 `mergedIntoId` hops, logged (badge/tombstone/survivor/hops), then the scan proceeds *as the survivor* through every downstream step; only a 6-hop-required chain still 409s with the reissue message. You were right that rejecting a human for an admin's cleanup was backwards.
- **googleId radio shows the identity** (each side's email, googleId tail as fallback) — no more picking blind between two "Connected"s. **DOB conflict is now wrapped in a warning** ("different birth dates may mean these are NOT the same person") rather than sitting as a peer of a typo'd name.
- **Sessions are deleted, not inherited** — commented as the deliberate auth-artifact exception to the no-deletion principle; the zero-rows-deleted invariant test now asserts the deletion explicitly. And `MergedBadge` moved out of the page into `components/ui/`.

thpr:
I stopped my build, I didn't realize it said it was going to do it... after last night's ping pong I'm in comment and let one side do the work

jee7s:
Understood — we've got it. The guard is building now on this branch to your spec: pageRegistry-pattern scanner over person list/count query sites, your sweep-vs-display rule verbatim as the allowlist header, your three sweep sites seeded with the leftovers-still-get-released justification, and the filters.ts "no guard by design" paragraph deleted as superseded. Any human-facing site the guard catches unfiltered gets fixed in the same push and called out in the reply. Sorry about the ping-pong — single-writer per surface from here.

jee7s:
The guard is in (a469808a), built to your spec — and your objection is now quantified: making it pass required fixing **~25 real human-facing sites** that were leaking tombstones, including a full unfiltered roster on the physical kiosk display, the board directory and contact surfaces, the CSV import silently matching/updating tombstones instead of live records, the annual BG-check sweep opening obligations against tombstoned identities, the admin roles list (no where clause at all), enrollment pickers, and three inflated nav counts. matchAudit would have seen none of them. Good call.

Design notes: your sweep-vs-display rule is the allowlist header verbatim; two of your three seeds are in with your justification (`lifecycleDrift` deliberately isn't — verified it never queries Person, only scalar `personId`, so seeding it would be a stale entry the guard's own second assertion rejects; documented in the header). Boundary: `findUnique` by unique key is out of scope by name (one row, not a list — this is also why merge/analyze/scan need no entries); id-scoped `findFirst`s are individually allowlisted. The sibling-filter detection uses a brace-depth walk to the enclosing call's own argument object after a flat lookback window produced both a false negative and a false positive — and the guard was validated live by reverting one fix and watching it fail on exactly that line. The filters.ts "no guard by design" paragraph is gone.

thpr (approved):

PR #1104, opened 2026-07-18
roles: PersonRole table as source of truth, asymmetric matrix, holders-first editor

jee7s:
Reworked per thpr's review (Jeff concurred) — the conscious decision, made before role #6 ships as a boolean:

- **`PersonRole` table is the source of truth**, with the grant metadata booleans could never hold (`grantedAt`, `grantedById`). Session/JWT claims are **derived** from the table at token refresh (zero extra queries — one added include), so `withAuth`, `BusinessRole`, and the ~200 session-claim consumers don't move. The four legacy boolean columns remain as deprecated write-through mirrors for existing readers; they drop in a later contract-phase PR. `isOperations` ships table-native — no column ever had a consumer. One migration: enum + table + backfill, verified three ways including a populated-DB simulation of the migration-safety gate.
- **The asymmetric matrix, table-ported**: board grants/revokes all five roles including sysadmin and board; sysadmin may grant anything including adding board but never removes board; last-board removal blocked under a `FOR UPDATE` lock on the role table (race-tested). Per-flag before/after audit rows.
- **Holders-first roles page** (`/membership-ops/roles`) is the **sole editor** — sections per role, person picker, the edit modal extracted verbatim to a shared component; the participants page keeps read-only pills with a Manage link (thpr's preference; one editor, one API).
- **Gap fixed**: `isOperations` was in `BusinessRole` but missing from `security/core`'s `Role`/`VALID_ROLES`/`callerHoldsRole` — a registry route gating on it would have thrown. Fixed, with a `BusinessRole ⊆ VALID_ROLES` guard test so the one non-compiler-checked plumbing step can't drift again.
- Also carries #1107's sign-in rebuild (merged into this stack) and #1108's RoleBadge de-dup (cherry-picked — its content had never flowed back to this branch).

Verified: bare tsc, CI-exact lint, 2141 unit + 1128 integration green (full matrix + race + backfill correctness + dual-write consistency), route coverage clean. Sequenced behind this PR: #1110 (outreach) and the small ops-access follow-on for the contacts endpoint.


checkin-app/prisma/migrations/20260718180437_person_is_operations/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "Person" ADD COLUMN     "isOperations" BOOLEAN NOT NULL DEFAULT false;
```
thpr: I'd be cautious about storing all the roles as booleans... do we want to go to a role table now before we add like 4 more roles? or do we want to add a boolean with each new role? - we should make this conscious not just Claude did something to meet the UI requriement

checkin-app/src/app/membership-ops/participants/page.tsx
```diff
@@ -23,6 +29,11 @@ type PersonRow = {
   isDeclaredAdult?: boolean;
   lastBackgroundCheck?: string | null;
   household?: HouseholdRef | null;
+  isSysadmin?: boolean;
```
thpr, line 29: Is this really the place we want to modify roles?  I really like the concept of a separate roles page vs a participant page with different flags visible.  Another point we should discuss

checkin-app/prisma/migrations/20260718180437_person_is_operations/migration.sql
jee7s: Conscious decision made (Jeff concurred with your caution): **we're converting to a role table now, in this PR** — rework in progress. Shape: `PersonRole { personId, role, grantedAt, grantedById }` as the source of truth (the grant metadata is the point — booleans could never hold who/when), with session/JWT claims **derived** from the table so the ~200-file consumer surface (withAuth gates, page predicates, session types) doesn't move. The four existing boolean columns stay as deprecated write-through mirrors so current readers are untouched; they drop in a later contract-phase PR once nothing reads them. isOperations ships table-native — its column never had a consumer. Bonus this analysis caught: isOperations was missing from security/core's VALID_ROLES (the one non-compiler-checked plumbing step) — fixed, with a BusinessRole⊆VALID_ROLES guard test so that class of drift can't recur.

checkin-app/src/app/membership-ops/participants/page.tsx
jee7s: Agreed, and it's in the same rework: role editing moves to a **dedicated holders-first roles page** (sections per role listing holders — answers 'who are our keyholders?' at a glance — plus a person picker), reusing the modal/matrix/confirm verbatim as a shared component. The participants page keeps the pills **read-only** for context; its edit button goes. One editor, one API — same single-surface principle we applied to the renewal send buttons.

jee7s:
Sequencing note: #1105 is being reworked to a **board-only** variant that targets main independently (Jeff's call — the contact-creation functionality shouldn't wait on the role-table review). A small **follow-on PR will stack on this one** to grant the operations role access to that functionality (role gates, directory admission with the ops-stripped search response, ops-403 tests) once the rework here lands. #1110 (outreach) remains sequenced behind this PR as before.

thpr:
Reviewed with an eye to whether this shape holds when the next four roles land. The table-as-truth call is right, and the compiler forces most of the per-role edit sites. Three things I'd fix before more roles go in:

**1. The invariants live in the route, not at the write choke point.**

`applyRoleFlag` is documented as the single write path, and it has three callers besides the API route — the first-user sysadmin bootstrap in `lib/auth-options.ts:356` and five grants in `lib/dev/seed-helpers.ts`. But the authority matrix (`api/roles/route.ts:106`), the `FOR UPDATE` serialization (:88), and the last-board-member guard (:129) all sit in the route, above that choke point. Any future writer — an import, a CLI, the outreach PR's ops-access follow-on — grants roles with no matrix check and no last-board protection.

With five roles and one editor that's tolerable. With nine roles and more grant surfaces it duplicates, and the copy that forgets the last-board guard is the one that bricks the org. Suggest pushing the guard down into a `grantRole()`/`revokeRole()` service that owns the lock + matrix + invariant, leaving the route as request parsing and audit.

**2. The `isX` boolean vocabulary is the debt that four new roles doubles.**

The table is `PersonRoleKind`-native, but every layer above it still speaks boolean claim names: `RoleFlag`, `BusinessRole`, `security/core`'s `Role`, the JWT claims, `ROLE_META`. `applyRoleFlag(db, id, "isKeyholder", true)` translates a boolean name back into a kind on every call.

That's a deliberate expand-phase choice and it's the right one for this PR — I'm flagging the sequencing. If the four new roles land before the contract PR, we write four more `isX` names into all of those places and then rename nine instead of five. Either the contract PR goes first, or we consciously accept that role names are permanently `isX`-shaped and stop treating the rename as pending.

Related, and the one silent-drift spot the PR leaves: `ROLE_FLAGS` (`lib/roles.ts:17`) is a hand-maintained array sitting next to the `RoleFlag` union. Adding a role to the union and forgetting the array compiles clean — the exact class of gap that let `isOperations` miss `VALID_ROLES`. Make `FLAG_TO_KIND` the single literal and derive the rest (`Object.keys(FLAG_TO_KIND) as RoleFlag[]`, `rolesToFlags` as a loop over it). Same treatment for the five hand-written lines in `assignParticipantClaims` and the three duplicated five-key literals in `RolesEditModal` (lines 43, 51, 89, 113). Removes roughly half the per-role edit sites and the only unguarded one.

**3. Audit rows are mis-labeled.**

`api/roles/route.ts:151` writes `tableName: "Participant"`. The write now targets `PersonRole`, and `Participant` is the pre-rename model name besides — so role-grant audit rows point at a table that no longer exists under that name. Should be `PersonRole`.

thpr:
Re-reviewed at 87dfc2c9. All three points landed, and the choke-point fix in particular is the right shape — `setRoleFlag` owning the lock, the matrix, and the last-board guard, with `"system"` bypassing the matrix but deliberately *not* the guard, is exactly the split I was hoping for. Nested `withTx` joining the route's outer transaction means the per-flag loop re-acquires a lock it already holds, which is a no-op; that's correct.

Two of the fixes each left one hole behind, plus a smaller thing in the seeds.

**A. The session callback is now the only hand-listed flag enumeration left in the claim path.**

`assignParticipantClaims` got the `ROLE_FLAGS` loop; `auth-options.ts:412-416` did not — it still assigns the five flags to `session.user` by hand. So adding flag #6 to `FLAG_TO_KIND` and `next-auth.d.ts` and forgetting this block gives you a JWT that carries the claim and a session that drops it: every `session.user.isNewRole` reads `undefined`, which is falsy, which is a silent deny with no compile error.

That's the same drift class the rest of the commit set out to kill — it's just concentrated in one place now instead of spread across six, which arguably makes it easier to miss. Same `for (const flag of ROLE_FLAGS)` treatment should work, assuming `SessionUser` takes the indexed write without a cast.

**B. `applyRoleFlag` is still exported, and still bypasses both invariants.**

After this commit it has zero callers outside `setRoleFlag` itself (roles.ts:157). Leaving it exported means the choke point is a convention rather than a boundary — the doc comment says every writer routes through `setRoleFlag`, but nothing stops a future caller from importing `applyRoleFlag` and getting the dual-write without the matrix or the last-board guard. Making it module-private moves that from "documented" to "enforced by the module system", which is the whole point of having consolidated the invariants.

**C. `seedRole` hands `setRoleFlag` a current-state it never read.**

seed-helpers.ts:33 passes `roles: []` unconditionally. Harmless as written — seeds only grant, and the underlying write is an idempotent upsert — but it means the no-op short-circuit is evaluating against a fabricated state. A seed that ever revokes would hit `current[flag] === on` with both sides false and return early without writing anything, silently.

Worth considering having `setRoleFlag` take a `personId` and read the roles itself inside the transaction, rather than accepting a caller-supplied `target`. That fixes this and also closes a smaller gap on the API path: the route reads `target.roles` before the lock is taken, so the no-op check and the delta computation both run against a pre-lock snapshot. The last-board guard re-counts under the lock so the critical invariant is safe either way — this is about the read being consistent with the write, not about a live race.

jee7s:
All three fixed and pushed (87dfc2c9), full gates green (2141 unit / 1128 integration):
- **Invariants now live at the write choke point**: new `setRoleFlag()` in lib/roles.ts owns the FOR UPDATE lock, the matrix, and the last-board guard; the route is parsing + audit. Bootstrap and seeds route through the same function with an explicit, commented system-actor bypass — which skips the matrix but **never** the last-board guard.
- **One hand-written literal**: `FLAG_TO_KIND` is now the single source; `RoleFlag`/`ROLE_FLAGS`/claims/modal all derive from it, so adding role #6 has exactly one unguarded edit. On your sequencing point: we're consciously accepting `isX` as the permanent claim vocabulary — renaming JWT claims buys churn without capability, and the derivation work just removed most of the places the shape could hurt. Flag it if you disagree.
- Audit `tableName` → `PersonRole`.

jee7s:
Pipeline fixed: the isolation red was this PR's hand-edits to src/security — extracted to a dedicated boundary PR per the AGENTS.md rule (see the new security-role-vocab PR; 13 lines, merge it first). This branch reverted those edits (isolation now passes) and temporarily dropped the ⊆-guard test, which returns via a rebase once the boundary PR lands — both halves of the check need to coexist. All gates green at e786e527 (2136 unit / 1128 integration).

jee7s:
#1114 merged, so the two-step is complete on this branch: merged main past it (clean — the branch had pre-reverted its security edits) and restored `src/security/__tests__/coreRoles.test.ts`, the BusinessRole ⊆ VALID_ROLES vocabulary guard that needed both halves in one tree. It passes with `isOperations` in both vocabularies. Head `8d4f1736`; full gates green locally (tsc, lint, 2141 unit / 1128 integration).

thpr:
I don't think my second set of feedback is addressed?

jee7s:
You're right — round 2 hadn't landed yet. It has now (head `e8ebb00e`, which also merges main post-#1103):

**A (session callback drops flag #6).** The session callback now runs the same `for (const flag of ROLE_FLAGS) session.user[flag] = token[flag]` loop as `assignParticipantClaims`. No cast needed — `Session["user"]` and `JWT` declare the identical five `boolean | undefined` flags, so the indexed write typechecks (mirror of the existing `token[flag]` loop). Flag #6 lands in `FLAG_TO_KIND` + `next-auth.d.ts` and both JWT and session pick it up with no hand-listed line to forget.

**B (`applyRoleFlag` still exported).** Confirmed zero callers outside `roles.ts` (repo-wide, tests included) and dropped the `export`. The choke point is now enforced by the module system, not a doc comment.

**C (fabricated current-state / pre-lock snapshot).** `setRoleFlag` now takes a `personId` and reads the person's roles itself, inside the tx *after* the `FOR UPDATE`, returning `{changed, before, after}`. `seedRole` drops its `roles: []`; the bootstrap and the route pass the id. The route builds delta + audit from the returned post-lock values and reads the final flag set once more under the same lock for the response — no-op check, audit, and write are all consistent with one under-lock read. (As you said, the last-board guard already re-counted under the lock, so this was read-consistency, not a live race. The trade: a no-op now takes the board lock before returning — acceptable, the board row set is tiny.)

Also in this push: `coreRoles.test.ts` moved out of `src/security/__tests__/` to `src/__tests__/` — a hand-authored file under `src/security/**` trips the boundary-isolation gate even as a test, which is what failed the last run. Pure `git mv`, alias imports unchanged. Gates locally: tsc/lint clean, 2152 unit / 1148 integration passing, drift guard green with this branch's two new person queries filtered via the merge resolutions.

thpr (approved):
Approving. All three round-2 items are fixed, and two of them are fixed more thoroughly than I asked for.

- **A (session callback)** — now loops `ROLE_FLAGS`. Because `RoleFlag` derives from `FLAG_TO_KIND`, adding a sixth flag without declaring it in `next-auth.d.ts` is a compile error on the index rather than a silent `undefined`. The drift path is closed, not relocated.
- **B (`applyRoleFlag`)** — private; `setRoleFlag` is the only reachable writer, enforced by the module boundary instead of a doc comment.
- **C (pre-lock read)** — `setRoleFlag(db, personId, ...)` takes the lock, then reads that person's roles under it, and returns `{changed, before, after}`. The route builds the delta *and* the audit row from those returns, and the response from a post-write read. That also fixes something I didn't call out: the audit row's before/after was previously derived from the same pre-lock snapshot.

Also checked: audit is skipped when nothing changed and written after all `setRoleFlag` calls, so a matrix or last-board throw on the second flag rolls back the first with no partial audit row. Lock ordering is invariant (BOARD set, always first), so the per-flag re-acquire can't deadlock. Merge to main is clean.

Two follow-ups, neither blocking, both needing a boundary-isolated PR so they can't ride along here:

1. `access-resolvers.ts:169` still has `(auth.user as { isOperations?: boolean })`. #1114 left it with a comment saying it expires when roles-foundation lands — it has now landed, and `types/participant.ts` declares the field. A cast that defeats type-checking on a security-boundary read shouldn't outlive its stated condition.
2. `coreRoles.test.ts` now lives in `src/__tests__/` to satisfy the isolation gate, two directories from the `VALID_ROLES` it guards — and it was already deleted once in this PR's history. A one-line pointer next to `VALID_ROLES` would keep the next person from re-orphaning it.

Note on what I did and didn't verify: I read the diff and the merged tree, and confirmed the invariants and the module boundary by inspection. I did not run the test suites myself — CI's integration, flow, migration, and isolation jobs are green, and the unit job was still pending when I approved.

Issue #1083, opened 2026-07-17
All tabs in finance-ops should be board only


thpr:
Should not be accessible to admin users  - only board

PR #1105, opened 2026-07-18
contacts: email-only account creation for board (ops follow-on after RBAC)

jee7s:
**Board-only variant targeting main independently** (Jeff's call — reworked from the ops-stacked original; see the scope-change comment below for history).

- `POST /api/membership-ops/contacts`: name + email → household shell + lead via the same builder first sign-in uses, so shells are claimable through the live email-linking path and appear in the unclaimed-households surfaces. Gated `['isBoardMember']` — **sysadmin deliberately excluded** (#1083 precedent; don't "fix" the roles array).
- The duplicate check **normalizes email** the way the write extension does, then 409s naming the owner (deliberate, documented for future audits — callers are already directory-trusted); the existing rich-create route's un-normalized check is the known latent 500 this refuses to copy, pinned by a case-variant test.
- Add-contact modal on the participants page (board-visible); audit CREATE row per creation.

**Follow-on after #1104 (role-table RBAC) merges**: a small stacked PR re-grants this to the operations role — the exact stripped inventory (ops in the route gate, layout/pageRegistry admission, ops-stripped people/search, the ops-403 test block, ops read-only directory RTL).

Verified: bare tsc, CI-exact lint, 2119 unit green, contacts integration 5/5 on real Postgres.


checkin-app/src/app/api/membership-ops/contacts/route.ts
```diff
@@ -0,0 +1,53 @@
...
+        // Dup check MUST normalize the key the same way the write extension lowercases
+        // (prismaEmailNormalize.ts) — else a case-variant misses here and 500s on the
+        // @unique constraint instead of returning this owner-named 409.
+        const existing = await prisma.person.findUnique({
+            where: { email: normalizeEmail(email) },
+            select: { name: true },
+        });
+        if (existing) {
+            return NextResponse.json(
+                { error: `This email already belongs to ${existing.name ?? 'an existing account'}`, fields: ["email"] },
```
thpr, line 36: Claude should probably comment that this is legal 409 ONLY because ops is trusted with this info that it's a duplicate - this is not a security leak.  This is likely to show up in a code audit at some point

jee7s: Done in 0a9562db — the comment now sits directly above the 409, stating it's deliberate (every caller role is already trusted with the full directory via people/search; the 409 reveals nothing they can't query) and flags it for future audits with an explicit do-not-redact.

thpr (dismissed):

jee7s:
**Scope change per Jeff: this PR is being reworked to a board-only variant targeting main independently.** The functionality (email-only contact creation via the shared first-sign-in builder, the normalized duplicate check that fixes the latent case-variant 500, the owner-named 409 with its audit note, the add-contact modal) lands now, gated `['isBoardMember']` — sysadmin remains deliberately excluded per the standing decision. Every `isOperations` reference is being stripped: role gates, layout/pageRegistry admission, the ops-stripped people/search response, and the ops-403 tests.

**A follow-on PR will re-grant this functionality to the operations role once the RBAC change (#1104's role-table rework) merges** — it will stack on #1104 and contain exactly the stripped inventory. Rework is in flight; base retargets to main when it force-pushes green.

thpr (commented):
Design is right — reusing `createParticipantWithHousehold` (not a hand-rolled create) is what makes the shell claimable and makes it fall out of `UNCLAIMED_OR_BROKEN_HOUSEHOLD_WHERE` for free, and the integration test asserts exactly that. Role gate checks out against `src/lib/auth.ts`; sysadmin exclusion is deliberate and tested, not flagging it.

Four items, one of them blocking.

---

### 1. Whitespace-padded email defeats both the dup check and the `@unique` constraint — medium

Same bug class this PR exists to prevent, one notch over from case.

- `isValidEmail` trims *before* testing (`src/lib/emergencyContacts/identity.ts:27`), so `" Foo@x.com "` passes validation.
- `normalizeEmail` only lowercases — no trim (`src/lib/prismaEmailNormalize.ts:31`). The dup lookup keys on `" foo@x.com "`, misses the stored `foo@x.com`. No 409.
- The raw (untrimmed) `email` is then passed to `createParticipantWithHousehold`. The write extension lowercases but also doesn't trim, so `" foo@x.com "` gets stored — a *different* string, so `@unique` doesn't fire either. Duplicate account, silently, no error.

Worse than a duplicate row: `getUserByEmail` normalizes Google's address to `foo@x.com` (`src/lib/auth-options.ts:99`) and will never match that shell. It's permanently unclaimable — which is the one thing this endpoint exists to produce.

The modal itself is safe (HTML `type="email"` strips surrounding whitespace on sanitization), but the endpoint is directly callable and this is the API's own validation contract.

Minimum fix — trim once at the top and use that value for both the lookup and the create:

```ts
const email = typeof rawEmail === 'string' ? rawEmail.trim() : rawEmail;
```

Root-cause fix — make `normalizeEmail` do `email.trim().toLowerCase()`. It's the declared single choke point for read *and* write, the trim hole is in the extension too, and every current and future write site gets the fix for free. Worth noting `cleanEmail` in that same identity module already does trim+lowercase — the codebase has the right function, this path just isn't reaching for it.

A whitespace-variant test alongside the existing case-variant one would fail today.

---

### 2. Dup check → create is not atomic — low

Two concurrent posts (or a race against a live Google sign-in on the same address) land a P2002 in the catch and return the generic 500, not the owner-named 409 the endpoint promises. Low volume here so genuinely low severity, but the fix is ~3 lines in the catch you already have:

```ts
if (error instanceof Prisma.PrismaClientKnownRequestError && error.code === 'P2002') {
    // lost the race — same answer the pre-check would have given
    return NextResponse.json({ error: "This email already belongs to an existing account", fields: ["email"] }, { status: 409 });
}
```

---

### 3. Optimistic row renders "No household" — low

`data.participant` is a bare `Person`; the list row type wants `household?: HouseholdRef | null` and renders `p.household?.name || "No household"`. So the prepended row claims no household immediately after creating one named after the person, until the next search refresh.

The RTL test mocks `participant: { ..., household: null }`, which encodes the mismatch rather than catching it. Either `include: { household: true }` on the response, or drop the optimistic prepend and re-run the search.

---

### 4. `useRequireRole([])` just to read the session — nit

Empty `allowed` means "any authenticated user", but the hook still installs its redirect effect on a page the ops layout already gates. `useSession()` is the smaller tool for "give me the role flags".

---

Tests are otherwise strong — real-Postgres coverage of happy path, case-variant 409, sysadmin 403, anon 401, plain 403, plus the audit row and the unclaimed-households predicate. The two-tag fixture lifetime, the `jest.unmock('@/lib/auth-options')` note, and the single-stable-`fetch`-mock reasoning about the 250ms debounce race are all correct and well-explained. Only other gap: no 400 coverage for blank name / malformed email.

The 409 name disclosure is fine as documented — board already holds full directory read.

thpr (commented):
Re-reviewed against `7a46ccab`. Three of the four are clean; the fix for #3 traded a cosmetic bug for a crash.

---

### 1. Whitespace email — fixed, root-cause version ✅

You took the `normalizeEmail` → `trim().toLowerCase()` option rather than patching the route. Blast radius checked: only three callers (`auth-options.ts:99`, `auth-options.ts:331`, and `lowerEmailOnData` inside the extension), and all three want trimming. The extension inherits it automatically, so every write site is covered rather than just this one.

The `stores a whitespace-padded email trimmed` test is the valuable one — it empirically proves the extension still fires inside `createParticipantWithHousehold`'s `$transaction`, which the file's own KNOWN LIMITATION note asserts but nothing previously pinned.

### 2. P2002 race — fixed ✅

Guard sits before `logBackendError`, so a lost race doesn't page anyone. Dropping the owner name in that branch is the right trade — re-querying to name them would just re-open the same race.

### 4. `useRequireRole([])` → `useSession()` — fixed ✅

---

### 3. The optimistic-prepend fix introduces a render crash — blocking

`route.ts:57` re-fetches with:

```ts
include: { household: true }
```

Prisma returns Household's **scalar fields only** — no `householdMembers`. But the client's `HouseholdRef` declares `householdMembers` as non-optional (`page.tsx:14`), and `page.tsx:176` computes at component top level, unconditionally, on every render once a participant is selected:

```ts
const canSubmitAssign = !selectedParticipant?.household || (selectedParticipant.household.householdMembers.length > 1);
```

Click **Assign** on the row you just added (`page.tsx:240` → `setSelectedParticipant(p)`) and `household` is truthy while `householdMembers` is `undefined` → `TypeError` during render. White screen. `page.tsx:270`'s `.householdMembers.map(...)` is the same story, and `canChangeHousehold` on :177 too.

Two reasons this slips through:

- tsc is blind to it — the response crosses an untyped JSON boundary, so nothing reconciles the route's actual `include` against `HouseholdRef`.
- The RTL mock now supplies `household: { id: 77, name: "Ada Lovelace", householdMembers: [...] }`, which is the shape the *type* wants but not the shape the *route* returns. The test stays green while prod crashes. The `// Honest shape:` comment is the one thing in this commit that isn't — worth fixing the comment along with the mock, since it'll otherwise read as verified next time someone touches this.

Fix:

```ts
const participant = await prisma.person.findUnique({
    where: { id: person.id },
    include: { household: { include: { householdMembers: { select: { id: true, name: true, email: true } } } } },
});
```

Then make the mock mirror that exactly. Alternative, if the extra nesting isn't worth it: drop the optimistic prepend and re-run the existing search — one line, no shape contract to keep in sync, at the cost of a round trip.

Worth a quick grep for other consumers of this response shape before settling on either.

---

### Minor

`normalizeEmail`'s contract changed, but `src/lib/__tests__/prismaEmailNormalize.integration.test.ts` got no padded-email case — the trim is currently only proven through the contacts route. That file is where the next person looks for the choke point's contract; one case there would keep it honest.

jee7s:
All four + the test gap fixed, pushed (7a46ccab), gates green (2119 unit / 1125 integration):
- **The whitespace hole, fixed at the root you pointed at**: `normalizeEmail` now trims before lowercasing — the single choke point covers the dup lookup, the write extension, and `getUserByEmail`, so the permanently-unclaimable-shell failure mode is closed everywhere at once. Your whitespace-variant test would have failed before; it passes now, alongside a stored-trimmed create test.
- P2002 race → the owner-style 409 (no more generic 500 for the race loser); the optimistic row now carries its household (route includes it; RTL mock made honest); `useRequireRole([])` → `useSession()`; blank-name and malformed-email 400 tests added.

checkin-app/src/app/api/membership-ops/contacts/route.ts
jee7s: Agreed on the audit-trail need — this landed with the board-only rework and is in the code at route.ts:30-33, in exactly those terms:

```ts
// Deliberate, not a leak: every caller role here (board) is already
// trusted with the full people directory and can look up this same
// name+email via people/search. The 409 reveals nothing they can't already
// query. Flagged for future audits — do not redact.
```

(The P2002 fallback catch a few lines down deliberately returns the same 409 shape for the TOCTOU race, same reasoning.) When #1111 later widens this route to operations, the comment's premise still holds — ops is granted directory read as part of that same PR, so the trusted-with-directory invariant travels with the access change.

jee7s:
Item 3 fixed at `02c18567`, nested-include version:

- The re-fetch now includes `household → householdMembers { id, name, email }`, matching what `HouseholdRef` demands, with a comment explaining why the bare include crashed the Assign render path.
- The RTL mock is pinned to the route's *actual* shape and the `// Honest shape:` comment now tells the truth (it previously claimed `household: true` above a mock that already carried `householdMembers` — your point exactly). The board-session test now walks the real crash path — create → Details → Move to Another Household — and asserts the Assign modal renders; that assertion throws pre-fix.
- Consumer grep done before settling: the POST response is consumed only by `page.tsx` (`authzRegistry.test.ts` references the route path but not the body; `HouseholdRef` is file-local). Additive shape, nothing else to reconcile.
- Your minor too: `prismaEmailNormalize.integration.test.ts` now has the padded-email create case, so the trim contract is proven at the choke point rather than only through this route.

Gates: tsc/lint clean, 2227 unit / 1182 integration passing.

thpr (commented):
Re-reviewed against `02c18567` (+ the two main merges). **All four original items are closed.**

#3 is fixed properly: the nested `include` matches `HouseholdRef` field-for-field, the mock was corrected, and the RTL test now drives the actual regression path — create → prepend → Details → "Move to Another Household" → assert the modal renders, which exercises `canSubmitAssign`/`canChangeHousehold`, the two top-level computations that threw. Documenting *why* it goes through Details (a solo household surfaces the Household button, not Assign) is the kind of note that stops someone simplifying the test back into uselessness. The `prismaEmailNormalize` trim case at the choke point covers the minor.

Merge hygiene is clean — no conflict markers in `checkin-app/src`, and both merges are main catch-ups that don't touch this PR's own files.

Two follow-on observations. Neither blocks.

---

### 1. The route's response shape still isn't pinned by a real test

`route.integration.test.ts` asserts against the database — it re-queries with `prisma.person.findUnique(...)` and never inspects `data.participant.household`. So the only place the serialized response shape is written down is the hand-maintained RTL mock in `page.test.tsx`.

Concretely: revert line 61 back to `include: { household: true }` today, and the full suite stays green. The RTL test would be asserting against a shape the route no longer returns, which is exactly the hole the original bug came through — the mock agreed with the client's *type* while the route disagreed with both.

One line in the happy-path integration test closes it permanently, at the only layer that sees the real serialized payload:

```ts
expect(data.participant.household.householdMembers).toHaveLength(1);
```

Worth adding before merge, given it's the specific guard against this regression recurring.

### 2. The "waiting on RBAC" comments are now stale

The main merge brought `isOperations` onto this branch — it's in `BusinessRole` (`src/types/auth.ts:7`), `SessionUser` (`src/types/participant.ts:15`), the next-auth augmentation, and `src/lib/roles.ts`. But `route.ts:12` and the matching comment in `page.tsx` both still say the grant lands "once the RBAC rework (the isOperations role) merges" — describing as pending something that is now present in this very tree.

The grant itself correctly lives in #1111, so this is comment accuracy, not behavior. Flagging it because these are the same comments carrying the "do NOT add 'isSysadmin' here" instruction — a future dev who catches one half being wrong has less reason to trust the other half. Suggest something like "Operations is granted separately in #1111" so the pointer stays true after this merges.

---

Nothing else outstanding from my side. Nice work on the #3 fix — the added Details→Assign traversal is a better test than the one-line `include` change strictly needed.

jee7s:
Both follow-ons in, `6b23c7ff` + `7bfe46f3`:

**Response shape pinned at the integration layer.** The happy-path test now asserts on the serialized response itself — `data.participant.household.householdMembers` has length 1 and the member matches `{id, name, email}`. Reverting the nested include now fails this test at the only layer that sees the real payload, which closes the mock-agrees-with-the-type-but-not-the-route hole for good.

**Stale RBAC-pending comments swept — all three.** `route.ts`, `page.tsx`, and the test file's own docblock (a third instance of the same phrase) now say Operations access is granted separately in #1111, with the sysadmin-exclusion instruction and #1083 precedent untouched. The pointer stays true after this merges.

thpr (approved):
Approving at `7bfe46f3`. Every item from the four review passes is closed:

| Item | Resolution |
|---|---|
| Whitespace email bypassed both the dup check and `@unique` | Fixed at the `normalizeEmail` choke point, so the write extension inherits the trim; pinned at both the route and the choke point |
| Dup check → create TOCTOU returned 500, not 409 | P2002 guard, placed before `logBackendError` so a lost race doesn't page anyone |
| `include: { household: true }` TypeError'd the Assign render path | Nested include matching `HouseholdRef`, plus a regression test driving the real create → prepend → Details → Assign traversal |
| `useRequireRole([])` used only to read the session | `useSession()` |
| Response shape unpinned by any real test | Integration assertion on the serialized payload, extended to prove the sole household member is the created person |
| Stale "once the RBAC rework merges" comments | Swept to zero, all three sites now point at #1111 |

The two things I'd call out as better than what was asked for: the #3 fix shipped the full click-path regression test rather than just the one-line `include` change, and the response-shape assertion checks member identity rather than just array length — so it catches a wrong-household regression too, not only a missing-array one.

Note on my basis for approval: this is static review — I read the diff and the surrounding code (the `withAuth` gate, `normalizeEmail`'s callers, `people/search`'s response shape, the `routeAuthDrift` rules) but did not execute the test suites. Unit/integration/flow were still pending on the last two commits when I approved, and this suite has already had one laptop-vs-CI divergence (`5cffd1a7`, missing Google/NextAuth env in CI), so the new integration assertions are worth a glance when they go green. Merge gating covers it.

PR #1110, opened 2026-07-18
outreach engine: templated renewal/join emails, ledger-drain delivery, unsubscribe

jee7s:
PR-2 — the final piece of the interviewed outreach/RBAC series (interview → adversarial challenge → reviewed opus spec → sonnet build → architect review).

**⚠️ Contains #1103's commits until it merges**: this branch is #1104's + #1103's branches merged (both are hard dependencies — `isOperations` and `LIVE_PERSON`). Merge order: #1101→#1102→#1103, #1104, then this — at which point this PR's diff collapses to just the outreach work. The assumed #1103 SHA (6d03626) was verified unchanged at ship time; if #1103 moves before merging, I reconcile.

- **Two templated emails** (Opening of renewals + Reminder) with `{{name}}/{{deadline}}/{{actionWord}}/{{actionLink}}` and per-recipient join-vs-renew; a new Settings → Outreach page (board + sysadmin + operations) with live both-variant preview, send-test-to-me, unknown-token rejection, and a hard block when the membership-year boundary is unset.
- **Delivery that survives deploys**: no cron, no background job — a client-driven chunked drain over a per-recipient ledger. Recipients are snapshotted at create (audience presets: unsettled+non-members / also-skip-in-flight / everyone — with the off-season "nobody is settled" behavior stated in the UI copy next to the count); each item is claimed (`queued→sending`) *before* `sendEmail` for at-most-once delivery under concurrent batch calls; abandoned sends resume; failures retry; one in-flight send at a time is a Serializable-transaction guarantee. The ledger doubles as the last-sent panel.
- **Unsubscribe (join variant only)**: HMAC-signed token; GET shows a confirmation page only — mail scanners prefetch GETs — POST flips the suppression flag; members re-enable from the communication page; an "unsubscribed" pill in the directory keeps exclusions legible. Renew recipients are never suppressed.
- **The machine stops emailing about renewals**: the cron's remind path and the bulk-open button's `sendReminders` option are gone (process opening unchanged) — the outreach page is the only send surface, per the manual-communication model.
- One additive migration; every new scalar sensitivity-tagged; no `src/security` policy edits.

Verified: bare-exit tsc, CI-exact lint, **2178 unit + 1166 integration** green locally against a real Postgres — including the concurrent no-double-send race, resume-after-abandon, preset semantics in and out of season, and the GET-doesn't-flip unsubscribe cases.

### DIFF
--- a/checkin-app/src/app/api/outreach/process-batch/route.ts
+++ b/checkin-app/src/app/api/outreach/process-batch/route.ts
@@ -60,7 +60,14 @@ export const POST = withAuth({ roles: ROLES }, async (req) => {
         const claim = await prisma.bulkSendItem.updateMany({ where: { id: item.id, status: "queued" }, data: { status: "sending" } });
         if (claim.count === 0) continue; // already claimed elsewhere — skip, don't send
 
-        const person = await prisma.person.findUnique({ where: { id: item.personId }, select: { name: true } });
+        const person = await prisma.person.findUnique({ where: { id: item.personId }, select: { name: true, emailSuppressed: true } });
+        // Honor an unsubscribe that landed AFTER the snapshot was frozen: the drain resumes
+        // arbitrarily later (survives deploys), so re-check suppression at send time. Join-only,
+        // recorded (not dropped) as the same status the snapshot path uses — spec §2.4.
+        if (item.variant === "join" && person?.emailSuppressed) {
+            await prisma.bulkSendItem.update({ where: { id: item.id }, data: { status: "skipped_unsubscribed" } });
+            continue;
+        }
         const { subject, html } = renderOutreachEmail(template.subject, template.body, {
             name: person?.name ?? "",
             variant: item.variant as OutreachVariant,
--- a/checkin-app/src/app/api/outreach/process-batch/route.ts
+++ b/checkin-app/src/app/api/outreach/process-batch/route.ts
@@ -1,0 +1,85 @@
+import { NextResponse } from "next/server";
+import prisma from "@/lib/prisma";
+import { withAuth } from "@/lib/auth";
+import { apiError } from "@/lib/api-response";
+import { sendEmail } from "@/lib/email";
+import { renderOutreachEmail, resolveBoundary, loadTemplate, BoundaryNotSetError, type EmailType, type OutreachVariant } from "@/lib/outreach/render";
+import type { BusinessRole } from "@/types/auth";
+
+export const dynamic = "force-dynamic";
+
+const ROLES: BusinessRole[] = ["isBoardMember", "isSysadmin", "isOperations"];
+const BATCH_SIZE = 40; // within the spec's ~25-50/round
+
+/**
+ * POST /api/outreach/process-batch — claim and drain one batch of a BulkSend's `queued`
+ * items. The browser calls this in a loop until `remaining` is 0 (see the send panel).
+ *
+ * No double-send: each candidate item is claimed with a guarded `updateMany where
+ * id=<id> AND status='queued'` BEFORE it is sent — the claim must land before sendEmail,
+ * not after, because the ledger can only prevent a double *write*, not a double *send*, if
+ * the guard ran afterward. A losing claim (count 0, e.g. a retried/duplicate POST racing
+ * another in-flight process-batch call for the same item) is simply skipped.
+ *
+ * ponytail: the claim writes a transient "sending" status — not one of the four the design
+ * doc names (queued/sent/failed/skipped_unsubscribed) — as the atomic marker, since Prisma's
+ * updateMany can't return which rows it touched. A crash in the narrow window between claim
+ * and finalize could strand a single item in "sending" (not resumable by count-of-queued or
+ * retry-failed); add a stale-"sending" reaper if that's ever observed in practice.
+ */
+export const POST = withAuth({ roles: ROLES }, async (req) => {
+    let body: { bulkSendId?: number } = {};
+    try {
+        body = await req.json();
+    } catch {
+        // empty body is fine — falls back to the current in-flight send below
+    }
+
+    const bulkSend = body.bulkSendId
+        ? await prisma.bulkSend.findUnique({ where: { id: body.bulkSendId } })
+        : await prisma.bulkSend.findFirst({ where: { completedAt: null }, orderBy: { id: "desc" } });
+    if (!bulkSend || bulkSend.completedAt) return apiError("No in-flight send.", 404);
+
+    let boundary;
+    try {
+        boundary = await resolveBoundary(new Date());
+    } catch (e) {
+        if (e instanceof BoundaryNotSetError) return apiError(e.message, 400);
+        throw e;
+    }
+    const template = await loadTemplate(bulkSend.emailType as EmailType);
+
+    const candidates = await prisma.bulkSendItem.findMany({
+        where: { bulkSendId: bulkSend.id, status: "queued" },
+        orderBy: { id: "asc" },
+        take: BATCH_SIZE,
+    });
+
+    let processed = 0;
+    for (const item of candidates) {
+        const claim = await prisma.bulkSendItem.updateMany({ where: { id: item.id, status: "queued" }, data: { status: "sending" } });
+        if (claim.count === 0) continue; // already claimed elsewhere — skip, don't send
+
+        const person = await prisma.person.findUnique({ where: { id: item.personId }, select: { name: true } });
+        const { subject, html } = renderOutreachEmail(template.subject, template.body, {
+            name: person?.name ?? "",
+            variant: item.variant as OutreachVariant,
+            boundary,
+            personId: item.variant === "join" ? item.personId : undefined,
+        });
+
+        const ok = await sendEmail(item.email, subject, html);
+        await prisma.bulkSendItem.update({
+            where: { id: item.id },
+            data: ok ? { status: "sent", sentAt: new Date() } : { status: "failed", error: "sendEmail failed" },
+        });
+        processed++;
+    }
+
+    const remaining = await prisma.bulkSendItem.count({ where: { bulkSendId: bulkSend.id, status: "queued" } });
+    if (remaining === 0) {
+        await prisma.bulkSend.update({ where: { id: bulkSend.id }, data: { completedAt: new Date() } });
+    }
+
+    return NextResponse.json({ bulkSendId: bulkSend.id, processed, remaining });
+});
--- a/checkin-app/src/app/api/outreach/status/route.ts
+++ b/checkin-app/src/app/api/outreach/status/route.ts
@@ -1,0 +1,60 @@
+import { NextResponse } from "next/server";
+import prisma from "@/lib/prisma";
+import { withAuth } from "@/lib/auth";
+import { apiError } from "@/lib/api-response";
+import type { BusinessRole } from "@/types/auth";
+import type { BulkSend } from "@/generated/prisma/client";
+
+export const dynamic = "force-dynamic";
+
+const ROLES: BusinessRole[] = ["isBoardMember", "isSysadmin", "isOperations"];
+const STATUSES = ["queued", "sent", "failed", "skipped_unsubscribed", "sending"] as const;
+
+async function countsFor(bulkSendId: number) {
+    const groups = await prisma.bulkSendItem.groupBy({ by: ["status"], where: { bulkSendId }, _count: { _all: true } });
+    const counts: Record<(typeof STATUSES)[number], number> = { queued: 0, sent: 0, failed: 0, skipped_unsubscribed: 0, sending: 0 };
+    let total = 0;
+    for (const g of groups) {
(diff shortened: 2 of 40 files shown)


checkin-app/prisma/migrations/20260718120000_person_merged_into/migration.sql
```diff
@@ -0,0 +1,5 @@
+-- AlterTable
+ALTER TABLE "Person" ADD COLUMN     "mergedIntoId" INTEGER;
```
thpr, line 2: This is merge notrenewal/join, so this is a mixed PR, can we separate it so the "what needs prisma changes" is better isolated?

thpr (changes requested):

jee7s:
On the mixed-PR concern (the `person_merged_into` migration showing up here): that migration is **#1103's**, not this PR's — this branch is deliberately built on #1103 + #1104 because both are hard dependencies (`mergedIntoId`/`LIVE_PERSON` keep tombstones out of bulk email; `isOperations` gates the send surface). GitHub can only diff against one base, so until those merge, their commits render as part of this diff.

**The isolation you're asking for arrives via merge order**: land #1101→#1102→#1103, then #1104 — this PR's diff then collapses to outreach-only, with exactly **one** prisma change of its own (`20260718194028_outreach_engine`: the four template columns, `emailSuppressed`, and the two ledger tables — nothing else). The #1103 SHA this branch assumed is pinned and checked (6d03626, unchanged as of the last verification); if #1103 moves before merging, I reconcile this branch rather than letting it drift.

If you'd rather review the outreach delta in isolation *right now*: `git diff origin/claude/roles-foundation...origin/claude/outreach-engine -- ':!*person_merged_into*'` still mixes #1103's code, so the honest isolated view is `git log --oneline claude/outreach-engine ^origin/claude/merge-redesign ^origin/claude/roles-foundation` — one commit, the outreach work.

checkin-app/src/app/api/outreach/process-batch/route.ts
```diff
@@ -0,0 +1,85 @@
...
+        orderBy: { id: "asc" },
+        take: BATCH_SIZE,
+    });
+
+    let processed = 0;
+    for (const item of candidates) {
+        const claim = await prisma.bulkSendItem.updateMany({ where: { id: item.id, status: "queued" }, data: { status: "sending" } });
+        if (claim.count === 0) continue; // already claimed elsewhere — skip, don't send
+
+        const person = await prisma.person.findUnique({ where: { id: item.personId }, select: { name: true } });
```
thpr: **Blocking: unsubscribe is not honored at send time.**

`emailSuppressed` is consulted only when the snapshot is built (`lib/outreach/recipients.ts:113`). It appears nowhere in this drain path — the loop sends every item still in `queued`, regardless of what the person's suppression flag says *now*.

That matters specifically because of this design's best property: the drain is built to survive deploys and resume arbitrarily later, and a campaign's recipient set is frozen at create. So the sequence that breaks it is an ordinary one — a recipient gets the opening email, clicks unsubscribe, and is then emailed anyway when the drain resumes, or when a reminder campaign snapshotted before their click finishes draining. An unsubscribe that a later batch ignores is the one failure mode this feature exists to prevent.

The loop already fetches the person on this line, so the fix rides along on a query that's happening regardless:

```ts
const person = await prisma.person.findUnique({
    where: { id: item.personId },
    select: { name: true, emailSuppressed: true },
});
if (item.variant === "join" && person?.emailSuppressed) {
    await prisma.bulkSendItem.update({
        where: { id: item.id },
        data: { status: "skipped_unsubscribed" },
    });
    continue;
}
```

Recording it as `skipped_unsubscribed` rather than dropping it keeps the ledger honest, matching what `recipients.ts` already does at snapshot time. Renew-variant items stay unaffected, per the spec §2.4 decision.

Worth a test alongside it: suppress a person after the send is created, then drain, and assert no email went out for them.

checkin-app/src/app/settings/outreach/page.tsx
```diff
@@ -0,0 +1,436 @@
...
+          onChange={(e) => onChange({ ...value, body: e.currentTarget.value })}
+        />
+      </Stack>
+
+      <Group mt="md" grow align="flex-start">
+        {(["join", "renew"] as const).map((variant) => (
+          <Card key={variant} withBorder padding="sm" radius="sm">
+            <Badge size="sm" mb="xs" variant="light">{variant === "join" ? "Join (non-member preview)" : "Renew (member preview)"}</Badge>
+            <Text fw={600} size="sm" mb={4}>{previewSubject(variant)}</Text>
+            <div style={{ fontSize: 13 }} dangerouslySetInnerHTML={{ __html: preview(variant) }} />
```
thpr: **Stored XSS with a privilege escalation path (operations → board/sysadmin).**

This is the first `dangerouslySetInnerHTML` in production code — the only other occurrence in the repo is `src/app/dev/sent-mail/page.tsx`, a dev-only page.

The body it renders is authored through `PUT /api/settings/outreach`, which is deliberately open to `isOperations` (that's the stated point of keeping it separate from `settings/email`). Board members and sysadmins then open this same page and render that markup in their own authenticated session, same origin. `` in a template body executes with their cookies. Token validation only checks `{{...}}` names — it never looks at the markup around them, and it shouldn't have to, since the field is HTML by design.

The email being HTML is fine. The *preview* doesn't need to share an origin with the admin session to do its job:

```tsx
<iframe
  sandbox=""
  srcDoc={preview(variant)}
  style={{ border: 0, width: "100%", fontSize: 13 }}
/>
```

`sandbox=""` with no `allow-scripts` blocks execution while still rendering the markup, which is closer to how a mail client will show it anyway.

If you'd rather accept the risk instead — on the grounds that operations is a trusted internal role — that's a defensible call, but it should be an explicit comment here rather than an implicit one, since the role boundary is exactly what the separate-route split was designed to enforce.

checkin-app/src/app/api/outreach/status/route.ts
```diff
@@ -0,0 +1,60 @@
...
+    } catch {
+        return apiError("Invalid JSON", 400);
+    }
+    if (body.action !== "retry" || !body.bulkSendId) return apiError('Body must be { action: "retry", bulkSendId }', 400);
+
+    const bulkSend = await prisma.bulkSend.findUnique({ where: { id: body.bulkSendId } });
+    if (!bulkSend) return apiError("Send not found", 404);
+
+    await prisma.bulkSendItem.updateMany({ where: { bulkSendId: bulkSend.id, status: "failed" }, data: { status: "queued", error: null } });
+    const reopened = bulkSend.completedAt
```
thpr, line 64: **Retry can break the one-in-flight invariant.**

This reopens a completed send by clearing `completedAt`, without checking whether another send is already in flight. The send panel only surfaces the retry button when nothing is in flight, so the UI won't produce this — but the API will, and a stale tab or a direct POST is enough.

Once two rows have `completedAt: null`, two things go wrong quietly. `send/route.ts:64` finds one of them and 409s every new campaign. And `process-batch`'s no-`bulkSendId` fallback (`route.ts:40`, `orderBy: { id: "desc" }`) resolves to whichever happens to be newest, so a Resume can start draining a send the operator wasn't looking at.

The create path already guards this correctly — same check here:

```ts
if (bulkSend.completedAt) {
    const other = await prisma.bulkSend.findFirst({ where: { completedAt: null }, select: { id: true } });
    if (other) return apiError("A send is already in progress.", 409);
}
```

A Serializable transaction around the check-and-reopen would match `send/route.ts` more closely, though the exposure here is much narrower than on create, so the plain check is probably proportionate.

checkin-app/src/app/api/outreach/test-send/route.ts
```diff
@@ -0,0 +1,55 @@
...
+        if (e instanceof BoundaryNotSetError) return apiError(e.message, 400);
+        throw e;
+    }
+
+    for (const variant of ["join", "renew"] as const) {
+        const { subject, html } = renderOutreachEmail(subjectTemplate, bodyTemplate, {
+            name: SAMPLE_NAME,
+            variant,
+            boundary,
+            personId: variant === "join" ? auth.user.id : undefined,
```
thpr: **The test email carries a live unsubscribe token for the tester.**

Passing `auth.user.id` means `renderOutreachEmail` signs a real token for the caller's own person, so the footer link in the test message genuinely works. A board member checking their copy who clicks through — reasonably, to verify the link isn't broken — lands on the confirm page and is one button from suppressing their own outreach email. The confirm step softens it, but the failure mode is a privileged user quietly opting themselves out while doing exactly what testing invites them to do.

The client preview already solves this by rendering a dead footer (`settings/outreach/page.tsx:202`). Matching that here keeps the two paths consistent:

```ts
const { subject, html } = renderOutreachEmail(subjectTemplate, bodyTemplate, {
    name: SAMPLE_NAME,
    variant,
    boundary,
    // No personId: a test render must not mint a working unsubscribe token for the tester.
});
```

That drops the footer from the test entirely. If seeing the footer is the point of the test, appending an inert placeholder — same markup, `href="#"` — gets the visual check without the live token.

thpr (changes requested):
Reviewed the diff as it now stands (deps #1103/#1104 merged, so this is outreach-only — one commit, one additive migration). The earlier mixed-PR comment is resolved.

Design is solid: claim-before-`sendEmail` ordering, GET-doesn't-flip, HMAC keyed off a distinct derived key with a length-safe compare, a single shared `tokens.ts` so the preview can't drift from what sends, and a `routeAuthDrift` allowlist entry that actually justifies itself. Test coverage is real.

Four things to address — one of them blocking. Details inline.

1. **Blocking:** unsubscribe is not honored at send time, only at snapshot time.
2. First production `dangerouslySetInnerHTML`, on content an `isOperations` user can author and a board/sysadmin renders.
3. Retry can break the one-in-flight invariant at the API level.
4. Test-send emails a live unsubscribe token for the tester's own person.

Nits (non-blocking, no need to hold the PR): the per-item `person.findUnique` in the drain loop could be one `findMany` over the batch's ids; and `BulkSendItem.personId` has no FK, which looks deliberate (surviving merges/deletes, with `person?.name ?? ""` handling the gap) but is the only uncommented departure from the surrounding schema.

thpr (approved):
Approving on `e6b8e873` — all four points addressed, and addressed at the right layer rather than papered over.

- **process-batch**: the suppression re-check sits *after* the `queued → sending` claim, so it closes the gap without opening a double-send window. Recording `skipped_unsubscribed` instead of dropping the item keeps the ledger honest, matching `recipients.ts`.
- **preview**: `sandbox=""` with neither `allow-scripts` nor `allow-same-origin` is the correct form — opaque origin, scripts dead, markup still visible.
- **retry guard**: correctly scoped to `bulkSend.completedAt` being set, so retrying the genuinely in-flight send still works. The test asserting A's failed items are *not* re-queued on a 409 is the detail I'd have asked for.
- **test-send**: `personId` dropped and the inert footer appended; the test pinning both `href="#"` and `not.toContain('/api/unsubscribe')` is the right shape.

Two cosmetic leftovers, explicitly not blocking: the iframe's hard-coded `height: 160` will scroll a long template where the old div auto-grew, and the unsubscribe-footer literal now lives in three places (page, test-send route, test) instead of two. Fold into a follow-up if either bothers you.

The two nits from my previous review body (batching the per-item `person.findUnique` into one `findMany` over the batch, and a comment on `BulkSendItem.personId` having no FK) weren't taken — they were flagged non-blocking, so that's fine by me, but the FK one is worth a line of schema doc eventually since it's the only unexplained departure from the surrounding models.

Note on verification: I reviewed this by reading, and did not execute the suite — the new cases are integration tests needing a live Postgres. Relying on CI and your local run for green.

jee7s:
All four addressed at `e6b8e873`:

**Unsubscribe at send time (blocking).** Adopted as sketched — the drain's existing person fetch now selects `emailSuppressed`; join-variant items for a suppressed person get `skipped_unsubscribed` (verified it's the same status string the snapshot path writes in `recipients.ts`, so the ledger reads one vocabulary) and the loop continues. Renew untouched. Your test is in `ledger.integration.test.ts`: suppress after snapshot → drain → no email, item marked.

**Preview XSS.** `dangerouslySetInnerHTML` → `<iframe sandbox="" srcDoc>`. One adaptation: your `fontSize: 13` on the iframe element can't cascade into the framed document, so the 13px moved into a `<style>` inside `srcDoc`; iframe gets height + an a11y `title` including the card name (both cards render a join preview, so bare titles collided in the test queries).

**Retry reopen.** Plain guard as suggested — reopening a completed send 409s if any other row has `completedAt: null`, before the re-queue `updateMany`. Test asserts the 409 and that the original send's failures are not re-queued. Skipped Serializable per your proportionality note.

**Test-send token.** No `personId` — no live token minted. The join-variant test email appends the same inert `href="#"` footer the client preview renders, so both no-token paths match and the visual check survives. Test asserts the footer contains no `/api/unsubscribe`.

On the older mixed-migrations comment: resolved by the rebase — the PR is now based on main and its diff is pure outreach (no merge-redesign migrations). 40 files, all feature scope.

PR #1111, opened 2026-07-18
ops access: grant the contacts functionality to the operations role

jee7s:
The follow-on promised in #1105's scope-change: re-grants the contact-creation functionality to the operations role now that the RBAC foundation (#1104) carries it. **Merge after BOTH #1104 and #1105** — this branch cherry-picks #1105's commit because the contacts route wasn't on the roles branch; once both parents merge, this diff collapses to just the ops grants.

- `isOperations` joins the contacts route gate (sysadmin still excluded) and `people/search` — where an ops-only caller gets the stripped response (no `lastBackgroundCheck`, no membership/finance data).
- Membership-ops nav admits ops to the **Participants tab only**; the page becomes a read-only directory for ops (Bulk Import, +New Person, Household-Assign, Details, and the roles Manage link are staff-gated) with Add-contact as the sole ops affordance — the role's only write path in the app.
- Tests: ops-succeeds on contacts, ops admission + field-stripping assertions, a table-driven ops-403 block across all seven participant edit/rich-create endpoints, and a read-only-directory RTL suite.

Verified: bare tsc, CI-exact lint, 2147 unit green, 292 targeted integration tests green on real Postgres.


thpr (changes requested):
Design is sound and the commenting is unusually good — the `normalizeEmail` dup-check with its case-variant regression test, the FK-ordering fix in `afterAll`, and the table-driven ops-403 block are all the right calls. Two blocking items, both native to this PR's own commit (`67139ff3`), not inherited from #1104/#1105.

## 1. The ops strip is bypassed by the nested include

`src/app/api/people/search/route.ts` includes `householdMembers: true`, which returns **full Person rows**. The new strip only clears the top-level field:

```ts
lastBackgroundCheck: opsOnly ? undefined : p.lastBackgroundCheck,
household: p.household ? { ...p.household, orgMembership: opsOnly ? undefined : p.household.orgMembership } : null,
```

`household.householdMembers[]` still carries every member's `lastBackgroundCheck` (`@sensitivity:internal`), plus `googleId` (pii), `emailVerified`, and `emailUndeliverableAt`. An operations-only caller reads the exact field this PR sets out to withhold, one level down, for the searched person *and* everyone in their household.

The `householdMembers: true` include is pre-existing (it's on `main`), but it leaked nothing before now — sysadmin and board were the only callers, and both are entitled to `lastBackgroundCheck`. This PR introduces the first caller class that isn't, so the incomplete strip is new.

The new test doesn't catch it because it asserts only the top-level key. In `authzRoutes.integration.test.ts` the target is a member of `Test HH`, so `hit.household.householdMembers[0].lastBackgroundCheck` is truthy in that very fixture while the assertion above it passes.

Suggested fix: replace `householdMembers: true` with an explicit `select` of what the UI actually consumes (id/name/email/phone), and add `expect(hit.household.householdMembers[0].lastBackgroundCheck).toBeUndefined()` to the ops case. The explicit select also stops the next field added to `Person` from auto-joining this response.

## 2. `isMember` contradicts the strip it sits beside

The comment promises operations gets "no membership/finance standing (Household.orgMembership)". One line above the `orgMembership` strip:

```ts
isMember: personRecordIsActiveOrgMember(p),
```

That derived boolean is membership standing. Either strip it under `opsOnly` as well, or keep it and narrow the comment to say the dates and finance detail are withheld but active-membership status is not. As written, the code and the stated intent disagree, and the comment is what the next auditor will read.

The line itself predates the PR, but nothing contradicted it until this strip and its comment landed.

## Non-blocking

- **Stale justification comment** in `membership-ops/contacts/route.ts`: "every caller role here (board) is already trusted with the full people directory." Callers are now board **+ operations**, and operations explicitly does *not* get the full directory — that's this PR's premise. The conclusion still holds (ops can reach name+email via `people/search`), but the block is marked "do not redact" for future audits, so it should name the role it's now actually about.
- **Role flags reach operations unreviewed**: `...rolesToFlags(p.roles)` ships every person's sysadmin/board/keyholder/bg-reviewer flags, and the page renders `RoleBadge` for ops — only the "Manage" link is gated. Intentional? It's the one field class the strip doesn't weigh in on.
- **Optimistic prepend shape drift**: `setResults((r) => [data.participant, ...r])` inserts the raw `Person` returned by `createParticipantWithHousehold`, not a search-result row — no role flags, no `household`. It renders (falls through to "Assign"), but the new row looks different from every other row until the next search. The RTL test mocks the richer shape, so it won't surface this.

## Separately

Three pages under `membership-ops/` have no page-level role check at all — `participants/import`, `participants/merge`, `volunteer-memberships`. Adding `isOperations` to the layout gate widens who can direct-URL into those write UIs, but this is genuinely pre-existing: `isBackgroundCheckReviewer` is already on that gate on `main` and can reach them today. The APIs all 403 correctly, so it's UI-shell exposure only. Being fixed on its own branch off `main` — not this PR's to carry.

## Before merge

Merge ordering after #1104 and #1105 is right. Since `people/search` changed its role set, worth confirming the full CI coverage run rather than the 292 targeted integration tests alone — role-flag changes have historically been invisible to `tsc` in `scopeBindings` and `routeAuthDrift`.

jee7s:
Both blockers + the non-blocking items fixed, pushed (74ee4369), and per your before-merge note the FULL suites ran, not the targeted set: 2147 unit / 1142 integration green.
- **Nested-include leak closed**: `householdMembers: true` → explicit `select {id, name, email}`, with regression assertions on the nested `lastBackgroundCheck`/`googleId` being absent for ops callers. Your point about the test asserting only the top-level key was exactly right — the fixture would have caught it one level down.
- **`isMember` stripped under opsOnly** — code now matches the comment's promise, and the comment enumerates precisely what ops is denied (BG dates, membership standing, finance) vs gets (names, contact, role pills).
- The contacts justification comment now names board **+ operations** with the ops rationale. On role flags reaching ops: **intentional and staying** — `isBoardMember` is public-tier and the pills are org structure, not PII; the Manage link is what's gated. The optimistic-prepend shape is aligned with #1105's fix (household + defaulted flags).

thpr (changes requested):
Re-reviewed `74ee4369`. Both blockers are correctly resolved, and the two non-blocking notes were picked up as well. One regression introduced by the `select` narrowing — small fix, but neither `tsc` nor the test suite can catch it, so flagging before merge.

## Resolved

- **Nested leak** — `householdMembers: { select: { id, name, email } }` closes it, and the new assertions on `householdMembers[0].lastBackgroundCheck` / `.googleId` pin it against the same fixture that would have gone green before. Right fix in the right place.
- **`isMember`** — now stripped under `opsOnly`, with the comment rewritten to enumerate what ops is and isn't denied, including the explicit note that role flags are `@sensitivity:public` org structure rather than an oversight. That answers the open question from the last round.
- Contacts-route audit comment and the optimistic-prepend shape are both handled.

## New — the narrowed select drops a field the merge page depends on

`participants/merge/page.tsx` reads `isHouseholdLead` off `household.householdMembers`, which the new three-field select no longer returns:

```ts
// merge/page.tsx:239
const isLead = (mergeParticipant.household?.householdMembers as { id: number, isHouseholdLead?: boolean }[] | undefined)
  ?.find((m) => m.id === mergeParticipant.id)?.isHouseholdLead;
isLeadWithOthers = !!isLead && othersCount > 0;
```

With `isHouseholdLead` gone, `isLead` is always `undefined`, so `isLeadWithOthers` is permanently `false`. Two consequences:

- `<Button disabled={isLeadWithOthers}>` (line 282) no longer disables, and the red blocking alert — "Error: This participant is the lead of a household with other members. You cannot delete them." (line 201) — never renders. The operator gets no warning, clicks through preview, and hits a bare 400 from the merge route instead.
- The `[Lead]` marker disappears from the household-member list (line 194).

No data-integrity risk: `api/membership-ops/participants/merge/route.ts:54` independently blocks it server-side with a 400. So this is a broken client-side safety rail, not a corruption path.

Worth noting why the existing checks don't cover it. The merge page casts through `Record<string, unknown>[]` and re-asserts with `as { id: number, isHouseholdLead?: boolean }[]`, so **`tsc` is blind** to the missing field. And the merge page's tests hardcode it in their mocks (`page.test.tsx:182` and `:211` both pass `isHouseholdLead: true`), so they stay **green against a shape the endpoint no longer returns**.

Fix is one line — `isHouseholdLead` is `@sensitivity:public` in the schema, so it's safe to return to every caller including ops:

```ts
householdMembers: { select: { id: true, name: true, email: true, isHouseholdLead: true } },
```

The comment above that select claims "id/name/email is exactly what every consumer of this endpoint's household.householdMembers reads (the Assign-household picker and its household-member list)" — the merge page is the third consumer and should be named there once the field is back.

Also worth a look: `merge/page.tsx:17` types `householdMembers: Record<string, unknown>[]`, which is what made this invisible. Tightening it to the real shape would let `tsc` catch the next narrowing — but that's a cleanup for its own PR, not this one.

## Nit

The RTL mock in `participants/__tests__/page.test.tsx` is now labelled "Honest shape", but it includes `household.householdMembers` while the contacts route's re-fetch uses `include: { household: true }`, which doesn't return that nested relation. The assertion it supports (`findAllByText("Ada Lovelace")` → 2, from person name + household name) is still valid against the real response, so nothing false-passes — the extra key is just cosmetic drift in the opposite direction.

checkin-app/src/app/api/people/search/route.ts
```diff
@@ -30,7 +30,15 @@ export const GET = withAuth(
...
-                            householdMembers: true,
+                            // Explicit select, not `true` — a plain include returns full
+                            // Person rows one level down (lastBackgroundCheck, googleId,
+                            // emailVerified, emailUndeliverableAt, ...), leaking every
+                            // household member's sensitive fields regardless of the
+                            // opsOnly strip below, which only touches the top-level person.
+                            // id/name/email is exactly what every consumer of this
+                            // endpoint's household.householdMembers reads (the
+                            // Assign-household picker and its household-member list).
+                            householdMembers: { select: { id: true, name: true, email: true } },
```
thpr: This drops `isHouseholdLead`, which `participants/merge/page.tsx` depends on.

`merge/page.tsx:239` reads it off these rows to compute `isLeadWithOthers`:

```ts
const isLead = (mergeParticipant.household?.householdMembers as { id: number, isHouseholdLead?: boolean }[] | undefined)
  ?.find((m) => m.id === mergeParticipant.id)?.isHouseholdLead;
isLeadWithOthers = !!isLead && othersCount > 0;
```

With the field gone, `isLead` is always `undefined`, so `isLeadWithOthers` is permanently `false`. The blocking alert at :201 ("Error: This participant is the lead of a household with other members. You cannot delete them.") stops rendering, `<Button disabled={isLeadWithOthers}>` at :282 stops disabling, and the `[Lead]` marker vanishes from the member list at :194. The operator clicks through to preview and gets a bare 400 from the merge route instead of the inline guard.

Not a corruption path — `api/membership-ops/participants/merge/route.ts:54` blocks it server-side independently. It's the client-side safety rail that breaks.

Neither `tsc` nor CI catches this: the merge page casts through `Record<string, unknown>[]` (`merge/page.tsx:17`) before re-asserting the shape, and the merge tests hardcode `isHouseholdLead` in their mocks (`merge/__tests__/page.test.tsx:182` and `:211`), so they stay green against a shape this endpoint no longer returns.

`isHouseholdLead` is `@sensitivity:public` in the schema, so returning it to every caller including ops is fine:

```suggestion
                            householdMembers: { select: { id: true, name: true, email: true, isHouseholdLead: true } },
```

The comment just above says id/name/email is "exactly what every consumer ... reads (the Assign-household picker and its household-member list)" — the merge page is the third consumer and is worth naming there.

jee7s:
Fixed at `37864fcf` — `isHouseholdLead: true` is back in the `householdMembers` select (it's `@sensitivity:public`, and the `opsOnly` strip only overrides `orgMembership`, so ops callers get it too — verified). The comment above the select now names the merge page as the third consumer.

Also closed the drift hole you pointed out: the merge page's own tests mock this response shape, so nothing pinned the contract. The route's integration test now asserts `householdMembers[0]` has `isHouseholdLead` — in the ops-only case specifically, which pins both the field's presence and that the strip doesn't remove it. Dropping it from the select fails a test next time instead of silently killing the client-side lead guard.

thpr (dismissed):
Approving — every item from both review rounds is resolved.

**Round one**
- Nested leak via `householdMembers: true` closed with an explicit select, and the assertions on `householdMembers[0].lastBackgroundCheck` / `.googleId` are pinned to the same fixture that would have gone green before the fix.
- `isMember` stripped under `opsOnly`, with the comment rewritten to enumerate exactly what operations is and isn't denied — including the explicit note that role flags are `@sensitivity:public` org structure rather than an oversight, which answers the open question I'd raised.
- Contacts-route audit comment and the optimistic-prepend shape both handled.

**Round two**
- `isHouseholdLead` restored to the select, comment updated to name the merge page as the third consumer.
- The added integration assertion (`toHaveProperty('isHouseholdLead', false)`) is better than the fix I asked for: it pins the contract in the one place that can catch a future narrowing, since the merge page's own tests mock this response. `toHaveProperty` also distinguishes present-and-false from absent, which a truthiness check would have missed.

The net effect on `people/search` is a tighter endpoint than before this PR — the explicit select means the next field added to `Person` won't silently join the response for any caller, not just operations.

Two things outstanding that aren't mine to block on:

- CI was still pending on unit and integration when I reviewed (build, flow, isolation, s-read all green). Worth confirming those land before merge — `people/search` changed its role set, and role-flag changes have historically been invisible to `tsc` in `scopeBindings` and `routeAuthDrift`. The pre-rebase results quoted in the PR description predate the current base.
- #1105 is still unmerged (`ba2a1aac` remains in this branch), so the stated merge ordering still applies.

One cosmetic nit, non-blocking and safe to ignore or fold into any later commit: the "Honest shape" RTL mock in `participants/__tests__/page.test.tsx` includes `household.householdMembers`, which the contacts route's `include: { household: true }` doesn't return. The assertion it supports is still valid against the real response, so nothing false-passes.

Separately, the ungated-pages issue I flagged in round one shipped as #1112 and is in this branch's history.

thpr (approved):

PR #1113, opened 2026-07-18
perf(security): scope CallerContext prefetch to each route's derived ctxNeeds

thpr:
## What

`buildCallerContext` ran up to 5 sequential queries (led programs, core-vol programs, in-scope households, in-scope events, all open visits) on **every** `handler()` request — including routes whose policy never consults any of it. That standing cost was the perf argument for keeping new routes on `withAuth` instead of the audited registry path.

- `defineRoute` now derives **`CtxNeeds`** statically from the spec (authorize + orderedView roles + granted token scopes) and stores it on the registered spec. Derivation runs after token validation and overwrites any author-supplied value.
- `buildCallerContext(auth, needs)` fetches only the needed fields, in 2 parallel levels instead of 5 serial awaits.
- `handler()` passes `spec.ctxNeeds` — one-line runtime change.

## Effect

- Routes granting only session-derived scopes (`their_own` / `their_households` / `keyholders` / `everyones`) — admin/board, self, and household routes, i.e. the main `withAuth`→`handler()` migration targets — now pay **zero** context queries.
- Program-lead routes keep their queries, in 2 round trips instead of 5.
- The all-open-visits scan (previously ran for every keyholder request) now runs only on routes granting `all_current_visitors` — currently **none**.

## Why it's safe

Skipping a fetch is fail-closed by construction: an unfetched field is an empty set, which can only *shrink* `scopesHeld` — and a scope no view token grants can never flip `fieldVisible`.

Proven, not just argued:

- **`tests/security/ctxNeeds.test.ts`** — for every registered route × 10 personas × param sets, asserts admission (`resolveAccess`), the role walk, and `stripValue` output (every view × 20 models × representative rows) are identical under a fully-populated context vs one masked to the route's `ctxNeeds`. Personas are deliberately adversarial: program-scope sets are populated even for personas that wouldn't earn them from the DB, so a missed consumer would surface as a diff.
- **Drift guard** — `deriveCtxNeeds` is built from three *total* switches (`assertNever`), so adding a `Scope`, `Role`, or `Authorize` variant fails compile until the derivation handles it.
- The handler fn never sees `CallerContext` (its `HandlerContext` is `{req, auth, params, role}`), so admission, role walk, and stripper are provably the only consumers.

## Reviewer notes

- Boundary-only PR (core.ts, access-resolvers.ts, handler.ts + test) per the boundary-isolation rule — no app/feature code.
- `getRoute`/`allRoutes` now return `RegisteredRoute` (`RouteSpec` + derived `ctxNeeds`); existing consumers treat it as `RouteSpec`, no changes needed.
- Verified: tsc clean, security suites 548/548, full unit suite 2105/2105, lint clean.

### DIFF
--- a/checkin-app/src/security/access-resolvers.ts
+++ b/checkin-app/src/security/access-resolvers.ts
@@ -15,7 +15,7 @@
  */
 import prisma from '@/lib/prisma';
 import type { AuthResult } from '@/types/auth';
-import type { Authorize, Role } from './core';
+import type { Authorize, CtxNeeds, Role } from './core';
 import { LIVE_PERSON } from '@/lib/person/filters';
 
 export interface CallerContext {
@@ -39,7 +39,15 @@ export interface CallerContext {
     activeVisitorIds: Set<number>;
 }
 
-export async function buildCallerContext(auth: AuthResult): Promise<CallerContext> {
+/**
+ * Build the per-request context, fetching ONLY what `needs` says this route's
+ * policy can consume (see deriveCtxNeeds in core.ts). An unfetched field stays
+ * an empty set — strictly fewer scopes than a full fetch, and provably
+ * output-identical for any scope the route's tokens don't grant (the
+ * ctxNeeds equivalence test asserts this per registered route). Session-derived
+ * fields (selfId/householdId/isKeyholder) are always populated.
+ */
+export async function buildCallerContext(auth: AuthResult, needs: CtxNeeds): Promise<CallerContext> {
     const ctx: CallerContext = {
         selfId: undefined,
         householdId: undefined,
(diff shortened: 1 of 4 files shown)


jee7s (approved):

dkaygithub (approved):

jee7s:
Heads-up on branch activity (Jeff asked me to clear the conflicts): pushed merge commit `8533db4a` bringing in main's `d61d7850` (#1111) — conflict-free, no force, nothing of yours touched.

Timeline honesty: mid-resolution you rebased `34230506` → `6d719540`; my first push was rejected non-fast-forward and I rebuilt on your new tip rather than forcing. Worth noting your rebase and my discarded resolution placed the #1124 LIVE_PERSON filters at identical sites (relation filter on the core-volunteer query in the upper Promise.all, spread on the in-scope-participants query in the lower) — two independent passes agreeing on filter placement in the new prefetch structure is decent evidence it's right.

Verified on the pushed tree: your tests/security proof suite 527/527, drift guard green, tsc/lint clean, full unit+integration green, diff vs main still exactly your 4 boundary files. If you'd rather own further catch-up merges on this branch solo, say so and I'll keep hands off.

Issue #1133, opened 2026-07-19
Boundary-isolation check is not merge-blocking: 'isolation' job runs but is not a required status check on main


thpr:
## Summary

**The `Security boundary isolation` workflow is not a required status check on `main`, so it can flag a boundary-isolation violation but cannot block the merge.**

`.github/workflows/security-boundary-isolation.yml` (job id `isolation`) enforces the rule in `AGENTS.md` (Security section): a change to the security boundary — registry grants, token grammar, handler/stripper, scope bindings, or an existing field's `@sensitivity` re-tier — must ship in its own PR, with only tests/docs/schema alongside (no app/feature code). The workflow logic is correct and runs on every PR to `main`. But because the check is not in the branch's required-contexts list, a red `isolation` result does not gate the merge button — it relies on a human noticing.

Sibling issue #1132 covers the CODEOWNERS half of the same guarantee; this issue is the status-check half. Both are needed for the boundary to actually be enforced rather than merely observed.

## Evidence

**The check is not required.** `main` branch protection required contexts:

- `Deploy build (no deploy)`
- `Coverage gate (merged)`
- `Run Tests (unit)`
- `Run Tests (integration)`

`isolation` is absent from that list.

**It fires correctly but soft.** During the board-contacts work (2026-07-19) the `isolation` check failed on two real PRs — a stacked boundary PR that had accumulated app code (`checkin-app/scripts/migrated-routes.txt`, `checkin-app/src/app/api/directory/board/route.ts`, `checkin-app/src/app/api/safety/board-contacts/route.ts`) alongside the registry change. The failure was accurate: the PR genuinely was not isolated. It turned the PR red, but nothing prevented a merge — the stack was only corrected because it was being watched by hand.

**The workflow itself is sound.** `security-boundary-isolation.yml` triggers `on: pull_request: branches: [main]`, classifies each changed file as boundary vs. companion via the `is_boundary`/`is_companion` shell functions, and `exit 1`s on a violation. The gap is purely that GitHub does not treat that exit code as merge-blocking.

## Fix

Add `Security boundary isolation` (job `isolation`) to the required status checks on `main`:

```
gh api -X PATCH repos/innovationtreehouse/checkin/branches/main/protection/required_status_checks \
  -f 'contexts[]=Deploy build (no deploy)' \
  -f 'contexts[]=Coverage gate (merged)' \
  -f 'contexts[]=Run Tests (unit)' \
[...]

Issue #1132, opened 2026-07-19
Security CODEOWNERS gate is inert: false code-owner-review + wrong owner handle + root-anchored paths under checkin-app/


thpr:
## Summary

**The `.github/CODEOWNERS` security gate is not in force.** The policy doc (`docs/security/SECURITY-POLICY.md:29`), `AGENTS.md`, and 8 `checkin-app/src/security/*.ts` headers all claim the security layer requires maintainer approval to change. It does not. Three independent, each-sufficient failures mean the mechanism cannot fire on any PR.

## Three failures (any one is fatal)

1. **`require_code_owner_reviews: false`** on `main` branch protection. GitHub never consults CODEOWNERS at merge time. Voids the entire file regardless of paths/owner.
2. **Owner handle wrong.** CODEOWNERS names `@dkay` — a real but unrelated GitHub user (id 113407) who is **not a collaborator** (`collaborators/dkay` → 404). GitHub's validator flags every line "Unknown owner". The intended owner is `@dkaygithub` (id 3522000), who *is* assignable. Off by the `github` suffix.
3. **Paths dead.** App lives under `checkin-app/`; patterns are root-anchored. `/src/security/`, `/prisma/schema.prisma`, `/src/lib/auth.ts`, `/tests/security/`, `/scripts/*` match **0 tracked files**. Only `/.github/` (12) and `/.husky/` (1) match anything at root.

## Evidence

**Validator** — `gh api repos/innovationtreehouse/checkin/codeowners/errors` → 10 errors, all `"Unknown owner … make sure @dkay exists and has write access"`.

**Pattern resolution** (`git ls-files` ground truth):

| Pattern | Root match | Live? | Real target |
|---|---|---|---|
| `/prisma/schema.prisma` | 0 | DEAD | `checkin-app/prisma/schema.prisma` |
| `/src/security/` | 0 | DEAD | `checkin-app/src/security/` (18) |
| `/src/lib/auth.ts` | 0 | DEAD | `checkin-app/src/lib/auth.ts` |
| `/src/lib/auth-options.ts` | 0 | DEAD | `checkin-app/src/lib/auth-options.ts` |
| `/tests/security/` | 0 | DEAD | `checkin-app/tests/security/` (10) |
| `/scripts/check-route-coverage.ts` | 0 | DEAD | `checkin-app/scripts/…` |
| `/scripts/security-generator.js` | 0 | DEAD | `checkin-app/scripts/…` |
[...]

PR #1135, opened 2026-07-19
feat(security): legacy-authz ratchet — no new routes outside the registry; revive CODEOWNERS gate

dkaygithub:
## What

Makes `defineRoute()` + `handler()` the only way to ship a **new** API route.

1. **`scripts/legacy-authz-routes.txt`** — frozen baseline of the 161 route method keys that predate the registry (139 `withAuth`, 9 `withCron`, 1 `withKiosk`, 12 bespoke: webhooks/HMAC, signed unsubscribe, health, dev-personas, 2 manual-session reads). Shrink-only by contract: migrating or deleting a route removes its line; **adding** a line is an explicit decision to ship surface outside the audited policy layer (CODEOWNERS-gated once #1136 + the #1132 settings flip land).
2. **Lint rule `new-route-old-authz`** (scripts/check-route-coverage.ts, check #6) — any exported route method in neither `src/security/registry.ts` nor the baseline is a hard error, **blocking even in advisory mode** (grandfathering is the baseline's job; a new route on old authz has no advisory tier). Stale baseline entries warn so a deleted route can't be resurrected on old authz. A missing baseline file is itself an error, so the ratchet can't be deleted to bypass it.
3. **CI runs the lint `--strict`** — 0 errors on the tree at flip time, so this lands green; warnings (137 unmigrated / 2 partially-migrated) are unaffected. Strict also arms `direct-json` / `third-party-fetch` / registry-drift as real gates. The lint step is inside **Run Tests (unit)**, already a required status check, so violations mechanically block merge.

CODEOWNERS revival split out to **#1136** per review feedback.

## Verified

- `npm run check-route-coverage -- --strict` → exit 0 on the clean tree.
- Synthetic `withAuth` route (`src/app/api/__ratchet_test__/route.ts`) → `new-route-old-authz` error, exit 1 in **both** advisory and strict modes; clean again after removal.
- Baseline generated with the lint's own verb-extraction regex against the live route tree; 161 keys = 175 exported methods − 14 registry entries, matching the audit exactly.

## Post-merge (repo settings, not code — #1132 / #1133)

- Enable `require_code_owner_reviews: true` on `main` after #1136 merges (#1132).
- Add the `isolation` job (Security boundary isolation) to required status checks (#1133) — pin the exact context string from a recent run first.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -6,33 +6,28 @@
 # DO NOT remove paths from this file without understanding which guarantee
 # you are weakening. See docs/SECURITY-POLICY.md.
 
-# NOTE: paths were root-anchored from Sprint 1 (#140) while the app lives
-# under /checkin-app — root-anchored patterns matched NOTHING, so none of
-# these gates were active until the prefix fix below (2026-07-19).
-
 # Schema annotations — the source of truth for field sensitivity.
-/checkin-app/prisma/schema.prisma                 @dkaygithub
+/prisma/schema.prisma                 @dkay
 
 # The policy layer itself.
-/checkin-app/src/security/                        @dkaygithub
+/src/security/                        @dkay
 
 # Authentication & session machinery.
-/checkin-app/src/lib/auth.ts                      @dkaygithub
-/checkin-app/src/lib/auth-options.ts              @dkaygithub
+/src/lib/auth.ts                      @dkay
+/src/lib/auth-options.ts              @dkay
 
 # Contract tests that enforce the policy in CI.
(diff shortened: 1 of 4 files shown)


jee7s:
Let's split the code owners change into a dedicated PR

thpr (approved):

dkaygithub:
Done — CODEOWNERS change moved to its own PR: #1136 (code half of #1132). This PR is now ratchet-only: baseline + `new-route-old-authz` lint rule + CI `--strict`. No ordering constraint between the two.

PR #1136, opened 2026-07-19
fix(security): revive the CODEOWNERS gate — re-anchor paths, correct owner handle (#1132)

dkaygithub:
## What

Code half of **#1132**, split out of #1135 per review feedback. The CODEOWNERS security gate has been inert since Sprint 1 (#140) — two independent, each-sufficient failures fixed here:

1. **Dead paths**: patterns were root-anchored (`/src/security/`, `/prisma/schema.prisma`, `/tests/security/`, `/scripts/*`) while the app lives under `checkin-app/` — they matched **0 tracked files**. Only `/.github/` and `/.husky/` were path-live. All re-anchored under `/checkin-app/`.
2. **Wrong owner**: `@dkay` is a real but unrelated GitHub account (not a collaborator), so every line failed GitHub's validator ("Unknown owner"). Corrected to `@dkaygithub` — verified an **admin collaborator** via API.

Also adds `/checkin-app/scripts/legacy-authz-routes.txt` (the authz ratchet baseline from #1135) — a harmless dead pattern until that PR lands, then load-bearing. No ordering constraint between the two PRs.

## Not in this PR (repo settings, post-merge)

- `require_code_owner_reviews: true` on `main` — the third failure from #1132. Must flip **after** this merges: flipping while the broken `@dkay` handle is on main leaves the gate inert or wedges `/.github/`-touching PRs (including this one).
- Adding `isolation` to required status checks (#1133).

## Evidence

#1132's behavioral proof stands: PRs #1120/#1128 modified `checkin-app/src/security/` with no code-owner review requested. After this + the settings flip, that can't recur.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -6,28 +6,33 @@
 # DO NOT remove paths from this file without understanding which guarantee
 # you are weakening. See docs/SECURITY-POLICY.md.
 
+# NOTE: paths were root-anchored from Sprint 1 (#140) while the app lives
+# under /checkin-app — root-anchored patterns matched NOTHING, so none of
+# these gates were active until the prefix fix below (2026-07-19).
+
 # Schema annotations — the source of truth for field sensitivity.
-/prisma/schema.prisma                 @dkay
+/checkin-app/prisma/schema.prisma                 @dkaygithub
 
 # The policy layer itself.
-/src/security/                        @dkay
+/checkin-app/src/security/                        @dkaygithub
 
 # Authentication & session machinery.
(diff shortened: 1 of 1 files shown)


dkaygithub:
Expanded ownership per Daniel: every trust-anchor line is now `@dkaygithub @jee7s @thpr` (all verified admin collaborators). Any one of the three satisfies the code-owner review — which also makes the future `require_code_owner_reviews` flip workable in practice: the author of a security-layer PR can't approve their own, so with a single owner, Daniel's own boundary PRs would have been unmergeable.

thpr (approved):

Issue #1134, opened 2026-07-19
Security layer: declarative row visibility (query-side row filtering)


thpr:
# Security layer: declarative ROW visibility (query-side), closing the class handler() can't cover

## Problem

The security policy layer (`docs/security/SECURITY-POLICY.md`) enforces **field** visibility: `@sensitivity` tiers → generated classifications → registry views → `stripBag()` at response time. That made the field-over-return leak class (#129, #127, #122) unmergeable, and it works because sensitivity is intrinsic to the data — one annotation, enforced everywhere.

**Row** visibility has no equivalent enforcement. Which rows a caller may see stays hand-rolled in each route's `where`-clause, and the only guard is human review plus prose: the `EDGE_INCLUDE_ALLOWLIST` in `tests/security/routeAuthDrift.test.ts` is full of entries like *"query-shaped — self + own household only"* — a row policy written in English, checked by eyeball.

This class has already shipped real bugs:

- **P0-C** — denied-household members could see `orgMemberOnly` programs (row leak; fixed by collapsing denied sessions to anonymous in `getOptionalSessionUser`).
- **#575** — `GET /api/programs/[id]` included participant/volunteer rosters for non-staff callers (row/edge leak; fixed by query-shaping the include).

`handler()` structurally cannot catch these: it strips fields per row, not rows. And it shouldn't — response-time row stripping breaks pagination and counts, and parks invisible rows in process memory (the exact concern the 60s catalog cache's explicit `select` defends against). Rows must be filtered **query-side**, in the `where`.

Row visibility is also not intrinsic the way field tiers are: the same `Program` row is visible to anonymous on the catalog (published + not member-only), to its lead while draft, and to everyone in ops views. It's contextual — route purpose + business lifecycle + membership state. So a per-model annotation can't capture it; a per-route declaration can.

### Step 1 — Persona row-contract test (detection; no runtime change, no migration risk)

SECURITY-POLICY.md already names the gap: *"Per-row scope correctness in the generic contract test … would require route-specific personas."* We have the pieces: `seedBaseline` debug personas, and contract walkers in `tests/security/policy.contract.integration.test.ts` / `registryAuthz.integration.test.ts`.

Build:
- A standard seeded persona world (admin, board, program lead, core vol, household member, plain member, denied, anonymous, kiosk) with fixture rows arranged so each scope relationship exists (own row, household row, program-participant row, unrelated row).
[...]

PR #1137, opened 2026-07-19
feat(security): wave-1 registry entries — 3 remaining settings routes (inert)

dkaygithub:
## What

Boundary-only PR (per the AGENTS.md isolation rule + `security-boundary-isolation.yml`): `defineRoute` entries for migration wave 1 — **inert** until each route's `handler()` conversion lands in its own feature PR, which will be stacked on this branch and retargeted to `main` when this merges.

Ten reads whose responses are (near-)raw model rows, so the declared views are exact, not aspirational:

| Endpoint | Admission | View |
|---|---|---|
| `GET /api/events` | sysadmin/board | full everyones band |
| `GET /api/facility/visits` | sysadmin/board | full everyones band (nested Person email = pii) |
| `GET /api/facility/badges` | sysadmin/board | full everyones band (RawBadgeLog internal/personal) |
| `GET /api/system-status/errors` | sysadmin/board | full everyones band (ErrorLog all internal) |
| `GET /api/system-status/links` | sysadmin/board | full everyones band (IntegrationErrorLog all internal) |
| `GET /api/settings/email` | sysadmin/board | full everyones band (BoardSettings) |
| `GET /api/settings/membership` | sysadmin/board | full everyones band (BoardSettings incl. internal Shopify ids) |
| `GET /api/settings/membership/volunteer-designations` | sysadmin/board | full everyones band (email = pii) |
| `GET /api/admin/settings/localization` | **sysadmin only** | all-public AppSettings singleton |
| `GET /api/profile/visits` | authenticated | **`their_own:personal` only** — member-facing, deliberately tight |

Notes for the CODEOWNERS reviewer:

- Admin surfaces use the registry's standard full-everyones band (same rationale as the finance-ops queues: the win is declared policy — any role later added to a view gets field-stripped automatically).
- `GET /api/profile/visits` is the one member-facing entry: `their_own:personal` + member + public. The Visit binding keys on `row.personId`, so the feature PR adds `personId` to the route's select — without it the scope fails closed and arrivedAt/departedAt strip (over-restriction, not a leak).
- Registry-first is CI-safe here: `policy.contract` skips array bodies and these routes' allowed personas hold full bands; `registryAuthz` admission expectations match the current `withAuth` gates exactly.

## Verified

- 7 security suites: **522/522** (incl. the ctxNeeds masked≡full equivalence over the grown registry)
- `check-route-coverage --strict`: 0 errors
- Boundary isolation: only `src/security/registry.ts` changed

## Sequencing

Feature PRs (one per route/area) convert each route to `handler()`, preserving every row filter and inline gate verbatim (#1134: field stripping ≠ row visibility). They stack on this branch until it merges.


jee7s:
## Code review — 11 findings

Boundary-only diff: 140 added lines in `registry.ts`, no other file. I verified each entry against its live handler rather than against the table in the description.

**What checks out:** all 10 declared `envelope` values match the handler's actual `NextResponse.json` shape (including `envelope: null` for the bare array on `/api/events`, and the deliberate `errors` reuse on `/system-status/links`). All 10 declared admission gates match the handler's real `withAuth({ roles })`. All named models resolve in generated classifications. `Visit.arrivedAt`/`departedAt` are genuinely `'personal'`, so `their_own:personal` is the right token.

Two findings are load-bearing; the rest are policy-precision and hygiene.

---

### 1. `registry.ts:392` — `/api/profile/visits` cannot work as declared: the handler's select omits `personId`

The entry's own comment states the requirement:

> *Visit binding keys on `row.personId` — the handler's select must include `personId` or the scope fails closed and the timestamps strip*

`scopeBindings.ts:96` confirms it: `Visit: { their_own: { field: 'personId', eqCtx: 'selfId' } }`. But the live handler selects:

```ts
select: { id: true, arrivedAt: true, departedAt: true, event: { select: { name: true } } }
```

No `personId`. On conversion, every row fails the `their_own` binding, `arrivedAt`/`departedAt` strip as `'personal'`, and the endpoint returns rows carrying only `id` and a nested event name — the entire payload of a member-facing history page, gone.

The comment identifies the hazard but nothing enforces it, and the code it describes does not currently satisfy it. `Visit.personId` is classified `'public'`, so adding it to the select costs nothing and leaks nothing. Worth doing in this PR, or the conversion PR starts from a broken state.

### 2. `registry.ts:334` — the entries are **not inert**; `registryAuthz.integration.test.ts` executes every registered route

The description's central claim is that these are inert until each `handler()` conversion lands. That holds for the stripper, but not for the contract suite:

```ts
const PLANS = Array.from(allRoutes()).map(([endpoint, spec]) => …);
for (const plan of PLANS) { … }
```

For every registered endpoint it derives an import path, **imports the real route module and calls it**, asserting 401 unauthenticated, 403 under-privileged, and **2xx for an admitted caller**. Registration alone is what enrols a route.

So this PR immediately adds 30 live assertions against ten unconverted handlers. They may well pass — the gates match — but the risk profile is not "inert", and it's an integration suite that needs Postgres, so neither of us can confirm it locally. Worth restating in the description, since a reviewer approving on "inert" is approving something different from what lands.

### 3. `registry.ts:430` — registering `/api/settings/email` makes the contract suite mutate the database

That handler's GET is not a read:

```ts
prisma.boardSettings.upsert({ where: { id: 1 }, create: { id: 1 }, update: {}, select: SELECT })
```

Via finding 2, the "admits an allowed caller (2xx)" case now runs it, so merely registering this route causes the security suite to **create the `BoardSettings` singleton** as a side effect. Any suite that asserts default-absent board settings, or runs after this one on a shared database, inherits a row it did not create. Check `/api/admin/settings/localization` for the same pattern before merge.

### 4. `registry.ts:334` — the declared view for `/api/events` is wider than the response, contradicting the stated principle

The description says these are *"reads whose responses are (near-)raw model rows, so the declared views are exact, not aspirational."* This entry isn't. The handler selects five public columns; the entry grants the full everyones band including `everyones:internal`, which covers `Event.attendanceConfirmedAt`, `attendanceConfirmedById` and `postEventEmailSent`.

The inline comment defends this as future-proofing — *"a later select widening is still policy-covered"* — but that is the inverse of what the registry is for. Under this entry, widening the select silently ships internal fields with **no boundary PR and no CODEOWNERS review**, because the policy already permits them. Declaring `['everyones:public']`-equivalent and forcing a boundary change when the surface grows is the behaviour the isolation rule is built around.

### 5. `registry.ts:377` — same over-grant on `AppSettings`, where it is most likely to bite

The comment concedes the data is *"wholly public-tier"*, yet grants all five tokens. `AppSettings` is the app-config singleton — the natural place for a future integration key or webhook secret. The day such a field is added, it ships to sysadmin through an entry nobody revisited. A public-only view here would fail closed instead.

### 6. `registry.ts:334-467` — the same five-token band is copy-pasted across eight entries

Eight of ten `orderedView`s are character-identical two-line blocks. Extracting one `const ADMIN_FULL_VIEW: readonly OrderedViewEntry[]` would make a future narrowing a single edit, and — more usefully for a boundary file — make any entry that *deviates* visible in the diff instead of hidden among identical text.

### 7. `registry.ts:392` — kiosk callers reach `authorize: 'authenticated'` and fail closed silently

If a kiosk session satisfies `'authenticated'` admission, the caller is admitted with no `selfId`, so `their_own` binds nothing and the response is an empty/stripped payload rather than a 401 or 403. A confusing success is worse than a clean rejection here. `core.ts` has a `kiosk` role — an explicit entry (or a comment recording that kiosk cannot reach this route) would settle it.

### 8. `registry.ts:392` — no admin entry in `orderedView`, deliberately, but nothing records why

A sysadmin calling this gets `their_own:personal` only, because first-match-wins stops at `'authenticated'`. That is intentional per the description, but the *file* doesn't say so. The next person debugging "why can't an admin see a member's visit history" has a one-line fix that silently widens a member-facing endpoint. The rationale belongs next to the entry.

### 9. `registry.ts:430` and `:443` — two entries, same model, same view, materially different responses

`/api/settings/email` returns a narrow `SELECT`; `/api/settings/membership` returns the full `BoardSettings` row including internal Shopify ids. Both declare the identical full band, so the registry — the artifact an auditor reads to answer "what does this endpoint expose?" — reports them as equivalent when they are not.

### 10. Zero test changes

A CODEOWNERS-gated policy change adding ten entries ships with no new assertions of its own. The generic enumeration suite covers admission, but nothing pins the *views* — e.g. that `/api/profile/visits` strips another member's rows, which is finding 1's failure mode and the one thing here with member-facing privacy consequences.

### 11. `registry.ts:417` — `/system-status/links` reuses the `errors` envelope key

Called out as back-compat, and correct against the handler. Noting only that two distinct endpoints now share an envelope name, so a payload alone can't identify its source — worth a rename when the UI can absorb it.

---

Findings 1 and 2 I'd want resolved before merge: the first means the endpoint ships broken on conversion, the second means the PR's risk description doesn't match its effect.

dkaygithub:
I'm breaking this up into smaller PRs

jee7s (commented):

**This PR and the #1209–#1215 re-split are mutually exclusive, and one family must be explicitly closed before anything merges.** Both register the same endpoints; `defineRoute` throws on duplicates (core.ts:316-318), and because each PR inserts against the same surviving context anchor, the merge queue will land both without a textual conflict — the duplicate is only discovered at the next cold start, when module init throws and **every** registry-governed route 500s.

**Recommendation: keep this PR; close #1209–#1215.** Three reasons: (1) the conversion stack #1138–#1142 is based on this PR's branch (`sec/registry-wave1`), so it's the family the work was designed around; (2) the re-split covers only 7 of these 10 entries — it drops `GET /api/admin/settings/localization`, `GET /api/settings/email`, and `GET /api/settings/membership`, which #1141's conversions require at request time; (3) #1212 in the re-split is a broken concat (five entries, one duplicated) that hard-fails at boot.

Nits for this PR (AGENTS.md comment rules): the per-entry "Registered ahead of its handler() conversion (inert until then…)" narration goes stale the moment #1138–#1142 merge — nothing in the stack deletes it, leaving the registry permanently claiming live routes are inert. Same for the 6–8-line speculative rationale blocks on the events/localization entries ("if the select later grows…", "natural home for a future integration key") — one line stating the fail-closed invariant keeps the constraint; the speculation belongs in the PR body.

jee7s:
**Linkage map** (companion to the family-conflict review above):
- Conversion stack based on this branch: #1138 (events) · #1139 (facility visits+badges) · #1140 (profile/visits) · #1141 (settings ×4) · #1142 (system-status errors+links) — each conversion needs its entry here at request time.
- Duplicate family to be closed if this PR is kept: #1209, #1210, #1211, #1212, #1213, #1214, #1215 (re-split of 7 of these 10 entries; #1212 is a broken concat).
- Related tracked work: #1321 (unbuilt scopable+sensitive routes pending scoping), #1320 (GAP-2 drift-guard CI ban) — this wave is the migration those issues assume.
- Not to be confused with #146/#1348 ('dangerously_allow' elimination) — that mechanism doesn't exist on main; this registry-first migration is the real stripper-coverage work.

thpr:

jee7s:
Merged `main` into this branch and trimmed the diff down to what this PR still uniquely contributes.

7 of the original 10 entries landed on `main` via the one-route re-cuts (#1209–#1215) — including the deliberately **tightened** public-only `GET /api/events` view, which this branch's older draft would have widened back had its copy survived. The resolution takes main's copies verbatim and keeps only the three routes that exist nowhere else:

- `GET /api/settings/email`
- `GET /api/settings/membership`
- `GET /api/admin/settings/localization`

Net diff is now registry-only, +43 lines, exactly those three `defineRoute` blocks. Merging this unblocks #1141 (its conversions of these routes currently fail CI with `[security] No registry entry` — the red is accurate and flips green once this lands; #1141 will need a CI re-trigger after).

jee7s (approved):

PR #1138, opened 2026-07-19
feat(security): migrate GET /api/events to handler()

dkaygithub:
Feature half of the wave-1 pair for `GET /api/events`. The registry entry landed on `main` with the wave-1 batch (#1209), so this PR is no longer stacked — it targets `main` directly and its diff does not touch `registry.ts` at all, per the boundary-isolation rule.

- Same query, same tight select (all public-tier), same bare-array response (envelope `null` + single-key bag)
- Admission: `withAuth({roles})` → registry `anyRole` (identical gate), stripper now the declared backstop
- Row filter `programId: null` preserved verbatim (#1134: row visibility stays query-side)
- `migrated-routes.txt` +1; POST stays `withAuth` for a later wave
- Drops the now-stale `'events'` entry from `AUTHZ_TESTED` — that guard only covers routes with a `roles:` gate, which this route no longer has (same bookkeeping #1127 did for `safety/board-contacts`). The deny-path integration case is kept.
- Tests: program-ops/events pages + routeAuthDrift, 436/436

Note on the merge: `main`'s registry entry for this endpoint is **stricter** than the draft this branch was written against — `orderedView` is narrowed to `['public']` for both roles instead of the full `everyones:*` bands. The route's select is exactly the five public-tier `Event` columns (`id`/`name`/`startAt`/`endAt`/`description`), so the tighter view covers the response exactly, and any later select growth into an internal field (`attendanceConfirmedAt`/`ById`, `postEventEmailSent`) now fails closed and forces a boundary PR.


jee7s (commented):

**CONFIRMED: this PR fails its own CI.** `src/__tests__/authzRegistry.test.ts` keeps `'events'` in `AUTHZ_TESTED`, and its stale-entry assertion requires every listed file to still match a `roles: [` gate. Converting the GET to `handler()` leaves POST `withAuth({})` with no roles array, so `'events'` drops out of the role-gated set and `expect(stale).toEqual([])` fails on the unit tier that runs on every PR. The test needs updating in the same commit as the conversion.

Also applies here (and to the rest of the #1138–#1142 stack): once this conversion merges, the registry entry's "Registered ahead of its handler() conversion (inert until then…)" comment is stale — per AGENTS.md, delete it when you touch the code; nothing in the stack currently does.

Sequencing note: this stack depends on the wave-1 registry entries. See #1137 — that PR (not the #1209–#1215 re-split) is the family this stack is based on.

jee7s:
**Linkage:** converts the route registered by #1137's `GET /api/events` entry (duplicate entry in #1209). Blocked on the family resolution at #1137; carries its own CI blocker (stale `authzRegistry.test.ts` entry — see review above). Sibling conversions: #1139, #1140, #1141, #1142.

thpr:

jee7s (approved):

PR #1140, opened 2026-07-19
feat(security): migrate GET /api/profile/visits to handler()

dkaygithub:
Feature half of the wave-1 pair for `GET /api/profile/visits` — registry entry in #1137 (**stacked**; retarget to `main` after #1137 merges).

The one member-facing route in wave 1, so the view is deliberately tight: `their_own:personal` + member + public (not an admin band).

- Row filter preserved verbatim (own rows only, ±7-day window) — row visibility stays query-side per #1134
- `select` adds `personId`: the Visit `their_own` binding keys on `row.personId`; without it the scope fails **closed** and `arrivedAt`/`departedAt` (personal tier) would strip from the caller's own visits. Additive, public-tier field — the integration test asserts status/array/length only
- In-handler `auth.type` check kept (defense-in-depth convention)
- Tests: routeAuthDrift + stripper 477/477; `check-route-coverage --strict` clean


jee7s (commented):

Two CONFIRMED defects in the rewritten GET (carried logic, but re-shipped by this conversion — this is the moment to fix them):

**1. Unvalidated `?date=` → opaque 500.** `new Date(filterDateStr)` on an unparseable value yields Invalid Date; `setDate(NaN)` keeps both bounds invalid, prisma throws on the NaN DateTime, and the route 500s (plus a stack in the console) where it should 400. One `isNaN(baseDate.getTime())` check → `badRequest('Invalid date')`.

**2. The ±7-day window is timezone-skewed.** A date-only param ("2026-07-25") parses as UTC midnight, then `setDate()` does day arithmetic in server-local time — boundary-day visits are silently dropped/included depending on the server's UTC offset. Compute the window with UTC accessors (or the repo's `lib/time` helpers per the 1149 design).

Nit (AGENTS.md): the route comment citing "#1137 finding 1" is a PR-review artifact — the constraint itself belongs in the comment, the provenance doesn't (and the reference dangles once the family conflict resolves).

Sequencing: depends on the wave-1 `GET /api/profile/visits` entry — see the family-conflict analysis on #1137. FYI: `/attendance/my-visits` (PR #1357) is now a live UI consumer of this route, so the 400-vs-500 behavior is user-visible.

jee7s:
**Linkage:** converts the route registered by #1137's `GET /api/profile/visits` entry (duplicate in #1215). Two consumers/constraints to be aware of: `/attendance/my-visits` (PR #1357) now renders this route's data, making the 400-vs-500 behavior user-visible; and the timezone-skewed ±7-day window from the review is exactly the class of defect tracked by #1346 (date/time canonical-layer remediation) — fixing it here per the #1149 design (merged in #1338) keeps this route off #1346's backlog. Blocked on the family resolution at #1137.

thpr:

jee7s:
The conversion is faithful and the `personId` addition to the select is right — `scopeBindings.ts:96` binds `Visit.their_own` to `personId`, so the row would strip to nothing without it. No `AUTHZ_TESTED` breakage here either, since `profile/visits` used `withAuth({})` and is not in that list.

Two pre-existing defects get carried across unchanged, and this is the natural moment to fix them rather than re-ship them:

**An unparseable `?date=` returns 500, not 400.** The new code keeps:

```ts
const baseDate = new Date(filterDateStr);
startDate.setDate(baseDate.getDate() - 7);
```

`new Date("garbage")` is Invalid Date, `getDate()` is NaN, both bounds become NaN, prisma throws, and `handler()`'s catch turns it into an opaque 500. `badRequest()` is already imported one line away.

**The ±7-day window mixes UTC parsing with local arithmetic.** A date-only `?date=` parses to UTC midnight, then `setDate(getDate() ± 7)` operates in local time — so on any non-UTC server the boundary-day visits shift. This is F8 in `1149_DATE_TIME_TZ_DESIGN.md`, and #1423 fixes exactly this pattern in the sibling `household/visits` route by switching to `baseMs ± 7 * DAY_MS`. Worth matching so the two routes do not diverge.

Both are small, and #1357 makes this endpoint user-visible via `/attendance/my-visits`, which promotes the 500 from cosmetic to something a member can actually hit.

---
_Posted by Claude._

jee7s (approved):

PR #1141, opened 2026-07-19
feat(security): migrate settings GET routes to handler() (email, membership, volunteer-designations, localization)

dkaygithub:
Feature half of four wave-1 pairs — registry entries in #1137 (**stacked**; retarget to `main` after #1137 merges).

- `GET /api/settings/email` — same narrow 3-column SELECT (defense in depth)
- `GET /api/settings/membership` — full BoardSettings row; internal-tier Shopify ids now declared policy
- `GET /api/settings/membership/volunteer-designations` — designation email (pii) covered by admin band
- `GET /api/admin/settings/localization` — sysadmin-only; all-public AppSettings singleton
- Create-on-first-read upserts preserved; envelopes unchanged; all write siblings stay `withAuth`
- `migrated-routes.txt` +4
- Tests: settings suites + routeAuthDrift 496/496; `check-route-coverage --strict` clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1142, opened 2026-07-19
feat(security): migrate system-status GET routes (errors, links) to handler()

dkaygithub:
Feature half of two wave-1 pairs — registry entries in #1137 (**stacked**; retarget to `main` after #1137 merges).

- `GET /api/system-status/errors` — ErrorLog, take-100, all internal tier
- `GET /api/system-status/links` — IntegrationErrorLog, take-200, unresolved-first ordering and the read-triggered 90-day purge preserved; response key `errors` kept for the existing UI (matches the registry envelope note)
- Admission: identical anyRole sysadmin/board gates; `migrated-routes.txt` +2
- `PATCH /api/system-status/links/[id]` stays `withAuth` (later wave)
- Tests: system-status suites + routeAuthDrift 449/449; `check-route-coverage --strict` clean


jee7s (commented):

**Same class of CI failure as confirmed on #1138 — verify before merge:** `src/__tests__/authzRegistry.test.ts`'s stale-entry assertion requires every `AUTHZ_TESTED` file to still match a `roles: [` gate. If any route file this PR converts is listed there and its remaining handlers carry no roles array, this PR's own unit CI goes red; update the test alongside the conversion.

Also: delete the registry entries' "inert until conversion" comments as part of this conversion (AGENTS.md — stale the moment this merges), and note the stack's dependency on #1137's wave-1 entries (see the family-conflict analysis on #1137 — the #1209–#1215 re-split is missing entries and should not be the base).

jee7s:
**Linkage:** converts the routes registered by #1137's system-status entries (duplicates in #1212 — broken concat — and #1213). Blocked on the family resolution at #1137; check `authzRegistry.test.ts` for the same stale-entry class confirmed on #1138. Sibling conversions: #1138–#1141.

thpr:

jee7s:
**Area of concern: a deterministic unit failure that no CI run will show you.**

Both converted files lose their `roles: [` gate, while `system-status/errors` and `system-status/links` stay in `AUTHZ_TESTED` in `src/__tests__/authzRegistry.test.ts`. Its stale-entry assertion fails on the unit tier. The precedent is #1127, which removed `safety/board-contacts` from that set in the same PR as its migration.

This is invisible today because the base is `sec/registry-wave1`, so `gh pr checks` reports "no checks reported" and the `CLEAN` status is measured against an unprotected branch. The first real signal comes only after retargeting to main.

The rest of the conversion is faithful — both handlers return correctly-keyed bags, and the 90-day purge and `nulls: "first"` ordering survive.

---
_Posted by Claude._

jee7s (approved):

PR #1147, opened 2026-07-19
ci: prod releases must be cut from a rel/* branch

jee7s:
## What

Adds a `branch-guard` job to `deploy-prod.yml`: the released tag's commit must be **contained in a `rel/*` branch** or the deploy run fails immediately.

- Real ancestry check (`git branch -r --contains` against explicitly fetched `rel/*` refs) — deliberately not a check on `release.target_commitish`, which can be a branch name **or a bare SHA** and is trivially re-taggable; ancestry is the property we actually care about.
- `validate` (the full-CI revalidation) now needs `branch-guard`, so an off-branch release dies with one clear error instead of burning a CI pass in parallel; `deploy` lists both needs explicitly so the gate can't silently loosen in a refactor.
- Guard runs before and outside the `production` environment — reviewers are only ever asked to approve a release that has proven both its branch origin and a green tag.
- Header comment and the release-cutting doc updated: cut/fast-forward a `rel/*` branch, then `gh release create --target rel/…`.

## Context

`rel/1.0` now exists at `ecf7e3f8` (v1.0.7's commit, the current prod deploy), so the released state is retroactively conformant: a re-publish of v1.0.7 would pass the guard.

## Notes

- Historical tags (v1.0.5 and earlier) are not on `rel/1.0`'s history-of-record path? They are — `rel/1.0` branched from main's tip, so every prior release commit is an ancestor and would also pass. The guard constrains the *future* shape: work merges to main, prod ships only after landing on a `rel/*` branch.
- Complementary (not in this PR, repo-settings side): the `production` environment's deployment branch/tag policy could additionally restrict by tag pattern — belt-and-braces if wanted, but the workflow guard is the auditable, in-repo enforcement.

### DIFF
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -4,22 +4,28 @@ name: Deploy prod
 # published in this repo. Mirrors deploy-dev.yml step-for-step (build an ARM64
 # image, run DB migrations, roll out the ECS service) with these differences:
 #   - trigger: a published GitHub release instead of a merge to main
-#   - guard:   RE-RUNS the entire CI pipeline (tests + the deploy build/smoke)
+#   - guard 1: the released commit must live on a rel/* release branch — see
+#              the `branch-guard` job. Prod ships only from release branches;
+#              a release cut from main (or any other ref) fails fast here,
+#              before any CI or human approval is spent on it.
+#   - guard 2: RE-RUNS the entire CI pipeline (tests + the deploy build/smoke)
 #              against the released tag — see the `validate` job. A release can
-#              be cut from ANY commit, and stale or partial check status on that
-#              commit is not trustworthy, so we revalidate from scratch rather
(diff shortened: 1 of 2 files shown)


jee7s:
Second commit `4ec35782` adds the companion visibility piece: a **pr-target-label** workflow that keeps a `target:<base>` label on every PR (`target:main` neutral, `target:rel/*` in a loud color), so the PR list shows at a glance what ships where now that rel/* branches exist.

Details: `edited` events are filtered to actual base retargets (title/body edits don't run it); stale `target:*` labels self-heal on retarget; a `workflow_dispatch` sweep backfills the PRs that were open before this landed (run it once after merge). Fork PRs would fail label writes with the read-only token — noted in the header as fine for this repo and a deliberate reason not to reach for `pull_request_target`.

thpr (approved):

PR #1148, opened 2026-07-20
scholarship ack email: configurable through settings

jee7s:
## Summary

The scholarship / payment-plan request ACK (subject "We received your scholarship / payment-plan request") is now configurable on **Settings → Email**, same section as `scholarshipNotifyEmail`.

- **Three new nullable `BoardSettings` columns** (additive migration `20260719200000_scholarship_ack_settings`): `scholarshipAckSubject`, `scholarshipAckMembershipBody`, `scholarshipAckProgramBody`.
- **Fallback semantics**: NULL or blank/whitespace → falls back verbatim to today's copy. Defaults live as exported constants (`DEFAULT_ACK_SUBJECT`, `DEFAULT_ACK_MEMBERSHIP_BODY`, `DEFAULT_ACK_PROGRAM_BODY`) in `scholarshipEmails.ts` so both routes and the UI (placeholder text) share one source of truth.
- **Plain-text rendering, not another HTML-authoring surface**: bodies are rendered by `renderAckBody()` — HTML-escape (`escapeHtml` from `@/lib/email-templates/base`) → token-replace with escaped values → wrap each non-empty line in `<p>…</p>`. No `dangerouslySetInnerHTML`, no preview iframe. This is a deliberate simplification vs. the outreach campaign templates, which had to add a sandboxed, scriptless preview iframe specifically because they let the board author raw HTML — see `settings/outreach/page.tsx`. The scholarship ACK sidesteps that whole class of bug by never accepting HTML in the first place.
- **`{{programName}}` token** on the program-variant body only, substituted with its escaped value.
- **The Shopify-seat-hold-failure branch keeps its hard-coded body** (operational-incident copy, deliberately not configurable) — but the *configured subject* still applies to that send.
- **Contract unchanged**: the ACK is still the applicant's only automatic email (`docs/PROGRAM_CAPACITY_AND_SCHOLARSHIPS.md` §5), just with configurable copy.
- **Migration is additive/nullable** — no backfill, no destructive step; authored via `prisma migrate diff` against a scratch DB with the full existing chain applied (never `migrate dev` against the shared dev DB), verified to apply cleanly and drift-free.

## Files

- `prisma/schema.prisma` + `prisma/migrations/20260719200000_scholarship_ack_settings/` — the three columns
- `src/lib/scholarshipEmails.ts` — `renderAckBody`, `resolveAckCopy`, default constants
- `src/app/api/membership/request-payment-plan/route.ts`, `src/app/api/programs/[id]/request-payment-plan/route.ts` — wired to read settings + resolve copy
- `src/app/api/settings/email/route.ts`, `src/app/settings/email/page.tsx` — new fields (same role gate, same validation/audit pattern as `scholarshipNotifyEmail`; bodies capped at 5000 chars)
- `docs/PROGRAM_CAPACITY_AND_SCHOLARSHIPS.md` §5 — one-line note

## Test plan
- [x] `npx prisma generate`
- [x] `npx tsc --noEmit` — exit 0
- [x] `npm run lint` — clean
- [x] `npm run test:ci` — 249 suites / 2256 tests passed
- [x] `INTEGRATION_DB=1 npm run test:integration` — 127 suites / 1204 tests passed
- [x] New unit tests: `renderAckBody` (XSS-inert escaping, token substitution, paragraph splitting) + `resolveAckCopy` fallback resolution (null/blank/whitespace)
- [x] Extended both routes' integration tests: configured copy used (subject + body, token substituted), unset falls back verbatim to defaults, Shopify-failure branch keeps hard-coded body under a configured program template
- [x] Extended `emailSettingsAPI.integration.test.ts` with a PUT round-trip + length-cap rejection test
- [x] Migration verified: applies cleanly on top of a scratch DB carrying the full existing chain, and `prisma migrate diff --exit-code` shows no drift afterward


thpr:
**`{{programName}}` is silently dead in the subject**

`src/lib/scholarshipEmails.ts` — `resolveAckCopy`:

```ts
const subjectRaw = settings?.scholarshipAckSubject?.trim();
const subject = subjectRaw || DEFAULT_ACK_SUBJECT;
```

The body substitutes `{{programName}}`; the subject doesn't. A board member who types `Your {{programName}} request received` into the subject field gets that literal token in the applicant's subject line. Nothing in the field's `description` warns them, and the subject sits directly above the program-body field (which does document the token) in the same UI section — so the expectation is a natural one to form.

Fix is one line, and the subject isn't HTML so no escaping is needed:

```ts
const subject = (subjectRaw || DEFAULT_ACK_SUBJECT).replaceAll("{{programName}}", vars.programName ?? "");
```

If leaving the subject token-free is deliberate, then say so in the subject field's `description` on `settings/email/page.tsx` instead.

thpr:
**The default ACK copy is duplicated verbatim in the client, and nothing catches the drift**

`src/app/settings/email/page.tsx` re-declares `DEFAULT_ACK_SUBJECT` / `DEFAULT_ACK_MEMBERSHIP_BODY` / `DEFAULT_ACK_PROGRAM_BODY` as placeholder text. The comment explains *why* the direct import fails (`scholarshipEmails.ts` pulls in prisma at module top level, which a `"use client"` page must never bundle) — that reasoning is correct, but a third module solves it without the copy/paste.

`renderAckBody` and the three `DEFAULT_ACK_*` constants only depend on `escapeHtml` from `@/lib/email-templates/base` — all client-safe. Moving them into e.g. `src/lib/scholarshipAckCopy.ts` and re-exporting from `scholarshipEmails.ts` keeps every current import path working and gives the page a real import.

Worth doing because the drift ships silently: the page test only asserts the placeholder *contains* `"Treehouse membership dues"` and `"{{programName}}"`, so rewording a default in the lib leaves the two copies out of sync with a green suite. Making it a shared import removes the need for a drift test at all.

thpr (changes requested):
see comments

jee7s:
Both addressed in `3ff47d1`.

1. **Subject `{{programName}}`** — took the one-line fix, mirroring `renderAckBody`'s missing-value fallback (`?? ""`) so subject and body agree when there's no program in scope (membership acks). Plain text, no escaping, per your note. The subject field's description on the settings page now documents the token, and two new test cases pin the substitution (with a value, and the empty-fallback).

2. **Duplicated defaults** — extracted `DEFAULT_ACK_SUBJECT` / `DEFAULT_ACK_MEMBERSHIP_BODY` / `DEFAULT_ACK_PROGRAM_BODY` and `renderAckBody` into `src/lib/scholarshipAckCopy.ts` exactly as you suggested (`escapeHtml` in `email-templates/base` is import-free, so the module is client-safe). `scholarshipEmails.ts` re-exports everything, so all existing import paths still work — verified by grepping every importer of the moved symbols. The settings page now imports the real constants, and its test asserts against the imports rather than hardcoded substrings, so the drift class is gone rather than guarded.

Both suites green (29 tests); `tsc --noEmit` shows no new errors (the two pre-existing `@aws-sdk/client-lambda` errors reproduce identically on the base commit).

thpr:
**Nit, follow-up to the subject-token fix (3ff47d1):** the subject is shared across both ACK variants, but the token only has a value on one of them.

A board member who sets the subject to `Your {{programName}} request` gets the right thing on program ACKs — and `Your  request` on membership-dues ACKs, since the membership variant substitutes the token with `""` (deliberate and unit-tested: `"Re:  request"`). The updated field description on `settings/email/page.tsx` says the token "is replaced with the program's name" without mentioning that membership-dues sends have no program, so the natural reading is that it's always safe to use.

One clause fixes it, e.g.: `"{{programName}}" is replaced with the program's name (empty for membership-dues requests).`

Not blocking — behavior is intentional and covered; this is purely about the description setting the right expectation.

thpr (approved):

PR #1156, opened 2026-07-20
sync core: schema + Google/Slack clients + desired-state engine (inert)

jee7s:
## Summary

PR1 (of 4) of the Google Groups + Slack membership sync feature. Fully inert — schema, migration, and `lib/sync/**` only. Nothing outside `lib/sync/`, `config.ts`, and the schema/migration changes. **No route, cron, or UI calls any of this yet** — PR2 wires it in.

- **`SyncState` ledger + desired-state-as-pure-function design.** `computeDesiredState(now)` reads live rosters/memberships and emits the full "who should be in what group" tuple set every time it runs — it is NOT event-driven and keeps no bookkeeping of its own. The ledger (`SyncState`) is purely an applied-state cache + error/retry record + invite-sent marker. Because desired state is recomputed live, program-boundary removal, membership-boundary removal, and the youth reason-union all fall out of a plain diff between "desired now" and "applied previously" — no special-case removal logic needed anywhere.
- **Warn-then-remove Slack semantics.** Per binding review (addendum A2), Slack access is not silently add-only: a lapsed Slack-channel row gets a warning email first (`removalWarnedAt` set), then a `conversations.kick` 7 days later if still lapsed. Re-enrollment before the 7-day mark clears the pending warning automatically (tested). Google Group removal stays immediate on boundary — only Slack warns.
- **Population rules (addendum A3/A4):** the direct-sync threshold is age 13+ (not "adult"), stacking with the existing rule that a synced minor's household leads are *also* synced. Program leads and core `ProgramVolunteer` rows ("staff") get the program's group/channel under the same age/lead rule, via new `staff_lead:program:<id>` / `staff_corevol:program:<id>` reason tokens.
- **`ProgramSlackAuth` — an isolated per-program Slack bot-token table**, deliberately not a column on `Program`: every `program.findMany/findUnique` without an explicit `select` would otherwise risk loading the secret into memory; a separate table means a caller must explicitly join to ever touch it. It also lets a future read-only DB consumer exclude this one table with a single `REVOKE`, rather than a column-level revoke on the hot `Program` table — the migration includes a defensive (currently no-op — s-read has no external reader of checkin data today) `REVOKE SELECT ... FROM PUBLIC` anchor for that.
- **Integration-off-when-unconfigured** throughout: `config.googleDirectoryConfigured()` is false unless both `GOOGLE_DIRECTORY_SA_KEY` and `GOOGLE_DIRECTORY_ADMIN_SUBJECT` are set (mirrors the `resendApiKey`/`shopifyReadDatabaseUrl` null-⇒-off pattern); a program with no Slack bot token simply leaves its rows `applied:false` for the nightly retry once configured. No behavior is possible to trigger from this PR regardless of env.

## Deviations from the design doc (flagged, with reason)

1. **`applyAdd`/`applyRemove` return a small result (`{ok, retryAfterMs?}`) instead of `Promise<void>`.** The doc's own §5.1 reconcile algorithm requires budgeting ops and honoring Slack's 429 `retryAfterMs` mid-run — undoable cleanly with a void return without fragile DB-state re-reads or string-parsing the persisted `error` field. Every DB write the design specifies still happens exactly as spec'd; this only adds a return value on top.
2. **`getGoogleDirectoryClient()` reuses one module-level token cache across every call** (not a fresh cache per factory invocation). The design's own text says the token is "cached in-module until ~60s before exp" — a fresh cache per call would defeat that, re-minting a JWT for every single external call in a reconcile run (up to `MAX_OPS`=200/run) instead of once. `resetDirectoryTokenCache()` (test-only) resets it.
3. **`security/scopeBindings.ts`: added `ProgramSlackAuth` to the existing `OPT_OUT_PENDING_ROUTE` work queue** (1 line + comment). This is the one `src/security/**` touch in the PR, and I want to flag it explicitly since the instruction was "no `src/security/**` edits." It was not optional: `ProgramSlackAuth` has an `internal`-tier field (`updatedAt`) and a `programId` field, which the `validateBindings` CI gate (`test:ci`) treats as "sensitive + scopable, needs a binding or a queued exemption" — an unconditional consequence of the spec's own schema (§2.2's exact field tiers), not a choice I made. The queue is the designed escape hatch for exactly this "model ships before its route" situation — see the adjacent `PaymentException` entry, added under the identical circumstance ahead of #1031. It resolves to a real `SCOPE_BINDINGS` row in PR3, alongside the `sync-status` route registry entry. Reviewers who consider this over the line can revert just that hunk; `test:ci` will fail again pending an alternative fix.

## Test plan
- [x] `npx prisma generate` (client + `classifications.ts` regenerated, committed)
- [x] bare `npx tsc --noEmit; echo $?` → 0
- [x] `npm run lint` → clean (`--max-warnings 0`)
- [x] `npm run test:ci` → 251 suites / 2271 tests passed
- [x] `INTEGRATION_DB=1 npm run test:integration` → 130 suites / 1230 tests passed (35 new, in `lib/sync/__tests__/`)
- [x] `livePersonDriftGuard` passes with **zero new allowlist entries** — every `desired.ts` Person/join-table query (`programParticipant`, `programVolunteer`) filters `LIVE_PERSON` directly; `apply.ts`/`reconcile.ts`'s person lookups are all `findUnique`-by-id (the guard's own documented exclusion), so no sweep-allowlist entry was needed either
- [x] Migration authored on a scratch database (`scratch_group_slack_sync`, dropped after review), never against shared `checkmein`

### DIFF
--- a/checkin-app/prisma/migrations/20260720024844_group_slack_sync/migration.sql
+++ b/checkin-app/prisma/migrations/20260720024844_group_slack_sync/migration.sql
@@ -1,0 +1,66 @@
+-- CreateEnum
+CREATE TYPE "SyncTargetKind" AS ENUM ('google_group', 'slack_channel', 'newsletter');
+
+-- AlterTable
+ALTER TABLE "BoardSettings" ADD COLUMN     "membersGoogleGroupEmail" TEXT,
+ADD COLUMN     "newsletterGoogleGroupEmail" TEXT;
+
+-- AlterTable
+ALTER TABLE "Program" ADD COLUMN     "googleGroupEmail" TEXT,
+ADD COLUMN     "slackChannelId" TEXT,
+ADD COLUMN     "slackWorkspaceInviteUrl" TEXT;
+
+-- CreateTable
+CREATE TABLE "ProgramSlackAuth" (
+    "programId" INTEGER NOT NULL,
+    "botToken" TEXT NOT NULL,
+    "updatedAt" TIMESTAMP(3) NOT NULL,
+    "updatedById" INTEGER,
+
+    CONSTRAINT "ProgramSlackAuth_pkey" PRIMARY KEY ("programId")
+);
+
+-- CreateTable
+CREATE TABLE "SyncState" (
+    "id" SERIAL NOT NULL,
+    "personId" INTEGER NOT NULL,
+    "targetKind" "SyncTargetKind" NOT NULL,
+    "targetRef" TEXT NOT NULL,
+    "scope" TEXT NOT NULL,
+    "desired" BOOLEAN NOT NULL DEFAULT false,
+    "applied" BOOLEAN NOT NULL DEFAULT false,
+    "lastAttemptAt" TIMESTAMP(3),
+    "error" TEXT,
+    "inviteEmailedAt" TIMESTAMP(3),
+    "removalWarnedAt" TIMESTAMP(3),
+    "reasons" JSONB NOT NULL DEFAULT '[]',
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updatedAt" TIMESTAMP(3) NOT NULL,
+
+    CONSTRAINT "SyncState_pkey" PRIMARY KEY ("id")
+);
+
+-- CreateIndex
+CREATE INDEX "SyncState_error_idx" ON "SyncState"("error");
+
+-- CreateIndex
+CREATE INDEX "SyncState_desired_applied_idx" ON "SyncState"("desired", "applied");
+
+-- CreateIndex
+CREATE UNIQUE INDEX "SyncState_personId_targetKind_targetRef_scope_key" ON "SyncState"("personId", "targetKind", "targetRef", "scope");
+
+-- AddForeignKey
+ALTER TABLE "ProgramSlackAuth" ADD CONSTRAINT "ProgramSlackAuth_programId_fkey" FOREIGN KEY ("programId") REFERENCES "Program"("id") ON DELETE CASCADE ON UPDATE CASCADE;
+
+-- AddForeignKey
+ALTER TABLE "SyncState" ADD CONSTRAINT "SyncState_personId_fkey" FOREIGN KEY ("personId") REFERENCES "Person"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
+
+-- ponytail: defensive anchor only — there is no read-only DB role granted SELECT on
+-- this database today (s-read mirrors Shopify INTO checkin, it does not export
+-- checkin data), so this REVOKE is a no-op against the app's own DML role. If a
+-- read-only consumer of the checkin DB is ever added (the NOLOGIN grant-holder
+-- pattern, infra/modules/checkin-bootstrap/bootstrap.sh does GRANT SELECT ON ALL
+-- TABLES), that migration must also exclude this table by name for that role and
+-- omit it from ALTER DEFAULT PRIVILEGES — see spec §9. PUBLIC is revoked now so no
+-- future default-privilege grant silently includes it.
+REVOKE SELECT ON "ProgramSlackAuth" FROM PUBLIC;
--- a/checkin-app/prisma/schema.prisma
+++ b/checkin-app/prisma/schema.prisma
@@ -157,6 +157,7 @@ model Person {
   rawBadgeLogs        RawBadgeLog[]
   visits              Visit[]
   eventsConfirmedBy   Event[]                      @relation("EventAttendanceConfirmer")
+  syncStates          SyncState[]
 
   trustedAdultRecordsAsAdult TrustedAdult[]       @relation("TrustedAdultPerson")
   trustedAdultsDisclosed      TrustedAdult[]       @relation("TrustedAdultDiscloser")
@@ -593,6 +594,12 @@ model BoardSettings {
   /// withdrawal or payment; it never auto-expires).
   /// @sensitivity:public
   scholarshipDenialGraceDays Int?
+  /// Pre-existing "members" Google group email. Membership approval adds; membership
+  /// boundary removes. Null = members-group sync off. @sensitivity:internal
+  membersGoogleGroupEmail    String?
+  /// Pre-existing newsletter Google group email. Membership approval adds; NEVER
+  /// auto-removed (R3). Null = newsletter sync off. @sensitivity:internal
+  newsletterGoogleGroupEmail String?
   /// @sensitivity:internal
   updatedAt                  DateTime  @updatedAt
 }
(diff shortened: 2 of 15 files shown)


jee7s:
Sequencing note: the one hand-authored `src/security/scopeBindings.ts` line (the `ProgramSlackAuth` pending-route opt-out, required by the bindings coverage gate) has been extracted to boundary PR **#1157** per the AGENTS.md rule — the #1114→#1104 pattern. This PR's **isolation check is expected-red until #1157 merges**; after that, updating this branch from main collapses the line out of the diff and isolation goes green. Everything else here is expected-green (full local gates: 2271 unit / 1230 integration).

thpr:
Reviewed this merged locally with #1157 on top. The two `scopeBindings.ts` hunks are byte-identical (`md5 3ca47f3f…`), so the merge is clean in either order and #1157 is a strict subset — no action needed there.

The design holds up: desired-state-as-a-pure-function is the right call, and boundary removal genuinely does fall out of the diff with no event bookkeeping. Everything below is in the **apply** layer, not the design, and all of it is a small diff.

I'd want 1 and 2 fixed before PR2 makes any of this reachable, and 3 and 4 before it points at a real Workspace/workspace.

---

### 1. Shared `targetRef` across scopes causes wrongful removal

`reconcile.ts:134` → `apply.ts:193`.

Ledger rows are keyed `(personId, targetKind, targetRef, scope)`, but the external op is keyed on `targetRef` alone. Two programs pasted with the same `googleGroupEmail` — two sections of one class, say — or a `membersGoogleGroupEmail` that equals a program's group: when the person lapses out of program A, `removeMember(targetRef, email)` fires and they lose the group they still hold via program B's still-`desired && applied` row. Same hazard for a shared `slackChannelId`.

Nothing in the removals loop checks for this. Suggested fix: before removing, confirm no other `desired && applied` row shares `(personId, targetKind, targetRef)`.

### 2. `applyAdd` is documented "Never throws" — it can

`apply.ts:85` (and `applyRemove`/`applySlackRemoval` alongside it).

There's no `try`/`catch` anywhere in `apply.ts` or `reconcile.ts`, and three paths throw:

- `getToken()` throws by design on a failed token exchange (`googleGroups.ts:97`)
- bare `fetchFn` throws on `ECONNRESET` and friends
- `res.json()` at `slack.ts:59`, `:78`, `:108` is unguarded and throws on a non-JSON error page (Slack does serve HTML on some 5xx/429s)

So one transient failure aborts the reconcile mid-loop, and the row it died on keeps `error: null` — the ledger says "fine" about the op that killed the run. Once PR2 wires this into `cron/nightly`, it takes the rest of nightly with it. That route already guards against exactly this shape with `Promise.allSettled` (`cron/nightly/route.ts:29`).

No coverage for it either — there's no `mockRejected` anywhere in `lib/sync/__tests__/`.

### 3. Adds starve removals under the op budget

`reconcile.ts:109-147`.

The adds loop runs first and can consume all 200 ops; the removals loop then hits `opsUsed >= MAX_OPS` on every row and defers. A standing backlog of failing adds — Google unconfigured, people not yet in the Slack workspace — means **revocation never runs at all, indefinitely**, while grants keep retrying nightly.

That's backwards for an access-control sync: removals are the direction with a security consequence. Either run removals first, or reserve a slice of the budget for them.

### 4. Email-send failures are swallowed, but the "sent" marker is set anyway

`sendEmail` returns `boolean`; both call sites ignore it.

- `apply.ts:147` — send fails, `inviteEmailedAt: now` is written regardless, and the path is explicitly never-resend. That person never gets a workspace invite, and their row sits `applied:false` forever with no error to show in the status view.
- `apply.ts:232-240` — worse. The `if (person?.email)` guard skips the send, but the `removalWarnedAt: now` update runs unconditionally, outside it. Send failure *or* missing email both set the marker, and the kick fires 7 days later. Warn-then-remove silently degrades to remove.

Existing convention checks the return value (`notifications.ts:61`, `postEventEmails.ts:127`).

### 5. No circuit breaker on mass-lapse

`reconcile.ts:75`.

`undesireLapsed` flips *every* `desired:true` row not in today's computed set. That's correct and intended for real boundaries — but the same code path fires when an admin typos `Program.googleGroupEmail` or clears `BoardSettings.membersGoogleGroupEmail`, and then it means "remove everyone from the old target," up to `MAX_OPS` per night, against a live Google Workspace.

The pure-function design is what makes this powerful and also what makes it unbounded. Worth a guard before the removals loop — abort and alert if the desired count dropped by more than some fraction versus the last run — rather than trusting that every config edit is deliberate.

---

Didn't re-run tsc/lint/jest (worktree had no `node_modules`); taking the PR's green report at face value. Findings above are static analysis.

thpr:
We can debate if #1 needs to be fixed or not, but the others deserve a pass

thpr (changes requested):
see comments

jee7s:
Resolved the merge conflict against `main` (caused by #1162 landing).

**Conflict:** exactly one file, `checkin-app/src/security/scopeBindings.ts`, inside the `OPT_OUT_PENDING_ROUTE` set literal. Both sides only *appended* new entries at the same insertion point — no shared value was actually in dispute:
- **main (#1162)** added `PersonRole` and `BulkSendItem` (models the personId-in-`SCOPABLE_FIELDS` fix newly surfaced as scopable-but-unbound) plus a comment-only fix on the existing `Corporation` entry (`participantId`→`personId`), which auto-merged cleanly outside the conflict hunk.
- **this PR** added `ProgramSlackAuth` (schema-only, fully inert, no route yet).

Resolution: kept all three additions (`ProgramSlackAuth`, `PersonRole`, `BulkSendItem`) with their original comments intact.

**One follow-on fix beyond the raw conflict:** after merging, `npx jest tests/security` failed one assertion — `validateBindings`'s coverage gate now flags `SyncState` (this PR's schema) as "sensitive and scopable but has no binding and is not in OPT_OUT_PENDING_ROUTE." This is a direct, mechanical consequence of combining both branches: `SyncState.personId` was invisible to the coverage check before #1162 taught `SCOPABLE_FIELDS` about `personId`; now it is, and the model has no route yet (same inert state as `ProgramSlackAuth`). I added `SyncState` to `OPT_OUT_PENDING_ROUTE` with a short comment pointing at the same eventual route (`GET /api/settings/sync-status`, spec §7.4/§12 PR3) that will give it a real `SCOPE_BINDINGS` entry.

No other files needed manual resolution — everything else (schema.prisma, config.ts, classifications.ts, etc.) auto-merged.

**Verification (in the merged worktree):**
- `npx prisma generate` — clean; regenerated `classifications.ts` byte-identical to the auto-merged version (no drift).
- `npx tsc --noEmit --pretty false` — 2 errors, both the pre-existing `@aws-sdk/client-lambda` environmental baseline, nothing new.
- `npx jest tests/security` — **7/7 suites, 541/541 tests passing** after the `SyncState` addition.
- `*.integration.test.ts` not run (no local Postgres available).

Nothing here was a genuine disagreement between the two branches — resolved without needing to escalate.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01F6TfaRtHA5C76d6A8yU4Nm

jee7s:
Closing in favour of the design doc in #1197 (`docs/designs/MEMBERSHIP_SYNC.md`), which covers both the email (Google Groups) and Slack sides of this feature.

Nothing here is discarded — the engine design is carried forward largely unchanged. Desired-state-as-a-pure-function, `SyncState` as an applied-state cache rather than a source of truth, boundary removals falling out of a plain diff, the 13+/household-lead population rules, and the isolated `ProgramSlackAuth` table are all restated in the doc.

What sent it back to design: the engine is add/remove symmetric and has no opinion about **who** did the removing. A person who leaves a Google group has made a decision, and today it survives only by accident — the ledger still reads `applied: true`, so nothing re-adds them, but any event flipping `applied` back to false would silently override them. Making that a real guarantee needs an observation capability the client does not have (`insertMember`/`removeMember` only, no `listMembers`), plus audit and a sticky opt-out. That is a model change, not a patch on top of this branch.

Re-opening this to rebase would have meant reviewing the new consent semantics inside a 2,500-line diff. The doc goes first; implementation follows its §7 sequence, and the boundary-isolation constraint this PR pair existed to satisfy still applies to that sequence's PR 3.

PR #1157, opened 2026-07-20
security: ProgramSlackAuth pending-route binding opt-out (boundary PR)

jee7s:
One line + comment in `scopeBindings.ts`: queue `ProgramSlackAuth` (the isolated Slack-bot-token table arriving with #1156) in `OPT_OUT_PENDING_ROUTE`, exactly like the existing `PaymentException` entry — schema-only model, no reader route yet, real binding lands with the sync-status route's boundary PR (the feature's PR3).

Why a separate PR: hand-authored `src/security/**` changes ship alone (AGENTS.md boundary rule). #1156 currently carries this same line and its isolation check is expected-red until this merges; then its diff collapses and isolation goes green — the #1114→#1104 pattern.

Validators verified green on main with the entry present (no stale-model coupling); tsc/lint/test:ci all clean on this branch.

### DIFF
--- a/checkin-app/src/security/scopeBindings.ts
+++ b/checkin-app/src/security/scopeBindings.ts
@@ -162,6 +162,17 @@ export const OPT_OUT_PENDING_ROUTE = new Set<string>([
     // served via withAuth on finance-ops/payments; a household-facing "your payment
     // problem" route is plausible later, and that is when this earns a real binding.
     'PaymentException',
+    // Same shape as PaymentException above: the model (ProgramSlackAuth — the
+    // isolated Slack-bot-token table, group-slack-sync PR1) ships schema-only,
+    // fully inert, with no route yet. `programId` makes it isScopable(); its
+    // `updatedAt`/`updatedById` internal-tier fields make it "sensitive" by the
+    // coverage check even though the actual secret (`botToken`) is tier `secret`
+    // and thus never reaches a GET select regardless (security/core.ts's stripper
+    // drops `secret` unconditionally — see spec §9). GET /api/settings/sync-status
+    // is the model's real reader; its registry entry lands in the boundary PR
+    // (spec §7.4/§12 PR3), which is when this earns a real SCOPE_BINDINGS row and
+    // leaves this queue.
+    'ProgramSlackAuth',
 ]);
 
 /**


jee7s:
## Merge conflict resolved

Merged `origin/main` in; the only conflict was in `checkin-app/src/security/scopeBindings.ts`, inside the `OPT_OUT_PENDING_ROUTE` set literal. Both sides had independently appended a new entry right after `'PaymentException'`:

- **This PR (HEAD)** added `'ProgramSlackAuth'` — the isolated Slack-bot-token table (group-slack-sync PR1), ships schema-only/inert with no route yet; opts out pending its real `GET /api/settings/sync-status` binding.
- **`main` via #1162** added `'PersonRole'` and `'BulkSendItem'` — backfilled entries surfaced by the `personId` fix to `SCOPABLE_FIELDS` (the old `participantId` name had been silently exempting every personId-only model).

There was no actual disagreement on any shared value — it was two independent additions to the same set, not a conflicting edit to a shared line. Resolution keeps **both sides in full**, with their original explanatory comments intact: `PaymentException` → `PersonRole` → `BulkSendItem` (main's, kept as-is) → `ProgramSlackAuth` (this PR's, appended after). No entry from either side was dropped or altered.

### Verification (in a worktree, Node 24)
- `npx tsc --noEmit --pretty false` in `checkin-app/`: **2 errors**, both the pre-existing/environmental `@aws-sdk/client-lambda` module-resolution baseline — nothing else.
- `npx jest tests/security`: **7 suites / 541 tests passed**, 0 failed.
- `*.integration.test.ts` were not run (no local Postgres available in this environment).
- `git diff --name-only origin/main...HEAD` → **only** `checkin-app/src/security/scopeBindings.ts`. Satisfies `.github/workflows/security-boundary-isolation.yml` (the file is itself a boundary file and its own companion; no unrelated files were pulled in by the merge).

No refactors or scope changes beyond the conflict resolution itself.

jee7s:
Closing in favour of the design doc in #1197 (`docs/designs/MEMBERSHIP_SYNC.md`), which covers both the email (Google Groups) and Slack sides of this feature.

Nothing here is discarded — the engine design is carried forward largely unchanged. Desired-state-as-a-pure-function, `SyncState` as an applied-state cache rather than a source of truth, boundary removals falling out of a plain diff, the 13+/household-lead population rules, and the isolated `ProgramSlackAuth` table are all restated in the doc.

What sent it back to design: the engine is add/remove symmetric and has no opinion about **who** did the removing. A person who leaves a Google group has made a decision, and today it survives only by accident — the ledger still reads `applied: true`, so nothing re-adds them, but any event flipping `applied` back to false would silently override them. Making that a real guarantee needs an observation capability the client does not have (`insertMember`/`removeMember` only, no `listMembers`), plus audit and a sticky opt-out. That is a model change, not a patch on top of this branch.

Re-opening this to rebase would have meant reviewing the new consent semantics inside a 2,500-line diff. The doc goes first; implementation follows its §7 sequence, and the boundary-isolation constraint this PR pair existed to satisfy still applies to that sequence's PR 3.

Issue #1154, opened 2026-07-20
batching of large e-mails


jee7s:
When an event triggers a large e-mail list, batch it in chunks of 5 per second.

thpr:
I think this is the wrong solution to the problem.  This system should not be batching emails at any more than 2/event, so I don't see a point in developing this - the problem is not quoted here at the right level, so this is framed as solving the wrong problem IMO.  The problem is we aren't using our Google Groups - so we're building a full email system not using assets we already have correctly.

jee7s:
The direction to take here, as discussed, is to use google groups for all communication. This means that resend issues one e-mail at a time in most cases -- either it sends to a group or it sends to an individual person.

PR #1158, opened 2026-07-20
fix(programs): opt-in announce + covered-member audience + paced fan-out (#1153, #1154)

### LINKED ISSUE #1153: Program creation e-mail controls
Make program creation notifications both an optional setting by a program leader and only send to members who have an active membership for the duration of the program year.

jee7s:
## ⚠️ Behavior change (read before merging)

**Every program that previously auto-announced on reaching UPCOMING + OPEN will now be silent by default.** Leaders must tick the new "Announce to members when enrollment opens" checkbox (Program editor → General tab) to re-enable the announce email for that program. Existing programs get `announceOnOpen = false` on migration — nothing sends until a lead mentor opts in.

## What changed

**#1153 — opt-in + covered-member audience**
- New `Program.announceOnOpen` boolean, default `false` (Jeff's call — no program announces unless a leader turns it on).
- The announce email now goes only to members whose household membership covers the **program's full duration** — reuses the existing #1061 predicate (`isActiveOrgMemberThrough` in `lib/orgMembership.ts`), one definition, not two.
- Toggle is a lead-mentor setting; no new roles (existing gates on `/api/programs/[id]` and `/api/programs/[id]/settings` already allow lead mentors).

**#1154 — paced fan-out**
- New `runPaced()` in `lib/email.ts`: runs tasks in chunks of 5, 1s between chunks.
- `emailRecipients.ts`'s `fanOutEmails` (used by `emailHouseholdLeads`/`emailBoardMembers`) now routes through it — shared by both the announce path and existing fan-outs, for free.
- The announce path fires **without awaiting** from the PATCH route (a big recipient list × 5/sec would otherwise block the PATCH response for tens of seconds); `notifyNewProgramAnnounced` swallows its own errors, `.catch` is belt-and-suspenders.

## Files touched
- `prisma/schema.prisma` + migration `20260719120000_program_announce_on_open` (single additive `ALTER TABLE ... ADD COLUMN`)
- `src/security/generated/classifications.ts` (regenerated)
- `src/lib/notifications.ts` (`notifyNewProgramAnnounced` now takes the program row, applies the coverage gate, sends paced)
- `src/lib/email.ts` (`runPaced`)
- `src/lib/emailRecipients.ts` (`fanOutEmails` → `runPaced`)
- `src/app/api/programs/[id]/route.ts` (toggle persists; fire-without-await gated on `announceOnOpen`)
- `src/app/api/programs/[id]/settings/route.ts` (toggle persists via the lead-mentor settings surface)
- `src/app/api/programs/route.ts` (public catalog `PUBLIC_PROGRAM_SELECT` mirror updated — not in the original spec, needed to keep the anonymous-catalog projection test green)
- `src/app/program-ops/programs/[id]/page.tsx` (new checkbox on the General tab)

## Tests
- Unit: `src/lib/__tests__/email.test.ts` (new `runPaced` pacing suite, fake timers), `src/lib/__tests__/notifications.test.ts` (opt-in filtering + coverage gate)
- Integration: `programAnnounceNotification.integration.test.ts` (toggle gates the fire), new `programAnnounceRecipients.integration.test.ts` (covered/uncovered/opted-out/tombstoned + the #1061 midyear-boundary case), `programsSettingsAPI.integration.test.ts` (lead-mentor round-trip)
- Drift guard (`livePersonDriftGuard.test.ts`) stays green — no allowlist entry needed.

## Gates (all from `checkin-app/`)
- `npx prisma generate` — clean
- `npx tsc --noEmit` — clean
- `npm run lint` (`--max-warnings 0`) — clean
- `npm run test:ci` — 249 suites / 2247 tests passing
- `INTEGRATION_DB=1 npm run test:integration` — 128 suites / 1200 tests passing

Refs #1153. Fixes #1153. Refs #1154.

### DIFF
--- a/checkin-app/prisma/migrations/20260719120000_program_announce_on_open/migration.sql
+++ b/checkin-app/prisma/migrations/20260719120000_program_announce_on_open/migration.sql
@@ -1,0 +1,2 @@
+-- Additive, NOT NULL with a default so existing rows backfill to false (#1153 opt-in).
+ALTER TABLE "Program" ADD COLUMN "announceOnOpen" BOOLEAN NOT NULL DEFAULT false;
--- a/checkin-app/prisma/schema.prisma
+++ b/checkin-app/prisma/schema.prisma
@@ -819,6 +819,10 @@ model Program {
   enrollmentStatus               EnrollmentStatus @default(CLOSED)
   /// @sensitivity:public
   orgMemberOnly                  Boolean          @default(false)
+  /// Opt-in (#1153): when true, reaching UPCOMING+OPEN emails covered members once.
+  /// Default false — no program announces unless a leader enables it.
+  /// @sensitivity:public
+  announceOnOpen                 Boolean          @default(false)
   /// @sensitivity:public
   minAge                         Int?
   /// @sensitivity:public
--- a/checkin-app/src/app/__tests__/membershipBgNonBlocking.integration.test.ts
+++ b/checkin-app/src/app/__tests__/membershipBgNonBlocking.integration.test.ts
@@ -20,7 +20,10 @@ import { beginRenewal } from '@/lib/membership/renewal';
 import prisma from '@/lib/prisma';
 import { sendEmail } from '@/lib/email';
 
-jest.mock('@/lib/email', () => ({ sendEmail: jest.fn().mockResolvedValue(true) }));
+jest.mock('@/lib/email', () => ({
+    sendEmail: jest.fn().mockResolvedValue(true),
+    runPaced: (tasks: Array<() => Promise<unknown>>) => Promise.all(tasks.map((t) => t())),
+}));
 
 const TAG = 'bg-nonblocking-test';
 
--- a/checkin-app/src/app/__tests__/membershipPaymentAPI.integration.test.ts
+++ b/checkin-app/src/app/__tests__/membershipPaymentAPI.integration.test.ts
@@ -17,7 +17,10 @@ import { getServerSession } from 'next-auth/next';
 import { sendEmail } from '@/lib/email';
 
 jest.mock('next-auth/next', () => ({ getServerSession: jest.fn() }));
-jest.mock('@/lib/email', () => ({ sendEmail: jest.fn().mockResolvedValue(true) }));
+jest.mock('@/lib/email', () => ({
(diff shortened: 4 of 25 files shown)


thpr (approved):
Approving. Migration is safely additive, the coverage predicate is reused rather than reimplemented (`isActiveOrgMemberThrough`, one definition), and the tests cover the real edges — default-off, boundary in/out, tombstone, opt-out.

Non-blocking follow-ups, tracked separately:

1. **`/api/programs/[id]/settings` accepts `announceOnOpen` but has no announce trigger.** `notifyNewProgramAnnounced` has exactly one call site, in `/api/programs/[id]`. A lead mentor who ticks the box and opens enrollment through the settings surface gets a 200 with `announceOnOpen: true` and no email. The missing trigger predates this PR; what's new is that the surface now accepts the toggle, which implies it's a place announcing is configured. Spinning a separate fix for this.

2. **N+1, serialized.** `notifications.ts:96` awaits `isActiveOrgMemberThrough` inside the loop — ~6 sequential queries per person (`isActiveOrgMember` → `person.findUnique` → `membershipValidThrough`'s household + boardSettings + renewal window + process probe), over every active member, with `boardSettings` id=1 re-fetched once per person. The `ponytail:` comment names the fix; grouping by `householdId` is a smaller diff than the comment.

3. **`announceOnOpen` is classified `@sensitivity:public`,** which forces it into `PUBLIC_PROGRAM_SELECT` (the projection test derives its expected keys from the classifications — that's why the catalog change was needed). The anonymous catalog now advertises each program's internal notification setting. `internal` is the more honest tier and drops the `programs/route.ts` change entirely.

Minor: ~20 test files still mock `@/lib/email` with `sendEmail` only. A future test reaching `fanOutEmails` throws `runPaced is not a function` *inside* `emailHouseholdLeads`' try block, which swallows and logs it — silent zero-emails, not a loud failure. `src/lib/__mocks__/email.ts` already has the passthrough; bare `jest.mock('@/lib/email')` kills the class.

Heads up on merge order: `origin/fix/program-announce-disable-rel-1.0` (`ba7aac6a`) deletes the announce call from the PATCH route entirely as a rel/1.0 mitigation for the same #1153. If that lands first this PR conflicts directly on route.ts:220-227. Worth settling which is the intended fix.

PR #1129, opened 2026-07-19
ui: green → primary token swap (no-op split from #1109)

jee7s:
## Summary

thpr, on #1109: *"Let's take the Green -> Primary into its own PR. That's effectively a no-op swap and is easy to clear"* ... *"do the green one now"*.

This PR extracts exactly that slice from #1109's color sweep (source commits `cb4a977a` + `e309d66e` on `claude/ui-audit-c`), so the remaining #1109 diff shows only the contested visual changes (cyan/family-info, orange/severity, categorical badges, hard collapses, spacing/P2 polish, the deleted off-palette drift-guard test).

## No-op verification

The app's theme (`checkin-app/src/brand/treehouse.ts`) sets `primaryColor: 'treehouseGreen'`. Checked Mantine 7.17.8's source directly (`node_modules/@mantine/core`) for `Button`, `Badge`, `Alert`, `ThemeIcon`, and `Notification`: each resolves an unset `color` prop to `theme.primaryColor` (e.g. `Button.mjs`: `color: color || theme.primaryColor`; CSS fallback `background: var(--button-bg, var(--mantine-primary-color-filled))`; same pattern in `Badge.css`, `Alert.css`, `ThemeIcon.css`, and `Notification.css`'s `--notification-color: var(--mantine-primary-color-filled)`).

So every hunk here is one of:
- **Drop the `color="green"` prop entirely** — the component falls back to `theme.primaryColor` (treehouseGreen), same rendered color.
- **Literal `'green'` → `'treehouseGreen'`** in a conditional/ternary or a status-map, where dropping isn't possible because the other branch needs an explicit color.

Both render identically to the brand's green swatch. Verified by diffing our result against the full `8f24851c..e309d66e` diff (script-checked): every line we changed exists verbatim in that diff, and every line in that diff we did *not* carry is confirmed to touch a color other than green (see below).

Also verified no drift: every "before" snippet was checked against current `main` before editing — none of these lines had moved since the source commits, so no hunk needed adaptation.

## Left behind (in #1109, not here)

Everything below stays in #1109 because it isn't a green→primary swap — it's a different color family, a contested UX change, or unrelated polish:

- `attendance/certifications/page.tsx`, `attendance/current/page.tsx` — teal badge/alert, unrelated to green.
- `facility-ops/print-badges/page.tsx` — Sticker button `grape`→`treehousePurple`.
- `finance-ops/shopify-holds/page.tsx` — two `blue`→`treehousePurple` buttons (manual-hold actions).
[...]

PR #1163, opened 2026-07-20
fix(ui): stop section-tab count pills from collapsing to empty capsules (#1155)

### LINKED ISSUE #1155: ensure all pills, including those in section tabs/headers properly display numbers

thpr:
Fixes #1155.

## Root cause

Not a colour bug. The number is in the DOM with good contrast — it is being **clipped to zero width**.

`ScrollableTabsList` set `flexWrap: "nowrap"` but nothing to force the row to overflow its `ScrollArea`. `nowrap` only stops wrapping; the flex row still shrinks to the viewport width. In that squeeze the count `Badge` is the only child that can collapse — Mantine's Badge root sets `overflow: hidden`, which zeroes its automatic minimum size, while the tab labels have `white-space: nowrap` and keep their min-content floor.

The pill's background and padding still paint, so it reads as an **empty capsule** rather than a missing badge.

## Measurement

Rendered `MembershipOpsLayout` with real counts, dumped the markup, and measured it in headless Chrome against Mantine's actual stylesheet:

| viewport | badge label width | needs |
|---|---|---|
| 1400px | 5.8px | 6px |
| 1200px | 5.8px | 6px |
| **1100px** | **0px** | 6px |
| 900px | 0px | 6px |
| 768px | 0px | 6px |

Membership Ops trips first because it has the most tabs (7) and the longest labels — which is why **Applications** and **Background-check Review** were the two reported.

The `ScrollArea` also never actually scrolled. It was dead code; the tabs squeezed instead.

## Fix

```diff
-<Tabs.List style={{ flexWrap: "nowrap", ...style }} {...props}>
+<Tabs.List style={{ flexWrap: "nowrap", minWidth: "max-content", ...style }} {...props}>
```

Verified in the same harness: label holds 5.8px at 1400 / 1100 / 900 / 768 / 390px, and the viewport becomes genuinely scrollable below 1155px.

One line in the shared component covers every section-tab bar — all seven ops layouts and `SettingsTabs` route through it. `AttendanceTabs` uses a raw `Tabs.List` but renders no badges.

## Ruled out

The original suspicion was pill text matching pill fill. Computed every Badge fg/bg pair from Mantine's own resolver against the treehouse theme, light and dark. Nothing is same-on-same — the two reported pills measure **9.70:1** (gray) and **7.21:1** (green). The brand-token swap in #1129 was contrast-neutral (mantine `green` light 2.17:1 → `treehouseGreen` light 2.21:1).

Separate pre-existing finding, not addressed here: `variant="light"` badges run 1.75–2.9:1 in light mode (yellow worst, e.g. `Maybe {tally.maybe}` on My Programs → Attendance). Worth its own issue.

## Tests

jsdom performs no layout, so the regression test asserts the style that prevents the collapse rather than the resulting width. 94 tests pass across `SectionTabs` + all `membership-ops` suites.

### DIFF
--- a/checkin-app/src/components/ui/ScrollableTabsList.tsx
+++ b/checkin-app/src/components/ui/ScrollableTabsList.tsx
@@ -20,7 +20,15 @@ interface ScrollableTabsListProps extends TabsListProps {
 export function ScrollableTabsList({ children, style, mb, ...props }: ScrollableTabsListProps) {
   return (
     <ScrollArea type="auto" scrollbarSize={6} mb={mb}>
-      <Tabs.List style={{ flexWrap: "nowrap", ...style }} {...props}>
+      {/*
+        min-width: max-content is what actually makes this scroll. nowrap alone only
+        stops wrapping — the flex row still shrinks to the ScrollArea's width, and the
+        first thing to collapse is the count Badge, whose `overflow: hidden` zeroes its
+        automatic minimum size (labels have `white-space: nowrap` and keep their
+        min-content floor). That rendered the tab pills as empty capsules below ~1150px
+        (#1155). With max-content the row overflows and the ScrollArea scrolls instead.
+      */}
+      <Tabs.List style={{ flexWrap: "nowrap", minWidth: "max-content", ...style }} {...props}>
         {children}
       </Tabs.List>
     </ScrollArea>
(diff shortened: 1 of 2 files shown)


dkaygithub (approved):
We gotta find a way to tell Claude to not tell the bugs life story in every comment

PR #1164, opened 2026-07-20
fix(programs): fire the announce trigger from the settings PATCH too

thpr:
~~Stacked on #1158.~~ **Update:** #1158 squash-merged (`80bfefef`); this branch is rebased onto `main` (dropped the pre-squash stacked commit) and the diff is now exactly this PR's changes. Post-rebase gates re-run: tsc clean, lint clean, settings + announce integration suites 26/26.

Review disposition (12 findings): 1, 4, 5, 8–12 fixed in `97ba3cc1` (formerly `af82adc3`); 2, 3, 6, 7 travel together in an in-progress follow-up PR (`announcedAt` idempotency + race-safe claim + blast audit + helper extraction). Details in the comment thread.

## The gap

Two endpoints can edit a program; only one announced.

- `programs/[id]` PATCH — has the trigger.
- `programs/[id]/settings` PATCH (the lead-mentor surface) — accepts `phase`, `enrollmentStatus`, and as of #1158 `announceOnOpen`, writes all three, returns 200, and contains no announce code at all.

So a lead mentor who ticks the announce box and opens enrollment through the settings endpoint gets `announceOnOpen: true` back in the response body and no email is ever sent. Silent.

The missing trigger predates #1158. What #1158 changed is that the settings endpoint now accepts the toggle, which implies to any caller that it's a place where announcing is configured. No frontend calls it today — this is API-contract risk, not a live user bug.

The alternative fix — dropping `announceOnOpen` from the settings endpoint — is the shorter diff, but #1158 deliberately added the field there *and* added a passing round-trip test for it, so supporting it on the lead-mentor surface reads as intentional.

## The fix

Mirrored the trigger, semantics identical:

- fires only on the transition INTO (`UPCOMING` && `OPEN`) — not on every save while already there, not when only one of the two is set
- gated on `announceOnOpen`, log-and-skip when off
- fire-without-await (`void ... .catch(...)`) — the paced send is ~(recipients/5) seconds and must not block the response

Both sides of the edge were already in scope (the route loads `currentProgram` before its update), so no extra query.

**Duplicated rather than extracted.** A shared helper would have to live in `@/lib/notifications`, and both announce suites mock that module with a factory supplying only `notifyNewProgramAnnounced` — the helper would come back `undefined` and every mocking suite would need new plumbing. More plumbing than the six lines save. Marked with a `ponytail:` comment naming the reason.

## Tests

5 cases added to `programsSettingsAPI.integration.test.ts`, mirroring `programAnnounceNotification.integration.test.ts`: fires on the crossing with `announceOnOpen: true`; does not fire on the same crossing at the `false` default; does not re-fire while already UPCOMING+OPEN; does not fire on phase-only; does not fire on enrollment-only.

Verified non-vacuous — stashing the route change alone fails `fires once when the settings PATCH crosses INTO UPCOMING + OPEN` (`1 failed, 15 passed`), and it passes with the change.

Also widened that file's `afterAll` program cleanup from `id: targetProgramId` to delete by name — the new cases create their own programs and would otherwise leak and block the person deletes on FK.

## Gates

- `npx tsc --noEmit` — clean
- `npm run lint` — clean (`--max-warnings 0`)
- integration run scoped to `programsSettingsAPI|programAnnounce` — 3 suites, 24 tests, all passing

## Noted, not fixed

23 existing test files mock `@/lib/email` with a factory and don't supply `runPaced`. Pre-existing from #1158, none on this path. If any of them reach `fanOutEmails` it throws inside a try that swallows and logs — a silent zero-emails pass, not a loud failure. Worth a separate sweep.

### DIFF
--- a/checkin-app/src/app/__tests__/programsSettingsAPI.integration.test.ts
+++ b/checkin-app/src/app/__tests__/programsSettingsAPI.integration.test.ts
@@ -10,14 +10,17 @@ import { PATCH } from '@/app/api/programs/[id]/settings/route';
 import prisma from '@/lib/prisma';
 import { getServerSession } from 'next-auth/next';
 import { notifyNewProgramAnnounced } from '@/lib/notifications';
+import { logger } from '@/lib/logger';
 // Mock NextAuth
 jest.mock('next-auth/next', () => ({
     getServerSession: jest.fn(),
 }));
 jest.mock('@/lib/notifications', () => ({
-    // The settings route's fire-without-await edge does
-    // `notifyNewProgramAnnounced(...).catch(...)` — the mock must resolve so
-    // `.catch` is defined (a bare jest.fn() returns undefined and throws).
+    // Keep the real module (sendNotification etc.) and stub only the announce
+    // fan-out. The fire-without-await edge does `notifyNewProgramAnnounced(...)
+    // .catch(...)` — the stub must resolve so `.catch` is defined (a bare
+    // jest.fn() returns undefined and throws).
+    ...jest.requireActual('@/lib/notifications'),
     notifyNewProgramAnnounced: jest.fn().mockResolvedValue(undefined),
 }));
 
@@ -84,12 +87,22 @@ describe('Program Settings API Integration Tests', () => {
     afterAll(async () => {
         const existingUserIds = [adminId, leadId, newLeadId, commonId].filter(id => id !== undefined);
 
+        // Participant cleanup mirrors the program sweep below — id-scoped alone
+        // would FK-block deletion if an announce-case program ever enrolls anyone.
+        await prisma.programParticipant.deleteMany({
+            where: { program: { name: { contains: 'Settings API Test' } } }
+        });
         if (targetProgramId) {
             await prisma.programParticipant.deleteMany({
                 where: { programId: targetProgramId }
             });
+            // Id anchor: survives even if a test renames the program out of the
+            // naming convention the sweep below depends on.
+            await prisma.program.deleteMany({
+                where: { id: targetProgramId }
+            });
         }
-        // Covers targetProgramId plus the per-test programs the announce cases create.
+        // Name sweep: the per-test programs the announce cases create.
         await prisma.program.deleteMany({
             where: { name: { contains: 'Settings API Test' } }
         });
@@ -283,6 +296,23 @@ describe('Program Settings API Integration Tests', () => {
              const res = await PATCH(req as unknown as import("next/server").NextRequest, createParams(targetProgramId) as unknown as never);
              expect(res.status).toBe(400);
         });
+
+        it('should reject bad announceOnOpen / phase / enrollmentStatus values with 400, not 500', async () => {
+             (getServerSession as jest.Mock).mockResolvedValue({ user: { id: adminId, isSysadmin: true } });
+
+             for (const body of [
+                 { announceOnOpen: 'yes' },
+                 { phase: 'upcoming' },          // enum values are uppercase
+                 { enrollmentStatus: 'ajar' },
+             ]) {
+                 const req = new Request(`http://localhost:4000/api/programs/${targetProgramId}/settings`, {
+                     method: 'PATCH',
+                     body: JSON.stringify(body)
+                 });
+                 const res = await PATCH(req as unknown as import("next/server").NextRequest, createParams(targetProgramId) as unknown as never);
+                 expect(res.status).toBe(400);
+             }
+        });
     });
 
     // The settings PATCH is the lead-mentor edit surface and accepts phase /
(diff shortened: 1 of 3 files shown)


jee7s:
## Code review — 12 findings

Scope note: reviewed **`de73433d`** (this PR's own commit) with enclosing functions in scope. `df4d6d35` (#1158) is excluded as separately reviewed, per the PR description.

The change does what it says and the four transition tests are the right four. Findings below are ordered most severe first.

### Correctness

**1. `settings/route.ts:127` — a throw in the announce block turns a committed write into a 500.**
The call sits inside the route's main `try`. `void notifyNewProgramAnnounced(...).catch(...)` only routes *async* rejection; if the reference is ever `undefined` (partial module mock, refactor, bad import) the call throws **synchronously**, before `.catch` attaches, and lands in the outer `catch (error)` at :135. The program update and audit log have already committed, so the caller sees a 500 for a write that succeeded. This isn't hypothetical — this PR's own test file documents the shape at :17: *"a bare `jest.fn()` returns undefined and throws"*. Wrap the announce block in its own `try`/`catch`.

**2. `settings/route.ts:122` — no idempotency marker; announce re-fires on every re-entry.**
The guard is only "the previous row wasn't UPCOMING+OPEN". Nothing records that a program was ever announced. Close enrollment and reopen it — or move phase away and back — and the **entire active membership is emailed again**. `schema.prisma` has `announceOnOpen` but no `announcedAt`. This PR doubles the number of endpoints that can trip it.

**3. `settings/route.ts:120` — read-then-write race double-sends.**
`currentProgram` is read at :20, long before the update at :100. Two concurrent PATCHes (or one here and one on `programs/[id]`) both observe `wasAnnounced === false` and both fire. Two full-membership blasts. A conditional update (`updateMany` with the pre-state in `where`) or a unique `announcedAt` write would collapse it.

**4. `lib/notifications.ts:81` — doc comment is now wrong.**
> `The announceOnOpen opt-in gate lives at the CALL SITE (programs/[id] PATCH)`

There are now **two** call sites. Anyone auditing "where can this fan-out fire from?" reads this and stops at one.

**5. `settings/route.ts:82` — `announceOnOpen` / `phase` / `enrollmentStatus` written without validation.**
Assigned straight into `updateData` with no type or enum check. A non-boolean `announceOnOpen` (or a lowercase `phase`) produces a Prisma validation error caught by the generic handler → **500 instead of 400**. Pre-existing in the touched function, but this PR makes `announceOnOpen` gate an email blast.

**6. `settings/route.ts:105` — the fan-out isn't audited.**
`AuditLog` records the field edit, but nothing records "an email to every active member was triggered, by this actor, at this time". Given the reach, the send deserves its own trail rather than being inferred from a `phase`/`enrollmentStatus` diff.

### Altitude

**7. `settings/route.ts:120` — the transition rule is duplicated verbatim; the stated justification doesn't hold.**
The `ponytail:` comment defends the copy with *"a shared helper in `@/lib/notifications` would need mock plumbing in every suite that stubs that module"* — true of that module, but a helper in a **new** module (`lib/programAnnounce.ts`) has no such problem, and suites that don't touch announcing never mock it. Two copies of a four-line state-transition rule is where the third editor of a program forgets one. This is the fix being applied at call-site depth when it belongs one level down.

### Test coverage

**8. `programsSettingsAPI.integration.test.ts:291` — the most likely real flow is untested.**
Every added test pre-sets `announceOnOpen` at creation. Nothing covers **enabling `announceOnOpen` in the same request as the transition** — tick the box and open enrollment in one save — which is exactly what the lead-mentor UI would send. That path works (both land in `updateData` before the read-back), but it's the case a regression would break first.

**9. `programsSettingsAPI.integration.test.ts:291` — the skip branch is unasserted.**
The default-off test asserts only `not.toHaveBeenCalled()`. The `else` at :130 emits a `logger.info` audit breadcrumb that nothing verifies, so it could be deleted or corrupted silently.

**10. `programsSettingsAPI.integration.test.ts:93` — teardown lost its id anchor.**
The id-based `program.deleteMany({ id: targetProgramId })` was replaced by `name: { contains: 'Settings API Test' }`. Correct today (I checked: the only non-matching rename, `'Hacked Settings Program'` at :127, is a 403 that never persists). But teardown now depends on every test keeping the program inside a naming convention — the first successful rename out of that prefix leaks a row silently. Keeping the id delete alongside the name sweep costs one line.

**11. `programsSettingsAPI.integration.test.ts:88` — participant cleanup didn't follow the widening.**
`programParticipant.deleteMany` is still scoped to `targetProgramId`, while program deletion is now name-wide. The first announce test that enrolls a participant will FK-fail teardown and take the whole suite's cleanup with it.

**12. `programsSettingsAPI.integration.test.ts:17` — the module mock is wider than the need.**
The factory replaces **all** of `@/lib/notifications` with a single export, so `sendNotification` and `sendCheckinNotifications` are `undefined` for every test in the file. Nothing reachable calls them today; the day something does, the failure is a `TypeError` at call time rather than an obvious missing mock.

---

Findings 1–3 are the ones I'd want resolved before merge; 1 is a wrong response for a successful write, and 2–3 are duplicate mass emails.

dkaygithub (approved):

thpr:
Disposition of the 12 findings, per discussion:

**Fixed here, `af82adc3`:**

| # | Finding | Fix |
|---|---------|-----|
| 1 | sync throw → 500 after committed write | announce block owns its errors (own `try`/`catch`) |
| 4 | stale doc comment | `notifyNewProgramAnnounced` doc names both call sites |
| 5 | unvalidated `announceOnOpen`/`phase`/`enrollmentStatus` → 500 | up-front validation → 400, scoped to the three fields that now gate the blast; enums from the generated client, no new source of truth |
| 8 | same-request enable+transition untested | test added — fires once |
| 9 | skip branch unasserted | default-off test asserts the breadcrumb via `logger.info` spy |
| 10 | teardown lost id anchor | id delete kept alongside the name sweep |
| 11 | participant cleanup didn't follow the widening | widened to the same name scope |
| 12 | module mock wider than the need | `...jest.requireActual`, only the fan-out stubbed |

Gates after fixups: tsc clean, lint clean, settings + announce integration suites 26/26.

**Deferred to one follow-up PR (in progress):** 2, 3, 6, 7 travel together — `announcedAt` column (idempotency), conditional `updateMany` claim at the fire point (race), AuditLog row with actor there (audit), all landing inside the extracted helper in a new `lib/programAnnounce.ts` (your finding-7 counterpoint was right: a *new* module dodges the mock-plumbing objection). Extraction rides with the rewrite rather than preceding it so the helper API is designed once, in its final idempotent shape. Endpoint A's copy of the finding-1 pattern dies there too.

Not moved on finding 7's "duplicate now" reading only in sequencing, not substance — the duplication this PR leaves is scaffolding the follow-up deletes.

thpr:
@jee7s Updated & #1194 Spun out due to it covering pre-existing issues to avoid confusion about what this PR does (equality not fix)

jee7s (approved):

PR #1166, opened 2026-07-20
feat(security): ops-stg access gate — canAccessStaging, 4-surface enforcement

jee7s:
> ⚠️ **Split for `security-boundary-isolation`:** the `resolveAccess` boundary edit moved to the stacked companion **#1168**. Merge **this PR first**, then #1168. Together they are the complete 4-surface gate.

The **access gate** for staging, and the security-critical PR of the batch. ops-stg holds a copy of prod data, uses prod's Google OAuth client, and is deliberately **not** restricted to the org workspace (so a sysadmin-settable role can admit an outside collaborator). Any Google account on earth can complete sign-in → this gate is the only thing between that stranger and the data.

Policy: `allowed iff (hd == innovationtreehouse.org && emailVerified) || canAccessStaging`.

### What's here
- `Person.canAccessStaging` (additive migration, hand-verified `ADD COLUMN … DEFAULT false`), sysadmin-settable, **deliberately outside** the #1104 board-symmetric role matrix (that matrix lets a board member grant almost anything).
- Enforced on **four** surfaces, not the three originally planned — adversarial review found `authenticateRequest` *downgrades* to anonymous rather than denying, so bare/public routes (`/api/programs`, `/api/attendance`) leaked. Gate now sits in middleware (pages), `authenticateRequest` (API), `resolveAccess` (the `authorize:'public'` path), and explicit denies on the two bare routes.
- **Drift-guard test** (`routeAuthDrift`) — every `getOptionalSessionUser` route must gate or be explicitly allowlisted, so route #8 can't reintroduce the hole.
- Fuse hardened: `isStaging()` is `CHECKIN_STAGING === '1'` **or** the `ops-stg.` host (from `NEXTAUTH_URL`, can't be forgotten) — it fails *open*, unlike `CHECKIN_ENV`, so it's derived as well as declared. `staging:` surfaced in `/api/health`.
- Zoho "NOT A LEGAL AGREEMENT" watermark now gated `isProd() && !isStaging()` — staging would otherwise emit documents indistinguishable from binding agreements.

### Tests
Fail-closed matrix incl. the regression case (anonymous `GET /api/programs/[id]` on staging → 401, roster name absent), run against the **real** route chain, not a gate mock. Full unit (2304) + integration (1215) green; `tsc` + lint clean.

### ⚠️ Merge ordering
**This must be live before the first `rel/` branch is cut**, or the first staging deploy serves prod data ungated. CODEOWNERS-gated (`schema.prisma`, `lib/auth.ts`, `src/security/`).


thpr (changes requested):
Request changes on one usability item (logic otherwise looks solid).

### Staging dead-ends unauthenticated org members

On ops-stg an anonymous org member who hits any page is redirected to the bare `/access-denied`, never to `/signin` — unlike the dev-instance gate a few lines down in `middleware.ts`, which sends unauthenticated callers to Google login with a `callbackUrl`.

The staging redirect target is deliberate (documented in the comment + covered by a test), and the rationale holds for *authorization* failures: any Google account can complete sign-in, so re-prompting login would not help a stranger. But it also catches the *un-authenticated* case: a legitimate org member who simply has not logged in yet gets a dead-end page with no login affordance and has to know to hand-type `/signin`.

Two ways to resolve:
- If `/access-denied` already renders a visible sign-in link, that is enough — please confirm it does.
- Otherwise, either point the anonymous case (no token at all) at `/signin` like the dev gate, or add a sign-in affordance to `/access-denied`. Signing in does not by itself grant access (still gated on verified-org-member || canAccessStaging), so routing an anonymous caller to login stays safe.

Not a security issue — no data exposure — but it makes staging hard to actually log into for the intended users.

checkin-app/src/app/api/roles/route.ts
```diff
@@ -37,6 +40,9 @@ export const GET = withAuth(
                     name: true,
                     dateOfBirth: true,
                     roles: { select: { role: true } },
+                    // NOT one of the five ROLE_FLAGS — a plain column exposed alongside them
```
thpr, line 43: Why NOT a role?

checkin-app/src/app/api/roles/route.ts
thpr, line 156: Does this NOT audit staging being changed or does it incorrectly write staging under PersonRole?

jee7s:
Fixed in d5b78fea — and to answer the question directly: **no, `/access-denied` does not render a sign-in link.** It renders the words "Access Denied" and nothing else. The file comment is explicit that this is deliberate:

> Landing screen for a household whose membership is DENIED. By deliberate product decision this page leaks nothing — no reason, no contact, no membership wording, no sign-out affordance. Just the words "Access Denied".

So option one was out, and adding an affordance there is the wrong lever — that page is **shared with the household-denial gate**, which specifically must not offer one. Took your other option instead.

**The gate now splits on authentication, not authorization:**

```ts
if (isStagingAccessAllowed(token)) return NextResponse.next();
if (!token) {
    const signInUrl = new URL('/signin', req.url);
    signInUrl.searchParams.set('callbackUrl', req.nextUrl.pathname + req.nextUrl.search);
    return NextResponse.redirect(signInUrl);   // anonymous — same as the dev gate
}
return NextResponse.redirect(new URL('/access-denied', req.url));  // authenticated, unauthorized
```

Your reasoning for the `/access-denied` target still holds where it applies, so that path is unchanged: any Google account can complete sign-in on ops-stg, so re-prompting a stranger would just loop them through Google to the same wall. It only ever applied to the *authenticated* case.

**No loop — pinned rather than argued.** A caller returning from sign-in has a token, so they take the authenticated branch. Three tests cover it:

```
✓ sends an ANONYMOUS visitor to /signin with a callbackUrl (never next()) — /access-denied has no sign-in affordance
✓ does not loop: the caller returning from sign-in still fails the predicate and lands on /access-denied
✓ still sends a DENIED household to /access-denied, not /signin, even with no staging access
```

That third one is the case worth guarding: the denial gate runs *before* the staging block, so a denied session must never be handed a login round-trip by the new anonymous branch.

Also refreshed the middleware header and the test-file docblock, both of which stated the old "a failing caller is bounced to /access-denied, not /signin" rule flatly.

**21/21** in `middlewareDenied.test.ts`, `eslint` clean, `tsc --noEmit` 0 errors.

checkin-app/src/app/api/roles/route.ts
jee7s: Because it isn't org authority — it's environment access, and the two systems answer different questions.

The five `ROLE_FLAGS` are `PersonRole` rows governed by the authority matrix in `lib/roles.ts`: `setRoleFlag` decides *who may grant what* and throws `RoleMatrixError`, with board members able to grant some of them. `canAccessStaging` has none of that shape — it's sysadmin-only, it grants no authority inside the app, and its typical holder is an outside collaborator with **no** org role at all. Modelling "can reach ops-stg" as a role would mean inventing a role that confers nothing.

The deciding reason is claims. Roles flow through `rolesToFlags` into `assignParticipantClaims`, so they become part of the authority claim set the app authorizes on **in every environment**. Making this a role would put an ops-stg door key into prod's authorization claims — a much wider blast radius than a boolean column read by one middleware gate and one API chokepoint.

So it stays a plain column, exposed alongside the flags for UI convenience but deliberately not one of them — which is what the comment on this line is marking.

Your other comment turned out to be the sharper end of exactly this distinction; fixed in cbffdea0.

checkin-app/src/app/api/roles/route.ts
jee7s: The second one — it **did** audit the change, but wrote it under `tableName: "PersonRole"`, which is wrong. Good catch. Fixed in cbffdea0.

`canAccessStaging` is a column on `Person`; `PersonRole` rows are a different table. Folding both into one row filed a Person change under the role table — the same metadata drift #1179 is cleaning up elsewhere — and the practical cost is that a staging grant is invisible to anyone querying the audit log for `Person` changes, which is the query you'd actually run to ask *who was let into staging*.

Now the accumulators are split: role flags keep their `PersonRole` row, `canAccessStaging` gets its own with `tableName: "Person"`. Both are written in the same transaction, so a PATCH that changes both still audits both atomically — each is just filed under the table it touched. A request changing only one still writes only one row, so the existing no-op test holds.

Worth noting *why* this drifted quietly: the tests asserted `oldData`/`newData` but never `tableName`. They do now — the single-change test pins `"Person"`, and a new test drives one request through both tables and asserts two rows with the right `tableName` and neither carrying the other's field.

These are integration tests and I could not run them locally (no Postgres on this box; they're gated behind `npm run test:integration`), so they need a CI run to confirm. `eslint` clean and `tsc --noEmit` 0 errors.

thpr (approved):

PR #1168, opened 2026-07-20
feat(security): staging gate at the resolveAccess boundary (isolated, stacked on #1166)

jee7s:
The **security-boundary half** of the ops-stg access gate (PR #1166), split out because `security-boundary-isolation.yml` requires any change under `src/security/**` to land in isolation — a widening/narrowing of what data leaves the server should never be buried in a feature diff.

**Stacked on #1166** (base is `feat/staging-access-gate`) — it imports `isStagingAccessAllowed` and the `canAccessStaging` column from there. **Merge #1166 first**, then this.

### The one change
`security/access-resolvers.ts` — `resolveAccess` denies on staging **ahead of every `authorize` branch, including `'public'`**. This is the surface the other three (middleware, `authenticateRequest`, the two bare routes) structurally cannot cover: a registered `authorize:'public'` route returns `allowed` regardless of session, so without this an authenticated non-org caller — or the anonymous public path — reaches copied prod data through `handler() → registry → resolveAccess`. Adversarial review ranked deleting this block as the #1 blast radius.

### Companions (tests only)
- `tests/security/resolveAccess.test.ts` — the predicate at the boundary.
- `programsIdAPI.integration.test.ts` — anonymous `GET /api/programs/[id]` on staging → 401, enrolled name absent from the body, driven through the **real** route chain (not a gate mock). Delete the gate and this fails — that's the point.

### DIFF
--- a/checkin-app/src/app/__tests__/programsIdAPI.integration.test.ts
+++ b/checkin-app/src/app/__tests__/programsIdAPI.integration.test.ts
@@ -286,15 +286,15 @@ describe('Individual Program API Integration Tests', () => {
     // this describe is that regression test, run against the REAL route (handler() ->
     // registry -> resolveAccess -> stripBag), not a mock of the gate.
     describe('GET /api/programs/[id] — ops-stg access gate (regression: authorize:"public" must not bypass staging)', () => {
-        const CHECKIN_STAGING_BEFORE = process.env.CHECKIN_STAGING;
+        const CHECKIN_ENV_BEFORE = process.env.CHECKIN_ENV;
 
         beforeEach(() => {
-            process.env.CHECKIN_STAGING = '1';
+            process.env.CHECKIN_ENV = 'stg';
         });
 
         afterAll(() => {
-            if (CHECKIN_STAGING_BEFORE === undefined) delete process.env.CHECKIN_STAGING;
-            else process.env.CHECKIN_STAGING = CHECKIN_STAGING_BEFORE;
+            if (CHECKIN_ENV_BEFORE === undefined) delete process.env.CHECKIN_ENV;
+            else process.env.CHECKIN_ENV = CHECKIN_ENV_BEFORE;
         });
 
         it('DENIES an anonymous caller — the regression case for the defect that shipped real minors\' data to curl', async () => {
@@ -347,8 +347,8 @@ describe('Individual Program API Integration Tests', () => {
             expect(res.status).toBe(200);
         });
 
-        it('is inert outside staging (CHECKIN_STAGING unset): the same anonymous request that gets denied above still succeeds', async () => {
-            delete process.env.CHECKIN_STAGING;
+        it('is inert outside staging (CHECKIN_ENV not stg): the same anonymous request that gets denied above still succeeds', async () => {
+            process.env.CHECKIN_ENV = 'prod';
             (getServerSession as jest.Mock).mockResolvedValue(null);
 
             const req = new Request(`http://localhost:4000/api/programs/${publicProgramId}`, { method: 'GET' });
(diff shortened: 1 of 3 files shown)


jee7s:
**Retargeted to `main` deliberately, while #1166 is still open. This PR is knowingly wrong until #1166 merges — read this before reviewing or merging it.**

## Why the base moved

Stacked PRs lose their target when the base branch is deleted: GitHub closes PRs whose base no longer exists, and auto-retarget doesn't reliably rescue them. Rather than depend on that, the base is pinned to `main` now. The cost is that this PR is temporarily misleading; the benefit is that it cannot be silently closed or orphaned when `feat/staging-access-gate` is deleted on merge.

## What is wrong with it right now

- **The diff is a superset of #1166.** Against `main` it shows **32 files / ~1245 lines**, because this branch carries #1166's commits. Its real content is **3 files / 345 lines**.
- **It does not compile against `main`.** `access-resolvers.ts` imports `isStagingAccessAllowed` from `@/lib/config`, which #1166 introduces and `main` does not have yet.
- **`security-boundary-isolation` will fail, correctly.** `access-resolvers.ts` is a boundary file, so every other changed file must be a test/doc/schema annotation — and the borrowed #1166 files (`middleware.ts`, `auth.ts`, `config.ts`, `RolesEditModal.tsx`, route handlers) are not. That check failing is the expected state, not a thing to work around.

**Converted to draft** so it can't be merged in this state — merging it would land all of #1166 without #1166's own sign-off. One click to undo once it's cleaned up.

## Recovery, after #1166 merges

#1166 will be **squash-merged**, so `main` gets one new commit whose content matches this branch's borrowed history but shares no ancestry with it. Merging `main` back in therefore conflicts add/add on every shared file — the standard squash-merge trap, and not the way to fix this.

Reset onto `main` and replay only this PR's own work instead:

```
git fetch origin
git checkout feat/staging-gate-boundary
git reset --hard origin/main
git cherry-pick 44518a2b   # feat(security): staging gate at the resolveAccess boundary (isolated)
git cherry-pick 3ed6b937   # test(staging): engage the resolveAccess gate with CHECKIN_ENV=stg
git push --force-with-lease
```

(Skip `6845a022` — it is the merge of `feat/staging-access-gate` into this branch, i.e. exactly the borrowed #1166 content.)

Pre-retarget head, if anything needs to be recovered: `3ed6b93714d6b92834b9e5b2e1ddae5002f19dd9`.

After that the diff returns to its intended 3 files, `isStagingAccessAllowed` resolves, and the boundary-isolation check passes.

jee7s:
Rebuilt on `main` now that #1166 has merged. **`MERGEABLE`, back to its intended 3 files.**

Reset onto `main` and replayed only this PR's own two commits, rather than merging `main` in. #1166 was squash-merged, so `main` carries one commit whose *content* matches the borrowed history on this branch but shares no ancestry with it — merging would have conflicted add/add on every shared file. (Same trap that bit infra #141 earlier; the recovery was written down in the comment above before it was needed.)

Dropped `6845a022` — the merge of `feat/staging-access-gate` that was carrying #1166's content.

**Now:**

| | before | after |
|---|---|---|
| diff | 32 files / ~1245 lines | **3 files / 345 lines** |
| `tsc --noEmit` | failed — `isStagingAccessAllowed` absent from `main` | **0 errors** |
| boundary isolation | would fail (non-exempt files) | one boundary file + two tests |

`isStagingAccessAllowed` and `isStaging` both resolve from `main` now, so the import that couldn't compile is fine.

**Verification:** `tests/security/resolveAccess.test.ts` **28/28**, `eslint` clean, `tsc` 0 errors. `programsIdAPI.integration.test.ts` is unrun — no Postgres on this box and it's gated behind `npm run test:integration` — so that one needs CI.

`BLOCKED` is just pending review, not conflicts.

dkaygithub (approved):

PR #1181, opened 2026-07-21
feat(footer): stamp the build with git describe, and the branch when known

jee7s:
Makes the footer answer the question people actually ask of a non-release build: **what is running, relative to the last release?**

Today it renders `v1.0.0 abc1234` on a release and `abc1234` on everything else. A bare SHA can't tell you it's two commits past `v1.0.7`.

## What it stamps

`git describe --tags --long --abbrev=7 --always` → **`v1.0.8-2-g0123456`** — the tag it descends from, the distance from it, and the commit.

- **`--long`** keeps the shape identical for an exact tag (`v1.0.8-0-ge790e17`). The `-0-` is itself worth having on a release: it proves the deployed commit **is** the tag, not a descendant of it.
- **`--always`** degrades to the bare SHA rather than failing a deploy when no tag is reachable.
- **`--abbrev=7`** matches the 7 chars the footer already used, instead of git's variable default.

## Branch prefix, without touching #1167

The label prefixes the branch when known → `rel/1.0 v1.0.8-0-ge790e17`.

This PR **does not introduce `NEXT_PUBLIC_GIT_BRANCH`** — #1167 adds that build arg. The footer reads it defensively, so it starts rendering the branch the moment #1167 lands, with **no change here and no edit to that PR**. Until then the label is the describe string alone.

## The part that's easy to lose

Both deploy workflows now check out with **`fetch-depth: 0`**. That's load-bearing: a shallow clone has no tags, and `git describe` degrades **silently** to a bare SHA rather than failing. Dropping it wouldn't break a build — it would just quietly stop stamping the version. Called out in a comment at both sites for that reason.

## Backward compatibility

`buildLabel` keeps its tag+sha and bare-sha branches instead of requiring describe. Not dead code: an image built before this change, or any build whose checkout was shallow, still renders a sensible label rather than regressing to `dev`. Precedence is describe → tag+sha → sha → `dev`.

## Verification

**11/11** in `BuildInfoFooter.test.tsx` (6 new: describe-wins precedence, branch prefix, describe-without-branch, the pre-describe fallback, describe-without-sha, and branch-with-fallback-version). `eslint` clean, `tsc --noEmit` 0 errors, both workflow YAMLs parse.

## Conflict note

Textually conflicts with **#1167** on `Dockerfile`, `deploy-dev.yml`, `deploy-prod.yml` and `BuildInfoFooter.tsx` — both add build args in the same places. The resolutions are all "keep both": #1167's `NEXT_PUBLIC_GIT_BRANCH` alongside this `NEXT_PUBLIC_GIT_DESCRIBE`. Whichever merges second takes the fixups; no behavioural disagreement between them.

### DIFF
--- a/.github/workflows/deploy-dev.yml
+++ b/.github/workflows/deploy-dev.yml
@@ -68,6 +68,22 @@ jobs:
         uses: actions/checkout@v5
         with:
           ref: ${{ steps.vars.outputs.sha }}
+          # Full history + tags: `git describe` below needs them. The default
+          # shallow clone has neither, and describe degrades SILENTLY to a bare
+          # SHA rather than failing, so a missing fetch-depth shows up only as a
+          # footer that quietly stopped carrying its version.
+          fetch-depth: 0
+
+      - name: Resolve build describe
+        id: describe
+        run: |
+          set -euo pipefail
+          # --long so an exact tag still reports its distance ("-0-g"), keeping
+          # the shape identical for tagged and untagged builds; --always so a
+          # repo with no reachable tag yields the SHA instead of failing the run.
+          DESCRIBE=$(git describe --tags --long --abbrev=7 --always)
+          echo "describe=$DESCRIBE" >> "$GITHUB_OUTPUT"
+          echo "build describe: $DESCRIBE"
 
       - name: Configure AWS credentials (OIDC)
         uses: aws-actions/configure-aws-credentials@v6
@@ -92,6 +108,7 @@ jobs:
           # ECS task defs require (runtime_platform.cpu_architecture = ARM64).
           docker build \
             --build-arg NEXT_PUBLIC_GIT_SHA="${{ steps.vars.outputs.sha }}" \
+            --build-arg NEXT_PUBLIC_GIT_DESCRIBE="${{ steps.describe.outputs.describe }}" \
             -f checkin-app/Dockerfile \
             -t "$IMAGE" .
           docker push "$IMAGE"
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -106,6 +106,9 @@ jobs:
         uses: actions/checkout@v5
         with:
           ref: ${{ github.event.release.tag_name }}
+          # Full history + tags for `git describe` below (see deploy-dev.yml's
+          # note: a shallow clone degrades describe to a bare SHA silently).
+          fetch-depth: 0
 
       - name: Resolve released commit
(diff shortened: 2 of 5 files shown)


dkaygithub (approved):

jee7s:
Merged `main` into this branch and resolved the 3 conflicts.

**Cause:** #1167 landed on `main` and added a `NEXT_PUBLIC_GIT_BRANCH` build arg in the exact same spots this PR adds `NEXT_PUBLIC_GIT_DESCRIBE` — no behavioral disagreement, just two complementary additions touching the same lines.

**Resolution — keep both everywhere:**
- `checkin-app/Dockerfile`: kept both `ARG`/`ENV` blocks (`NEXT_PUBLIC_GIT_DESCRIBE` and `NEXT_PUBLIC_GIT_BRANCH`).
- `.github/workflows/deploy-dev.yml`: kept both `--build-arg` lines; the `fetch-depth: 0` checkout and "Resolve build describe" step had already auto-merged cleanly (no markers).
- `checkin-app/src/components/BuildInfoFooter.tsx`: kept this PR's 4-arg `buildLabel(sha, tag, describe, branch)` and wired it to `NEXT_PUBLIC_GIT_BRANCH`, updating the branch comment since main's Dockerfile now actually sets that var (it's no longer "nothing sets this yet").

One follow-on fix beyond the 3 flagged files: `BuildInfoFooter.test.tsx` auto-merged without a textual conflict, but 3 of main's tests called `buildLabel` with its old 3-arg convention (`sha, tag, branch`), which under the merged 4-arg signature silently read `branch` as `describe`. Updated those 3 calls to pass branch positionally as the 4th arg. All 16 tests now pass (up from the original 11 — main's branch-only test cases carried over too).

**fetch-depth: 0 confirmed intact** on every checkout preceding a `git describe` call — deploy-dev.yml's checkout step and deploy-prod.yml's (both unaffected by the conflict, verified post-merge).

**Verification:**
- `npx tsc --noEmit --pretty false`: 2 errors, both pre-existing `@aws-sdk/client-lambda` baseline (unrelated to this change).
- `npx jest src/components/__tests__/BuildInfoFooter.test.tsx`: 16/16 passed.
- `deploy-dev.yml` and `deploy-prod.yml` both parse as valid YAML.
- No conflict markers remain anywhere in the tree.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01F6TfaRtHA5C76d6A8yU4Nm

PR #1182, opened 2026-07-21
ci(release): a v<major>.<minor>.x release must be cut from rel/<major>.<minor>

jee7s:
Requires a `v<major>.<minor>.x` release to be cut from `rel/<major>.<minor>` — `v1.1.*` only from `rel/1.1`.

## The gap

The version and the branch are **two independent claims** about which release line a build belongs to, and nothing made them agree. `v1.1.0` could be tagged on `rel/1.0`, or on `main`, and ship to prod carrying a version number that lies about its lineage — worst exactly when it matters, during an incident, when the version string is what people reason about.

## The check

Derives the line from the tag and requires the released commit to be **contained in that branch**:

| tag | derived line |
|---|---|
| `v1.1.4` | `rel/1.1` |
| `v1.1.99` | `rel/1.1` |
| `v1.1.0-rc.1` | `rel/1.1` |

`PATCH` is deliberately not part of the branch name — a release *line* is major.minor, and every patch on it ships from the same branch.

**Ancestry, not `release.target_commitish`.** That field can be a branch name *or* a bare SHA, is chosen by whoever cut the release, and is re-taggable. Containment is the property we actually care about, and `--is-ancestor` counts a commit as an ancestor of itself, so a tag on the branch tip passes.

**A missing release line is a hard failure, not a skip** — a typo'd minor or an uncut branch is exactly the mistake being caught. On rejection it prints the branches that *do* contain the commit, so the error says where the release actually came from rather than only that it was wrong.

## Placement

Runs **before** `validate`, so a mis-cut release fails in seconds instead of burning a full CI revalidation in parallel. `deploy` lists **both** needs explicitly rather than inheriting through `validate`, so a later refactor can't drop the gate silently:

```yaml
needs: [release-branch-match, validate]
```

## Testable, and tested

The logic lives in `scripts/release-branch-match.sh` rather than inline YAML so it can actually be run. `scripts/test-release-branch-match.sh` builds a throwaway repo with a bare origin and two release lines, then drives all eight outcomes — **no network, no CI**:

```
1. a v1.1.x tag on rel/1.1 passes
2. the SAME commit tagged v1.0.x is REJECTED — this is the whole point
3. a patch release on the same line passes (PATCH is not part of the branch)
4. a pre-release tag derives the same line
5. an ancestor of the branch tip passes (tag need not be the tip)
6. a commit on no release line at all is rejected
7. a version whose release line was never cut is a hard failure, not a skip
8. a malformed tag is refused before any git work
ALL CHECKS PASSED
```

Case 2 is the one that matters: the identical commit passes as `v1.1.4` and is refused as `v1.0.4`. A release gate whose failing paths have never been executed isn't a gate.

## Overlap with #1147

#1147 adds a `branch-guard` job to the same file, checking containment in **any** `rel/*`. This check is strictly stronger — containment in the *matching* line implies containment in some line — so the two should collapse into one job rather than both running. Whichever merges second should fold them; I'd keep this one and drop `branch-guard`, but that's a call for whoever merges. Textual conflict is limited to the `jobs:` block and the `needs:` lines.

### DIFF
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -67,6 +67,32 @@ env:
   APP_URL: https://ops.innovationtreehouse.org
 
 jobs:
+  # A release must ship from its OWN release line: v1.1.x may only be cut from
+  # rel/1.1. The version and the branch are two independent claims about "which
+  # line is this", and nothing previously made them agree — so v1.1.0 could be
+  # tagged on rel/1.0, or on main, and deploy to prod carrying a version number
+  # that lies about its lineage. That is worst exactly when it matters: during an
+  # incident, when the version string is what people reason about.
+  #
+  # Checks ANCESTRY, not `release.target_commitish`. That field can be a branch
+  # name OR a bare SHA, is chosen by whoever cut the release, and is re-taggable —
+  # containment in the branch is the property we actually care about.
+  #
+  # Runs before `validate` so a mis-cut release fails in seconds with one clear
+  # error, instead of burning a full CI pass in parallel first.
+  release-branch-match:
+    name: Release version matches its rel/ branch
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout released tag
+        uses: actions/checkout@v5
+        with:
+          ref: ${{ github.event.release.tag_name }}
+          fetch-depth: 0   # ancestry check needs real history, not a shallow clone
+
+      - name: Assert the tag's line matches the branch containing it
+        run: ./scripts/release-branch-match.sh "${{ github.event.release.tag_name }}"
+
   # Re-run the FULL CI pipeline (lint, type-check, migration-drift, Jest, the
   # security contract tests, AND the deploy build + boot/smoke) against the exact
   # released tag. This is the H1 fix: a release tag can point at any commit, and
@@ -74,6 +100,7 @@ jobs:
   # the released code green from scratch before a human is even asked to approve.
   validate:
     name: Revalidate CI on the released tag
+    needs: [release-branch-match]   # never burn a CI pass on a mis-cut release
     uses: ./.github/workflows/ci.yml
     with:
       ref: ${{ github.event.release.tag_name }}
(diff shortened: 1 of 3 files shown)


dkaygithub (approved):

jee7s:
## Merge conflict resolved (deploy-prod.yml)

Main picked up #1147's `branch-guard` job in the interim, which conflicted with this PR's `release-branch-match`. Both check the same thing at different precision:

- `branch-guard` (#1147): released commit is contained in **some** `rel/*` branch.
- `release-branch-match` (this PR): released commit is contained in **the specific** `rel/<major>.<minor>` branch matching the tag's own version.

Containment in the matching line implies containment in some `rel/*` line, so `branch-guard` checks nothing `release-branch-match` doesn't already cover — running both was redundant. Resolution: **dropped `branch-guard`, kept `release-branch-match`** as the sole guard, and rewired `needs:`.

Job/needs graph after the merge:

```
release-branch-match   needs: (none)
validate                needs: [release-branch-match]
deploy                  needs: [release-branch-match, validate]
```

`deploy` lists both explicitly (not just `[validate]`) so a future refactor of `validate`'s needs can't silently drop the branch gate.

Test suite (`./scripts/test-release-branch-match.sh`) after the merge:

```
1. a v1.1.x tag on rel/1.1 passes
   ✓ OK: v1.1.4 is contained in rel/1.1
2. the SAME commit tagged v1.0.x is REJECTED — this is the whole point
   ✓ rejected, and reported where the commit actually lives
3. a patch release on the same line passes (PATCH is not part of the branch)
   ✓ v1.1.99 -> rel/1.1
4. a pre-release tag derives the same line
   ✓ v1.1.0-rc.1 -> rel/1.1
5. an ancestor of the branch tip passes (tag need not be the tip)
   ✓ ancestry, not tip-equality
6. a commit on no release line at all is rejected
   ✓ main-only commit refused
7. a version whose release line was never cut is a hard failure, not a skip
   ✓ missing release line refused
8. a malformed tag is refused before any git work
   ✓ v1.1 / 1.1.0 / release-1.1.0 / v1..0 all refused
ALL CHECKS PASSED
```

`python3 -c "import yaml; ..."` confirmed every `needs:` in deploy-prod.yml names a job that exists — no dangling references to the removed `branch-guard` job.

PR #1197, opened 2026-07-22
docs: membership sync design — Google Groups + Slack, consent-first

jee7s:
**Design doc only — no code.** Supersedes #1156 and #1157, both closed in favour of this.

`docs/designs/MEMBERSHIP_SYNC.md` covers **both** sides of the feature: Google Groups (program groups, members list, newsletter) and Slack channels.

## What's carried forward unchanged

#1156's engine was sound and is restated, not redesigned: desired state is a **pure function** recomputed live each run, `SyncState` is an applied-state cache rather than a source of truth, and boundary removals fall out of a plain diff. Population rules (13+, household-lead stacking, leads/core volunteers) and the isolated `ProgramSlackAuth` table are unchanged.

## What changed, and why this is a doc rather than a rebase

The engine is add/remove symmetric and **has no opinion about who did the removing**. That's the gap.

A person who leaves a Google group has made a decision. Today that decision survives only *by accident*: the ledger still reads `applied: true`, so nothing re-adds them — but any event flipping `applied` back to false (a failed op, an admin reset, a row rebuild) would silently override them.

The doc specifies:

- **Every add is audited** — except the newsletter, which would be the highest-volume audit source in the app for no investigative value.
- **A self-removal is detected, audited, and permanently respected.** No in-app clear path: someone who wants back on a list is re-added on the Google side by hand. An accidental unsubscribe is unfixable from inside the app — accepted deliberately as the price of the guarantee.
- **A members-list self-removal raises an admin follow-up.** Program-group and newsletter removals don't.

Detection needs a capability the earlier design **lacked entirely**: the Google client can only `insertMember`/`removeMember`, never *observe*. So `listMembers` is part of the model, not an optimisation.

## Two things I've deliberately left for you rather than decided

**1. Slack symmetry is proposed, not asserted** (§3.4). The same detection applies cleanly and `conversations.members` makes it cheap — but leaving a Slack channel is far more casual than unsubscribing from a mailing list, so a *permanent* opt-out may be too strong there. This extends a decision that was taken about email; I'd rather you confirm it than inherit it silently.

**2. The admin follow-up has nowhere good to live** (§6). The intent was to reuse an existing queue. I checked all three and none fits:

- **`PaymentException`** is closest structurally — but its key is `("kind", "shopifyOrderId")` with no `NULLS NOT DISTINCT`, so a sync row with no order id would insert a **fresh duplicate every nightly run**. The one property that makes the table right for a detector is the one that fails. Its badge would also file the item under **Finance Ops → Payment problems**.
- `TrustedAdultReview` is background-check specific; `OrgMembershipProcess` models the membership lifecycle.

So a new table or a field is unavoidable, and the doc writes up both options with the tradeoff. Worth noting: because §3.2 leaves no way to reverse an opt-out in-app, this flag is the *only* admin-facing trace of a self-removal — which argues for the table if an admin should be able to record what they did about it.

## Next

Review and settle §3.4 and §6, then implementation follows the four-PR sequence in §7 (the `src/security/**` isolation constraint from #1157 still applies to PR 3).

### DIFF
--- a/docs/designs/MEMBERSHIP_SYNC.md
+++ b/docs/designs/MEMBERSHIP_SYNC.md
@@ -1,0 +1,151 @@
+# Membership Sync — Google Groups (email) + Slack channels
+
+**Status:** Design, for review. No code. Supersedes PRs #1156 and #1157, both closed in favour of this doc.
+**Date:** 2026-07-22
+**Purpose:** Keep Google Group membership (program groups, the members list, the newsletter) and Slack channel membership in step with who is actually enrolled and current — while treating a person's own decision to leave a list as a first-class, auditable, permanent choice.
+
+This doc is written so an implementation session can pick it up cold. Read §1 for what is true today, §2 for the model, §3 for the part that is new and most likely to be argued with, then §7 for the PR sequence.
+
+---
+
+## 1. Current state
+
+**Nothing ships today.** There is no Google Groups or Slack integration in `main`. Two PRs carried an earlier version of this feature and are now closed:
+
+- **#1156** — schema (`SyncState`, `ProgramSlackAuth`), Google/Slack clients, desired-state engine. Inert; nothing called it.
+- **#1157** — one line in `src/security/scopeBindings.ts` queueing `ProgramSlackAuth`, split out under the boundary-isolation rule.
+
+Their engine design is carried forward here largely unchanged (§2). What changed is §3 — consent and audit — which is why the doc exists rather than a rebase.
+
+Relevant existing machinery this design reuses rather than reinvents:
+
+- `AuditLog` (`actorId`, `action`, `tableName`, `affectedEntityId`, `oldData`, `newData`) — the app's existing audit trail. Note the `tableName` accuracy rule: a change to a `Person` column is filed under `Person`, not under whatever table the *route* is about (#1175, #1179).
+- `BoardSettings.membersGoogleGroupEmail` / `.newsletterGoogleGroupEmail` — the two org-wide list addresses.
+- `config.googleDirectoryConfigured()` — integration is OFF unless both the service-account key and the admin subject are set. Same null ⇒ off pattern as Resend and the s-read mirror.
+
+## 2. Core model (carried forward from #1156)
+
+### 2.1 Desired state is a pure function
+
+`computeDesiredState(now)` reads live rosters and memberships and emits the complete "who should be in what" tuple set on every run. It is **not** event-driven and keeps no bookkeeping of its own.
+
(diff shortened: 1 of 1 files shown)


thpr:
Should this be an issue not a PR with a doc (which will have a lot of churn and go stale if it stays in the repo?)

thpr:
...or maybe designs go in a separate repo and the doc is distilled to be put into this repo?  I'm finding stale docs are confusing claude and causing it to make bad decisions (it assumes the docs are correct and ours... aren't any more)... so there's a "design area gets everything, once committed the doc is reduced to key decisions and the code is the ground truth" we should consider as an appropriate mode (IMHO)

jee7s:
The staleness concern is well-founded — today's doc-simplification sweep (#1200–#1204) exists because exactly that happened: committed design docs drifted from the tree and the agents reading them inherited the drift (three docs claimed "proposed / no code written" for shipped features; one plan pointed at a directory layout that no longer exists). And your HOUSEHOLD_LEAD_MODEL review on #1201 is now the worked example of the mode you're describing: 368 lines of investigation reduced to 68 lines of decision record, with shipped code as ground truth for everything else.

On the concrete question for THIS PR, the options as I see them:

1. **Keep the PR, adopt "distill on ship" as the convention** — full design doc merges now for line-anchored review (which a big design genuinely benefits from — the #1156 review happened inline against sections), then shrinks to key decisions + constraints when the implementation lands, same as HOUSEHOLD_LEAD_MODEL just did. Cheapest change; the convention is one paragraph in AGENTS.md, and the sweep PRs already enforce the end state retroactively.
2. **Convert to an issue** — no churny doc in the repo at all; review happens in issue comments. Loses line-anchored review and the doc's merge record; agents also can't be pointed at an issue as canonical the way they can a file.
3. **Separate designs repo** — full history without polluting this repo, but adds a second place to look and cross-repo links agents don't follow well.

My lean is (1): it matches what you articulated ("design area gets everything, once committed the doc is reduced to key decisions"), needs no new infrastructure, and the failure mode it leaves (a design doc that ships and never gets distilled) is exactly what the sweep demonstrated is recoverable. But this is a repo-policy call that's Jeff's to make — flagging it to him; happy to convert this PR to an issue, or write the AGENTS.md convention paragraph, whichever way it lands.

thpr:
@dkaygithub You okay with using (1) as the methodology going forward?

dkaygithub:
Yep, design-as-code & using PRs sounds good to me

thpr (approved):

thpr:
and being clear here - approving the PR isn't approving the design... it's approving it to be on disk so I can comment on the design and do PRs against it

Issue #237, opened 2026-06-11
Enhance s-read-function to be able to use production credentials


thpr:
Requires PR#233 to be merged (or this needs to be built on top of that branch). 

The Correct Architecture for Token Refreshing
When Shopify returns an expiring token, your background worker or API server should handle the refresh cycle entirely in-memory and database-side.

Here is the correct workflow for managing a token refresh without hitting your cloud secret manager every time:

[ Application RAM ] 
   └── Has SHOPIFY_CLIENT_SECRET (Loaded once at boot)
   
[ Encrypted DB ] 
   └── Stores the temporary Access Token + Refresh Token per store
   
[ Token Expires ]
   └── App takes CLIENT_SECRET (from RAM) + Refresh Token (from DB)
   └── Hits Shopify token endpoint to get new Access Token
   └── Updates Encrypted DB with new token
The Best Practice Lifecycle
Boot: Your server spins up and reads client_id and client_secret from your cloud vault, saving them into memory.

Execution: Your code detects a store's token has expired (or is about to expire).

Refresh: Your server grabs the client_secret from its own memory, grabs the refresh_token from your database, and makes a POST request to Shopify to exchange them for a new access token.

Storage: The new access token is saved back into your database.

dkaygithub:
A couple of corrections after digging into the current code + infra:

**The `refresh_token` lifecycle above doesn't apply.** We use Shopify's **client-credentials grant**, which returns a short-lived (~24h) Admin access token and **no refresh token**. To renew, you just re-run the `client_id` + `client_secret` exchange — there's no encrypted token DB and no refresh-token rotation to build. The real implementation is simpler than the diagram:

- ❌ encrypted DB storing access + refresh tokens
- ✅ re-exchange `client_id` + `client_secret` for a fresh `shpat_…` when missing/near expiry, cached in-memory for the invocation

**What actually has to change in code.** Today `s-ingest-core/src/config.ts` requires a static `SHOPIFY_ADMIN_TOKEN` and `s-read-function/src/shopify/client.ts` uses it verbatim — there's no exchange path. Add the client-credentials exchange (POST `https://<shop>/admin/oauth/access_token` with `grant_type=client_credentials`). Keep static `SHOPIFY_ADMIN_TOKEN` as a fallback for local dev / legacy custom apps.

[...]

PR #1201, opened 2026-07-22
docs 2/5: tighten app design docs; fold webhook PLAN into its design doc

jee7s:
Part of the five-PR doc sweep (opus team + fable adversarial verify; see PR 1/5 for method). Independent chunk.

**This chunk (−221 net, 1 file deleted):**
- `SHOPIFY_DEV_STORE_WEBHOOK_PLAN.md` **deleted** — ~90% restated the design doc. Its unique content was folded in: the cred-verify `curl`, the Bogus-Gateway acceptance test, the superseded-branch warning (`claude/wonderful-tesla-621492`), and the OIDC region pointer (the last two were initially dropped by the sweep and restored by the adversarial pass).
- `HOUSEHOLD_LEAD_MODEL.md` 368→286 — option (a1) is shipped, so §7's open questions became a decision-status list; §5's rejected alternatives (a2/b/c) kept in full.
- `ZOHO_SIGN_DEV_MOCK.md` — §7 was a wholesale restatement of §2+§4a+§5+§6; deleted, internal refs repointed.
- `data-fetching-hook-migration.md` / `resilient-load-swr.md` — verified complementary (audit+engine-decision vs hook build), NOT duplicates; cross-referenced instead of merged.
- `UNFINISHED.md` untouched: every entry checked against the tree; none verifiably done.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01RQ22akgo6H6vRiJ4iE3Emg

PR #1204, opened 2026-07-22
docs 5/5: s-read/s-replay READMEs — one canonical monitoring story

jee7s:
Part of the five-PR doc sweep (opus team + fable adversarial verify; see PR 1/5 for method). Independent chunk.

**This chunk (−34 net):** the freshness heartbeat-vs-watermark-lag explanation existed in triplicate (README, FUTUREWORK, MONITORING-PRD); it now lives once in MONITORING-PRD §4.1 with pointers. FUTUREWORK 153→121: replay/reset-watermark (shipped in s-replay-function, verified) and token acquisition (shipped, #237) now point at their homes instead of re-describing them; the still-unshipped skip-and-flag DLQ design kept in full. s-replay's idempotency section points at the shared model, keeping only its replay-specific key.

Flagged, untouched: s-read README's superseded Lambda deployment sketch still carries infra facts not confirmed to exist elsewhere (NAT egress, RDS Proxy params, IAM DML-only) — left pending confirmation against infra.

https://claude.ai/code/session_01RQ22akgo6H6vRiJ4iE3Emg

### DIFF
--- a/s-read-function/FUTUREWORK.md
+++ b/s-read-function/FUTUREWORK.md
@@ -88,8 +88,8 @@ set up.
 
 ## 3. Monitoring — alarm wiring
 
-The stale-run reaper (`reapStaleRuns`, called at handler startup) and reserved-concurrency
-= 1 are **done** (README "Operations & monitoring"). What remains is wiring the alarms in
+The stale-run reaper (`reapStaleRuns`, called at handler startup) and the per-store run
+lock are **done** (README "Operations & monitoring"). What remains is wiring the alarms in
 deployment: emit both the freshness heartbeat and watermark lag as CloudWatch metrics,
 alarm on the **heartbeat** (not watermark lag). Full design — including why heartbeat over
 watermark lag — in [MONITORING-PRD.md](MONITORING-PRD.md) §4.1. No auto-reprojection; a dead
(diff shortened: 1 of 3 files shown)


jee7s:
Second pass per the #1201 standard. The flagged "unconfirmed infra facts" in the superseded Lambda sketch got their verdict: checked against `infra/modules/s-read`, they were **contradicted, not unrecorded** — no NAT (public-subnet ENI, and NAT cost is why that design died), explicitly "NO RDS PROXY", no log retention set. Sketch deleted; the one durable invariant (code reads only `process.env`, platform injects secrets) kept with a DEPLOY.md pointer.

Also fixed three live tree-contradictions: "reserved concurrency = 1" is actually a per-store Postgres advisory lock (`withAdvisoryLock`/`ConcurrentRunError`) — reframed in all three places; stale "Lambda"/"ECR" vessel words corrected to the shipped ECS/GHCR shape; the false "any RUNNING row guaranteed dead" corollary dropped (the reaper is threshold-based). `FUTUREWORK` items all stay (live worklist); §4 trimmed to its durable rationale with the §4/§5 numbers preserved because infra terraform cites them. `s-replay`'s `(store_id, refund_gid)` idempotency claim verified against `schema.prisma` and kept.

thpr (approved):

PR #1139, opened 2026-07-19
feat(security): migrate facility GET routes (visits, badges) to handler()

dkaygithub:
Feature half of the wave-1 pairs for `GET /api/facility/visits` + `GET /api/facility/badges` — registry entries in #1137 (**stacked**; retarget to `main` after #1137 merges).

- Same queries, same `take` limits, same `{visits}`/`{badges}` envelopes
- Admission: `withAuth({roles})` → registry `anyRole` (identical gate); nested `Person.email` (pii) + RawBadgeLog internal/personal now stripped by declared policy for any future role
- No row filters existed on these admin logs (take-N recency windows only) — preserved as-is
- `migrated-routes.txt` +2; badges file is now fully migrated (direct-json lint applies and passes); visits PATCH/DELETE stay `withAuth`
- Tests: facility-ops suites + routeAuthDrift 447/447; `check-route-coverage --strict` clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1210, opened 2026-07-23
feat(security): register GET /api/facility/visits (inert)

dkaygithub:
## What

Wave-1 read route, re-cut as a small **registry-only PR to `main`** (one route per PR). Adds a single `defineRoute` entry for `GET /api/facility/visits`. Handler stays on `withAuth` → **inert** until its `handler()` conversion lands separately.

## View

Admin surface (`sysadmin`/`board`), full everyones band. The response is the near-raw `Visit` rows with nested `Person` (email = pii, role flags), so the declared view is exact — not an over-grant. GET is a pure `findMany`; the visit update/delete are on other HTTP methods and out of scope.

## Verified

- `tsc --noEmit`: clean (after regenerating the Prisma client for the current base).
- Mocked security suites (`ctxNeeds`, `authzRegistry`, `scopeBindingsEquivalence`): **12/12**.
- Boundary-isolation: `registry.ts` only. Built in an isolated worktree off `origin/main`.

## Sequencing

Sibling to the other per-route wave-1 registry PRs; as they land, later ones get a trivial keep-both rebase at the shared insertion point. `handler()` conversion is a separate follow-up.


jee7s (commented):

**Family conflict — see the full analysis on #1137.** This PR duplicates one of #1137's ten wave-1 registry entries; `defineRoute` throws on duplicate endpoints, and the merge queue would land both families without a textual conflict, so the failure appears only at the next cold start as an app-wide 500 (module-load throw in registry.ts). One family must be closed before either merges.

**Recommendation: close this PR in favor of #1137** — the #1138–#1142 conversion stack is based on #1137's branch, and this re-split covers only 7 of the 10 entries (the three settings entries #1141 requires were dropped). If the re-split is kept instead, the missing entries must be restored and #1212's broken concat fixed first.

jee7s:
**Linkage:** duplicates the `GET /api/facility/visits` entry in #1137; paired conversion #1139. See family analysis on #1137.

thpr:

jee7s (dismissed):

jee7s (approved):

PR #1209, opened 2026-07-23
feat(security): register GET /api/events (public-only view, inert)

dkaygithub:
## What

First of the wave-1 read routes, re-cut as a **small registry-only PR to `main`** — replacing the 10-routes-in-one boundary batch (#1137) with one route per PR, per the request to keep these small and land them directly on `main`.

Adds a single `defineRoute` entry for `GET /api/events`. The handler stays on `withAuth`, so the entry is **inert** (the stripper isn't wired until the handler's `handler()` conversion lands in its own follow-up PR).

## View

Declared **public-only** (`['public']`) deliberately — the handler selects only public-tier `Event` columns (id/name/startAt/endAt/description), so the declared view is exact, not aspirational. If the select later grows to an internal field (`attendanceConfirmedAt/ById`, `postEventEmailSent`), the strip **fails closed** and forces a boundary + CODEOWNERS review rather than shipping silently under a pre-granted band. This folds in #1137 review findings 4/5 at the point of registration.

## Verified

- `tsc --noEmit`: clean (registry/security).
- Mocked security suites (`ctxNeeds`, `authzRegistry`, `scopeBindingsEquivalence`): **12/12**.
- Boundary-isolation: `registry.ts` only.
- Built in an isolated git worktree off fresh `origin/main`.

## Sequencing

Remaining wave-1 routes (facility, profile-visits, settings, system-status) follow as their own small registry-only PRs to `main`. Each route's `handler()` conversion is a separate follow-up PR once its entry lands.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1212, opened 2026-07-23
feat(security): register GET /api/system-status/errors (inert)

dkaygithub:
## What
Wave-1 read route, small **registry-only PR to `main`** (one route per PR). Adds a single `defineRoute` for `GET /api/system-status/errors`. Handler stays on `withAuth` → **inert** until its `handler()` conversion.

## View
Admin surface (`sysadmin`/`board`). `ErrorLog` rows are all internal-tier and the response is near-raw, so the full everyones band is exact. GET is a pure `findMany`.

## Verified
- `tsc --noEmit`: clean (validated in the 4-entry union pass).
- Mocked suites (`ctxNeeds`, `authzRegistry`, `scopeBindingsEquivalence`): **12/12**.
- Boundary-isolation: `registry.ts` only. Isolated worktree off `origin/main`.


jee7s (commented):

**Blocker, independent of how the family conflict resolves: this PR's diff does not match its title.** It adds **five** `defineRoute` entries — `GET /api/system-status/errors` **twice**, plus copies of #1213's links, #1214's volunteer-designations, and #1215's profile/visits entries (visible concat accident: no blank line between the profile/visits entry's closing `});` and the repeated errors comment block). `defineRoute` throws on duplicate endpoints (core.ts:316-318), so merging this PR **alone** breaks module load of registry.ts and 500s every registry-governed route in prod at the next cold start.

Family context: the #1209–#1215 re-split duplicates #1137, which the #1138–#1142 conversion stack is based on, and the re-split is missing 3 of the 10 wave-1 entries that #1141 needs. Recommendation on #1137 is to keep that family and close this one — but if the re-split is kept instead, this PR must be reduced to exactly one `GET /api/system-status/errors` entry and the three missing settings entries added somewhere before #1141.

jee7s:
**Linkage:** intended entry `GET /api/system-status/errors` pairs with conversion #1142 — but this diff also carries copies of #1213 (links), #1214 (volunteer-designations), and #1215 (profile/visits) plus a duplicated errors entry (boot-breaking; see review above). If the re-split family is kept over #1137, this PR must be cut back to exactly one entry.

thpr:

jee7s:
This one is broken independently of the #1137-vs-per-route-family question, so flagging it separately.

**The diff registers `GET /api/system-status/errors` twice.** Counting the added `endpoint:` lines:

```
2  endpoint: 'GET /api/system-status/errors'
1  endpoint: 'GET /api/system-status/links'
1  endpoint: 'GET /api/settings/membership/volunteer-designations'
1  endpoint: 'GET /api/profile/visits'
```

`defineRoute` throws `Duplicate route registration` at module init (`security/core.ts:316-318`), and `handler.ts` side-effect-imports `./registry` — which is why this fails at load rather than in a single test. The unit job fails 8+ suites on that throw and `Deploy build` dies during `next build` page-data collection. Not environmental and not a stale base.

There is no blank line before the repeated block, so it reads as a concat accident rather than an intentional edit.

**Secondary:** the title says one route, but the diff also carries `system-status/links`, `volunteer-designations` and `profile/visits` — the entries belonging to #1213, #1214 and #1215. If those three stay open alongside this, they duplicate content this PR already adds.

Worth rewriting to just the one endpoint regardless of how the family question lands.

---
_Posted by Claude._

jee7s:
The five failing checks are caused by this PR, not by the branch being behind `main`. It registers the same endpoint twice.

## The defect

`checkin-app/src/security/registry.ts:329` and `:386` both contain `endpoint: 'GET /api/system-status/errors'` — the second block is glued directly onto the previous `});` with no blank line, which reads like a conflict or cherry-pick resolution that kept both sides.

`defineRoute` refuses that (`checkin-app/src/security/core.ts:317` on this branch):

```ts
if (_routes.has(spec.endpoint)) {
    throw new Error(`Duplicate route registration: ${spec.endpoint}`);
}
```

## Why the CI signature looks strange

```
Test Suites: 21 failed, 107 passed, 128 total
Tests:       817 passed, 817 total
```

**Zero test failures.** Nothing asserted wrong — 21 suites died at import, before a single test ran. That is the tell for a module-load throw rather than a logic bug, and it is why the failure list looks unrelated to the diff (`payment-plans-strip`, `shop-org-members-strip`, `emergency-contact-program-scope`, and so on).

It is also not a test-only problem: **`Deploy build (no deploy)` fails too**, with the same error at Docker build time.

## Why this matters beyond CI

`handler.ts` side-imports `./registry` at module load. So a throw in the registry does not take down one endpoint — it takes down **every already-migrated route**, because none of them can import their own handler. This is the one way a registry-first PR is genuinely *not* inert.

## It is not staleness

All eight PRs in this wave sit exactly 71 commits behind `origin/main` — identical drift — and the other seven are green. Drift is held constant across the group, so it is not the variable.

## Scope

The title says one endpoint. The diff registers four: `system-status/errors` (twice), `system-status/links`, `settings/membership/volunteer-designations`, and `profile/visits` — the last three duplicating #1213, #1214 and #1215 wholesale.

## Suggestion: close this one

Everything here exists elsewhere and is already green:

- **#1137** is a strict superset — it carries all four of these endpoints plus six more, with byte-identical `defineRoute` bodies. I compared the endpoint lists directly.
- **#1213 / #1214 / #1215** carry the three non-duplicated entries individually.

Worth knowing for whichever path is taken: all eight PRs in this wave insert at the same anchor (immediately before `// ─── Outbound surfaces ───`), so **every pair conflicts textually** and they must be merged strictly serially with a rebase between each. This PR is what a "keep both sides" resolution of one of those conflicts looks like after it lands — which argues for merging #1137 alone rather than sequencing seven.

---
_Posted by Claude._

jee7s (dismissed):

jee7s:
This branch had been mis-cut from a stacked branch and carried duplicate copies of the routes from #1213 (`GET /api/system-status/links`), #1214 (`GET /api/settings/membership/volunteer-designations`), and #1215 (`GET /api/profile/visits`), plus a doubled `GET /api/system-status/errors` entry. This merge (main → branch) trims it back to the single `GET /api/system-status/errors` entry matching the PR title. The sibling routes stay in their own PRs.

jee7s (dismissed):

jee7s (approved):

PR #1324, opened 2026-07-24
fix(notifications): real program-lead welcome email (#1220)

### LINKED ISSUE #1220: Program Lead welcome email
Program leads get notification on program creation (good)
notification text is really bad (bad)

<title> Treehouse Notification
<body> System Action: PROGRAM_ASSIGNMENT

Should be 

<title> Your Innovation Treehouse Program has been Created!
<body> Congrats! Your program has been approved and initialized by Innovation Treehouse.

You can manage it <a href>here</a>

We recommend adding the first couple of events to the program, and then marking it open for enrollment as soon as you can so our community can start signing up!

Thank you again for your commitment to teaching and fostering community with this upcoming program,
Sincerely: Innovation Treehouse
</body>

dkaygithub:
Closes #1220.

## Problem
Program leads received a notification on program creation, but `sendNotification` had no `case` for `PROGRAM_ASSIGNMENT`, so it fell through to the `default` branch — body `System Action: PROGRAM_ASSIGNMENT`, subject `Treehouse Notification`.

## Fix
- **`email-templates/program-assignment.ts`** (new) — welcome-email template using the shared `baseEmailLayout`, with the copy from the issue and a "manage it here" link.
- **`notifications.ts`** — added the `PROGRAM_ASSIGNMENT` case (real subject + templated HTML). `sendNotification` gains an optional pre-built `html` body path so this case renders a linked template instead of the escaped one-liner; the other cases are untouched.
- **`api/programs/route.ts`** — passes `programId` in the payload so the link targets `/program-ops/programs/<id>` (the lead's manage page). URL base from `config.baseUrl()`, matching `postEventEmails.ts`.
- **`notifications.test.ts`** — asserts the welcome subject, the manage link, program name, and that the body no longer contains "System Action".

## Verification
- `tsc --noEmit`: no new errors in touched files (remaining failures are pre-existing: missing generated Prisma client, `xlsx`, test `any`s).
- Targeted jest: 7/7 pass, including the new case.

### DIFF
--- a/checkin-app/src/lib/notifications.ts
+++ b/checkin-app/src/lib/notifications.ts
@@ -5,6 +5,8 @@ import { formatTime, formatDate } from "./time";
 import { checkinReceiptTemplate } from "./email-templates/checkin";
 import { householdMemberTemplate } from "./email-templates/household";
 import { escapeHtml, type VisitSource } from "./email-templates/base";
+import { programAssignmentTemplate } from "./email-templates/program-assignment";
+import { config } from "./config";
 import { LIVE_PERSON } from "./person/filters";
 import {
     ACTIVE_ORG_MEMBER_PERSON_WHERE,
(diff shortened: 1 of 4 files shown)


checkin-app/src/lib/notifications.ts
```diff
@@ -36,15 +38,25 @@ export async function sendNotification(userId: number, eventType: NotificationEv
...
         switch (eventType) {
             case 'PROGRAM_ENROLLMENT':
                 subject = `Confirmed: Enrollment in ${payload.programName}`;
                 message = `Hi ${user.name}, you have been successfully enrolled in ${payload.programName}.`;
                 break;
+            case 'PROGRAM_ASSIGNMENT':
+                subject = 'Your Innovation Treehouse Program has been Created!';
+                html = programAssignmentTemplate({
+                    programName: String(payload.programName ?? 'your program'),
+                    manageUrl: `${config.baseUrl()}/program-ops/programs/${payload.programId}`,
```
dkaygithub, line 57: Minor robustness: `programName` gets a `?? 'your program'` fallback but `programId` has none. If any future caller fires `PROGRAM_ASSIGNMENT` without `programId`, the manage link renders as `/program-ops/programs/undefined` (a dead link) rather than degrading gracefully. Today the sole caller in `api/programs/route.ts` always passes it, so this is not currently reachable — but consider guarding (e.g. skip the link, or assert the id) so the helper can't silently emit a broken URL.

thpr (approved):

PR #1325, opened 2026-07-24
docs(agents): wire AGENTS.md into every agent + add concise comment policy

dkaygithub:
## Why

Two problems with agent guidance in this repo:

1. **AGENTS.md wasn't actually being read by Claude.** Claude Code auto-loads `CLAUDE.md`, not `AGENTS.md`, and there was no `CLAUDE.md` (or symlink) — so the intended shared orientation was silently ignored by every Claude session.
2. **No commenting guidance existed anywhere**, which showed up as multi-line comments reading like a history of tech debt rather than a short description of what/why a function exists.

## What

- **`CLAUDE.md`** (new) — a thin loader that imports AGENTS.md via `@AGENTS.md`, so every Claude session (including subagents) picks it up. AGENTS.md stays the single source of truth.
- **`AGENTS.md`** — new **Comments** section: a comment describes what/why the code exists *now*; history, bug/PR numbers, and past attempts belong in PR descriptions and commit messages, not source. "Why" comments are fine when the reason is non-obvious and still true.
- **`GEMINI.md`** / **`.jules/sentinel.md`** — pointers back to AGENTS.md as the canonical shared instructions, satisfying AGENTS.md's own cross-consistency rule.

Docs-only; no app/code changes.

### DIFF
--- a/.jules/sentinel.md
+++ b/.jules/sentinel.md
@@ -1,3 +1,9 @@
+# Jules Security Sentinel Log
+
+> **Start with [AGENTS.md](../AGENTS.md)** — the shared, canonical orientation
+> for all agents in this repo. This file is Jules's running log of security
+> learnings, not general instructions; keep it cross-consistent with AGENTS.md.
+
 ## 2024-05-24 - [Timing Attack in Webhook Verification]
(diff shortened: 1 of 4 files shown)


thpr (approved):

thpr:
@dkaygithub You are stuck behind the bug in PR 1339, the merge queue here will hang for an hour again due to logic issues.

Issue #975, opened 2026-07-09
Do we need the legacy product shape (with 2 variants in Shopify?)


thpr:

thpr:
asked because claude flagged that we still have this in our schema so it misleads claude as to how things actually operate now

thpr:
Design doc merged: [`checkin-app/docs/designs/975-LEGACY_VARIANT_CONTRACT.md`](https://github.com/innovationtreehouse/checkin/blob/main/checkin-app/docs/designs/975-LEGACY_VARIANT_CONTRACT.md) (via #1327).

**Answer: no** — both legacy two-variant pairs (`Program.shopify{OrgMember,NonOrgMember}VariantId` and `BoardSettings.shopify{Normal,Volunteer}VariantId`) are superseded and safe to remove. Doc covers the blast radius, the write path that must be closed first, the money-audit hazard, and the expand/contract deploy sequence (Release 0 write-strip → Release 1 code → Release 2 DROP COLUMN). Implementation not yet started.

jee7s:
**Status: READY** (not In progress). Moving to Iteration 2 / **Ready** — #1464 covers only half the work, and the other half is unowned.

The design question is answered (`checkin-app/docs/designs/975-LEGACY_VARIANT_CONTRACT.md`, merged in #1327: **no**, both legacy pairs are dead). #1464 is the code-only Release 1 and is approved and green. **Release 2 has no PR and no issue tracking it**, so the open PR set does not close this issue — which is why this goes to Ready rather than In progress.

**Fix in #1464 before it merges.** Its body says "Residual references … Zero in live code." That is not true — `GET /api/programs/[id]` still selects both legacy columns:

```
checkin-app/src/app/api/programs/[id]/route.ts:80-81
    shopifyOrgMemberVariantId: true,
    shopifyNonOrgMemberVariantId: true,
```

Selected, never read, no consumer. Harmless while the columns exist, but it is the last live read, on the route #1464 itself names among those that would 500 if the `DROP` shipped early. Left in, Release 2 must delete a live read **and** drop the column in one release, which is exactly the drain-window hazard #1464 opens by proving is unavoidable here (`deploy-prod.yml` runs migrations at line 210 and the ECS update at line 283). Two-line fix.

[...]

PR #1327, opened 2026-07-24
docs(design): plan to drop legacy two-variant Shopify shape (#975)

thpr:
## What

Design doc (`checkin-app/docs/designs/LEGACY_VARIANT_CONTRACT.md`) planning removal of the two superseded two-variant Shopify product pairs. **No implementation — design only**, per #975.

The two legacy pairs:

| Pair | Model | Superseded by |
|---|---|---|
| `shopifyOrgMemberVariantId` + `shopifyNonOrgMemberVariantId` | `Program` | `shopifyVariantId` (single pool + minted discount code, #930) |
| `shopifyNormalVariantId` + `shopifyVolunteerVariantId` | `BoardSettings` | `orgMembershipVariantId` + `volunteerDiscountCode` |

## Why

#975: the legacy columns still sit in the schema and mislead readers (human and Claude) about how checkout works today. Both pairs are **write-dead** — nothing creates new legacy rows — and, per board confirmation, **no live prod row** depends on them.

## What a reviewer should know

- The doc maps the full blast radius, including the legacy-only **dead code that collapses** once the columns go: the `purchasedOrgMember` / sibling-inventory-mirror path in `activateEnrollment.ts`, `createShopifyProgramVariants` in `shopify.ts`, the `isLegacy` fork in `sync-shopify`, and the legacy ids in the webhook + reconciler matchers.
- **Deploy sequencing is the one hazard.** Columns cannot drop in the same release as the code that stops selecting them — old pods still `SELECT` them during a rolling deploy. Plan is expand/contract: Release 1 code-only, Release 2 `DROP COLUMN`. Collapses to one PR only under a maintenance-window deploy.
- Open decisions for the follow-up code PR(s): rolling (two PRs) vs window (one PR).


jee7s (commented):

Three CONFIRMED findings and one plausible gap. All are fixable in the doc, but the first invalidates the plan's stated safety premise.

**1. (line ~26) The "no write path creates new legacy rows" premise is false.** `api/programs/[id]/route.ts:197-198` accepts `shopifyOrgMemberVariantId`/`shopifyNonOrgMemberVariantId` from the body and writes them for any sysadmin/board caller — including onto a program that never had them. A board member can create a live legacy-dependent row between the board confirmation and Release 1; Release 1 then removes the checkout ternary/webhook matcher for that program → wrong variant/price at checkout and a paid webhook that fails to activate. The plan needs either a Release-0 step that strips these fields from the PATCH, or a re-verify-at-cutover gate.

**2. (line ~65) Release-1 blast-radius inventory misses three live consumers.** `lib/programCheckout.ts` (`isProgramCheckoutBroken` + `PROGRAM_CHECKOUT_BROKEN_WHERE`, used by `api/nav/todo-counts`, both program-ops pages, and sync-shopify), `lib/lifecycleDrift.ts:126-127`, and `lib/program/capacity.ts:69`. Left as-is, Release 2's DROP COLUMN breaks nav/todo-counts and system-status/lifecycle and aborts the lifecycle-reconcile cron during the drain window — the exact hazard the doc's own two-release section exists to prevent.

**3. (line ~58) Dropping the legacy ids from the match set silently un-audits legacy-era orders.** `runMatchAudit` builds its order universe from `mirror.ordersForVariants(allVariants)`; removing the legacy membership ids and program columns means historical legacy orders stop being audited entirely (not "stay MATCHED"). A legacy order whose claim later breaks can never surface as UNCLAIMED_PAID — reintroducing #1074-class invisible money by construction. The doc should keep legacy ids in the *audit* universe (read-only) even after checkout stops offering them.

**4. (plausible, worth a line in the doc):** the safety argument is DB-only — no Shopify-side step archives/unpublishes the legacy products or voids stale carts, so a stale cart can still complete checkout against a legacy variant after Release 1.

thpr:
Verified all four against the tree — all real. Folded into the doc (commit pushed):

1. **Write-dead premise was wrong.** Confirmed `api/programs/[id]/route.ts:197-198` PATCH writes the program pair for any sysadmin/board caller. Added **Release 0** — strip those two conditional writes first — plus a cutover re-verify SQL gate (`Program` rows with a legacy id + null `shopifyVariantId` must be 0; `BoardSettings` legacy pair must be null) before Release 2.

2. **Missed consumers.** Added `lib/programCheckout.ts` (`isProgramCheckoutBroken` + `PROGRAM_CHECKOUT_BROKEN_WHERE`, consumed by nav/todo-counts, both program-ops pages, sync-shopify), `lib/lifecycleDrift.ts:125-127`, and `lib/program/capacity.ts:69` to the Release-1 blast radius.

3. **Audit hazard (the important one).** Confirmed `matchAudit` builds its whole order universe from `ordersForVariants(allVariants)`, so dropping the legacy ids un-audits legacy-era orders entirely — #1074 invisible-money class, not "stays MATCHED". Doc now keeps the legacy ids in the audit universe read-only via a `LEGACY_AUDIT_VARIANT_IDS` static-const snapshot captured at drop time; only the *live reconcile* set loses them.

4. **DB-only caveat.** Added: archive/unpublish the legacy Shopify products (or void stale carts) as part of Release 1, else a stale cart can complete against a legacy variant post-matcher-removal → silent paid-but-stuck order.

Thanks — #1 invalidated the stated safety premise and #3 would have reintroduced a money bug.

jee7s (approved):

Issue #1224, opened 2026-07-24
Age-based individual agreement per adult child (≥18, child-of-signer)


thpr:
**Backlog item M6** (from the backlog baseline, PR #1192 `docs/backlog/INDEX.md`).

Today the household signs **one household-level agreement**. A child of the signer who is **≥18** (an adult, but NOT a spouse) should have their **own individual agreement**, distinct from the household lead's.

- Origin: ENHANCE · Readiness: NEEDS-DESIGN · Size: M
- Distinct from the spouse case (spouse stays on the household agreement).
- Relates to the SA1 18+ BG trigger (already fires on activation) — this is the *agreement* side of turning 18, not the BG side.

Design needed: who signs, when the obligation opens (18th birthday mid-year vs renewal), and how it renders alongside the household agreement.

thpr:
Design doc merged to `main` (PR #1328):

📄 [`checkin-app/docs/designs/1224_PERSON_AGREEMENT.md`](https://github.com/innovationtreehouse/checkin/blob/main/checkin-app/docs/designs/1224_PERSON_AGREEMENT.md)

Proposes `PERSON_AGREEMENT` — a per-person membership-agreement obligation (mirror of `PERSON_BG`) for an adult child (≥18, non-lead, program-attached now or within the last ~year). No new columns; additive enum migration; unblocks the lead-only signing gate so a non-lead adult signs their own agreement; non-blocking on activation, enforced by hand to start.

One item is intentionally left open in the doc for a follow-up pass: how to flag/explain the two false-positive cases (aged-out overshoot; a non-lead spouse who should be marked a lead). Implementation still pending — leaving this issue open.

thpr:
**Backlog reconciliation (2026-08-01)** — nuance from the merged #1328 design: M6 needs a **family-facing 'why this + what to do' surface** for the two false-positive edges — a graduating senior flagged a cycle late, and spouse-vs-child can't be told apart in data (message: ignore, or mark spouse a lead); copy+placement TBD. Source: INDEX M6.

thpr:
Related: #1462 — the underlying data gap behind this design's "overshoot" caveat (no exit signal for program attachment; withdrawal hard-deletes the row, no class year, no graduation state). Not blocking #1224, which ships accepting the overshoot.

jee7s:
[...]

Issue #289, opened 2026-06-21
Add membership pdf to checkin repo


dkaygithub:
Note: this will be internet-public as stuff exists today

This will be sent to zoho to dynamically generate a signed document

dkaygithub:
this moved to s3 bucket

PR #1328, opened 2026-07-24
docs: design for individual adult-child membership agreement (#1224)

thpr:
## What

Design doc for **backlog item M6** ([#1224](https://github.com/innovationtreehouse/checkin/issues/1224)): an individual membership agreement for an adult child (≥18, non-lead) who can no longer be covered by the household lead's signature.

`docs/designs/PERSON_AGREEMENT.md` — no code, design only.

## Why

Today one household = one agreement, signed by the lead. A minor is covered by that signature; at 18 the child is a legal adult and needs to sign their own. A spouse stays on the household agreement.

## Design shape (agreed with requester)

- **Model:** new `OrgMembershipProcessKind.PERSON_AGREEMENT`, mirror of `PERSON_BG` — `subjectPersonId` set, `orgMembershipId` null, reuses `contractSignedAt`/`zohoEnvelopeId`/`zohoActionId`. **No new columns**; additive enum migration.
- **Population:** ≥18 as-of boundary ∧ not a household lead ∧ program-attached **now or within the last ~year** (proxied via `Program.startAt`/`endAt` — `ProgramVolunteer` has no per-attachment timestamp). Deliberately overshoots aged-out seniors.
- **Triggers:** activation + annual boundary sweep (clone `personBgTriggers.ts`) + **two manual board buttons** ("needs an agreement now" / "needs a BG check now"), off the membership boundary.
- **Signer:** unblock the lead-only signing gate so a non-lead adult signs their **own** agreement; household flow stays lead-gated. Signed `PERSON_AGREEMENT` flips straight to `ACTIVE`, skipping the payment/BG advance.
- **Gating:** none — non-blocking on household activation, enforced by hand to start.

## For reviewers

One open question is parked in the doc: how to flag/explain the two false-positive cases (aged-out overshoot; a non-lead spouse who should be marked a lead). Copy + placement TBD.

Corrected along the way: `#289` (agreement PDF) is **closed** — the feature is live; the `TODO(#289)` comments in `agreementDocument.ts` / `sign/route.ts` are stale.


jee7s (commented):

Four CONFIRMED findings — two spec gaps that break the feature's completion path, two that contradict the doc's own decided semantics.

**1. (line ~75) The attachment-overlap proxy silently drops NULL-dated programs.** `Program.startAt`/`endAt` are both nullable; any overlap where-clause (`startAt lte now AND endAt gte yearAgo`) excludes undated/half-dated programs via NULL three-valued logic. An 18+ non-lead attached only to an ongoing program with no `endAt` never gets a PERSON_AGREEMENT opened — a silent compliance gap. The spec needs an explicit NULL rule (e.g. null endAt = still active; both null = treat as attached).

**2. (line ~106) The signing flow's third leg is missed: `syncContractStatus`.** The return-from-signing sync (lib/membership/external.ts:280) resolves processes via `household.orgMembership.processes`, which can never contain a PERSON_AGREEMENT (orgMembershipId = null). An adult child who signs and returns with `?signed=1` completes only via the inbound Zoho webhook — which the code documents as unreliable against a scale-to-zero instance. The card keeps demanding a signature that was executed; re-clicking mints another ceremony.

**3. (line ~106) A stray open PERSON_AGREEMENT blocks the household contract.** `getOrCreateContractSigningUrl` resolves an open PERSON_AGREEMENT first, unconditionally, and the manual board button has no NOT-isHouseholdLead guard. Board opens one on someone who is (or becomes) a lead → at renewal the resolver never reaches the household INITIAL/RENEWAL path → activation/renewal stalls with no UI explanation. The resolver needs an ordering/exclusion rule.

**4. (line ~149) The migration note contradicts the Population section.** "Reuses PROGRAM_ATTACHED_WHERE from personBgTriggers.ts" — but that helper is attached-to-any-program-EVER (no time bound), while the Population section decided a 12-month overlap proxy. Followed verbatim, "flagged for one cycle" becomes "flagged forever". The note should say the predicate is derived-from-but-narrowed, and spell out the bound.

thpr:
Thanks — all four CONFIRMED findings addressed in `1d947cf3`.

1. **NULL-dated programs dropped by the proxy** — added an explicit NULL-date rule to the Population section. Predicate is now `(endAt IS NULL OR endAt >= yearAgo) AND (startAt IS NULL OR startAt <= now)`: no `endAt` = still active = attached, both NULL = currently attached. Closes the silent gap for ongoing/undated programs.

2. **`syncContractStatus` third leg missed** — the Signer section now requires the return-from-signing sync to resolve the subject's own `PERSON_AGREEMENT` (by `subjectPersonId = userId`) with the same precedence as the signing resolver. Without it, completion hinged on the unreliable webhook and a re-click minted a second ceremony.

3. **Stray open PERSON_AGREEMENT blocks the household contract** — two guards now: (a) the resolver checks `isHouseholdLead` *first* and always takes the household path for a lead, so a PERSON_AGREEMENT can never shadow it; (b) the manual agreement button carries a NOT-`isHouseholdLead` guard (BG button keeps none — leads need BG checks).

4. **Migration note contradicts Population** — reworded: the trigger *derives-and-narrows* the predicate rather than reusing `PROGRAM_ATTACHED_WHERE` verbatim (which is attached-ever, no time bound). Spells out the 12-month window + NULL rule + lead exclusion, so "flagged for one cycle" doesn't become "flagged forever."

The one deliberately-open item (flagging/messaging for the aged-out overshoot and the non-lead-spouse false positive) stays parked under Open Questions for a follow-up pass.

jee7s (approved):

PR #1329, opened 2026-07-24
fix(merge): resolve login identity (email+googleId+emailVerified) as one unit

### LINKED ISSUE #1225: Person/record merge can leave a valid but non-controlled email address
**Backlog item M18** (from the backlog baseline, PR #1192 `docs/backlog/INDEX.md`).

The person/record merge (#1103 LIVE_PERSON) can leave a **valid but non-controlled email** — the address still resolves (delivers mail) but **nobody owns it** after the merge. Contact-integrity risk: the org may believe it can reach someone at an address no member controls.

- Origin: FIX · Readiness: NEEDS-DESIGN · Size: S
- On merge, **invalidate or reassign** the orphaned address.
- Ties the RB10 sole-contact residual ("if a corporate/admin address is the only contact and it's deleted, the org can't reach them").
- Confirmed this is the **person/record** merge (#1103), not household-merge (household-merge does not exist).

thpr:
> [!NOTE]
> **Not #1225.** Independent correctness bug found while working near the merge resolver — unrelated to that issue, just surfaced along the way. Does not close #1225.

## Bug
Participant merge resolved `email` and `googleId` as two independent conflict radios and never touched `emailVerified`. A merge could therefore produce a login identity impossible in normal ops:
- one side's `email` seated on the other side's `googleId`, or
- a swapped-in `email` carrying the keeper's stale `emailVerified` stamp.

Readers that trust `emailVerified` (`lib/impersonation.ts`, `lib/config.ts` staging gate, directory pills) then treat an address nobody proved they control as verified.

## Fix
Resolve `{email, googleId, emailVerified}` as ONE `identity` unit. A record holds only one of each (all `@unique`), so the only coherent outcomes are "keep the keeper's whole identity" or "adopt the merge side's" — never a cross-side split.

- **route.ts** — drop `email`/`googleId` from `CONFLICT_FIELDS`; add one `identity` key. Both sides carry an identity → client must send `fieldChoices.identity` (`keep`|`merge`); on `merge` (or empty-keeper backfill) write `email`+`googleId`+`emailVerified` together. Missing `identity` → 400. Tombstone CAS also nulls `emailVerified` so no tombstone wears a verified stamp for the `merged-*@deleted` sentinel. Login-strand guard kept as defense-in-depth.
- **analyze/route.ts** — select `emailVerified` (one field; explicit-select discipline preserved) so the UI can label the verified side.
- **page.tsx** — collapse the two email/Google radios into one "which login identity survives?" radio showing each side's email, Google account, and verified/unverified marker.

## Scope
Login identity only. `emailSuppressed`/`emailUndeliverableAt` left untouched (owned by a separate branch). `name`/`phone`/`dateOfBirth`/`image` handling unchanged. No schema change, no migration.

## Tests
- `tsc --noEmit`: clean
- Page unit (`page.test.tsx`): 17/17
- Integration (`--runInBand`, throwaway Postgres): full suite **1249/1249** — merge route + audit-log suites green, no collateral breakage
- New identity-unit integration cases: both-Google `merge` (asserts no cross-side split, tombstone stamp nulled), magic-link→Google, one-sided auto-adopt, missing-`identity` → 400

### DIFF
--- a/checkin-app/src/app/__tests__/auditLog.integration.test.ts
+++ b/checkin-app/src/app/__tests__/auditLog.integration.test.ts
@@ -327,10 +327,10 @@ describe('AuditLog Integration Tests', () => {
 
         const req = new Request('http://localhost:4000/api/membership-ops/participants/merge', {
             method: 'POST',
-            // name/email differ between the two fixtures (real conflicts) — resolve
-            // both to "keep" so this audit-log assertion doesn't depend on the
-            // field-picker's validation, which is exercised by its own suite.
-            body: JSON.stringify({ keepId: keep.id, mergeId: merge.id, fieldChoices: { name: 'keep', email: 'keep' } }),
+            // name differs and both carry a login identity (email) — real conflicts.
+            // Resolve name + the identity unit to "keep" so this audit-log assertion
+            // doesn't depend on the field-picker's validation (its own suite covers that).
+            body: JSON.stringify({ keepId: keep.id, mergeId: merge.id, fieldChoices: { name: 'keep', identity: 'keep' } }),
         });
 
         const res = await mergeParticipants(req as never);
--- a/checkin-app/src/app/api/membership-ops/participants/merge/__tests__/route.integration.test.ts
+++ b/checkin-app/src/app/api/membership-ops/participants/merge/__tests__/route.integration.test.ts
@@ -24,11 +24,12 @@ jest.mock("@/lib/shopifyRead/client", () => ({
 import { runMatchAudit } from "@/lib/finance/matchAudit";
 
 // Base fixture participants differ on name+email by design (realistic collision
-// data for the field-picker tests below), so a merge that doesn't care about that
-// conflict needs *some* resolution — default to "keep" for both; tests that
-// specifically exercise fieldChoices validation pass their own map (including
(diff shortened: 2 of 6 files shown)


thpr:
### Follow-up notes (non-blocking)

**1. Wholesale pick can drop the keeper's Google sign-in — by design, made explicit in the UI.** If the keeper has `email`+`googleId` and the admin picks the email-only merge side, the keeper's `googleId` is nulled — correct under "adopt the whole identity," but it is real login loss on one explicit click. The radio shows each side's identity (email + Google line + verified mark), and the group description now spells the consequence out: *"picking a side replaces the kept record's whole login, so choosing an email-only side drops the other's Google sign-in."*

**2. Ordering dependency with the suppression branch.** When the email swaps to the merge side, `emailSuppressed`/`emailUndeliverableAt` stay as the keeper's — now describing the *old* address. That's owned by the separate suppression branch (`claude/fix-merge-email-suppression`) and deliberately out of scope here, but flagging it: **this PR shipping alone leaves that orphaned-flags case live.** Land the suppression branch alongside (or first) so the two stay consistent. Not a defect in this diff — a merge-order thing.

jee7s (approved):

Issue #167, opened 2026-05-27
Need to define enrollment rules for youth


thpr:
For those under 18, their ability to enroll (e.g. http://localhost:4000/programs/1 ) ... perhaps should be limited?  We should discuss the design here, i.e. does this send a message to parents and reserves a slot for N hours or days... or is just impossible or whatever.

thpr:
Design doc merged to `main` (via #1330): [`checkin-app/docs/designs/167-youth-enrollment-rules.md`](https://github.com/innovationtreehouse/checkin/blob/main/checkin-app/docs/designs/167-youth-enrollment-rules.md)

Summary: Option A, fail closed — only a known adult (declared 25+ or 18+ by DOB) may self-enroll; a minor is routed to a household lead. Self-declared adulthood is trusted, with a minor misdeclaring accepted as a documented residual risk. One open question remains flagged in the doc (does the self-gate apply to confirmed admin overrides?).

Design only — implementation is a follow-up, so this issue stays open.

jee7s:
**Status: READY, PENDING #1474.** Moving to Iteration 2 / **In progress** — the design landed, the implementation is open and green, and it needs a reviewer rather than a decision.

The design doc (`checkin-app/docs/designs/167-youth-enrollment-rules.md`, merged in #1330) left one open question: whether a confirmed board/sysadmin `override` should skip the known-adult self-gate. #1474 resolves it as **moot**, and the resolution checks out on `main`: #1463 changed `enforceLimits` to `!override || !isSysAdminOrBoard || isConflicted` (`checkin-app/src/app/api/programs/[id]/participants/route.ts:97`), and `isConflicted` derives from `isSelfEnrollment` (line 73). A self-enroller is always conflicted, so the override exemption is already unreachable on the only path the gate covers. That dependency landed first, so the gate's placement inside `enforceLimits` is safe rather than merely equivalent.

The gap is still open on `main` — no `isKnownAdult` or `ageBand` exists in `checkin-app/src`, so prod (v1.1.1) still admits self-enrollment on `actor === target` alone.

**One consequence worth an explicit yes/no before merge:** a sysadmin or board member with no DOB on file can no longer self-enroll, even with a confirmed override. That follows from the resolution above rather than from the doc's original proposal, and nobody has signed off on it.

**Two out-of-scope items from #1474 that aren't tracked anywhere:**
[...]

PR #1330, opened 2026-07-24
Design: youth enrollment rules (fail-closed self-enroll) — #167

thpr:
Design PR for #167. **Docs only — no code, nothing to ship yet.**

Adds `checkin-app/docs/designs/167-youth-enrollment-rules.md`.

## Decision — Option A, fail closed
Only a **known adult** (declared 25+ or 18+ by DOB) may self-enroll into a program:
- Under 18 by DOB → a household lead must enroll them.
- Age unverifiable (no DOB, not declared 25+ — e.g. a fresh Google sign-up) → set a DOB or confirm 25+ first, then self-enroll.
- Leads / sysadmin / board enrolling someone else are unaffected.

## Why fail closed
- Blocking only *known* minors is a hole — omit your DOB and the gate opens.
- New accounts auto-provision with no DOB, so the block must be **self-serviceable** (the existing intake / "over 25" checkbox), not a lead dead-end.

## Needs sign-off
One open question in the doc: does the known-adult self-gate apply to confirmed **admin overrides**? Proposed — no; it sits under the same override exemption as the other soft limits.

Design only — refs #167. Deliberately no "closes/fixes" keyword: the issue stays open until the implementation lands in a follow-up PR.

### DIFF
--- a/checkin-app/docs/designs/167-youth-enrollment-rules.md
+++ b/checkin-app/docs/designs/167-youth-enrollment-rules.md
@@ -54,9 +54,11 @@ Two facts drove this shape:
    the schema (schema.prisma:85) intends to be *lead*-set, but which a member,
    as their own household lead, can set on themselves. A 15-year-old can hit the
    refusal, tick "over 25", satisfy the "known adult" definition, and self-enroll
-   + pay. Self-entered DOB is no different. **Do not claim this design stops a
-   determined minor; it does not.** What "fail closed" buys is that a member of
-   *unknown* age is refused by default — not that self-declared age is trusted.
+   + pay. Self-entered DOB is no different. **This design does not stop a
+   determined minor, and that is an accepted risk (see below).** What "fail
+   closed" buys is narrow but real: a member of *unknown* age is refused by
+   default. A member who *declares* adulthood — by DOB or the over-25 tick — is
+   taken at their word, by decision.
 
 Rejected alternatives (unchanged from first pass):
 - **Parent-notify + hold-a-slot-for-N-hours + parent-confirms** — a real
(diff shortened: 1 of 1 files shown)


jee7s (commented):

One CONFIRMED finding, but it invalidates the design's central safety claim.

**(line ~50) "a minor who omits DOB still can't get past it" is false — the gate reduces to self-attestation.** The unverifiable case routes to the intake panel whose `primaryOver25` checkbox (FirstTimeIntakePanel.tsx:48) is an unverified self-tick that sets `isDeclaredAdult` — a flag the schema comment (schema.prisma:85) says is meant to be lead-set. A 15-year-old with a Google account hits the refusal, ticks "over 25", becomes a "known adult" per this doc's own definition (declared 25+ OR 18+ by DOB), and self-enrolls + pays via Shopify with zero guardian involvement — exactly the outcome #167 exists to prevent, while the doc asserts to reviewers this cannot happen.

The fail-closed model is only real if the declared-adult leg is verified (lead/board attestation, ID-adjacent signal, or excluding self-set `isDeclaredAdult` from the enrollment gate). The doc should either specify that verification or drop the claim and treat self-attestation as an accepted residual risk explicitly.

(An earlier candidate about the 18+ DOB leg inheriting calculateAge's local-time birthday off-by-one was **refuted** in verification and is not raised.)

thpr:
Confirmed and valid finding — thanks. The "over 25" checkbox self-sets `isDeclaredAdult` (`api/household/member/route.ts:57`, `api/household/route.ts:109`) and a new sign-up is their own household lead, so the gate does reduce to self-attestation. The doc's "a minor can't get past it" claim was wrong.

Resolution (owner decision): we **accept** self-attested adulthood as a documented residual risk rather than harden it. Rationale in the doc — small trust-based org; the gate's job is to stop the casual/accidental case (a minor who simply lands on the enroll page), not to defeat deliberate misrepresentation, which no self-service flow can. Anyone wanting the stronger guarantee uses the lead-enrolls-the-minor path, which already exists.

Fixed in 7be2015a (dropped the false claim) and 4b58fa20 (added the "Accepted residual risk — self-attestation" section). The birthday off-by-one you refuted is not in play.

Still flagged for a separate decision: whether the self-gate applies to confirmed admin overrides.

jee7s (approved):

Issue #1333, opened 2026-07-24
Deploy-free email templates (EmailTemplate + sendTemplated + admin tab)


thpr:
Move email subject/body copy out of code into a board-editable `EmailTemplate` table so content changes need no deploy, and enforce a single templated send path so no flow ships bare inline HTML.

Design: `checkin-app/docs/designs/EMAIL_TEMPLATES.md` (this PR, docs only — no code yet).

Scope: 27 template keys, `sendTemplated` choke-point + grep-to-zero enforcement test, seed-from-current-copy (copy-neutral sweep), new board-gated Settings → Email Templates tab. Folds the two existing deploy-free systems (outreach, scholarship ACK) in.

Open sub-decisions in the doc: blurb-as-token vs 11 keys; link labels in registry vs DB; wrap-all layout.

thpr:
Design doc merged to `main` (via #1334):

[`checkin-app/docs/designs/1333-EMAIL_TEMPLATES.md`](https://github.com/innovationtreehouse/checkin/blob/main/checkin-app/docs/designs/1333-EMAIL_TEMPLATES.md)

Docs only — no code yet. Three sub-decisions still open before build: blurb-as-token vs 11 keys; link labels in registry vs DB; wrap-all layout.

jee7s:
**PROMOTE** → Planning.

It meets the bar on every axis: the user is the board, the problem is that any copy change today needs a deploy, the outcome is 27 keys in an `EmailTemplate` table behind a board-gated Settings tab, and the boundary is explicit — one `sendTemplated` choke-point with a grep-to-zero enforcement test, seeded copy-neutrally from current copy, folding the outreach and scholarship-ACK systems in. The design is not speculative: `checkin-app/docs/designs/1333-EMAIL_TEMPLATES.md` merged to `main` via PR #1334.

Verified on `origin/main`: neither `EmailTemplate` nor `sendTemplated` exists yet, so this is a clean build against a merged design rather than a partially-shipped one.

The three open sub-decisions (blurb-as-token vs 11 keys; link labels in registry vs DB; wrap-all layout) are shape choices inside the merged design and are resolvable in Planning. Worth settling blurb-as-token first, since it determines how many editable fields the board actually sees.

---
_Posted by Claude._

PR #1334, opened 2026-07-24
docs: deploy-free email templates design (#1333)

thpr:
Docs only. Design for #1333.

Moves email subject/body copy out of `.ts` files into a board-editable `EmailTemplate` table so content changes need **no deploy**, and enforces a single `sendTemplated` send path (grep-to-zero test) so no flow ships bare inline HTML. Today only 3 of ~20 flows use the shared branded layout; the rest hand-build `<p>` copy inline.

- **27 template keys**, seeded from current copy → the sweep is copy-neutral.
- Generalizes the two systems that already prove the pattern here (outreach, scholarship ACK) into one; folds them in.
- New board-gated **Settings → Email Templates** tab with live preview + send-to-self.
- Fixes a latent bug found while mapping: `PROGRAM_ASSIGNMENT` currently ships `"System Action: PROGRAM_ASSIGNMENT"` as its body.

Design doc: `checkin-app/docs/designs/1333-EMAIL_TEMPLATES.md`

**3 open sub-decisions** (in the doc) before build: blurb-as-token vs 11 keys; link labels in registry vs DB; wrap-all layout.

No code — design review only.

### DIFF
--- a/checkin-app/docs/designs/1333-EMAIL_TEMPLATES.md
+++ b/checkin-app/docs/designs/1333-EMAIL_TEMPLATES.md
@@ -1,0 +1,274 @@
+# Email Templates — deploy-free email content
+
+**Status:** design, not built.
+**Goal:** change any email's subject/body *without* a code change or deploy, and make it
+structurally impossible to ship an email that bypasses the shared template + branding.
+
+## Problem (as it stands today)
+
+- **The shared layout is used by 3 of ~20 flows.** `baseEmailLayout` wraps only checkin,
+  household, and post-event. Everything else (`membership/*`, `trusted-adult/*`, scholarship,
+  shopify-error, cron warnings, `notifications.ts`) assembles `<p>...</p>` by hand at each
+  call site with no branded shell. A new flow has to *remember* to use the wrapper — nothing
+  enforces it, so drift is the default.
+- **Subjects and body copy are hardcoded in `.ts` files.** Rewording "Welcome to Treehouse"
+  is a code edit, PR, and deploy.
+- **Two deploy-free systems already exist** and prove the pattern works here:
+  - **Outreach** — subject/body in `BoardSettings`, `{{token}}` substitution
+    (`src/lib/outreach/`), edited at `settings/outreach`.
+  - **Scholarship ACK** — subject/body in `BoardSettings`, edited at `settings/email`
+    (`scholarshipAckCopy.ts`, `renderAckBody`).
+
+This design **generalizes those two into one system** that every email goes through.
+
+## Design
+
+### 1. Storage — `EmailTemplate` table
+
+One row per email type, keyed by a stable string `key`:
+
+| column      | notes                                                   |
+|-------------|---------------------------------------------------------|
+| `key`       | unique, e.g. `membership.congrats` (see registry below) |
+| `subject`   | token string                                            |
+| `body`      | **plain text**; blank line = paragraph                  |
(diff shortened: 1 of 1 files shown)


jee7s (commented):

Four CONFIRMED findings: one deploy-safety violation, two dropped behaviors, one stale spec.

**1. (line ~265) Build-order step 5 drops the moved BoardSettings columns in the same release as the code sweep.** No expand/contract sequencing — the invariant #1327/#1337 preserve and DEPLOY_MIGRATION_ORDER_OF_OPERATIONS.md mandates. During the drain window, old pods still select the dropped columns from scholarshipEmails.ts, api/settings/email, the settings page, and outreach render/settings — every scholarship-ack send and outreach send/preview 500s. Split the DROP into a follow-up release.

**2. (line ~107) The one-door pipeline silently loses the opt-out gate and the retry contract.** `sendNotification` checks `notificationSettings.email !== false` before sending and returns a boolean callers branch on (participants/route.ts:148, programs/route.ts:241); `sendTemplated`'s spec (row → substitute → layout → sendEmail) has neither. Post-sweep, opted-out members get emailed and failed sends stop being retried. The spec needs a preference check + defined return semantics.

**3. (line ~79) HTML-escaped tokens in plain-text subjects.** `subj = substitute(row.subject, ctx, …)` with escaped text tokens — subjects aren't HTML, so "Arts & Crafts" arrives as "Arts &amp; Crafts" and "O'Brien" as "O&#39;Brien" in every inbox. This generalizes the existing outreach wart (outreach/render.ts:76) to all ~27 flows. Subjects need raw (unescaped) substitution.

**4. (line ~142) The PROGRAM_ASSIGNMENT "latent bug" claim is stale and the key spec regresses #1220.** PR #1324 (merged 07-25) added a real `case 'PROGRAM_ASSIGNMENT'` using `programAssignmentTemplate` with a `manageUrl` link; the `program.assignment` spec here (tokens: name, programName — no actionLink) would replace the just-shipped lead welcome email with one lacking the manage-program link. Update the registry row to include the link token.

thpr:
All four verified against code and folded into the design (commit 33bce0ec).

1. **Deploy-safety (BoardSettings DROP).** Right. Split into expand/contract: sweep reads from `EmailTemplate` but **leaves the columns populated** (step 5); the DROP + old `settings/email`/`settings/outreach` controls move to a **follow-up release** (new step 8), after the drain window per `DEPLOY_MIGRATION_ORDER_OF_OPERATIONS.md`.
2. **Opt-out gate + retry contract.** Right. Pref checks (`notificationSettings.email`, `emailCheckinReceipts`, etc.) **stay at the call sites** — `sendTemplated` owns only copy→render→send and **returns the `sendEmail` boolean`** so callers keep branching for retry. Enforcement test checks the door, not the opt-out. Spelled out in §4.
3. **Escaped subjects.** Right. Subject now substitutes **raw**; only the body escapes + wraps to `<p>`. Fixes the `outreach/render.ts:76` wart rather than spreading it.
4. **Stale PROGRAM_ASSIGNMENT.** Confirmed — `origin/main` has `programAssignmentTemplate` + `manageUrl` (#1220); my worktree was behind. Dropped the 'latent bug' framing; `program.assignment` now carries the required `{{actionLink}}` token so the sweep can't regress the shipped link.

Still docs-only. The 3 original sub-decisions (blurb, link labels, wrap-all) remain open.

jee7s (approved):

Issue #1336, opened 2026-07-25
PL "paid?" signal — derived from ProgramParticipant.status (no Fee/FeePayment)


thpr:
## Summary

Give a **Program Leader** a per-participant "has this family paid?" indicator so they can remind
parents — **without** revealing full-pay vs scholarship vs payment-plan. Derived from existing
`ProgramParticipant` state; **no new schema**, and specifically **not** built on `Fee`/`FeePayment`
(being killed in #354 / FR7).

## Why derived, not `FeePayment`

A scholarship recipient never generates a payment row, so a payment-record-based "paid?" check would
show scholarship families as **unpaid** → PL nags a scholarship parent. That inverts the privacy
requirement. `ProgramParticipant.status` already collapses all three settled paths (paid Shopify
webhook `activateEnrollment.ts`, scholarship comp, payment-plan approval) into `ACTIVE` **by
construction**, and the sensitive *how* lives at tiers PL never receives.

## Collapsed states (compute server-side per `ProgramParticipant`)

| State | Rule | PL sees |
|---|---|---|
| **covered** | `status == ACTIVE` | paid/covered — no action |
| **in process** | `status == PENDING` AND (`isPaymentPlanRequested` OR `inventoryHeldAt != null`) | "in process" — do NOT nag (scholarship/plan under review) |
| **reminder due** | `status == PENDING` AND not in-process (incl. `paymentPlanDeniedAt != null` — denial means they now owe, may pay normally) | "reminder due" — PL nudges the parent |

All three settled paths land in **covered**; PL cannot distinguish full-pay from scholarship from plan.

## Tier / privacy boundary

- Raw `status` is `@sensitivity:public`, but a *"who owes"* flag is more sensitive than raw enrollment
  status — it must **not** ride the public band. Emit the derived field at the **PL/board-scoped band**
  (`their_program_households:personal`, which PL already gets), so the public envelope never carries it.
- The inputs `isPaymentPlanRequested` (personal) / `inventoryHeldAt` (internal) stay above PL — only the
  **collapsed** field crosses to PL.

## Open questions

1. **Computed-field classification:** the derived field is not a schema column, so the security
   generator won't classify it. Decide how a computed response field gets a band (explicit allowlist
   entry vs. a small typed wrapper). Check how the route stripper handles non-column fields today.
2. **Surface:** existing program roster page vs. the planned **My Programs** staff home
   (attendance-inbox MVP)? Likely a per-participant chip on the roster.
[...]

Issue #354, opened 2026-06-28
Fee / FeePayment appear to be unwritten dead schema


thpr:
## Summary

The `Fee` and `FeePayment` tables (`checkin-app/prisma/schema.prisma:601-637`) have no production write path. Nothing in the app ever creates a `Fee` or a `FeePayment`, so the columns can never hold real data. They should be flagged for deletion rather than carrying coverage.

## Evidence

Grep across `checkin-app/src`, `prisma`, `packages`:

- **`Fee.create`** — only in `src/lib/dev/seed-helpers.ts:282` (dev/test fixture, not a route or service).
- **`FeePayment.create`** — **zero hits anywhere in the repo.** Nothing ever creates a payment row.
- **`FeePayment` references** — only `src/app/api/admin/participants/merge/route.ts`, which `update`s/`delete`s already-existing rows during a participant merge (and nothing creates them in the first place), plus generated security classifications.
- **`Fee` reads** — `src/app/api/programs/[id]/route.ts` includes `fees: true`, but no UI component renders them.

So both tables are write-dead: a route reads `fees`, but no code path ever populates them outside the dev seed helper.

## Why it matters

- `Fee.memberPriceCents` / `nonMemberPriceCents` and `FeePayment.paidOn` / `shopifyLink` / `quickBooksInvoice` are dead columns — schema surface, migrations, and generated security classifications all carry weight for tables that can never be populated.
- The merge route contains logic to reconcile `feePayments` that can never run with real data.

## Recommendation

Either:
1. **Delete** `Fee` + `FeePayment` (and the `fees` include in `programs/[id]/route.ts`, the `feePayments` handling in the merge route, and the seed helper), **or**
2. **Wire up** the route/service intended to write them, if program fees are a planned feature.

## Context

Found while adding program-price round-trip integration tests (`checkin-app/src/app/__tests__/programPriceRoundTrip.integration.test.ts`). Per task instructions, no tests were written for `Fee`/`FeePayment` since nothing writes them — flagging here instead.

dkaygithub:
Yep, this was data schema i built out so that eventually Program Leaders and Board could track who has paid for what (in PL's case I envisioned only their program), I'll defer to you where and how you want to continue that work.

thpr:
[...]

PR #1337, opened 2026-07-25
docs(#354): plan to kill dead Fee/FeePayment schema (FR7)

thpr:
## What

Adds `checkin-app/docs/designs/PROGRAM_PAYMENT_VISIBILITY.md` — a plan to delete the `Fee` / `FeePayment` tables (backlog **FR7**, issue #354). Docs only, no code change.

## Why

`Fee`/`FeePayment` are **dead and redundant**:

- **No writer.** Zero non-test `FeePayment.create`; the only `Fee.create` is a dev seed fixture. `programs/[id]/route.ts` reads `fees: true` but no frontend renders it; the merge route reconciles `feePayments` that can never exist.
- **Redundant.** Program payment already lives, authoritatively, in the Shopify pipeline — `ProgramParticipant.status`+`shopifyOrderId` → `shopify_read` mirror → daily `reconcile.ts` → `PaymentException`. `FeePayment` is a parallel hand-maintained copy of that.

Even the schema's originally-intended consumer ("PL sees who paid") doesn't want `FeePayment` — that signal derives from `ProgramParticipant.status` instead, and is split out to #1336.

## What's in the doc

- §1 Removal rationale (dead + redundant; payment-storage map).
- §2 Two-release deletion plan — Release 1 removes all code refs (schema relations, route include, registry entry, merge-route reconcile, scopeBindings, seed helper, `prisma generate` regen of classifications), Release 2 does `DROP TABLE` after the drain window. Fires `migration-safety` + `safe-refactor-sweep`.
- §3 Backlog wiring (FR7 → kill, P13).

## Reviewer notes

- **Plan only** — no schema or code edited in this PR.
- Tables are empty in prod (no writer ever ran), so the eventual drop is data-safe; the two-release split exists only because old code reads `fees: true` during rolling-deploy drain.
- Follow-up feature: #1336.


jee7s (commented):

One CONFIRMED sequencing flaw plus one CONFIRMED process violation. Both are small doc edits.

**1. (line ~76) Release 1's schema edit collapses the two-release safety this doc mandates.** Removing `model Fee`/`FeePayment` from schema.prisma in Release 1 while deferring DROP TABLE to Release 2 leaves schema-vs-DB drift across the whole inter-release window — the first `prisma migrate dev` anyone runs for unrelated work auto-generates the table drop into Release 1, and old pods running `include: { fees: true }` (programs/[id]/route.ts:77) then 500 on every program-detail request for the drain window. Sibling plan #1327 sequences this correctly: keep the models in schema.prisma through Release 1; remove them together with the drop migration in Release 2.

**2. The Release-1 table bundles security-boundary rows with app code.** Rows #5 (`src/security/registry.ts` — drop 'Fee' from returns) and #8 (`src/security/scopeBindings.ts:92-95` — remove the FeePayment binding) sit alongside app-code rows #4/#6/#9 in one release. AGENTS.md's boundary-isolation rule (CI-enforced by security-boundary-isolation.yml) requires registry/scope-binding changes to ship in their own PR. The plan should split those two rows into a dedicated boundary PR within Release 1.

thpr:
Both findings resolved (commit 4ba580f7), each verified against the code before editing.

**1. Release sequencing.** Confirmed. `model Fee`/`FeePayment` removal moved out of Release 1 — it now lands **with** the `DROP TABLE` migration in Release 2. Release 1 removes code references only; the schema models and tables stay intact through the drain window, so no schema-vs-DB drift and nothing for an unrelated `prisma migrate dev` to fold a drop into. Matches the #1327 code-only → schema+drop pattern (this plan needs no "Release 0 — close the write" since no writer exists).

**2. Boundary isolation.** Confirmed against AGENTS.md:114 / `security-boundary-isolation.yml`. Release 1 is now two PRs — **1a** (app/test code) and **1b** (`registry.ts` + `scopeBindings.ts`, own PR, no app code).

One thing I checked on the split: after 1a removes `fees: true`, `'Fee'` still sits in `registry.returns` until 1b. That's an inert over-declaration — nothing asserts `returns` equals the actual include, and `Fee` isn't an EDGE_MODEL, so `routeAuthDrift` is unaffected. Safe either order.

Also added a cross-ref to #1327 as the sequencing precedent (commit 047e2296).

jee7s (approved):

Issue #1149, opened 2026-07-20
Time display misalignment


jee7s:
When creating a program, the program creation page accepts dates. When that program is viewed in 'Programs' those dates are a day earlier. Is this a UTC/Local time mismatch?

thpr:
1366 solved a tiny piece of the design, there are like 11 findings left to address

jee7s:
**Status: READY** (not In progress). Moving to Iteration 2 / **Ready** — the reported bug shipped, and the one unclaimed work item has no PR.

The program-date off-by-one in the title shipped in #1366. What remains of `checkin-app/docs/designs/1149_DATE_TIME_TZ_DESIGN.md` is **one unclaimed work item**, one approved PR awaiting merge, and one decision owed. Because nothing covers that work item, this goes to Ready rather than In progress.

Status of all twelve findings against `origin/main`:

| | Finding | Status |
|---|---|---|
| F1 | DOB written midnight-UTC vs noon-UTC | **open — unclaimed** |
| F2 | Program `startAt`/`endAt` | shipped (#1366) |
| F3 | `memberSince` displays | shipped (#1424) |
| F4 | DOB via `formatDate` | shipped (#1424) |
| F5 | DOB in the merge diff | shipped (#1424) |
| F6 | `lastBackgroundCheck` | shipped (#1424) |
| F7 | `calculateAge` read local fields | shipped (#1422) |
| F8 | Visit-filter window | #1423 — approved, green, unmerged |
| F9 | Attendance filter label | shipped (#1424) |
| F10 | Facility-trends bucketing | #1423 |
| F11 | `parseImportDob` | **open** — same work item as F1 |
| F12 | `@db.Date` | blocked on the storage decision |

Sequencing: 1 ✅ · 2 ✅ · **3 open** · 4 ✅ · 5 deferred by the doc until a second region deploys · 6 = #1423 · 7 gated on the decision.

**First step — Sequencing step 3, as the doc already specifies it.** Four sites, one convention:
1. Apply `parseDateOnly` inside `normalizeAdultDob` (`checkin-app/src/lib/person/adultDob.ts:28`), which fixes the four routes delegating to it at once.
2. Strip the `+ "T12:00:00Z"` at `checkin-app/src/app/api/household/member/route.ts:59`.
[...]

PR #1338, opened 2026-07-25
docs(design): date/time/timezone architecture & remediation design (#1149)

thpr:
## What

Standalone **design of record** for how `checkin-app` handles dates, times, and timezones: `checkin-app/docs/designs/1149_DATE_TIME_TZ_DESIGN.md`.

Seeded by the #1149 program-date off-by-one, but the scope grew — an app-wide audit found the same defect recurs wherever a **calendar date is stored as a `DateTime` and read through a wall-clock-zoned formatter**. This doc covers the whole class and stands on its own (no dependency on any unpushed branch).

## Contents

- **12 ranked findings** — `file:line`, concrete wrong-output scenario, severity (bug / latent / cosmetic), one-line direction. Plus a **confirmed-correct map** of the helpers/sites already handling tz properly.
- **Architecture root** — the two missing distinctions the findings are symptoms of:
  1. **instant vs calendar-date** classification (the spine),
  2. **one display-timezone source** from `AppSettings` (three zones disagree today),
  3. a **canonical `lib/time.ts` layer** — `formatDateOnly` / `parseDateOnly` / `getUTC*` age / org-tz-sourced instant formatters — so the class can't be reintroduced.
- **Open storage decision (left to the group):** `@db.Date` vs `DateTime`-by-convention (or hybrid), with per-field decision inputs and migration mechanics. **Not pre-decided** — minimize-migration and schema-honesty are both named as legitimate weights.
- **Sequencing** — split into layer/symptom fixes (identical under any storage choice) and the storage step (gated on the decision).

## Scope

**Docs only.** No product code, schema, or migrations changed. The doc is a plan; nothing here applies a fix.

### DIFF
--- a/checkin-app/docs/designs/1149_DATE_TIME_TZ_DESIGN.md
+++ b/checkin-app/docs/designs/1149_DATE_TIME_TZ_DESIGN.md
@@ -58,8 +58,8 @@ with **two incompatible conventions** for the same field (a third,
 
 | Convention | Where | Result |
 |---|---|---|
-| **UTC midnight** `new Date(dob)` | [intake.ts:221](checkin-app/src/lib/membership/intake.ts) (`toDate`, the shared signup/children/parents path), [profile/route.ts:53](checkin-app/src/app/api/profile/route.ts), [household/route.ts:107](checkin-app/src/app/api/household/route.ts) (add member), [membership-ops/participants/route.ts:83,93](checkin-app/src/app/api/membership-ops/participants/route.ts), [importDob.ts:22](checkin-app/src/lib/importDob.ts) | `2026-08-15T00:00:00Z` → displays as **Aug 14** in Chicago |
-| **UTC noon** `new Date(dob+"T12:00:00Z")` | [household/member/route.ts:54](checkin-app/src/app/api/household/member/route.ts) (edit member) | `2026-08-15T12:00:00Z` → displays as **Aug 15** everywhere west of UTC+12 — the only display-safe writer |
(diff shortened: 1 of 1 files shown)


jee7s (commented):

Two CONFIRMED findings — one would corrupt data if executed as written — plus one confirmed inventory gap.

**1. (line ~388) The prescribed Model-B migration is session-timezone-dependent, not "deterministic".** The calendar-date columns are `timestamp(3)` WITHOUT time zone (verified in the coalesced baseline, line 58; no `@db.Timestamptz` anywhere). `USING (col AT TIME ZONE 'UTC')::date` therefore builds a timestamptz and the `::date` cast converts it in the **session** TimeZone — on any connection west of UTC, every UTC-midnight value (2026-08-15 00:00:00 → 2026-08-14) is silently decremented: every DOB, program date, memberSince, year boundary. The correct form for these columns is plain `USING (col::date)`. Given the doc's "self-backfills, deterministic" framing invites running it unverified, this line is the highest-risk sentence in the seven docs under review.

**2. (line ~408) The writer inventory misses the shared parse point, `normalizeAdultDob`.** Four of the listed writers (profile, household, household/member, membership-ops/participants) delegate their DOB parse to `lib/person/adultDob.ts:28` (`new Date(dob)`), which the doc never names — its own docstring mandates every interactive DOB write route through it. Sweeping only the listed route lines leaves UTC-midnight parsing (and the helper's local-time `calculateAge` over-25 strip, absent from step 4's getUTC* list) in place after the remediation is declared done.

**3. (minor) The F1 table announces "three incompatible conventions" but contains two content rows plus a literal `| — | — |` placeholder.** Either the third convention is missing or the count is wrong — as the doc of record for this cleanup, the inventory should be complete.

thpr:
Rebased on `main`; fixes in `651ba13d`.

**1 —** Migration SQL corrected to `USING (col::date)`. These columns are `timestamp(3)` without time zone, so the `AT TIME ZONE 'UTC'` form would cast to date in the session TimeZone and decrement every row west of UTC; added a warning against it.

**2 —** Confirmed. F1 now names `normalizeAdultDob` (adultDob.ts:28) as the shared write choke-point, and F7 flags `adultDob.ts:29` — its `calculateAge` gates a *persisted* DOB strip, so the local-read bug there mutates data, not just display. `intake.ts:221` is noted as still bypassing the guard.

**3 —** Table fixed.

Thanks for the review.

jee7s (approved):

PR #1351, opened 2026-07-25
feat(programs): surface freed-seat notice + Shopify warnings on roster add/remove

### LINKED ISSUE #1345: Program roster: surface Shopify seat notices on add/remove (comp capacity)
**Scope narrowed by design** (`checkin-app/docs/designs/PROGRAM_COMP_CAPACITY_RECONCILIATION.md`). The original framing — "comp-add fires no Shopify call" — is already fixed by #1189 (a confirmed comp fires `adjustProgramInventory(-1)` on capped programs). What remains is D3: the human-in-the-loop pathway is never actually **triggered** at the UI. The absolute Shopify↔roster drift reconciler split out to #1349.

## Basis (ratified decisions)
- **Capacity models the physical room.** A board comp consumes a seat exactly like a sale (`-1`); a full program correctly shows sold-out. (#1189 behavior, ratified.)
- **No enrollment auto-restocks Shopify on removal — by design.** Removing an ACTIVE comp/approved-scholarship seat does NOT fire `+1`. It is deliberate friction: a removal is a customer-service moment where a human decides whether the freed seat goes back on sale, and acts on it by hand in Shopify (mirrors a paid seat, where the `+1` is a manual refund).
- **But that only works if a human is told a decision is pending** — and today they are not.

## The gap (D3)
The program-ops roster UI discards the API response on both add and remove (`src/app/program-ops/programs/[id]/ProgramRosterTab.tsx` → `confirmRemoveParticipant`, `doAddParticipant`). So:
- #1189's comp-add `-1`-failure `warning` never reaches anyone.
- The DELETE route's scholarship-hold-release-failure `warning` never reaches anyone.
- Nothing signals that a removed ACTIVE seat is now freed and needs a manual Shopify restock decision → under the physical-room model that seat stays sold-out in Shopify and wrongly blocks a paying family until someone remembers.

## Fix
1. **Backend** (`src/app/api/programs/[id]/participants/route.ts`, DELETE): return an informational `notice` when a removal freed a seat that was not auto-restocked — `released === false` AND `enrollment.status === 'ACTIVE'` AND `maxParticipants !== null` AND a Shopify variant is wired. Marker-free (fires for paid and comped ACTIVE removals; the staffer judges whether to restock). Distinct from the existing failure `warning`. No schema change, no auto-`+1`.
2. **Frontend** (`ProgramRosterTab.tsx`): surface `notice` (info) and the pre-existing `warning` (warn) via `notifications.show` on both remove and add.

## Tests
- Integration (`programsParticipantsAPI.integration.test.ts`): `notice` on ACTIVE capped+wired removal with no `+1`; none on PENDING; none on uncapped.
- Frontend: a DELETE response with `notice`/`warning` triggers `notifications.show`; a bare `{success:true}` does not.

Origin: ENHANCE · Readiness: READY · Size: S · Related: #1189, #1349

thpr:
## Problem

Program capacity models the physical room. A board comp consumes a Shopify seat (`-1`), and removing an ACTIVE enrollment does **not** auto-restock Shopify — a human decides whether to put the freed seat back on sale, by hand. But the program-ops roster UI **discarded the API response** on both add and remove, so:

- the comp-add `-1`-failure `warning` (already returned by POST) never reached anyone;
- the scholarship-hold-release-failure `warning` (already returned by DELETE) never reached anyone;
- a removed ACTIVE seat gave **no signal at all** that it is now freed and needs a manual Shopify restock decision.

## Change

**Backend** — `participants/route.ts` DELETE: add an advisory `notice` when an ACTIVE removal frees a tracked seat, i.e. all of: no scholarship hold released (`released === false`), `enrollment.status === 'ACTIVE'`, capped program (`maxParticipants !== null`), and a Shopify variant is wired. Purely derived from the row + program — **no auto-`+1`, no schema column, no Shopify call on removal**. The existing `released && !shopifyOk` warning is unchanged; both fields may appear together.

**Frontend** — `ProgramRosterTab.tsx`: surface `notice` (blue) + `warning` (yellow) on remove, and `warning` (yellow) on add, instead of dropping the response. Both are **sticky** (`autoClose: false`) — they prompt a manual Shopify action / flag an out-of-sync state, matching this page's existing load-error notification.

## Tests

- Integration (`programsParticipantsAPI.integration.test.ts`): ACTIVE removal from a capped Shopify-wired program → `notice`, no `+1`; PENDING no-hold removal → no `notice`; ACTIVE removal from an uncapped program → no `notice`, no Shopify call; plus asserting the existing scholarship-withdraw (`released:true`, `+1`) does **not** also get a `notice`.
- Frontend (`page.test.tsx`): a DELETE response with `notice` / with `warning` fires `notifications.show`; a bare `{success:true}` does not.

## Verify

- `npx tsc --noEmit` clean
- frontend `page.test.tsx` 23/23
- integration `programsParticipantsAPI` PASS

Closes #1345

### DIFF
--- a/checkin-app/src/app/__tests__/programsParticipantsAPI.integration.test.ts
+++ b/checkin-app/src/app/__tests__/programsParticipantsAPI.integration.test.ts
@@ -779,7 +779,11 @@ describe('Program Participants API Integration Tests', () => {
                     createParams(shopifyProgram.id) as unknown as never,
                 );
                 expect(res.status).toBe(200);
-                expect((await res.json()).warning).toBeUndefined();
+                const withdrawData = await res.json();
+                expect(withdrawData.warning).toBeUndefined();
+                // Scholarship release path already fires +1 (released:true); it is
+                // NOT the manual-restock notice case.
+                expect(withdrawData.notice).toBeUndefined();
                 expect(logSpy).toHaveBeenCalledWith(expect.stringContaining('Would adjust inventory by 1 for variant: dev-mock-variant-withdraw-partic'));
 
                 const row = await prisma.programParticipant.findUnique({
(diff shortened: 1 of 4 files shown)


jee7s (commented):
## Automated review — multi-agent high pass (comment-only)

9 verified findings (3 confirmed correctness, 2 plausible regressions, 4 cleanups). The first one undercuts the PR's stated purpose (#1345 — surfacing dropped API feedback).

### Correctness — confirmed

**1. `confirmRemoveParticipant` still discards error responses** (ProgramRosterTab.tsx:140). On a non-ok DELETE (403/400/500) the code reads the body but only checks `data.notice`/`data.warning` — `data.error` is never shown. A staffer whose rights were revoked (or any 500) confirms Remove, sees nothing, the roster repaints unchanged, and they retry indefinitely. This is the exact silent-drop this PR exists to fix, surviving on the most important path.

**2. The restock notice fires for enrollments that never consumed a Shopify seat** (participants/route.ts:271). Any ACTIVE row in a capped variant-wired program triggers it — including free-program enrollments (`isFree` short-circuits to ACTIVE, no −1), comps whose −1 failed, and rosters built before the variant was wired. Following the notice for those adds phantom seats: a program with null-price origins gets +1'd above its cap and the next sale oversells it. The notice needs to key off "this enrollment actually decremented Shopify," not "program is variant-wired and row is ACTIVE."

**3. The staff-only notice ships to non-staff callers** (participants/route.ts:272). The DELETE authorizes self-removal and lead-mentor removal; both now receive "add +1 to this program's inventory in Shopify" — actionable-looking instructions to people who can't reach Shopify, and a foot-gun for any future parent-facing surface that reads the same contract. Gate the notice on the caller's role (sysadmin/board/ops).

### Plausible regressions from unconditional `res.json()`

**4.** DELETE path: `await res.json()` now precedes `fetchProgram()` inside the swallow-all try — a non-JSON response (expired-session HTML redirect, proxy error page) throws, the empty catch eats it, and the refetch that used to reveal the true roster state is skipped.
**5.** POST ok-branch: the new parse runs with no catch — an unparsable 200 body skips the form reset + refresh and surfaces only as an unhandled rejection; the staffer resubmits into the 409 already-enrolled path.
One `try { data = await res.json() } catch { /* non-JSON */ }` around each parse (or a shared helper — see 6/8) covers both.

### Cleanup

**6.** `doAddParticipant` parses the same body twice (ok branch + else branch) — hoist a single `const data = await res.json()` above the `if`; two parse sites is the divergence pattern that produced the original dropped-feedback bug.
**7.** `hasShopifyVariant` is a character-for-character duplicate of programs/[id]/route.ts:233 and shadows the canonical variant selection in `adjustProgramInventory` (lib/shopify.ts:608-610) — extract a shared `programHasShopifyVariant()` next to `adjustProgramInventory`; the 975 legacy-variant removal (#1327 plan) will otherwise have three copies to keep in sync.
**8.** The notice/warning→notification mapping is hand-rolled and asymmetric (remove surfaces notice+warning, add surfaces warning only) — a tiny `showApiFeedback(data)` helper makes the two handlers symmetric and future fields land once.
**9.** The 5-line rationale block above the notice logic exceeds the repo comment convention (AGENTS.md: usually one line; rationale lives in the PR body).

(One candidate about `CHECKIN_ENV` restore in the new tests was refuted in verification — no issue there.)

thpr:
Thanks — worked through all nine. Applied the safe ones in the latest push (`4f6d8c86`), and I'm pushing back on #2 with the actual state machine below.

## Applied
- **#1** remove now surfaces `data.error` (red) on a non-ok DELETE — the silent-drop this PR exists to fix, on the most important path.
- **#4 / #5** added a `readJson()` that degrades a non-JSON body (session-expiry HTML, proxy page) to `{}`, so neither handler throws past the refetch / into an unhandled rejection.
- **#6** `doAddParticipant` now parses once, hoisted above the `if`.
- **#9** trimmed the 5-line rationale block to one line (rest is here in the PR).

## #2 — refuted (the free-program case is vacuous)

The claim is that the notice fires for enrollments that never consumed a Shopify seat (free programs, etc.). Tracing the actual inventory paths, the notice's `variant-wired` clause can never be satisfied by a free enrollment, because of a creation-time invariant.

**Inbound — when a Shopify seat is consumed**

| Enroll path | status | `inventoryHeldAt` | our `-1`? | seat actually taken? |
|---|---|---|---|---|
| Free self-enroll | ACTIVE | null | no | no — free ⟹ no variant exists |
| Paid self-enroll (normal) | PENDING→ACTIVE | null | no | yes, Shopify storefront `-1` at checkout |
| Scholarship apply (payment plan) | PENDING | set | **-1** | yes (held) |
| Comp (ext-admin override), capped | ACTIVE | null | **-1** if wired | yes |
| Comp, uncapped | ACTIVE | null | no (route gates `maxParticipants≠null`) | no |

**Outbound — removal.** `withdrawAndReleaseHold` fires `+1` iff `inventoryHeldAt` is set → `released:true`. The notice is then `!released && status==='ACTIVE' && maxParticipants!==null && variant-wired`.

| Removing | `+1`? | notice? | consistent? |
|---|---|---|---|
| Scholarship PENDING hold | yes | no | ✓ hold returned once |
| PENDING no-hold (unpaid) | no | no (not ACTIVE) | ✓ never took a seat |
| **Free ACTIVE** | no | **no** (free ⟹ no variant) | ✓ vacuous |
| Paid ACTIVE, capped+wired | no | yes | ✓ Shopify `-1` did happen |
| Comp ACTIVE, wired, `-1` ok | no | yes | ✓ |
| ACTIVE uncapped | no | no | ✓ no inventory tracked |
| ACTIVE capped, no variant | no | no | ✓ nothing to restock |

**The invariant:** variants are minted only when `basePriceCents > 0` (`programs/route.ts:209`), and the sync-repair route refuses free programs (`sync-shopify/route.ts:35` → `isProgramCheckoutBroken`, which explicitly treats "free" as checkout-OK). So **variant-wired ⟹ priced ⟹ not free** on every normal path. The free-program oversell in #2 can't occur.

Two residual edges, both pre-existing and neither an enrollment-logic bug:

1. **Comp whose `-1` failed at add, later removed** → notice fires, a `+1` would overshoot. But the staffer already saw the add-time `warning` ("capacity may be out of sync"), and the notice is explicitly advisory ("**IF** you want it available… add +1") — the human judges. This is exactly the marker-free, staffer-judges design #1345 chose (no `-1` marker column).
2. **A sysadmin hand-PATCHes `shopifyVariantId` onto a price-null program** via raw API (the field is not in the UI edit form). That produces a free+wired row. It's a misconfiguration the checkout-broken invariant assumes away, not a flow. If we want it enforced, the correct guard is symmetric on the PATCH final state (reject `variant-set AND both-prices-null` in either direction) — a separate hardening PR on that security-sensitive route, out of scope here.

Net: the permutations are self-consistent; keeping the notice condition as-is.

## #3 — fair, but out of D3 scope
The notice does reach self-removal / lead-mentor callers who can't act in Shopify. That's a delivery-audience concern (role-gate the `notice` field), not a state-machine one. Happy to do it as a follow-up; flagging here rather than widening this PR.

## #7 / #8 — deferring
Both are cleanups (extract a shared `programHasShopifyVariant()`; a `showApiFeedback()` helper). Low-value against this diff; the add/remove asymmetry (#8) is intentional (add has no notice concept). Will fold #7 into the #1327 legacy-variant removal, where the third copy actually lands.

jee7s (approved):

PR #1357, opened 2026-07-26
feat(attendance): self-correction of own visits — edit, tombstone delete, board flagging (AT5)

### LINKED ISSUE #1256: Hour correction + manual entry, incl. user self-correction of their own hours
**Backlog item AT5** — from the backlog baseline (`docs/backlog/INDEX.md`, PR #1192).

Hour correction + manual entry, incl. user self-correction of their own hours (confirmed live gap — hours are read-only today; attendance *visits* are self-backfillable but derived/volunteer *hours* are not). Scoped (self / leader / ops), audited.

jee7s:
Fixes #1256 — **AT5: hour correction / self-correction of own hours**, implementing the merged design (`checkin-app/docs/designs/1256_ATTENDANCE_CORRECTION_SURFACE.md` §2, PR #1352). Hours are a derived read over visits, so "correct my hours" = edit the underlying visit; the self-scoped edit path did not exist.

## What ships

**Self-correction API** — `PATCH`/`DELETE /api/attendance/manual/[id]` (sibling of the existing self-insert POST). Trust-first per the design: the edit always applies; ownership (`personId === session user`, 404 otherwise — no existence oracle) and validity (times parse, departure after arrival, ≤ 24 h, never reopen a closed visit) are the only gates. Edited fields become `WEB` — a corrected value is a self-report, whatever captured it before. Closing an open visit routes through `processVisitCheckout` so back-to-back event chunking still happens. Every write takes the same per-person advisory lock as scan/manual/attendance.

**Post-hoc oversight, not a gate** — each write audits (`type: "self_correction"`, `secondaryAffectedEntity` = subject per design §6.6) with a computed **significance** = Σ per-field minutes shifted × trust weight of the OLD value's source (SCANNER 3, roster-mark SYSTEM 2, WEB 1, auto-close SYSTEM 0.25 — the arrival-vs-departure SYSTEM disambiguation the AT3 rename will retire). Threshold 90 weighted-minutes, calibrated so all six worked examples in the design's §2 table come out right; **every delete flags** (the floor). Flagged changes email the board via `emailBoardMembers` (fire-and-forget). Thresholds are v1 constants in `lib/visit/significance.ts` pending the board calibration the design defers (§6.2) — promoting them to BoardSettings belongs with the AT12 screen.

**Visit tombstone** — `deletedAt`/`deletedById` on `Visit` (additive migration, wrapped in BEGIN/COMMIT). The one-open-visit partial unique index is recreated as `WHERE departedAt IS NULL AND deletedAt IS NULL`, so a deleted open visit can't squat the slot and block that member's next check-in. Delete = reversible update, never `visit.delete`.

**`LIVE_VISIT` sweep, CI-enforced** — every visit list/count/aggregate and nested `visits:` pull now filters tombstones (20 files), including three latent nested pulls the scanner surfaced beyond the design's own inventory (`profile` visit history, `postEventEmails` attendee set, `safety/emergency-contacts` on-site query). Enforced from here on by `src/__tests__/liveVisitDriftGuard.test.ts`, adapted from the LIVE_PERSON guard — with the stricter rule that sweeps/reconcilers filter too (a deleted visit holds no resources, unlike a merge tombstone).

**My Visits UI** — new attendance tab listing the member's own recent visits with inline edit and delete; first UI consumer of the previously-dormant `GET /api/profile/visits` (route response shape untouched — #1215/#1140 are in flight on it; only the tombstone filter was added).

## Deliberately out of scope (per the design's own split)
- **Boundary follow-up (required)**: `security/access-resolvers.ts`'s keyholder-scope query needs the same one-line filter, but it's a boundary file under `security-boundary-isolation.yml` — it ships in an isolated PR next, and the drift-guard allowlist entry documents/enforces that handoff.
- Household-lead / program-lead / ops scopes, the `SYSTEM → LEAD_MARKED/AUTO_CLOSE` enum rename, converting the staff routes' hard-deletes (`facility/visits` DELETE, events Absent `deleteMany` — the latter now at least skips tombstones), and the AT12 corrections screen: all AT3/AT12 (§3/§4).

## Verification
- 255/255 suites, 2392 tests, tsc clean. New tests: significance (the design's §2 worked-example table, verbatim), route invariants (ownership/tombstone 404s, validation, WEB re-sourcing, checkout routing, tombstone-not-delete, audit shape, board-flag firing), and the drift guard itself.
- Migration is expand-only (nullable columns + index recreate), rolling-deploy safe: old code ignores the new columns and only new code can create tombstones.

https://claude.ai/code/session_01KBeNiBRHhV7anXNQSWCTMv


thpr:
## Do AT3 + AT5 + AT12 in one shot — implement all of the 1352/1358 spec, not AT5 alone.

The design in `1256_ATTENDANCE_CORRECTION_SURFACE.md` is **one design over one substrate** — a permission matrix over the visit-edit model plus a single audit/oversight contract. AT5 was carved out and shipped by itself, and most of the findings below are direct consequences of that carve, not incidental bugs:

- The **significance flag** is the producer half of **AT12** with no consumer — so it ships as board-email-per-event, recreating the "buried in the log" problem AT12 exists to kill.
- The **tombstone invariant** is only half-applied: this PR makes `Visit` reversible-on-delete, but the **AT3** staff `facility/visits` DELETE still hard-deletes. One writer honors the invariant, another breaks it.
- The **significance code carries a `SYSTEM`→0.25 weight hack** purely because the enum rename lives in **AT3** — and it can't be fixed by weighting at all until AT3 stores the keyholder-close vs cron-close distinction (see #1358).
- **Household-lead** is dropped, which leaves **minors** — the population that can't self-serve, and the reason household-lead is in the spec — uncorrectable by anyone until AT3.

None of these close cleanly in isolation. The recommendation is to **implement the whole spec (AT3 + AT5 + AT12) as one deliverable** rather than re-slice it. The concrete findings below stand regardless of when that happens.

---

### What's right

Design conformance is high and the shape is correct: trust-first (the edit always applies, oversight is post-hoc), tombstone rather than row-removal, no new model, reuse of `parseVisitTime` / `departureAfterArrival` / `withinMaxDuration`, ownership as a 404 with no existence oracle. The `LIVE_VISIT` read sweep is complete for the verbs the guard covers, and the allowlist matches the flagged set exactly.

_(Findings merged from two independent review passes.)_

### Correctness

**1. Auto-close corrections flag the board — the design says they never should.** `src/lib/visit/significance.ts:26,29`. `AUTO_CLOSE_WEIGHT` 0.25 against `FLAG_THRESHOLD` 90 puts the break-even at a 360-minute correction. The nightly sweep stamps `departedAt = now` at cron time (`src/app/api/cron/nightly/route.ts:32`), so the routine "forgot to check out, auto-closed overnight" case corrects by far more than 6 h (arrive 9am, leave 1pm, cron 2am → 780 min × 0.25 = 195 → flagged). Design §2 lists that exact case as "no — expected, encouraged", and it is the single most common reason AT5 exists. The unit test picks a 2 h miss, which sits on the safe side of the threshold and hides the real distribution. A non-zero weight cannot express "never flag" while a threshold exists — suppress the auto-close departure by source (weight 0 / exempt from the sum), not by magnitude. This is entangled with the enum split (see #1358): the code currently can't tell a keyholder building-close from a cron close, so it can't suppress correctly until that distinction is stored.

**2. Facility close overwrites tombstoned open visits.** `src/lib/scan-service.ts:161`. `closeAllOpenVisits` does `updateMany({ where: { departedAt: null } })` with no `deletedAt: null`. The identical case in cron/nightly was fixed with the explicit rationale "closing it would rewrite a record the member chose to erase" — the facility-close sibling (called from `scan-service.ts:135` and `:196`) was missed. Not a read leak, but it corrupts the reversible "what was deleted" record, and silently re-adds a `SYSTEM` departure if the tombstone is later reversed.

**3. The drift guard is mutation-blind — which is why #2 slipped.** `src/__tests__/liveVisitDriftGuard.test.ts:86`. `VISIT_CALL_RE` covers `findMany|findFirst|count|aggregate|groupBy` but not `updateMany|deleteMany`. Re-running with those surfaces four unfiltered sites: `scan-service.ts:162` (the bug above), and `events/[id]/route.ts:143,169` + `membership-ops/participants/merge/route.ts:204` (both correct — FK-nulling before delete / tombstones travel with the person — but recorded nowhere). Adding the two verbs to the regex plus two allowlist entries closes the hole and documents the intent.

**4. Closing an open visit can leave the audit row pointing at a deleted id.** `src/app/api/attendance/manual/[id]/route.ts:105`. `processVisitCheckout`'s chunking path hard-deletes the original row and recreates chunks (`src/lib/attendanceTransitions.ts:126`). The route then writes `affectedEntityId: visitId` and returns `{ visit: updated }`, both referencing a row that no longer exists — on exactly the program-enrolled path. AT12 reviews by audit row, so the provenance link breaks there (the UI self-heals on reload; the audit does not).

**5. Ownership/tombstone check sits outside the lock.** `src/app/api/attendance/manual/[id]/route.ts:92-95`. `loadOwnVisit` reads before the transaction and the update re-asserts nothing, so a DELETE racing a PATCH writes times onto a tombstoned row. Same-user only (low severity) but free to close: `updateMany({ where: { id: visitId, personId: userId, deletedAt: null }, ... })` under the lock. Related wording: `processVisitCheckout` runs *outside* the lock, so the PR body's "every write takes the same per-person advisory lock" isn't quite true (it matches the sibling POST, so it's a doc fix).

### Security

**6. HTML injection into the board email.** `src/app/api/attendance/manual/[id]/route.ts:35`. `flagBoard` interpolates `auth.user.name` raw into the email body, and `name` is self-editable via `PATCH /api/profile` — so the person triggering the flag controls markup in the board's notification (anchor-tag phishing, not script). `emailRecipients.ts`'s header tells callers to build HTML with `escapeHtml`, which exists at `src/lib/email-templates/base.ts:5`. Wrap `actorName`.

### Coherence with AT3 / AT12 (track, don't let slide)

- **The tombstone invariant is half-applied.** This PR makes `Visit` tombstone-aware, but `facility/visits` DELETE (`route.ts:116`) still hard-deletes. Member-deletes are reversible, staff-deletes are permanent. Converting that one call is a one-liner and arguably belongs here, since this PR is what introduced the invariant.
- **The significance flag has no consumer surface yet.** It was designed as notification + AT12 lens; with AT12 not built it ships as board-email-per-event, which is noisy and hard to reconcile at volume. Fine as an explicit interim — but every delete emailing the board indefinitely is the "buried" pattern AT12 exists to replace. When AT12 lands, deletes should fall to the feed.
- **Household-lead is deferred, which leaves minors uncorrectable.** The spec put household-lead in the trust model specifically because minors can't self-serve. As shipped, no one can correct a minor's visit through this surface until AT3. Worth naming as an AT3 dependency, not silent.

### Tests

**7.** The migration's core claim is untested: the point of re-predicating the index is that a tombstoned open visit must not block the member's next check-in, and there's no integration test for it (`attendanceManualConcurrency.integration.test.ts` is the existing harness). The route test mocks `tx.visit.update`, so "tombstone not delete" only proves the call shape. No page test for `my-visits` (`facility-ops/visits/__tests__/page.test.tsx` sets the convention).

### Nits

- Malformed JSON body → generic `catch` → 500; should be 400.
- UI: clearing the departure on a closed visit is silently ignored server-side, yet toasts "Visit updated." Disable the field or explain.
- Deploy note: during the rolling drain, old pods don't filter `deletedAt`, so a fresh tombstone reads as live until they're gone.

### Verdict

Merge-blocking: **#1** (design conformance) and **#6** (email escape). **#2 + #3** are a natural pair — one regex change fixes the bug and prevents the next. **#4, #5** are cheap hardening. The coherence items resolve as AT3/AT12 land, but the staff hard-delete (half-invariant) is worth pulling in now.

thpr (dismissed):
See comment & PR 1358

dkaygithub (approved):

jee7s:
All seven findings plus both nits are in, and I pulled the staff hard-delete in as you suggested. Pushed as `2e7c59e3`.

**#1** — `MACHINE_CLOSE_WEIGHT = 0`; suppressed by source per the merged #1358 design, not a retuned threshold. Kept the SYSTEM-departure-on-non-SYSTEM-arrival inference until the enum split lands, so roster marks still weigh 2. The unit test is now the real cron distribution (arrive 9am, leave 1pm, cron 2am = 780 min → score 0), replacing the 2 h case that sat on the safe side and hid it.

**#6** — `escapeHtml(actorName)` inside `flagBoard`, covering edit and delete in one wrap; test asserts an anchor-tag name comes out escaped. `detail` is built only from `formatDateTime`, so it needed nothing.

**#2 + #3** — `closeAllOpenVisits` filters `deletedAt: null`; `VISIT_CALL_RE` gains `updateMany|deleteMany`. Allowlisted `events/[id]` (FK-null before the event delete) and the merge route (tombstones travel with the person); `scan-service` is not allowlisted and no longer hits. Two corrections to the finding: the merge site is `:251`, and `events/[id]:232`'s `deleteMany` already carried `LIVE_VISIT` — so the genuinely unfiltered pair is `:143`/`:169`, as you had it.

**#4** — `processVisitCheckout` only splits on the enrolled-with-overlapping-events path; all four exits return an array, in time order, with no canonical id. I took the **last** chunk — the segment holding the departure the member just entered, and the row `scan-service` already treats as the result of a checkout. `affectedEntityId`, the flag email and the response body all use it, and `oldData.id` keeps the replaced row's id so the split stays traceable from the audit row. An empty chunk list is now a 404 instead of a 200 reporting a close that never applied.

**#5** — Both verbs re-assert `(id, personId, deletedAt: null)` under the lock. DELETE uses `updateMany` + count as you wrote it. **PATCH uses a lock-scoped `findFirst` instead**: when the edit only closes an open visit the in-lock `data` is `{}` (the departure is routed to `processVisitCheckout`), and I did not want a 404 resting on `updateMany({data:{}})` count semantics I could not verify against a DB here. The lock is what serializes the two, so the read is an equivalent assert. Also corrected the in-code lock claim — `processVisitCheckout` runs outside it, same as the sibling POST.

**#7** — Integration test drives the real POST → DELETE → POST and asserts the tombstone survives, stays open, and does not block the next check-in (the P2002 the re-predicated index exists to prevent). `my-visits` page test added on the `facility-ops/visits` pattern. **Neither has executed locally** — no Postgres in my environment (5433 refused, no docker) — so they are typecheck/lint clean but unproven until CI's integration job runs them. Please weigh them accordingly.

**Nits** — malformed body → 400; Save is disabled when a closed visit's departure is cleared, so the "Visit updated." toast can no longer lie.

**Half-invariant** — `facility/visits` DELETE now tombstones with `deletedAt`/`deletedById`. Response shape unchanged, GET already filtered, and the re-predicated index means tombstoning an open visit still frees the member's next check-in. The `adminVisitsAPI` delete test asserts the tombstone.

AT3/AT12/household-lead deliberately not attempted — that re-scope is yours and Jeff's call, not something to decide inside a review pass.

---

**CI is still red for a reason neither of us can fix inside this PR.** `check-route-coverage --strict` hard-errors `new-route-old-authz` on the two new methods, so the unit job dies before jest. The fix is a registry entry, which AGENTS.md says must land first in its own boundary PR — but that is impossible as written: an entry whose route does not exist yet hard-errors `orphan-registry`, so the boundary PR fails its own CI, and merging it would red-line `main` for everyone until this PR landed. The earlier register-first waves (#1137, #1209-#1215) all registered routes that already existed on `main`, so nobody has hit this before. Genuinely new endpoints have no working sanctioned path today. Worth deciding how you want that resolved — I have a one-line option ready (`orphan-registry` → warning, matching how `stale-legacy-baseline` already treats the same "policy entry with no route" shape) but it changes a CI gate, so it is your call rather than mine.

thpr:
@dkaygithub Look at the CI related comment above - your input on the design here is welcomed given the security structure we have

jee7s:
@thpr @dkaygithub — the CI blocker is now unstuck, as two PRs. Merge order: **#1394 → #1395 → this**.

**#1394** is the decision you were weighing. Rather than downgrading `orphan-registry` wholesale, it splits the rule by which half is missing: **no route file → warn** (the register-first state, inert — nothing serves the endpoint), **file exists but exports no such verb → error** (real policy/code drift, a registered grant whose handler was removed). The first was blocking a workflow AGENTS.md and `security-boundary-isolation.yml` both document as sanctioned; the second is the case worth blocking and stays blocking. A blanket downgrade would have buried it as one warning among 176.

It also lines the rule up with its own sibling: `stale-legacy-baseline` is already only a warning for the strictly more dangerous direction, and says so in its comment. The hard error was the outlier.

Verified against the real tree in both directions, not just unit tests — entries alone fail, entries + the new checker pass as warnings, entries + #1357's actual route file go clean, and renaming `export const DELETE` away trips the new stale error. Details and the full matrix are in #1394.

**#1395** carries the registry entries. One thing worth a look there beyond the entries themselves: `registryAuthz.integration.test.ts` generates cases from the registry and dynamically imports each route module, so it would have failed on the missing module — and it would **not** have self-healed once this PR landed. These are the first non-GET entries in the registry, and its 2xx sanity case drives the endpoint with an empty `Request`, so PATCH would 400 on `req.json()` and DELETE would 404 on the ownership check. I narrowed that assertion to "not 401/403" for write verbs (a 400 or 404 still proves admission) rather than skipping it, since skipping would drop the only admission signal for every future write route. GET is unchanged.

Both are boundary-clean: #1394 touches only `scripts/`, #1395 only `src/security/` + `tests/security/`.

Caveat I want on the record: I could not run `registryAuthz.integration.test.ts` — no Postgres in my environment — so CI's integration job is its first real exercise. The path resolution it depends on I did verify directly.

jee7s:
@thpr — updated to current `main` (`fc2d3f77`), tsc/lint/touched suites green.

**Correction to what I posted here earlier:** I said the CI block needed both #1394 and #1395. That was wrong, and you caught it on #1395. `--strict` is never armed (root `package.json` re-enters `npm run` and swallows the flag), so `orphan-registry` is advisory — but `new-route-old-authz` blocks *even in advisory mode* by design, which is the one actually reding this PR. **#1395 alone unblocks it.** #1394 is orthogonal and I have corrected its description.

**Carrying over your register-without-migrate point, since you flagged it as a joint call with this PR.** Registering these two verbs while they stay on `withAuth` satisfies the ratchet with half the pair: the stripper never runs, so `their_own:personal` in the registry documents a policy that enforces nothing at runtime, and `direct-json` does not fire either. Every other registered-but-unmigrated entry is a pre-existing route being brought in — this is the only brand-new surface, and it carries PII.

My preference is to migrate the two verbs to `handler()` here rather than leave a grant that reads as enforcement and is not. Against that: this PR is already approved and that is real added diff, and the route currently returns `{ visit, flagged }` / `{ success, flagged }` shapes that would need to move behind the envelope. Both are fine with me — your call, and I will do whichever you prefer. If you would rather not grow this PR, the alternative I would want is a follow-up issue filed now rather than the gap inherited silently.

jee7s:
The new integration failure here is not from this branch — **`main` is red**. `programAnnounceRecipients.integration.test.ts` hardcodes its program end dates against an Aug 1 boundary, so the exclusion case inverted when the UTC clock crossed 2026-08-01 overnight (last green main run: 2026-07-31T23:08Z). It fails identically on #1395, which only edits `src/security/registry.ts`.

Fix is up as **#1413** — it derives both dates from `nextBoundary()` so they cannot expire again.

The remaining unit failure here is still the `new-route-old-authz` ratchet, which #1395 clears.

PR #1366, opened 2026-07-26
fix(programs): render program dates as calendar dates (UTC-pinned), per the date/time design

### LINKED ISSUE #1149: Time display misalignment

jee7s:
Program dates shown on the Programs list were one day earlier than the dates
entered on the creation form.

Implements the program-date slice (Finding 2) of the design of record,
`docs/designs/1149_DATE_TIME_TZ_DESIGN.md` (#1338): program `startAt`/`endAt`
are **calendar dates** — stored at UTC midnight, rendered UTC-pinned, never
shifted through a timezone.

> An earlier revision of this PR anchored the dates to the org timezone as
> instants and shipped a backfill migration. That diverged from the design's
> classification and unilaterally answered its open storage-model decision —
> reworked per review. The org-tz-instant approach, `fromDateInput`/
> `toDateInput`, and the migration are gone.

## Root cause

1. `program-ops/new/page.tsx` uses `<input type="date">`, which posts a
   date-only string — `"2026-09-01"`.
2. `POST /api/programs` stores it with `new Date(startAt)` → **midnight UTC**.
3. `programs/page.tsx` renders it through `formatDate`, which formats in
   `America/Chicago`. Midnight UTC is 7 PM the *previous* day there — so
   `2026-09-01` displayed as `8/31/2026`.

The stored value was fine under the calendar-date convention; the read applied
a wall-clock zone to a value that has no zone. Fix the read, not the storage.

## The fix (design Sequencing steps 1–3, program dates only)

**Canonical helpers** in `lib/time.ts`:

- `formatDateOnly(date, opts?)` — UTC-pinned `toLocaleDateString`; the
  calendar-date counterpart to `formatDate` (which stays for instants).
- `parseDateOnly(str)` — date-only input string → UTC-midnight `Date`. Same
  behavior `new Date(str)` already had; its value is naming the convention and
  giving the open storage decision a single write-side seam to change later.

**Reads** (the three display sites Finding 2 names) switch `formatDate` →
`formatDateOnly`: `programs/page.tsx`, `programs/[id]/page.tsx`,
`my-activities/programs/page.tsx`.

**Writes** (all three: `POST /api/programs`, `PATCH /api/programs/[id]`,
`PATCH .../settings`) route through `parseDateOnly` — behavior-preserving,
convention now named.

The edit form's `.split('T')[0]` input read is correct under this model (UTC
slice of a UTC-midnight value) and is untouched.

## No migration

Existing rows already conform to the calendar-date convention (UTC midnight) —
the display fix alone corrects them. The design's open decision (`DateTime`
by convention vs `@db.Date`, §Open decision) stays with the group; nothing
here constrains it.

## Tests

- `time.test.ts` — `parseDateOnly` stores UTC midnight; instant-carrying
  values parse as-is; `formatDateOnly(stored)` renders the picked date where
  `formatDate(stored)` shows the day before (pins the instant/calendar-date
  distinction); round-trips picked → stored → display/input across four dates.
- `programsCreateDate.test.ts` — create route stores UTC midnight; `it.each`
  round-trip asserts `formatDateOnly` and the input slice both return the
  picked calendar date.

## Follow-ups (design findings, not this PR)

The rest of the design's step-2/3 sweep: DOB reads (F4/F5) + writer
convention (F1), `lastBackgroundCheck` (F6), `memberSince` displays (F3),
filter label (F9), age via `getUTC*` (F7), visit windows (F8), trends
bucketing (F10).

Fixes #1149


thpr:
Isn't this part of 1352 an thus part of 1357? Im not sure doing this in pieces makes sense

jee7s:
They're disjoint — checked before opening this separately. #1352/#1357 is the visit-correction surface: **Visit** instants (arrival/departure timestamps) and who may edit them. This PR is a parse bug on **Program** calendar dates: `<input type="date">` values stored via bare `new Date()` land at UTC midnight and render a day early. Different model, different mechanism, and zero shared files with #1357 (33 files vs 8, empty intersection — it doesn't touch `lib/time.ts` either).

The real adjacency runs the other way: sweeping for this bug class found two date-arithmetic bugs in files #1357 already touches (the attendance ±7-day window labels, trends period bucketing). Those are deliberately **not** in this PR (see Follow-ups) so nothing collides. If you'd rather they ride #1357 since it owns those files, that works — say the word.

If sequencing is the concern, holding this until #1357 lands costs nothing — no rebase either way.

thpr:
Sorry crossing wires on what design docs exist, you're right

thpr (changes requested):
Really solid fix in isolation — root cause is dead-on, the `fromDateInput`/`toDateInput` pairing mirrors the existing `fromDatetimeLocal` helpers nicely, and the DST/Tokyo test coverage is better than most PRs in this area. The follow-ups you flagged (`broken.tsx:89`, the visit-window desync, `facility/trends`) are all real.

One heads-up that I think warrants a pause before merge — no fault here, it's easy to miss: this exact bug is written up as **one slice of an app-wide date/time/timezone design**, `checkin-app/docs/designs/1149_DATE_TIME_TZ_DESIGN.md` (in review as #1338). It catalogs the same defect across DOB, `memberSince`, `lastBackgroundCheck`, program dates, and the visit windows/trends you noted, and settles on a single model so all of them stay consistent.

The concern is a **model divergence**, not a bug:

- The design classifies program dates as **calendar dates** → fix is a UTC-pinned read helper (`formatDateOnly`), **no migration**. This PR instead treats them as **org-timezone-anchored instants** (like events) and migrates existing rows. Both correct the display, but they're different data models.
- After this change a program `2026-09-01` is stored as `2026-09-01T05:00:00Z` — **timezone-relative**. For a genuine calendar date (a program start isn't a moment in time) that's the anchoring the design deliberately avoids, and it makes the value depend on the org tz at write time.
- The design intentionally left the **storage model an open group decision** (UTC-convention vs `@db.Date` vs migrate). Shipping a backfill migration + org-tz anchoring for one field quietly answers that decision solo, and DOB/`memberSince`/BG are the same class — they'll look to this as precedent.

Ask: let's reconcile the model before this lands, so the whole surface stays consistent. Two clean paths —
1. Adopt the design's calendar-date approach for program dates (UTC-pinned read, drop the migration), or
2. If the org-tz-instant model is genuinely preferred here, let's fold that decision into the design doc so it's the chosen direction for the whole class, with the group's sign-off on the migration.

Happy to pair on either — most of the work here transfers directly. Thanks for digging into this one.

dkaygithub (approved):

jee7s:
Reworked to the design's model (path 1) — thanks for the catch, and apologies for the miss: `1149_DATE_TIME_TZ_DESIGN.md` was on main when I branched and I failed to check `docs/designs` for the issue number before implementing.

As of 77cc47c4:
- **Calendar-date model adopted.** Storage stays UTC midnight; the fix moved to the read side — `formatDateOnly` (UTC-pinned) at the three display sites Finding 2 names, `parseDateOnly` naming the write convention at the three writers (behavior-preserving; it's the single seam if the storage decision later picks `@db.Date`).
- **Migration dropped**, org-tz anchoring and `fromDateInput`/`toDateInput` removed, `date-fns-tz` no longer touched, edit-form `.split('T')[0]` read restored (correct under this model, as the design notes).
- Storage-model open decision is untouched and stays with the group.
- Tests rewritten to pin the convention: UTC-midnight storage, `formatDateOnly` vs `formatDate` contrast, picked→stored→display/input round-trips.

PR body rewritten to match. The DOB/memberSince/BG/window follow-ups from my sweep line up with the design's F1/F3–F10 — left to the design's sequencing.

thpr:
The rework looks great — `parseDateOnly` + `formatDateOnly`, three display sites swapped, migration dropped. Matches the design's model exactly.

One reader still slips through, same class as the rest: `src/app/program-ops/programs/page.tsx:39-40` has its own local `fmtDate`:

```
const fmtDate = (d) =>
  d ? new Date(d).toLocaleDateString(undefined, { month: "short", day: "numeric", year: "numeric" }) : null;
```

No `timeZone`, so it renders the UTC-midnight program dates a day early on the ops programs list (lines 43–44). Swapping it to `formatDateOnly` (the helper this PR already adds) closes it — one file.

With that in, I'll clear the change-request and approve. Nice work on the DST/Tokyo coverage.

jee7s:
Fixed in `ab25025e` — `fmtDate` now delegates to `formatDateOnly` (keeping the short-month options), so the ops list reads UTC-pinned like the other three sites. One file, as you said.

I re-swept every file that touches `startAt`/`endAt` and formats a date, to confirm nothing else of this class is left. It isn't — the remaining hits are all **Event** instants (`ProgramEventsTab`, `sessions/[id]`, `my-activities/events`, `program-ops/events`) or `pendingSince`/`attendanceConfirmedAt`, which are true instants and correctly use `formatDateTime`. Your find was the last calendar-date reader.

Two bare-`toLocaleDateString`-on-an-instant spots turned up in that sweep — `sessions/[id]/page.tsx:442` (`attendanceConfirmedAt`) and `program-ops/events/page.tsx:15` — but those are the *instant* zone-source issue (design Axis 2 / step 5, wiring org tz to the client formatters), not this class, so I left them out rather than widen the PR. Happy to file them if you want them tracked.

All checks green on the previous push; re-running now.

thpr (approved):
`fmtDate` now routes through `formatDateOnly` — that was the last gap. All four program-date display sites are UTC-pinned, writes go through `parseDateOnly`, migration's gone, and the DST/Tokyo coverage is solid. Aligned with the date/time design. Thanks for the reworks — LGTM.

thpr:
For future reference: PR text is wrong, doesn't fix 1149, it addresses one finding of the design

PR #1370, opened 2026-07-28
docs/designs: unified signup — one link, one form, one checkout (PR 0: design for peer review)

dkaygithub:
## What

PR 0 of the unified-signup stack: the design doc only, for peer review before any code lands. Adds `checkin-app/docs/designs/UNIFIED_SIGNUP.md`.

## Why

Today a new family that wants "join a program + become members" needs four disconnected steps and 2+ Shopify checkouts. The design gives the org ONE shareable link (`/join?program=N`) ending in ONE combined checkout, with **no new form to maintain** — the existing `/membership` intake gains an optional per-kid program question, and `/join` is a redirect-only entry point.

## Key design points (details + ground-truth citations in the doc)

- **Early payment as a parallel-track flag**: `paidAt` becomes stampable at `PENDING_EXTERNAL_ACTION`/`PENDING_BG_REVIEW`; a paid row never visits `PENDING_PAYMENT`; ACTIVE still gated on BG clearance. Two new TRANSITIONS edges; `intake-is-unpaid` and `active-is-bg-cleared` invariants unchanged.
- **Combined multi-program cart**: new `Program_Enrollments=<progId>:<personId>,…;<progId>:…` attribute beside `Membership_Process_ID`; webhook loses its membership early-return and runs both passes per order; reconciler + dev mock updated in lockstep; legacy attributes parsed forever. Variant-id line-item checks remain the fail-closed gate on both passes.
- **Member pricing** via one minted single-use amount-off-order code (`UNI…`); volunteer dues discount does not stack (one `discount=` slot per permalink).
- **Orchestrator** `POST /api/join/checkout` (withAuth + in-handler checks; registry migration deferred to an isolated optional PR per the boundary-isolation rule).
- **No schema changes** anywhere in the stack.

## Decisions already locked with the owner

Auth-first (no anonymous writes) · pay at end of form · multi-program in one cart · member pricing via minted code · one form only.

## Stack

0 (this) → 1 lifecycle → 2 attribute scheme (webhook+reconciler lockstep) → 3 discount mint → 4 profile + enroll-service extraction → 5 orchestrator → 6 UI (+7 optional registry migration). Each lands on main independently green.

### DIFF
--- a/checkin-app/docs/designs/UNIFIED_SIGNUP.md
+++ b/checkin-app/docs/designs/UNIFIED_SIGNUP.md
@@ -1,0 +1,243 @@
+# Unified Signup: one link → membership + multi-program enrollment + one checkout
+
+**Status: PROPOSAL — approved by owner, up for peer review. No code built.**
+Ground truth cited below (lifecycle, webhook, reconciler, intake profiles) was
+verified in-tree as of this doc's PR. Implementation is a 6-PR stack (§8);
+feedback on this doc can still reshape any of it.
+
+**Addresses:** the fragmented new-family funnel (sign in → membership at
+`/membership` → per-program enrollment at `/programs/[id]`, each priced program
+a **separate** Shopify checkout). **Related:** `LIFECYCLE.md` (membership state
+machine this extends), `SHOPIFY_MEMBER_SEGMENT_PRICING.md` (the anonymous-cart
+constraint all of this lives under), #278/#270 (honor-system cart attributes),
+`docs/PROGRAM_CAPACITY_AND_SCHOLARSHIPS.md` (single-pool capacity + minted
+discount codes), `src/lib/intake/README.md` (shared intake profiles).
+
+## 1. Problem
+
+A new family that wants "join a program, become members while we're at it"
+today needs four disconnected steps and two-or-more checkouts. There is no
+single link the org can hand out that collects household info once, applies
+for membership, enrolls kid A in program A and kid B in program B, and ends in
+**one** Shopify payment.
+
+## 2. Product decisions (locked with the owner)
+
+1. **Auth-first.** The link requires Google sign-in up front (existing
+   auto-household-on-first-sign-in, `auth-options.ts`). No anonymous writes —
+   consistent with the existing auth-first registration posture.
+2. **Pay at the end of the form.** The combined checkout is minted immediately
+   after intake submit (process at `PENDING_EXTERNAL_ACTION`) — *before* the
+   Zoho contract and background check complete. Money can arrive early;
+   **activation still waits on BG clearance** (unchanged two-track convergence).
+3. **Multi-program in one cart.** Different kids into different programs in a
+   single checkout.
+4. **Member program pricing in the combined cart**, via one server-minted
+   single-use amount-off-order discount code. The volunteer dues discount does
+   **not** stack (a cart permalink carries exactly one `discount=` code) —
+   volunteer households pay normal dues through this flow; folding the
+   volunteer delta into the minted amount is a possible later enhancement.
+5. **One form.** No parallel wizard. The existing `/membership` intake form
+   gains an optional per-kid program question; `/join` is a redirect-only
+   entry link, not a page with a form.
+
+## 3. Design at a glance
+
+```
+/join[?program=N]
+  └─ signed out → /signin?callbackUrl=…
+  └─ ACTIVE member → /programs/N (existing enroll)   [no program param → /membership]
+  └─ everyone else → /membership?program=N
+/membership intake (existing, resumable)
+  + optional "any of these programs?" per-kid picker
+  submit → POST /api/join/checkout
+             ├─ startIntake/submitIntake (context "unified-signup")
+             ├─ createPendingEnrollment per (program, kid)   [PENDING; free → ACTIVE]
+             ├─ mint UNI… amount-off-order code = Σ member discounts
+             └─ ONE cart permalink:
+                /cart/{memVariant}:1,{progA}:{nA},{progB}:{nB}
+                  ?discount=UNI…
+                  &attributes[Membership_Process_ID]={pid}
+                  &attributes[Program_Enrollments]={progA}:{kid1},{kid2};{progB}:{kid3}
+orders/paid webhook (one order, BOTH passes)
+  ├─ membership: variant-checked → paidAt stamped EARLY, status held
+  └─ programs:  per-program variant-checked → PENDING→ACTIVE
+contract + BG consent later → advanceExternalIfComplete
+  paid row skips PENDING_PAYMENT → PENDING_BG_CLEARANCE (or ACTIVE if cleared)
+```
+
+## 4. Lifecycle extension: early payment
+
+Today `activate()` (`src/lib/membership/payment.ts`) ignores payment for any
+status except `PENDING_PAYMENT` (and `BLOCKED` → record + refund alert). The
+"paid" notion already propagates elsewhere: `clearBackgroundCheck` flips a paid
+`PENDING_BG_REVIEW` row straight to `ACTIVE`. This design completes that:
+`paidAt` becomes a parallel-track **flag stamp** (same convention as
+`contractSignedAt`/`bgConsentAt`), legal from `PENDING_EXTERNAL_ACTION` onward.
+
+- **`activate()`** new branch for `PENDING_EXTERNAL_ACTION` / `PENDING_BG_REVIEW`:
+  H2 variant check first (`hasMembershipItem === false` → `underpaid` audit, **no**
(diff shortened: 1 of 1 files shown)


thpr:
## Partial review — design doc only

**Scope note:** this is a *partial* review. I read `UNIFIED_SIGNUP.md` end to end and verified its ground-truth citations against `main` (lifecycle, payment/external/review, webhook, reconciler, intake profiles, discount mint, security tiers). I did **not** review: the Shopify cart-permalink behaviour itself (§9's pre-PR-5 spike is still the only answer there), attribute-length limits on `note_attributes`, capacity/hold-ledger interaction under a multi-program cart, or anything about the UI in PR 6. Findings below are gaps in the plan, not corrections — the citations that are there hold up.

### Verified correct (no need to re-litigate)

`activate()` status handling · `clearBackgroundCheck` paid→ACTIVE · both INVARIANTS unchanged · `classify()` already tolerates a paid `PENDING_EXTERNAL_ACTION` · the webhook's membership early-return exists as described · a membership claim does suppress the program pass today · `resolveMembershipProcess` is `PENDING_PAYMENT`-only · `BoardSettings.orgMembershipVariantId` is `@sensitivity:internal` and client-invisible · `participantDob` is vacuous for membership-only submits · `withAuth` clears routeAuthDrift rule 2 · the registry deferral conforms to the AGENTS.md boundary-isolation rule · no schema changes needed.

---

## Design conflicts

**1. §4's note-hold path defeats what #907 exists for.**
`external.ts:104-113` states the reason for the hold explicitly: "payment must not run in parallel with a review that hasn't read it… so a note like 'treat us as a volunteer household' settles dues first." Under unified signup the cart is minted at submit, so a note-writing family pays full dues *before* any reviewer reads the note — the exact outcome #907 was built to prevent, now needing a manual refund. The doc treats a paid `PENDING_BG_REVIEW` row as a benign convergence and doesn't flag the regression.

Suggested resolution: the orchestrator omits the membership line when `household.intakeNotes` is non-empty (programs-only cart; membership pays through the classic `PENDING_PAYMENT` step after review). Otherwise it belongs in §10 as a named accepted risk.

**2. §6.3 silently drops the coverage-window rule.**
Member pricing today requires `isActiveOrgMember` **and** `isActiveOrgMemberThrough(userId, programCoverageDate(program))` — a program running past the membership-year boundary charges full price, with an explicit `membership_ends_before_program` reason (`src/app/api/programs/[id]/discount-code/route.ts:31-38`). `computeUnifiedDiscountCents` as specified has no such term. This bites hardest in the flow's own target case: a brand-new membership is covered only to the next boundary, so a long program should not get member pricing.

**3. §6.3 grants member program pricing against an unapproved membership application.**
`includeMembership` means "they're applying", not "they're a member". If review lands BLOCKED, or the family never signs the contract, the program seats are already ACTIVE at member price with no membership behind them. This isn't §10.2 (that's a membership stall); it's a pricing leak on already-activated seats, and it isn't listed.

## Reuse

**4. `mintUnifiedDiscountCode` duplicates `mintMemberDiscountCode` (`lib/shopify.ts:712`).**
Same `discountCodeBasicCreate` mutation, `usageLimit: 1`, 48h TTL, mock-mode short-circuit, never-throws/null-on-failure. The only deltas are the code prefix and `appliesOnEachItem` + variant scoping. Generalizing the existing function with an options param is a smaller PR 3 than a sibling. It also makes §9's fallback free: "scope the minted code to the program variants instead of amount-off-order" is precisely what `mintMemberDiscountCode` already does.

## Gaps in the plan

**5. §6.5's resume path 400s on `submitIntake`.**
`intake.ts:357` selects a process with `status === "INTAKE"` and throws `IntakeError("no_process")` otherwise. A re-POST after the first submit is at `PENDING_EXTERNAL_ACTION`. §6.1's skip condition ("already ACTIVE or paid in-flight") doesn't cover unpaid-but-already-submitted — which is the resume case itself. The condition wants to be "process is not at INTAKE".

**6. §5's reconciler change also needs the driver.**
`reconcile.ts:545` is `const claimed = await reconcileForwardMembership(order); if (!claimed) await reconcileForwardProgram(order);`. For an attribute-carrying combined order both passes must run. The doc names the two functions but not the loop that gates them.

**7. §5's `reconcileReversals` widening is too wide.**
"any `paidAt` + `shopifyOrderId` row" pulls in BLOCKED, which `reconcile.ts:446-449` excludes deliberately ("a refund on a paid-while-blocked application is the expected fix"). `status: { notIn: ["BLOCKED", "ARCHIVED"] }` keeps that intact.

**8. §5: the webhook's receipt outcome strings go stale.**
`api/webhooks/shopify/route.ts:128-140` branches on `proc?.status`. An early payment leaves the status at `PENDING_EXTERNAL_ACTION`, which falls through to the `else`: `"no state change for process N — already processed"` — logged for a *successful* payment. Needs an early-paid branch in PR 1/2.

**9. §8 misses a CI guard: `lifecycleStatusLiteralAllowlist.test.ts`.**
A new `src/lib/join/checkout.ts` carrying §6.1's multi-status probe must either consume a `StateSet.where` or earn a new allowlist entry, or PR 5 fails. Separately, PR 2 makes the existing `lib/finance/reconcile.ts` justification string stale — it still reads "status in {ACTIVE,PENDING_BG_CLEARANCE}".

**10. §3's diagram over-assumes the process is at `PENDING_EXTERNAL_ACTION` when the cart is minted.**
`submitIntake` ends by calling `advanceExternalIfComplete` (`intake.ts:423`), so a bgFresh household with a board-pre-marked contract is already at `PENDING_PAYMENT` (or `PENDING_BG_REVIEW` on a note) by the time the orchestrator builds the link. Harmless — `activate()` handles all three — but the doc should say the orchestrator tolerates it.

**11. §2.4 / §10.3: the volunteer delta has no refund path.**
`applyVolunteerStatus` still runs on every advance (`external.ts:129`), so the membership ends up `isVolunteer: true` against a full-price payment. Either name the recovery path or state there isn't one. Related: the new `PENDING_BG_CLEARANCE` branch sends nothing to the applicant, where the classic path sends `notifyPaymentOpen` — worth stating the silence is intended.

## Nits

- `src/lib/programs/enrollmentAttributes.ts` (plural) alongside `src/lib/program/enroll.ts` (singular). Both directories exist today, so this may be deliberate — worth one line either way.
- §6.2: "P2002 = 409 = already-enrolled = success" — say explicitly that an already-ACTIVE seat is never re-charged. §6.5 implies it via "still-PENDING seats"; §6.2 should too.
- `/join` as `src/app/join/route.ts` is correctly invisible to `pageRegistry.test.ts`, which walks `page.tsx` only. Worth a note so nobody later "fixes" it into a page and trips the drift guard.

dkaygithub:
Responses to @thpr's review (findings 1 and 2; rest to follow with the doc revision):

**Finding 1 (note-hold / #907 regression) — decision: remove the branching, don't work around it.** The whole "if note: pause for board review before dues; else pay now" split is confusing, and the unified flow would make it worse (cart contents depending on whether a free-text field is non-empty). New position: **no note-based branching — everyone pays full (normal) dues at submit.** When a family that should have been volunteer-designated signs up before their pre-designation exists (note or otherwise), that becomes a **refund situation** handled by the board — same recovery path as other overpayments. This also answers the volunteer half of finding 11: the refund IS the recovery path, and the doc revision will name it (and fold the #907 hold's removal/rescoping into PR 1's lifecycle changes rather than leaving two payment orderings alive).

**Finding 2 (coverage window) — agreed, and it's bigger than this PR.** Discussed offline (Daniel + Tom): the real fix is that a program must declaratively state which fiscal year it is in — automatic when start/end sit inside one membership year, board-set at creation when it spans the boundary. Filed as #1484. That issue likely **blocks the pricing/orchestrator PR of this stack**, which should consume the declared year (via #1421's `isDuesSettledThrough`) instead of re-deriving coverage from dates.

🤖 Posted via Claude Code on Daniel's behalf

thpr:
This needs a rebase and integrating capabilities merged in #1421 rather than re-deriving the same content

thpr (changes requested):
Reconcile with PR 1421

jee7s:
**Area of concern: a fresh changes-requested that has not been answered.**

thpr filed changes-requested today asking for a rebase that consumes the capabilities #1421 merged rather than re-deriving the same content. #1421 landed 2026-08-02, after this doc was written.

Your 08-04 reply accepts findings 1 and 2 — dropping note-based payment branching in favour of everyone paying full dues at submit, with refund as the recovery path — and promises a doc revision. The head is unchanged since `707c5d87`, so the doc as written now contradicts decisions already taken in-thread, with findings 3-11 unanswered.

Finding 2 also spawned #1484 (fiscal-year declaration), which by your own assessment blocks the stack's pricing/orchestrator PR.

---
_Posted by Claude._

PR #1377, opened 2026-07-29
docs: a documentation standard built on a domain rules register (proposal)

thpr:
## What

Two documents, both proposals. No code, no schema, nothing built.

- **`docs/DOCUMENTATION_STANDARD.md`** — the permanent standard. Document classes and what triggers each one's edits; `docs/rules/` as a maintained register of board and operations decisions; `docs/in-design/` for in-flight work with a delete-at-merge lifecycle; enforcement by PR review rather than tests.
- **`docs/in-design/DOCUMENTATION_MIGRATION.md`** — the working plan to get the existing corpus there. It deletes itself when complete, which is the standard practicing its own rule.

## Why

Reviewing [#1370](https://github.com/innovationtreehouse/checkin/pull/1370) surfaced a design proposing to change two shipped rules while citing neither — because neither had a home:

- **The intake-note payment hold.** An applicant note holds the application at BG review so dues settle before payment. Stated only in a code comment in `src/lib/membership/external.ts`, a parenthetical in the transition table, and one aside in `docs/backlog/CUJS.md`.
- **The member-pricing coverage window.** Member program pricing requires validity through the program's whole run, not merely ACTIVE today. Stated only in a JSDoc in `src/lib/orgMembership.ts`, tagged `POLICY (flag for veto)` and still sitting there.

Neither was skipped carelessly. There was no file either belonged in. The design cited `LIFECYCLE.md` (mechanics) and `SHOPIFY_MEMBER_SEGMENT_PRICING.md` (itself an unbuilt proposal), and found nothing factual to check against.

An existing PR-history analysis quantifies the pattern: **192 of 861 distilled PRs trace to SPEC-stage causes** — `scope-miss` (87), `domain-knowledge-gap` (63), `auth-unclear` (26). Roughly one PR in five exists because a rule was never written where anyone would look.

## The diagnosis

Docs today are organised by *what change produced them*, so a rule is only recorded if it was large enough to warrant its own doc, and a `Status: SHIPPED` doc is a record that decays into a wrong statement of current behaviour. `docs/rules/` is the one class that later changes are obliged to edit — that asymmetry is the whole proposal.

## Notes for reviewers

- **Enforcement is deliberately human.** Nothing catches an *omitted* rule; no test or lint can see a rule that was never written. The standard says so plainly and says not to build a test that pretends otherwise. If you think that is the wrong call, that is the thing to argue about.
- **No counts anywhere.** Earlier drafts said "a page per domain" and "extract the 2–3 rules." Both were replaced with a per-line test — *could a later change violate this?* — because fabricated numbers get read as quotas.
- **Migration §5 deletes both `designs/` directories**, sorting their contents into `in-design/`, `ops/`, or extract-and-delete. `LIFECYCLE.md` goes to `ops/`: it documents mechanism, which is exactly why citing it as a rules source did not help.
- **`CUJS.md` and `UNFINISHED.md` are flagged unresolved** rather than guessed at.
- Open questions are in migration §8, including whether the six proposed domains are the right cut.

### DIFF
--- a/docs/DOCUMENTATION_STANDARD.md
+++ b/docs/DOCUMENTATION_STANDARD.md
@@ -26,7 +26,7 @@ domain rules — and gives everything else a shorter life.
 | `docs/rules/` | Board and operations decisions; invariants new work must not violate | **Permanent, edited by later changes** |
 | `docs/ops/` | How to run, deploy, test, and mock things | Long-lived, edited as tooling changes |
 | `docs/security/` | Security policy and boundary rules | Long-lived |
-| `docs/generated/` | Machine-derived artifacts, drift-tested | Regenerated, never hand-edited |
+| `checkin-app/docs/generated/` | Machine-derived artifacts, drift-tested | Regenerated, never hand-edited |
 | `docs/backlog/` | Journey inventory and gap tracking | Long-lived |
 | `docs/in-design/` | Proposals and designs being worked right now | **Deleted at merge** |
 
@@ -66,9 +66,9 @@ empty ones in advance.
 Decisions the board or operations have made about how things work, and
 invariants that later work must not violate.
 
-**Not:** implementation mechanism, state-machine structure (already generated
-and drift-tested under `docs/generated/`), rollout sequencing, or anything a
-reader could derive by reading the code.
+**Not:** implementation mechanism, state-machine structure (already generated and
+drift-tested under `checkin-app/docs/generated/`), rollout sequencing, or
+anything a reader could derive by reading the code.
 
 ### 3.2 Format
 
@@ -78,13 +78,19 @@ so a reviewer can judge a diff against it without opening code:
 ```markdown
 ## Membership application
 
-- An intake note holds the application at background-check review. Payment does
-  not open until two reviewers have read the note — a family that writes
-  "treat us as a volunteer household" must have dues settled before paying.
+- An intake note holds the application at background-check review, so reviewers
+  read it before dues are settled — a family writing "treat us as a volunteer
+  household" must not pay first. A household that already holds a still-valid
+  background clearance is exempt and goes straight to payment.
 
 - Membership activates only after background-check clearance. Payment arriving
   first does not activate; clearance arriving first does not activate.
 ```
+
+Note what the first rule spends a whole clause on: the **exemption**. An earlier
+draft of this document stated it without the clearance carve-out, which would
+have marked shipped, correct code as a violation. State the guards and carve-outs
+the code actually has, or the register does damage rather than none.
 
 **Write rules as constraints, not descriptions.** If the sentence does not let a
 reviewer tell whether a change violates it, rewrite the sentence.
@@ -98,6 +104,17 @@ linked to a diff and stays accurate on its own.
 **No line-number citations.** They rot within weeks and pull the register toward
 implementation when what a reviewer needs is the rule. A bare file path is the
 most a "where this bites" pointer should carry, and most rules need none.
+
+**Never state an over-broad rule.** A rule that claims more than the code does is
+worse than no rule: it marks correct behaviour as a violation, and the likely
+response is someone "fixing" working code to comply. When in doubt, state the
+narrower version, or leave it out and raise it.
+
+**An unratified policy is a candidate, not a rule.** Code sometimes carries a
+decision that was never actually decided — a `POLICY (flag for veto)` comment, a
+choice made to unblock a PR and never confirmed. Writing it here converts an open
+question into a stated invariant nobody agreed to. Take it to the owner or the
+board first; record it only once it is answered.
 
 ### 3.3 How much belongs
 
--- a/docs/DOCUMENTATION_STANDARD.md
+++ b/docs/DOCUMENTATION_STANDARD.md
@@ -1,0 +1,203 @@
+# Documentation standard
+
+**Status: PROPOSED — for review. Becomes the standard on merge.**
+How documentation in this repo is organised, written, and retired. Read this
+before writing or moving any doc.
+
+---
+
+## 1. Why this exists
+
+Documentation organised by *what change produced it* tells you whether a change
+landed. It does not tell you **what is true now** — so a rule only gets written
+down if it happened to be large enough to warrant its own doc, and a doc
+describing a shipped change silently decays into a wrong statement of current
+behaviour.
+
+This standard adds one register that is maintained rather than frozen — the
+domain rules — and gives everything else a shorter life.
+
+---
+
+## 2. Document classes
+
+| Where | What | Lifetime |
+|---|---|---|
+| `docs/rules/` | Board and operations decisions; invariants new work must not violate | **Permanent, edited by later changes** |
+| `docs/ops/` | How to run, deploy, test, and mock things | Long-lived, edited as tooling changes |
+| `docs/security/` | Security policy and boundary rules | Long-lived |
+| `docs/generated/` | Machine-derived artifacts, drift-tested | Regenerated, never hand-edited |
+| `docs/backlog/` | Journey inventory and gap tracking | Long-lived |
+| `docs/in-design/` | Proposals and designs being worked right now | **Deleted at merge** |
+
+Most of these are edited over time — ops docs follow the tooling, security
+policy evolves, the backlog is maintained continuously. What sets `docs/rules/`
+apart is *what triggers* the edit.
+
+A doc in any other class changes when **the thing it describes** changes: new
+mock, new deploy step, new gap found. That is ordinary upkeep, and it falls on
+whoever touches that thing.
+
+`docs/rules/` changes when **any work in its domain** establishes, alters, or
+retires a rule — even work that has nothing to do with documentation. That makes
+it the one class carrying an obligation outward onto unrelated changes, which is
+why the extract step in §4 exists at all.
+
+---
+
+## 3. `docs/rules/` — the domain register
+
+One file per domain, named for the domain and not for any feature:
+
+| File | Covers |
+|---|---|
+| `membership.md` | org membership: application, review, dues, renewal, lapse |
+| `programs.md` | programs and enrollment: eligibility, capacity, pricing, scholarships |
+| `people-households.md` | households, leads, guardianship, youth/adult status |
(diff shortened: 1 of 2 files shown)


jee7s:
High-effort review. The diagnosis here is right and the quantified case for a rules register is the strongest argument I've seen for changing how this repo documents itself. Findings below are all verified against current `main`.

Two classes of defect: the seeded exemplar rules restate shipped code inaccurately, and the migration plan's destructive steps have no reference sweep.

## The seeded rules are wrong, and that is the load-bearing failure

Both rules the proposal uses to demonstrate the format — and which §2 then seeds `docs/rules/` from — drop a condition present in the code they cite. Since §5 makes the register the thing a reviewer judges a diff against, a wrong rule is worse than a missing one: it actively rules correct code non-compliant.

**1. Intake-note hold drops the `!bgClearedAt` guard.** `checkin-app/src/lib/membership/external.ts:112`:

```ts
const holdForNote = !process.bgClearedAt && !!membership?.household.intakeNotes?.trim();
```

A household with an intake note *and* a still-valid prior clearance goes straight to `PENDING_PAYMENT` — no two-reviewer hold. Both the §3.2 exemplar ("Payment does not open until two reviewers have read the note") and migration §1.1 state it unconditionally. A reviewer checking a renewal path against the seeded rule reads the fast path as a violation; the worse outcome is someone deleting the fast path to comply.

**2. Member-pricing window omits the ongoing-program carve-out.** `checkin-app/src/lib/orgMembership.ts:132` is `p.endAt ?? p.startAt ?? null`, and the JSDoc immediately above says: *"POLICY (flag for veto): an ongoing program (endAt null) requires validity only through its START, not indefinitely."* The migration states coverage through "the program's whole run" and omits the branch entirely. A member lapsing mid-run of an ongoing program is correctly still eligible under shipped code but violates the register.

Worth noting: that JSDoc says the rule is *awaiting a veto*. Enshrining it as settled in `docs/rules/` before §3.5's owner review runs converts an open policy question into a stated invariant — which is a failure mode the standard should probably name, since mined rules will hit it repeatedly.

## The migration's delete steps have no reference sweep

**3. Deleting both `designs/` directories dangles ~30 pointers.** `config.ts:13,:101,:185`, `lifecycleDrift.ts`, `membership/lifecycle.ts`, `cron/lifecycle-reconcile/route.ts`, `.github/workflows/shopify-live.yml:4`, and `README.md:47-52` all cite `docs/designs/<file>.md`. §7's "Done when" lists only AGENTS.md, so every one of those 404s afterward — including the README front-page links and the AGENTS.md instruction to "read `DEV_INSTANCE_DESIGN.md` before touching auth/env".

**4. Moving `LIFECYCLE.md` breaks a byte-equality guard.** `lifecycle/artifacts.ts:115` bakes the literal string `Generated from the machine's TRANSITIONS (docs/designs/LIFECYCLE.md)` into the generated artifacts, and `artifactsDrift.test.ts` asserts byte-equality. Either the mover edits the string without re-running `npm run generate:lifecycle-artifacts` (CI red on a confusing docs-path diff), or leaves it and ships drift-tested "do not hand-edit" files citing a deleted path. The one class §2 calls machine-verified ends up provably wrong.

**5. `PRODUCTION_PLAN.md` is bucketed for deletion but is the live prod runbook.** It is the ordered cutover checklist for ops.innovationtreehouse.org and still carries open items (APPLIED trust policy verification, release-gate reviewer roster, ECR/ECS task-def names). Extract yields zero rules — as §5 predicts — so it gets deleted, and the next deploy or rollback has no runbook. The plan's own ops-bucket definition already covers it; it should move, not delete.

**9. `HOUSEHOLD_LEAD_MODEL.md` has two destinations.** Step 1 promotes it into the seeded pair (`membership.md`/`programs.md`); step 4 sends it to `people-households.md`, which step 1 never creates. Run in order, guardianship rules land in `membership.md` — violating "named for the domain, not for any feature" — and then the source is deleted.

**10. `docs/generated/` is at the wrong root.** §2's table (and migration §5's "leave alone entirely") point at root `docs/generated/`, which does not exist; the artifacts are at `checkin-app/docs/generated/`. An agent following §2 writes to the wrong path and gets `Missing docs/generated/lifecycle/<name>.md` from the drift test while the file visibly exists.

## The evidence needs reconciling before it carries the argument

**6. Neither PR count reproduces.** `gh pr list --state merged`: 1028 merged total — 146 in #1–248, **816** in #249–1168, **66** since #1169. The doc says 861 and 80. Merged counts only grow, so both figures appear to include closed-unmerged PRs. Step 2's coverage-gap range is therefore wrong, and "100% distilled" is measured against an inflated denominator.

**7. The headline figure doesn't match its own table.** 192 SPEC-stage PRs vs 87 + 63 + 26 = **176**. That 192/861 "a little over a fifth" is the entire quantified case, and §3.2 sizes the sweep from the same three rows — so 16 rule-bearing PRs are either phantom or unmined.

**8. Both step-2 corpora are unreachable.** `~/Software/Checkin/pr-mining/` does not exist (no `/home/jeff/Software` at all), and the fallback commit `0107a949` is not in this clone. Step 2 is described as "the bulk of the value and the only step that cannot be done incrementally later", so whoever picks it up either stalls or re-derives the corpus from scratch — the one thing §3.1 says not to do.

## Suggestion

The standard and the migration have very different readiness. The standard needs findings 1, 2, and 10; the migration needs a reference-sweep step, the runbook re-bucketed, and the evidence reconciled. Splitting them would let the standard land while the migration keeps iterating — which is also what its own `in-design/` lifecycle implies.

jee7s:
☝️ This is claude's analysis. I don't think I agree with it fully, but it raises some challenges we should consider.

1. We should be more explicit about wanting those dangling pointers to also be deleted.
2. Claude got surprisingly specific at the start. I'm not sure what caused that.
3. We may need to start the build out of the change in docs/design pattern and let claude re-evaluate.

thpr:
## Updated — review findings addressed, plus a scope change

Thanks for the review. Every finding was verified against current `main` before acting; all were correct, and one was worse than reported. Four commits: `edaa4322`, `20b42fbb`, `edafbd3d`, `f171121e`.

### Findings 1 & 2 — the seeded rules

Both confirmed, and these were the important ones. Each dropped a condition from the code it cited:

- **Intake-note hold** omitted the `!process.bgClearedAt` guard (`external.ts:112`). A household with a still-valid prior clearance takes the direct path to payment and is exempt from the hold. The exemplar in §3.3 now carries the exemption, with a note explaining why it spends a clause on it — an earlier draft of the standard got this wrong, which is the most direct argument available for the rule it now states.
- **Member-pricing window** omitted the ongoing-program carve-out (`orgMembership.ts:132` resolves `endAt ?? startAt`). Also correct that the JSDoc tags it `POLICY (flag for veto)` — so it is **no longer seeded into `docs/rules/` at all**, pending an owner answer.

Both failure modes are now named as format rules rather than left implicit: *never state an over-broad rule* (it marks correct code as a violation, and the likely response is someone "fixing" working code to comply) and *an unratified policy is a candidate, not a rule*.

### Findings 3, 4, 5, 9 — the destructive steps

- **3.** Confirmed and understated: **63** references to `docs/designs/…` across `checkin-app/src`, `.github/`, `README.md`, `AGENTS.md` — not ~30. New §5.1 reference sweep runs *before* any move, with the grep in "Done when".
- **4.** Confirmed exactly. `artifacts.ts:115` bakes the path into byte-equality-tested artifacts. Documented as a single commit: edit the literal, regenerate, commit together. "Done when" now requires `generate:lifecycle-artifacts` to produce no diff.
- **5.** Confirmed — it is the live cutover runbook with open items. Moved to the `ops/` bucket with an explicit *do not delete despite the name*.
- **9.** Confirmed contradiction. Handled once, in step 4, where `people-households.md` is created; step 1 explicitly names it as not-a-source.

### Findings 6 & 7 — the evidence

Both correct. Rebuilt on the current corpus (941 PRs, #249–1372): **216** carrying a SPEC-stage cause, **243** rule-bearing.

The 192-vs-176 gap has a specific cause worth recording: cause families **overlap**, so the old figure summed columns that were not disjoint. The corrected number is the deduplicated union, and the doc now says so rather than presenting a sum. Merged-only counts (885) and the LLM-distillation caveat are stated rather than implied.

### Finding 8 — handled differently

The corpus being unreachable is by design; it stays owner-local. Rather than make it reachable, §3.1 now states the consequence you were really pointing at: **only the owner can execute step 2**, and handing it off makes publishing a prerequisite rather than a detail.

### Finding 10

Confirmed, fixed in both places.

---

## Scope change — policy vs procedure tiers

Separate from the review: the design was incomplete on a dimension it did not model at all.

Rules do not have uniform authority. Some exist because **the board adopted a policy** — the code implements them, it does not define them. Others are **working agreements** the team settled on because something had to be decided. Both are real rules; they are not equally negotiable.

Every domain file is now split, **Policy above Procedure**:

- A **Policy**-tier violation is not a design disagreement to settle in review. The policy has to change first, which is a board action, so the correct reviewer response is to stop and escalate rather than weigh trade-offs.
- **Procedure** rules can be renegotiated by the people doing the work, in a PR.

**Citations.** Policy-tier rules cite the governing policy by name and *structural* location — article, section, subsection, whichever that policy uses. `Bylaws Art. IV §3(b)`.

Explicitly barred: **page numbers** (they move on reformatting, differ between renderings, and do not exist in some formats), **filesystem paths**, and **Drive URLs** (rot on reorganisation, permission-gated). The name plus its article is the one identifier stable across every rendering — and when an amendment breaks a citation, that is correct behaviour, because the rule may have changed with it. A policy with no internal numbering is cited as such rather than given an invented locator; that gap usually means the policy itself needs structure.

**Where policies live.** Google Drive, canonical. That makes this tier verifiable rather than merely asserted: anyone with Drive access can check a policy-tier rule against the policy it names.

**Migration impact.** The mine cannot classify tiers on its own — a diff shows what code does, never whether a policy required it. So it drafts everything as **procedure by default**, flags likely candidates (dues, background checks, eligibility, refunds, volunteer status, access), and only the owner promotes them against Drive. A third outcome exists beyond promote/leave: a candidate that *contradicts* its policy is a compliance finding routed to the owner, not a documentation one.

---

## On splitting the PR

Reasonable suggestion, and it was right at the time. Since the migration's defects turned out to be fixable in one pass, splitting now would mostly add overhead — but happy to split if you would still rather review them separately.

jee7s (approved):

PR #1214, opened 2026-07-23
feat(security): register GET /api/settings/membership/volunteer-designations (inert)

dkaygithub:
## What
Wave-1 read route, small **registry-only PR to `main`** (one route per PR). Adds a single `defineRoute` for `GET /api/settings/membership/volunteer-designations`. Handler stays on `withAuth` → **inert** until its `handler()` conversion.

## View
Admin surface (`sysadmin`/`board`). `VolunteerDesignation` rows carry email (pii); near-raw response, so the full everyones band is exact. **GET is a pure `findMany`** — the create/delete are on POST/DELETE (out of scope). Unlike `settings/email` and `settings/membership` (both upsert-on-GET, held back per #1137 finding 3), this GET does not mutate.

## Verified
- `tsc --noEmit`: clean (validated in the 4-entry union pass).
- Mocked suites (`ctxNeeds`, `authzRegistry`, `scopeBindingsEquivalence`): **12/12**.
- Boundary-isolation: `registry.ts` only. Isolated worktree off `origin/main`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1387, opened 2026-07-29
feat(membership-ops): volunteer roster — current volunteers + pre-designated emails

### LINKED ISSUE #1374: Membership Ops -> Volunteer List is blank

dkaygithub:
Fixes #1374

## The report

"Membership Ops -> Volunteer List is blank." The Volunteers tab rendered only the designated-email table, so on an instance with no designations it read as an empty card — and the tab's name promised a roster of volunteers that was never there.

Two separate defects behind that:

1. **Wrong content.** The page listed `VolunteerDesignation` rows only. Households actually *on* volunteer dues (`OrgMembership.isVolunteer`) — the thing the tab is named after — appeared nowhere.
2. **Silent failure.** `load()` did `if (res.ok) setDesignations(...)`, so a 403/500/network failure fell through to the "No volunteer designations yet." empty state. A broken load and an empty list looked identical. Same bug class as #1385.

## What this does

The tab now shows the roster its name implies: every household on volunteer dues, plus emails designated ahead of signup, in one sortable/filterable table.

Rows are **household-grained**, because `isVolunteer` lives on `OrgMembership` (one per household) — volunteer status is a property of the family's dues, not of a person. A five-person volunteer family is one row showing its lead(s).

One status chip per row:

| Chip | Means |
|---|---|
| Volunteer member | `isVolunteer` + ACTIVE |
| Signup in progress | `isVolunteer`, not yet active, with an in-flight process |
| Application blocked | `isVolunteer` with a BLOCKED process — a BG reviewer filed REJECT; live and board-actionable |
| Pre-designated | a designated email with no volunteer membership yet |
| Takes effect next renewal | a designated email whose household already holds an ACTIVE full-price membership |
| Revoked | REVOKED / DENIED |
| No live application | `isVolunteer` with nothing in flight (archived/abandoned) — hidden behind a **Show inactive** toggle |

"Takes effect next renewal" surfaces standing what the add-form previously only warned about once, at write time — a designation that will not retroactively fix an already-active full-price membership.

Designations **fold into their household's row** rather than listing separately. A household can hold several (both parents designated is normal), so each keeps its own Remove action on that row, labelled with its own address — a designation is an email-level record but the row is household-level once the household exists, so a bare "Remove" there wouldn't say which address it drops. A row with no household yet *is* the email, so that one keeps the bare label. The `Show inactive` filter never hides a row carrying a designation — that row is the only place the designation can be removed.

A designated email resolves to its household through `canonicalizeEmail` against **household leads** — the same rule `applyVolunteerStatus` / `matchesVolunteerDesignation` use to actually grant volunteer dues — so the roster cannot show a standing the engine won't produce (plus-tagged and gmail-dotted addresses match; a designation on a non-lead person does not resolve a household).

Search spans household, lead, and designated emails; the status chips filter the roster; every column except the action column sorts. A failed load — or a failed Remove — now shows an error banner instead of masquerading as success.

## Implementation

- **New** `GET /api/membership-ops/volunteer-memberships` (sysadmin/board). Reuses `IN_FLIGHT_INITIAL`/`IN_FLIGHT_RENEWAL` from `lib/membership/lifecycle` for the in-progress test rather than restating the status list, and `LIVE_PERSON` for every Person read. `BLOCKED` is loaded *alongside* those sets, deliberately not added to them — they mirror the `membership_one_inflight_*` partial unique indexes. Registered in the negative-authz drift guard (`authzRegistry.test.ts` + `authzRoleRejection.integration.test.ts`).
- `GET/POST/DELETE /api/settings/membership/volunteer-designations` is **unchanged** — deliberately, so this does not collide with the handler() migration of that route in #1141 / #1214. The page still adds and removes designations through it.
- Page rebuilt on the shared `DataTable` (sorting, loading, empty state) + Mantine `Chip.Group`.
- Dates go through `lib/time`: `memberSince` is a calendar date (`formatDateOnly`, UTC-pinned per `1149_DATE_TIME_TZ_DESIGN.md`), a designation's `createdAt` is a true instant (`formatDate`).
- Tab renamed "Volunteer Memberships" → "Volunteers" (`membershipOpsNav.ts`, `pageRegistry.ts`): it is a roster now, and pre-designated rows are not memberships.

## Verification

- 26 tests — 13 page (roster render, empty vs. failed load, search filter, chip filter, sort, add, remove, failed remove, multi-designation row, per-row Remove labelling, inactive hide/reveal, inactive-with-designation stays, role guard) and 13 route (per-status row assembly including BLOCKED and INACTIVE, the fold-in, multi-designation fold-in, canonicalized lead matching, the lead-only filter, the skipped person lookup).
- `tsc --noEmit` and `npm run lint` clean.
- Full unit suite: 2396 passed, 2 failed — `membership page › fills every intake field…` and `HouseholdPage › covers the remaining server-error…`, both 5s timeouts that reproduce identically on main with this branch stashed. Pre-existing, not from this change.


thpr (changes requested):
## Review

Rebuilds Membership Ops → Volunteers from a designated-email list into a household-grained roster (volunteer `OrgMembership`s + pre-designated emails) behind a new `GET /api/membership-ops/volunteer-memberships`, and kills the `if (res.ok)` silent-load-failure. Right diagnosis, good test discipline, and the scoping around the #1141 / #1214 handler() migration is the correct call.

Findings below are against the diff plus current `main` for the schema and helpers.

---

### Blocking

**1. Fold-in silently drops designations, leaving them unremovable** — `volunteer-memberships/route.ts`

```js
if (existing) {
    existing.designationId = d.id;   // single slot
    existing.designatedAt = d.createdAt.toISOString();
```

One slot per household row. Two designated emails in the same volunteer household — both parents added, entirely normal — and the second overwrites the first. The first then has no row, no Remove button, and is invisible on the only page that manages designations.

Fix: `designationIds: number[]` on the row, with the Remove column rendering one button per designation. Note the `FULL_PRICE` rows have the same shape, so "just fold them by household" isn't the answer there either — it would drop Remove actions the same way.

**2. Designation → person match ignores `canonicalizeEmail`, and doesn't filter to household leads**

The roster matches with `email: { in: emails, mode: "insensitive" }` plus a JS `.toLowerCase()` — case-only. But the repo's email-identity rule is `canonicalizeEmail` (`lib/emailNormalize.ts:17` — lowercase, strip `+tag`, drop dots on gmail/googlemail), and that is exactly what `matchesVolunteerDesignation` uses to decide `isVolunteer` (`review.ts:326-328`). Separately, `applyVolunteerStatus` only ever matches against `isHouseholdLead: true` emails; the roster's lookup has no such filter.

So the roster contradicts the engine on precisely the cases the helper exists for:

- Designation `foo.bar+vol@gmail.com`, person `foobar@gmail.com` → `applyVolunteerStatus` grants volunteer dues; the roster fails to link them. Result is a stray "Pre-designated" row with no household *and* a separate volunteer household row, fold-in never fires, and the designation's Remove lands on the wrong row.
- A designation on a **non-lead** email (a youth, a second adult not marked lead) resolves that household and renders a status implying the designation is in play — when it can never confer volunteer dues.

One query rewrite fixes both. The pattern already exists in `api/webhooks/resend/route.ts:61-68`:

```js
const wanted = new Set(designations.map((d) => canonicalizeEmail(d.email)));
const people = await prisma.person.findMany({
    where: { email: { not: null }, isHouseholdLead: true, ...LIVE_PERSON },
    select: { email: true, household: { select: { /* … */ } } },
});
const householdByEmail = new Map(
    people.filter((p) => wanted.has(canonicalizeEmail(p.email!)))
          .map((p) => [canonicalizeEmail(p.email!), p.household]),
);
```

That drops the `in` + `mode: "insensitive"` query entirely. The webhook carries a `ponytail:` note that its O(n) scan is fine at this org's scale — the same tradeoff applies, and this route already loads every volunteer household anyway.

**3. Two dead-application states collapse into "Pre-designated"**

`OrgMembershipStatus` is only `NONE | ACTIVE | REVOKED | DENIED`, and `BLOCKED` / `ARCHIVED` processes are deliberately out of `IN_FLIGHT_INITIAL` / `IN_FLIGHT_RENEWAL`. So an `isVolunteer` household falls through to `DESIGNATED` in two very different situations, and neither is a designation:

| Case | What it is | Right treatment |
|---|---|---|
| BLOCKED process | a BG reviewer filed REJECT — live, board-actionable, possibly already paid (`PAID_WHILE_BLOCKED` → refund may be needed) | must stay visible, with a distinguishable chip |
| ARCHIVED / no process | board-disposed abandoned application | drop from the default roster, or put behind a "Show inactive" toggle |

No new enum value needed for this. One caution: please don't fix it by adding `"BLOCKED"` to `IN_FLIGHT_INITIAL` / `IN_FLIGHT_RENEWAL` — those mirror the DB partial unique index `membership_one_inflight_renewal` (`lifecycle.ts:91`), so widening them changes what the index forbids. Add BLOCKED as a separate term in the route's process filter.

**4. The "Full-price member" chip describes the wrong subject**

The row is a *designation* that is inert until the household's next renewal cycle — which the POST route already states at write time (`volunteer-designations/route.ts:45`). Labeling it "Full-price member" describes the household rather than the designation, so on a volunteer roster it reads as full-price families being listed. Suggest "Takes effect next renewal" or "Not yet applied".

---

### Worth fixing while here

**5. The silent-failure class survives in `removeDesignation`.** `await fetch(…, { method: "DELETE" }); await load();` ignores `res.ok`, so a 403/500 delete is indistinguishable from success — the same class as #1385 that `load()` just fixed, and this PR moves Remove into the new roster.

**6. Local `formatDate` duplicates and misnames a shared helper.** `formatDateOnly` in `lib/time.ts:95` is already the UTC-pinned formatter, and `formatDate` in that same module means the opposite (instants). So:
- `memberSince` → `formatDateOnly`. UTC pinning is correct here per `1149_DATE_TIME_TZ_DESIGN.md` Finding 3.
- `designatedAt` is `createdAt @default(now())` — a true instant. UTC-pinning it shows tomorrow's date for anything designated after ~7pm Chicago. That one wants `formatDate`.

---

### Tests

- With the #2 rewrite, the interesting cases become plain JS and need no DB: a plus-tagged designation, a gmail-dots designation, and a designation on a non-lead person. Changing the existing `'ANA@example.com'` fixture to something like `'A.na+vol@Gmail.com'` would have caught the mismatch.
- No coverage for two designations on one household (finding #1), or for the designation → in-flight-process → `IN_PROGRESS` branch.

### Security

`withAuth` is sanctioned under routeAuthDrift rules 1 and 2, no edge-sensitive model is read so rule 3 is clear, and the deny paths are registered in both drift guards. Correct as built. The route isn't registered in `src/security`, so no response-stripping applies to `Person.email` / `name` / `VolunteerDesignation.email` (all `@sensitivity:pii`) — consistent with the sibling membership-ops routes, and per the boundary-isolation rule a registry entry has to be its own PR. Follow-up, not a blocker.

### Minor

- Every Remove button shares the accessible name "Remove" — the test having to index `removes[0]` is the symptom. `aria-label={\`Remove ${r.email}\`}`. This compounds with #1, where one row will carry several.
- `processes: { take: 1 }` with no `orderBy` is fine for an existence check, but fragile if that select ever grows.
- `GET /api/settings/membership/volunteer-designations` now has zero app consumers (deliberate). Worth an issue so it isn't later read as live.

### Verified good

- `memberSince` is non-nullable (`@default(now())`), so `.toISOString()` on the ACTIVE branch is safe.
- `PERSON_BG` processes carry `orgMembershipId: null`, so `m.processes` can't be polluted by them — no kind filter needed.
- `DataTable` sorts nulls last in both directions, so the sort test's expectation holds.
- Type-import-from-route matches the existing `TodoCounts` convention.
- The empty-designations short-circuit before `person.findMany` is a nice touch, and tested.

---

**Overall:** solid direction and the right rebuild. #1 is a data-management bug (a designation nobody can see or delete), and #2–#4 each show board staff a status that isn't true. Those four before merge; the rest are cheap to fold in.

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

dkaygithub:
Thanks — all four blocking findings reproduced against the code, all four fixed in 8cb2556f, plus both "worth fixing while here" items and the `aria-label` minor.

### 1. Fold-in dropped designations

Confirmed: `existing.designationId = d.id` is a single slot, so the second designation in a household vanished with no Remove anywhere.

`VolunteerRow` now carries `designations: VolunteerRowDesignation[]` (`{ id, email, createdAt }`), and the action column renders one button per entry. Took your note about `FULL_PRICE` rows having the same shape — designation-derived rows now register in `rowByHousehold` too, so two designations on one full-price household fold to one row with two Remove actions rather than two rows.

Button text stays `Remove` when a row has one designation and becomes `Remove <email>` when it has several; `aria-label` is always `Remove ${d.email}`, which also settles the minor about the shared accessible name — tests now select by name instead of `removes[0]`.

### 2. Canonicalization + household-lead filter

Confirmed against `review.ts:325-346`. Rewritten to your pattern:

```js
const wanted = new Set(designations.map((d) => canonicalizeEmail(d.email)));
const leads = wanted.size ? await prisma.person.findMany({
    where: { email: { not: null }, isHouseholdLead: true, ...LIVE_PERSON },
    select: { email: true, household: { select: { /* … */ } } },
}) : [];
```

`in` + `mode: "insensitive"` is gone, with a `ponytail:` note mirroring the resend webhook's. Two new route tests cover it: `A.na+vol@Gmail.com` folding into the household whose lead is `ana@gmail.com` (with a decoy lead in a different household that a case-only match would have missed), and a designation on a non-lead resolving no household.

### 3. BLOCKED vs. ARCHIVED

Confirmed, and I did not widen the in-flight sets. `LIVE_PROCESS = [...IN_FLIGHT, "BLOCKED"]` is a separate term used only for this route's `processes` filter, which now selects `status` so the two are distinguishable:

```js
const processStatus = (processes) =>
    processes.some((p) => p.status === "BLOCKED") ? "BLOCKED"
    : processes.length > 0 ? "IN_PROGRESS" : null;
```

- BLOCKED → its own chip, **Application blocked** (orange). Stays on the default roster.
- nothing live → `INACTIVE`, chip **No live application**, behind a `Show inactive` switch (off by default).

One wrinkle worth calling out: hiding INACTIVE rows outright would have re-created finding #1 — an inactive household holding a designation would lose the only Remove button for it. So the filter is `!showInactive && status === "INACTIVE" && designations.length === 0`. Tested both ways.

### 4. "Full-price member"

Agreed, the chip named the household on a page about designations. `FULL_PRICE` → `NEXT_RENEWAL`, label **"Takes effect next renewal"** — took your first suggestion since it says what happens rather than what didn't.

### 5. `removeDesignation` silent failure

Fixed — checks `res.ok`, surfaces the error in the banner, leaves the row in place, and shows a confirmation toast on success. New page test asserts the failed-delete path.

### 6. `formatDate`

Local helper deleted. `memberSince` → `formatDateOnly`, designation `createdAt` → `formatDate`. The instant case was genuinely wrong — anything designated after ~7pm Chicago rendered tomorrow.

### Tests

16 → 26. Added: BLOCKED and INACTIVE row assembly, multi-designation fold-in, canonicalized lead match, non-lead non-match, the designation → in-flight-process → `IN_PROGRESS` branch you flagged as uncovered, failed remove, and the inactive hide/reveal pair. `tsc --noEmit` and `npm run lint` clean.

### Follow-ups not in this PR

- `src/security` registry entry for the route — agreed, boundary-isolation rule means its own PR.
- `GET /api/settings/membership/volunteer-designations` having zero app consumers. Happy to file that issue; say the word and I'll open it rather than guess at the disposition (delete vs. keep for the #1141/#1214 migration).
- `processes: { take: 1 }` with no `orderBy` — moot now, the select is `{ status: true }` over the filtered set with no `take`.

dkaygithub:
Follow-up on the Remove-label minor, from a round of feedback here: `c6b69d9d` makes the address **unconditional on household rows**, not just on rows carrying more than one designation.

The grain is the reason. A `VolunteerDesignation` is an **email**-level record, but a roster row is **household**-level as soon as the household exists — so a bare "Remove" on one of those rows doesn't say which of the household's addresses it drops, even when there's only one today. The label is now gated on `householdId`, not on `designations.length`:

```js
{r.householdId != null ? `Remove ${d.email}` : "Remove"}
```

A row with no household yet *is* the email — its Email column shows the same address — so that one keeps the bare label.

`aria-label` was already the full `Remove ${d.email}`, so every existing Remove selector was already green and would have stayed green if the visible text regressed. Added an assertion on the rendered text in both shapes to pin it: the unlinked row renders exactly `Remove`, the household-linked row renders `Remove bo@example.com`.

26 tests still passing, `tsc --noEmit` and lint clean.

thpr:
`c6b69d9d` looks right, and the reasoning for it is better than the minor I filed. Gating on `householdId` rather than `designations.length` is the correct seam: a row with no household *is* the email, so the bare label is unambiguous there; once the row is household-grained the address is load-bearing even at count 1. Pinning the rendered text in both shapes (not just the `aria-label`) is the right test, since that's exactly the regression the accessible-name selectors would have hidden.

Only cosmetic note: household rows with a single designation now render a full address in the action column, so that column gets wider on the common case. Fine as-is.

---

## Still blocked, and I owe you a correction

My first review said the `src/security` registry entry was a follow-up rather than a blocker. **That was wrong**, and it's the only thing keeping this red. Sorry — you've been iterating on my other findings while the actual merge blocker sat untouched.

`Run Tests (unit)` on `c6b69d9d`:

```
✗ [new-route-old-authz] checkin-app/src/app/api/membership-ops/volunteer-memberships/route.ts
  — GET /api/membership-ops/volunteer-memberships is not in the security registry and not
    in the frozen legacy baseline (scripts/legacy-authz-routes.txt). New routes must use
    defineRoute() + handler() from @/security.

1 error(s), 175 warning(s)
(1 new-route-old-authz error(s) block even in advisory mode)
```

`scripts/check-route-coverage.ts:243`. And `scripts/legacy-authz-routes.txt` closes the other door in its header — the list may only shrink, and adding a line "requires CODEOWNERS sign-off and is an explicit decision to ship new surface outside the audited policy layer. Don't."

So `withAuth` isn't an option for this route, even though routeAuthDrift accepts it. The sequence AGENTS.md prescribes:

1. A registry-entry-only PR adding `defineRoute()` for `GET /api/membership-ops/volunteer-memberships`. Inert on its own, and it satisfies the boundary-isolation workflow's "no app/feature code" rule.
2. That merges, then this PR converts the route from `withAuth` to `handler()`.

Both rules I cited are real; they just have an ordering between them that I didn't state. Worth checking what response shape the registry entry needs to declare for the `Person.name` / `Person.email` / `VolunteerDesignation.email` fields the roster returns — `@sensitivity:pii` across all three, so the `returns:` grant is the substantive part of that first PR, not boilerplate.

## Test signal is thinner than the numbers suggest

The unit job dies at the policy-lint step **before jest runs** — no `junit.xml`, no coverage artifact, which is also what's failing the coverage gate. So the 26 passing tests are a local result; CI has not executed them on any push in this PR. Integration and flow are green but neither touches this route.

Not a criticism of the tests — I read them and they match the claims. Just flagging that the green you're reporting and the green CI would report are different things right now, and the ordering fix above is what unblocks the second one.

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

thpr:
Two more minors, both at `c6b69d9d`. Neither is a merge blocker — the `new-route-old-authz` gate is still the only thing holding this.

## 1. The designation branch drops REVOKED / DENIED

`volunteer-memberships/route.ts`, membership rows vs. designation rows:

```js
// membership rows — REVOKED/DENIED is a first-class outcome
m.status === "REVOKED" || m.status === "DENIED" ? "REVOKED" : …

// designation rows — it isn't
membership?.status === "ACTIVE"
    ? "NEXT_RENEWAL"
    : ((membership && processStatus(membership.processes)) ?? "DESIGNATED")
```

A designation resolving to a household whose (non-volunteer) membership is REVOKED or DENIED, with no live process, falls to `DESIGNATED` — chip **"Pre-designated"**. The row shows the household name and a status implying the designation is queued up and waiting.

DENIED is the case that bothers me. Per the schema comment it blocks login for every member of the household, so "Pre-designated" tells the board this family is set up for volunteer dues at signup, when in fact they cannot sign in at all. REVOKED is milder but the same shape.

Cheapest fix is to hoist the check so both branches share it, since `REVOKED` already exists as a status and the membership branch's ordering is the one you want:

```js
membership?.status === "REVOKED" || membership?.status === "DENIED" ? "REVOKED" : …
```

## 2. INACTIVE badge renders with no filter chip for it

Two filters key off `showInactive`, but only one of them has the designation exception.

Chip bar:
```js
{STATUS_CHIPS.filter((c) => showInactive || c.value !== "INACTIVE").map(…)}
```

Row filter:
```js
if (!showInactive && r.status === "INACTIVE" && r.designations.length === 0) return false;
```

The row filter's carve-out is right — an INACTIVE row carrying a designation must stay, since it holds that designation's only Remove, and you have a test pinning exactly that shape. But the chip bar copied the flag without the exception. So with the switch off, that row renders badged **"No live application"** while no such chip exists to filter on.

Reachable whenever a designated household's application gets archived: designation added, family applies, application abandoned, board archives it.

What that costs a board member: they see a status, look for the chip, there isn't one. They flip *Show inactive* to get the chip and click it — now the result set also includes the rows that had been hidden, so the chip doesn't isolate what they were looking at. Flipping back off silently drops INACTIVE from `statuses` (your `onChange` guard, which is correct on its own — it prevents a filter that matches nothing).

Either gate the chip on a row being present:

```js
.filter((c) => c.value !== "INACTIVE" || showInactive ||
               rows.some((r) => r.status === "INACTIVE" && r.designations.length > 0))
```

…or drop the `.filter(…)` entirely. An unmatched chip is harmless, `statuses` already gets cleaned on toggle-off, and that's one fewer invariant to keep in sync with the row predicate.

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

jee7s:
**Area of concern: the standing changes-requested is stale, but a different blocker is live.**

Findings 1-4 were addressed in `8cb2556f` and `c6b69d9d` — `canonicalizeEmail`, the `isHouseholdLead` filter, multi-designation rows, the BLOCKED state and per-email Remove labels. That review can be cleared.

What blocks now is `Run Tests (unit)`: `✗ [new-route-old-authz] … GET /api/membership-ops/volunteer-memberships is not in the security registry and not in the frozen legacy baseline`. The sibling membership-ops routes are not a precedent — they are grandfathered in `scripts/legacy-authz-routes.txt`, which may only shrink.

#1459 is the registry-first entry for this endpoint. Merge that, then convert this route to `defineRoute()` + `handler()`. Also behind main.

---
_Posted by Claude._

jee7s:
Took this over on `feat/volunteer-roster-1374` (jee7s' call — same-repo branch, merge not rebase, no force-push). This revision is mechanical repair only; no behaviour beyond the two minors changed.

## 1. Merged `origin/main` in — 189 commits, one conflict

Only `volunteer-memberships/page.tsx` conflicted. `pageRegistry.ts` and `authzRoleRejection.integration.test.ts` both auto-merged, and the net diff against `main` is exactly the eight feature files, additive — no main-side work reverted.

Three main-side changes needed adapting, not just resolving:

- **#1585 vocab.** The conflicting hunk was the card blurb. Resolved to keep the branch's subject *and* main's vocabulary: "Households on the volunteer membership fee, plus emails designated ahead of signup. If one of those emails applies for membership, that whole household is treated as a volunteer only family (lower membership fee)." The branch's "volunteer-only dues" is gone; `grep -i dues` over the diff is clean, comments included.
- **`formatDate` now requires an explicit `timeZone`.** The designation instant goes through `useOrgTime()` — the client seam — instead of the raw `lib/time` formatter. `memberSince` stays on `formatDateOnly`; it is a `@db.Date` calendar day.
- **`1149_DATE_TIME_TZ_DESIGN.md` was deleted** (#1521). The comment citing it now points at `lib/time.ts`, where the rule actually lives.

## 2. The `new-route-old-authz` blocker is cleared

#1459 merged yesterday with the `defineRoute()` entry for `GET /api/membership-ops/volunteer-memberships`, so `check-route-coverage --strict` is **0 errors** here. The route file arriving also un-orphans that entry. The route keeps `withAuth`, which now scores as the ordinary `unmigrated` warning that ~170 other routes carry — converting to `handler()` is the next step in the sequence you laid out, not this PR.

## 3. Both 2026-07-29T22:45 minors

**(a) Designation rows dropped REVOKED / DENIED.** Confirmed — the branch only tested those two states on the membership branch. Hoisted into a shared `isRevoked`, called from both branches in the membership branch's order (ACTIVE first, then the board decision, then a live process). A designation on a DENIED household now reads **Revoked**, not "Pre-designated" for a family that cannot sign in.

**(b) INACTIVE badge with no chip.** Took your second option and dropped the `.filter(...)` entirely. The unmatched chip is harmless, `statuses` is already cleaned when the switch goes off, and it removes the invariant the chip bar had to keep in sync with the row predicate.

Tests: a parameterised REVOKED/DENIED designation case, plus the chip's presence asserted on the existing inactive-row-with-a-designation test. Both fail against the un-fixed code — checked by reverting each fix in place.

## Local verification

- `npx tsc --noEmit` — clean.
- `npm run check-route-coverage -- --strict` — 0 errors, 171 warnings (all pre-existing).
- `eslint --max-warnings 0` on the touched files — clean.
- `jest` over `membership-ops|nav|sidebar|pageRegistry|authzRegistry|time` — 28 suites, 243 tests, green. The roster's own two suites are 28 tests (26 → 28).
- `authzRoleRejection.integration.test.ts` is **not** runnable here: its `beforeAll` creates real `Household`/`Person`/`Program` rows, so it needs the integration tier's Postgres. It is typechecked locally and runs in CI's integration job, where the branch's one table row exercises the roster's 401/403.

The blocking review predates `8cb2556f`/`c6b69d9d`, which you confirmed on-thread — re-requesting so it can be cleared.

jee7s:
@thpr — this one is ready for your re-review. Recap of where it stands: the branch was taken over and mechanically repaired on jee7s' call (merge of `origin/main`, #1585 vocabulary adopted, the `new-route-old-authz` blocker cleared by #1459), and both of your 2026-07-29 minors are fixed — REVOKED/DENIED designations now read **Revoked**, and the INACTIVE chip filter is gone per your second option. CI is fully green and the branch is conflict-free.

The standing changes-requested predates all of that (you confirmed on-thread that findings 1–4 were addressed back in `8cb2556f`/`c6b69d9d`), so it's the only thing holding the PR — a re-review clears it either way.

thpr:
Re-reviewed `98d627b3`. The blocker is properly cleared — I checked rather than took it on the recap: the route is in `src/security/registry.ts:349` via #1459 as a registry-only PR, and it is **not** in `scripts/legacy-authz-routes.txt`, which is the shortcut I'd have objected to (that file's header requires CODEOWNERS sign-off and says "Don't"). Correct sequence. All nine checks green, and `Run Tests (unit)` executing means jest has actually run these tests for the first time on this PR.

Both minors are fixed the way I'd hoped. `isRevoked()` hoisted so the two branches share one board-decision check, with the membership branch's ordering preserved, and `it.each(['REVOKED','DENIED'])` covering it. Chip `.filter(...)` deleted, with the test asserting badge *and* chip together while the switch is off.

I also checked the `591a4174` merge for sweep damage, since it rewrote all four files for the fee vocabulary: no stale `dues`, `memberSince` is `@db.Date` and still non-nullable so `.toISOString()` on the ACTIVE branch holds, `useOrgTime()` sits above the early returns at 84/88 so hook ordering is fine, and the `memberSince`-UTC / `createdAt`-org-zone split survived intact.

---

## One small thing the chip fix left behind

`page.tsx:226-234` — the Switch's `onChange` still carries the guard from before:

```js
onChange={(e) => {
  const on = e.currentTarget.checked;
  setShowInactive(on);
  if (!on) setStatuses((s) => s.filter((v) => v !== "INACTIVE"));
}}
```

That clear made sense when the INACTIVE chip disappeared with the switch — it stopped a selected filter from stranding with nothing to match. Now the chip is always rendered, and with the switch **off** it still matches: `visibleRows` drops INACTIVE rows only when `designations.length === 0`, so the designation-carrying ones survive both predicates. The INACTIVE chip is a working filter in that state.

So the guard now discards a valid selection. Select "No live application", flip the switch on and back off, and the filter silently clears itself.

Dropping the `if (!on)` line is the whole fix — same reasoning that retired the chip `.filter(...)`: one fewer invariant tracking a condition that no longer exists.

Not blocking. Flagging it because it's a leftover of the change I asked for, not something you had before.

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

jee7s:
@thpr — the leftover is gone. `8ec1bc6b` drops the `if (!on)` clear from the Switch's `onChange`, exactly as you reasoned: the chip is unconditional now and stays a working filter with the switch off, so the clear was discarding a valid selection.

Pinned with a toggle-cycle test (select "No live application" on a designation-carrying inactive row → switch on → off → chip stays checked and the filter stays applied). Verified it fails against the pre-fix code — the unchecked chip after the cycle is what it catches — then green with the fix: 2 suites, 29 tests. eslint clean on both files.

That was the last open item; over to you for the formal re-review to clear the standing changes-requested.

thpr (approved):
Approving — every finding from this review thread is closed, and `8ec1bc6b` is the last of them.

The switch-off clear is gone, and the toggle-cycle test earns its place: it asserts the roster actually narrowed before cycling, then asserts both the checkbox state *and* the row set after, so it fails if the clear comes back rather than only if the checkbox resets.

Recap of what landed against the original review:

- Multi-designation rows, so a household's second designated email keeps its own Remove
- Designation matching moved onto `canonicalizeEmail` + `isHouseholdLead`, the same rule `applyVolunteerStatus` uses to grant the fee
- BLOCKED given its own chip without widening the in-flight sets, and no-live-application split out as INACTIVE behind the switch — with the carve-out that keeps a designation-carrying row visible for its Remove
- `NEXT_RENEWAL` naming the designation rather than the household
- `removeDesignation` checking `res.ok`
- `formatDateOnly` / org-zone `formatDate` split, which then survived the fee-vocabulary sweep in `591a4174`
- REVOKED/DENIED hoisted into `isRevoked()` so both branches share the board-decision check
- Every status keeping a chip

The `new-route-old-authz` blocker was cleared the right way — registry entry landed ahead of the route in #1459, and I confirmed it went into `src/security/registry.ts` and *not* into `scripts/legacy-authz-routes.txt`. That ordering was my error to begin with; thanks for absorbing the correction and routing it properly.

Two things deliberately left out of this PR: the empty-state wording is [#1691](https://github.com/innovationtreehouse/checkin/issues/1691), and the disposition of the now-UI-less `GET /api/settings/membership/volunteer-designations` is still @thpr's call — it is `handler()`-migrated and registered on main, so it is maintained surface either way.

`Run Tests (unit)` and `(integration)` were still queued when I looked; this approval is on the code, and the required checks remain the merge gate.

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

PR #1388, opened 2026-07-29
test(membership): split the intake mega-test so it fits the 5s jest timeout

dkaygithub:
`membership page › fills every intake field, adds/removes children, and saves with a secondary parent` fails on `main` with `Exceeded timeout of 5000 ms`. It reproduces in isolation, so it is not contention from a parallel run.

## Root cause

The test is not stuck — it is genuinely too big for the 5s default.

- No unmatched fetch URL, no missing `act()`, no swallowed rejection. Under `--testTimeout=60000` it passes first try, in ~10.5s.
- Step-by-step instrumentation showed the time spread evenly across ~30 interactions at ~300–350ms each. There is no single stall to fix.
- A CPU profile taken inside the test showed no hot spot. It is React dev-build element creation plus jsdom over a ~160-node tree, re-run on **every** `fireEvent` (the intake form's controlled Mantine inputs re-render the whole page), with each following `*ByRole`/`*ByLabelText` re-walking that tree. Measured here: `fireEvent.change` ≈ 260ms, `getByRole(…, {name})` ≈ 90–280ms.

So the fix is to make the test smaller, not to raise the timeout. **No timeout was bumped.**

**This is not an app bug** — no app code changed.

## The change

Split the one mega-test into four, by form section, behind a new `renderIntakeStep()` helper that also reads back the body of the PATCH that "Save progress" sends:

- fills every household and primary-parent field and saves them
- fills every second parent / guardian field and saves them
- adds and removes children, saving only the ones left
- leaves the Home link's navigation unblocked

Coverage is preserved: every intake `onChange` handler, `addChild`/`updateChild`/`removeChild`, and the `onNavigate` handler are all still exercised. Assertions got **stronger** on the way through — the payload checks are now `toMatchObject` over whole objects rather than one field each, and the Home-link case now asserts `fireEvent.click(...) === true` (i.e. `preventDefault` was not called) instead of just clicking and moving on.

A neighbouring comment that referred to the old test by name was updated to point at its replacement.

## Verification

- `src/app/membership/__tests__/page.test.tsx`: **57/57 pass**; slowest new test 3.5s, down from 10.5s.
- `npx tsc --noEmit` clean, `npm run lint` clean.

Independent of the companion PR from `fix/my-household-error-branches-test`, which fixes a second instance of the same pattern — different file, different directory, no shared helper touched (`rtl.tsx` and `jest.setup.js` are untouched by both). The two merge in either order; confirmed conflict-free by merging both into a scratch branch.

### DIFF
--- a/checkin-app/src/app/membership/__tests__/page.test.tsx
+++ b/checkin-app/src/app/membership/__tests__/page.test.tsx
@@ -811,8 +811,8 @@ describe("membership page", () => {
           household: { name: null, emergencyContactName: "Aunt Jo", emergencyContactPhone: "512-555-1234", emergencyContactEmail: null, line1: "1 Treehouse Ln", line2: null, city: "Austin", state: "TX", postalCode: "78701" },
           primaryParent: { id: 1, name: "Pat Parent", email: "pat@example.com", dob: "1980-01-01", allergies: null },
           // Sam has every field null except name/id — exercises the `sec.x ?? ""`
-          // fallback side of hydrate() (the "fills every intake field" test below
-          // already exercises the truthy side with a fully-populated secondary).
+          // fallback side of hydrate() (the "fills every second parent / guardian
+          // field" test below covers the truthy side).
           secondaryParent: { id: 9, name: "Sam Secondary", email: null, dob: null, allergies: null },
           children: [
             { id: 5, name: "Kid A", email: null, dob: "2015-01-01", allergies: "peanuts" },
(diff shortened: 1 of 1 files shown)


dkaygithub:
Companion: #1389 (same root cause, different file). Independent — merge in either order.

thpr:
Approving — verified the coverage claim independently rather than taking it on trust: ran `--coverage` over `src/app/membership/page.tsx` and `src/components/membership/ChildrenListForm.tsx` with the old test and with the four new ones. Identical %Stmts/%Branch/%Funcs/%Lines **and** identical uncovered-line lists on both files. Coverage is preserved exactly. 57/57 pass.

One stale back-reference the sweep missed.

`src/app/membership/__tests__/page.test.tsx:946`, in `omits blank secondary-parent and child fields from the saved payload`:

```
// fallback side of buildPayload() (the previous test covers the truthy side).
```

"The previous test" is now `leaves the Home link's navigation unblocked`, which exercises none of `buildPayload()`. The truthy side moved to `fills every second parent / guardian field and saves them` and `adds and removes children, saving only the ones left`. Worth repointing by name, the same way the `hydrate()` comment at :814 was.

Minor: `renderIntakeStep()` returns `{ fetchMock, savedPayload }`, but none of the four callers use `fetchMock` — `savedPayload` closes over it. Safe to drop from the return.

Non-blocking observation on the premise, since it may matter for how the follow-ups get prioritised: on my hardware the original mega-test runs in **240ms** at the default 5s timeout, not the ~10.5s measured here. So the failure is specific to slower machines and CI (the unit job is a 6m wall) rather than universal. Doesn't change anything about this PR — smaller tests with identical coverage are an improvement regardless — but "fails on `main`" won't reproduce everywhere.

thpr (approved):

PR #1389, opened 2026-07-29
test(my-household): split the error-branch mega-test so it fits the 5s jest timeout

dkaygithub:
`HouseholdPage › covers the remaining server-error and network-error branches (settings, contacts, member edit, lead)` fails on `main` with `Exceeded timeout of 5000 ms`. It reproduces in isolation, so it is not contention from a parallel run.

## Root cause

The test is not stuck — it is genuinely too big for the 5s default.

- No unmatched fetch URL, no missing `act()`, no swallowed rejection. Under `--testTimeout=60000` it passes first try, in ~9.0s.
- The time is spread evenly across ~25 interactions spanning four unrelated write routes, at ~300–350ms each. There is no single stall to fix.
- A CPU profile taken inside the test showed no hot spot. It is React dev-build element creation plus jsdom over a ~160-node tree, re-run on **every** `fireEvent` (controlled Mantine inputs re-render the whole page), with each following `*ByRole`/`*ByLabelText` re-walking that tree. Measured here: `fireEvent.change` ≈ 260ms, `getByRole(…, {name})` ≈ 90–280ms.

So the fix is to make the test smaller, not to raise the timeout. **No timeout was bumped.**

**This is not an app bug** — no app code changed.

## The change

Split the one mega-test into five, one write route each, behind a new `renderWithFailingRoutes()` helper that layers the failing route onto the two shared GET routes:

- address save — client-validation, server, then network failure
- add emergency contact — network failure
- remove emergency contact — server, then network failure
- edit household member — server, then network failure
- promote member to lead — network failure

Every error branch the original covered is still covered. Splitting also made the fixtures legible: the shared household and contact list are now named (`failureHousehold`, `failureLoadRules`) with a comment explaining why two contacts are on file — the last valid contact's Remove button is disabled, so a second one is needed for the remove-failure cases to be reachable at all. The three `let attempt = 0` counters that previously had to coexist in one scope are now one per test.

## Verification

- `src/app/my-household/__tests__/page.test.tsx`: **23/23 pass**; slowest new test 2.2s, down from 9.0s.
- `npx tsc --noEmit` clean, `npm run lint` clean.

Independent of the companion PR from `fix/membership-intake-test`, which fixes a second instance of the same pattern — different file, different directory, no shared helper touched (`rtl.tsx` and `jest.setup.js` are untouched by both). The two merge in either order; confirmed conflict-free by merging both into a scratch branch.

## Follow-up, not in this PR

Two tests in this same file are next in line on slow hardware — `"validates required fields, then adds a member…"` (4.1s) and `"edits a household member: cancels, validates…"` (4.2s). They pass today but have little margin against the 5s limit, and are worth splitting the same way if CI stays flaky.

### DIFF
--- a/checkin-app/src/app/my-household/__tests__/page.test.tsx
+++ b/checkin-app/src/app/my-household/__tests__/page.test.tsx
@@ -470,59 +470,92 @@ describe("HouseholdPage", () => {
     expect(screen.queryByText("Duplicate contact.")).not.toBeInTheDocument();
   });
 
-  it("covers the remaining server-error and network-error branches (settings, contacts, member edit, lead)", async () => {
-    const household = { ...householdData, householdMembers: [...householdData.householdMembers, { id: 12, name: "Casey Smith", email: "", isDeclaredAdult: true }] };
+  // ── Remaining server-error and network-error branches, one write route per
+  //    test ───────────────────────────────────────────────────────────────────
+  // Casey is a promotable non-lead adult; two contacts on file so either row's
+  // Remove is enabled (the last valid contact's is not).
+  const failureHousehold = { ...householdData, householdMembers: [...householdData.householdMembers, { id: 12, name: "Casey Smith", email: "", isDeclaredAdult: true }] };
+  const failureLoadRules = [
+    { url: "/api/household/emergency-contacts", method: "GET", result: { ok: true, body: { contacts: [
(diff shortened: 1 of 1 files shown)


dkaygithub:
Companion: #1388 (same root cause, different file). Independent — merge in either order.

thpr (approved):

PR #1391, opened 2026-07-30
fix(membership): hold every actor to the conflict-of-interest check, sysadmin included

### LINKED ISSUE #1390: No conflict-of-interest check on the sysadmin path for three membership decisions
## Summary

`hasHouseholdConflict` returns `false` whenever `opts.isSysadmin` is true. On
three surfaces the route gate admits `isSysadmin` as an alternative to
`isBoardMember`, so a sysadmin who is not a board member can decide their own
household's case with no conflict-of-interest check.

## The three call sites

| Route | Gate | Decision |
|---|---|---|
| `POST /api/membership-ops/applications/review-override` | `['isSysadmin','isBoardMember']` | Overrides a **blocked** background-check application |
| `POST /api/membership-ops/applications/certify-payment` | `['isSysadmin','isBoardMember']` | Activates a membership with **no payment** |
| `POST /api/membership-ops/households` | `['isSysadmin','isBoardMember']` | Grants or edits membership on any household |

Each resolves its subject household and calls
`hasHouseholdConflict(prisma, actorId, householdId, { isSysadmin: … })`
(`src/lib/membership/review.ts` → `overrideBlocked`,
`src/lib/membership/payment.ts` → `certifyPaymentPlan`, and the households
route's inline guard). The bypass is in `src/lib/conflictOfInterest.ts`.

## Not affected

The finance-ops payment-plan approve/refuse routes gate on `['isBoardMember']`
only, so a non-board sysadmin is rejected before reaching the check. There the
`isSysadmin` argument only has effect for an actor who is both board and
sysadmin.

## Why it matters

These are the decisions org policy reserves to the board: background-check
reviewers must be board members or authorised in writing by the board, and the
remedy for a contested check is an independent three-person appeals committee.
`sysadmin` is an application role with no counterpart in the policy corpus, so
the bypass places it above the board on exactly those decisions.

Mitigations that do exist: every action is audit-logged with a required written
reason, and the board retains the ability to revoke the sysadmin flag.

## Questions for the board, not for whoever picks this up

1. Should the bypass be removed outright, leaving these three decisions to a
   non-conflicted board member?
2. If a break-glass path is wanted, should it require two actors rather than one?
3. Should `isSysadmin` be dropped from the gate on these three routes, so the
   surfaces are board-only like finance-ops already is?

Narrowing the bypass changes who governs the organisation, so the shape of the
fix is a board decision. Please do not pick a variant and implement it without
that decision recorded on the issue.

dkaygithub:
We want to ensure that systems and processes relying on two-party control among board members are not accidentally bypassed by a board member who is also a sysadmin. `hasHouseholdConflict` is the shared guard behind every self-approval check, and it returned `false` immediately whenever a caller passed `isSysadmin` — so a board member holding that flag could decide their own household's case: force-clear a blocked background check, activate a membership with no payment, grant membership outright, or approve and refuse a payment plan. A non-board sysadmin could do the same on the three surfaces whose gate accepts `isSysadmin` in place of `isBoardMember`. This PR deletes the bypass from the primitive, so no actor is exempt regardless of the roles they hold. The `opts` parameter is removed rather than ignored, which turns every stale caller into a compile error and leaves no role-shaped hole to pass `true` into — and `tsc` used exactly that to surface an eighth decision path the issue had not inventoried.

The strongest argument for doing this at the primitive is an asymmetry inside our own code, independent of policy: `attest()` requires `isBackgroundCheckReviewer || isBoardMember` and has never had a sysadmin bypass on its same-household guard, so a bare sysadmin could not cast one of two votes on their own family's background check — yet `overrideBlocked` let them force-clear it outright, the strictly more powerful action. The bypass was inverted relative to the danger. **One consequence reviewers should weigh:** with no escape valve, a case whose only eligible deciders all live in the subject household can no longer be decided at all until the board adds a non-conflicted person. That is now a visible 403 instead of a silent bypass, and is the gap the planned admin auditing / break-glass mechanism is meant to fill — deliberately out of scope here.

Decision: remove the bypass outright (question 1 on the issue), not narrow the route gates (question 3). A non-board sysadmin can still act on *other* households' cases; only the self-approval path closes.

Fixes #1390

---

<details>
<summary><b>Appendix</b></summary>

### The eight decision paths (7 route files)

| Route | Gate | Reached the guard via |
|---|---|---|
| `POST /api/membership-ops/applications/review-override` | sysadmin **or** board | `overrideBlocked` |
| `POST /api/membership-ops/applications/certify-payment` | sysadmin **or** board | `certifyPaymentPlan` |
| `POST /api/membership-ops/households` (`active`) | sysadmin **or** board | inline guard |
| `POST /api/membership-ops/households` (`comingYear`) | sysadmin **or** board | `grantRenewalPayment` → `certifyPaymentPlan` |
| `POST /api/finance-ops/payment-plans` | board only | inline guard |
| `POST /api/finance-ops/payment-plans/refuse` | board only | inline guard |
| `POST /api/finance-ops/membership-payment-plans/refuse` | board only | inline guard |
| `POST /api/finance-ops/membership-payment-plans` | board only | `certifyPaymentPlan` ← **found by `tsc`, not by grep** |

The last row is the one the issue missed: it calls `certifyPaymentPlan` directly, so it never appeared in a `hasHouseholdConflict` search. Deleting the parameter is what caught it.

### Also updated, so client and server can't drift

Four UI mirrors dropped `me?.isSysadmin !== true &&` from their own-household predicate (`membership-ops/applications`, `membership-ops/households`, `finance-ops/payment-plan`, `finance-ops/shopify-holds`). Without this the client would keep offering a button the server now refuses. User-facing copy changed from "a sysadmin must" to "someone outside your household must", which is what is now true.

### Tests

Five tests asserted the bypass as correct behavior and now assert refusal; the route-level ones exercise `{isBoardMember: true, isSysadmin: true}` explicitly. In `membershipBgNonBlocking` the applicant's own household lead is promoted to sysadmin mid-test and still cannot certify or force-clear. That test sets only `isSysadmin`, not `isBoardMember` — the suite has "email all board members" fallbacks that a stray board member would inflate.

### Validation

`tsc --noEmit` clean. `npm run test:ci`: 2393 passed, 3 failed — all three pre-existing, confirmed by stashing this branch and re-running the same suites on clean `main` for an identical result (3 failed / 92 passed, same test names). Two are 5s render timeouts (`program-ops/programs/[id]`, `my-household`); one is a jsdom "navigation not implemented" error from `window.location.href` in `membership/page.tsx:525`. None are files this PR touches. Integration tests were not run locally (shared dev DB); they typecheck and CI covers them.

### Deliberately out of scope

`trusted-adult/service.ts` → `overrideReview` has a structurally identical bypass (`if (!opts?.isSysadmin)` around `isTrustedAdultConflict`, same "a sysadmin must" message). Different primitive, different approval regime, not covered by the #1390 decision — worth its own issue and PR.

</details>

### DIFF
--- a/checkin-app/src/app/__tests__/householdsMembershipAuditAPI.integration.test.ts
+++ b/checkin-app/src/app/__tests__/householdsMembershipAuditAPI.integration.test.ts
@@ -182,19 +182,21 @@ describe('POST /api/membership-ops/households — grant/revoke audit logging', (
         expect((audit!.newData as { status: string }).status).toBe('ACTIVE');
     });
 
-    it("refuses to grant the actor's OWN household (conflict of interest); sysadmin overrides", async () => {
+    it("refuses to grant the actor's OWN household (conflict of interest), sysadmin included", async () => {
         (getServerSession as jest.Mock).mockResolvedValue({ user: { id: boardId, isBoardMember: true } });
         const res = await post({ householdId: boardHouseholdId, active: true });
         expect(res.status).toBe(403);
         const before = await prisma.orgMembership.findUnique({ where: { householdId: boardHouseholdId } });
         expect(before?.status ?? 'NONE').not.toBe('ACTIVE'); // not granted
 
-        // Sysadmin is the deliberate remedy.
-        (getServerSession as jest.Mock).mockResolvedValue({ user: { id: boardId, isSysadmin: true } });
-        const ok = await post({ householdId: boardHouseholdId, active: true });
-        expect(ok.status).toBe(200);
-        const after = await prisma.orgMembership.findUnique({ where: { householdId: boardHouseholdId } });
-        expect(after?.status).toBe('ACTIVE');
+        // Holding sysadmin — alone or alongside board — buys no way through.
+        for (const roles of [{ isSysadmin: true }, { isBoardMember: true, isSysadmin: true }]) {
+            (getServerSession as jest.Mock).mockResolvedValue({ user: { id: boardId, ...roles } });
+            const denied = await post({ householdId: boardHouseholdId, active: true });
+            expect(denied.status).toBe(403);
+            const after = await prisma.orgMembership.findUnique({ where: { householdId: boardHouseholdId } });
+            expect(after?.status ?? 'NONE').not.toBe('ACTIVE');
+        }
     });
 
     it('coming-year grant: no in-flight renewal → 409, no mutation', async () => {
@@ -255,19 +257,22 @@ describe('POST /api/membership-ops/households — grant/revoke audit logging', (
         expect(process?.paidAt).not.toBeNull();
     });
 
-    it("refuses coming-year grant for the actor's OWN household (conflict of interest); sysadmin overrides", async () => {
+    it("refuses coming-year grant for the actor's OWN household (conflict of interest), sysadmin included", async () => {
         (getServerSession as jest.Mock).mockResolvedValue({ user: { id: boardId, isBoardMember: true } });
         const res = await post({ householdId: boardHouseholdId, comingYear: true, reason: 'Renewal grant test' });
         expect(res.status).toBe(403);
         const stillPending = await prisma.orgMembershipProcess.findUnique({ where: { id: boardProcessId } });
         expect(stillPending?.status).toBe('PENDING_PAYMENT');
 
-        // Sysadmin is the deliberate remedy — same renewal, now grantable.
-        (getServerSession as jest.Mock).mockResolvedValue({ user: { id: boardId, isSysadmin: true } });
-        const ok = await post({ householdId: boardHouseholdId, comingYear: true, reason: 'Renewal grant test' });
-        expect(ok.status).toBe(200);
-        const settled = await prisma.orgMembershipProcess.findUnique({ where: { id: boardProcessId } });
-        expect(settled?.status).toBe('ACTIVE');
+        // The comingYear branch reaches the same guard through grantRenewalPayment →
+        // certifyPaymentPlan, so sysadmin gets no separate door here either.
+        for (const roles of [{ isSysadmin: true }, { isBoardMember: true, isSysadmin: true }]) {
+            (getServerSession as jest.Mock).mockResolvedValue({ user: { id: boardId, ...roles } });
+            const denied = await post({ householdId: boardHouseholdId, comingYear: true, reason: 'Renewal grant test' });
+            expect(denied.status).toBe(403);
+            const unsettled = await prisma.orgMembershipProcess.findUnique({ where: { id: boardProcessId } });
+            expect(unsettled?.status).toBe('PENDING_PAYMENT');
+        }
     });
 
     it('coming-year grant: missing reason → 400, no mutation', async () => {
--- a/checkin-app/src/app/__tests__/membershipBgNonBlocking.integration.test.ts
+++ b/checkin-app/src/app/__tests__/membershipBgNonBlocking.integration.test.ts
@@ -321,7 +321,7 @@ describe('background check is non-blocking', () => {
         expect(await statusOf(processId)).toBe('PENDING_BG_REVIEW'); // held for the note
     });
 
-    it('conflict of interest: a certifier in the applicant household is blocked; sysadmin overrides', async () => {
+    it('conflict of interest: a certifier in the applicant household is blocked, sysadmin flag or not', async () => {
         const { processId, leadId } = await makeApplicant('PENDING_EXTERNAL_ACTION');
         await markContractSigned(processId);
         await markBgConsent(processId, revA);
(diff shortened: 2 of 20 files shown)


thpr (changes requested):
The approach is right. Deleting the `opts` parameter instead of ignoring it is what turned every stale caller into a compile error, and it's what surfaced the 8th decision path (`finance-ops/membership-payment-plans` → `certifyPaymentPlan` direct) that a `hasHouseholdConflict` grep couldn't see. The doc comment telling the next person not to re-add a role-shaped parameter is in the right place.

Verified independently against the head commit: no `hasHouseholdConflict` caller passes a 4th argument, no `isSysadmin !== true` survives in any UI, and the branch merges clean into `main` (7 commits behind).

Requesting changes for one item — the second and third are judgment calls.

---

### 1. Stale copy the sweep missed — `membership-ops/applications/page.tsx`

Two strings still name sysadmin as the remedy, on the exact two paths this PR closes:

- `page.tsx:395` — "You can't certify your own household's application — another board member (or a sysadmin) must."
- `page.tsx:422` — same, for override.

The `ownHousehold` predicate above them was updated; these weren't. A grep for `"a sysadmin must"` doesn't match `"(or a sysadmin) must"`, which is why they slipped past the sweep that caught the other four surfaces. These strings now render for a same-household sysadmin too, telling them to go find someone who also can't do it. Align with the wording used everywhere else in the diff:

```
— someone outside your household must.
```

### 2. The replacement unit test asserts arity, not behaviour — `conflictOfInterest.test.ts:30`

```js
expect(hasHouseholdConflict).toHaveLength(3); // db, actorId, subjectHouseholdId
```

`Function.length` stops counting at the first parameter with a default value, so this passes silently if someone reintroduces `opts: { isSysadmin?: boolean } = {}` — the exact regression it exists to catch. It only catches the bare-`opts?` form, which `tsc` already catches harder, at every call site, with a better error message.

Suggest either dropping it (the deleted-parameter guarantee is a compile-time property, and compile-time is where it's enforced) or replacing it with the assertion that actually has teeth: same household → `true`, regardless of what's on the actor row. Not blocking — it's a no-op test, not a wrong one.

### 3. Nit: single-field opts bag

`certifyPaymentPlan(processId, actorId, opts?: { reason?: string })` is down to one field and could be a plain `reason?: string`. Probably not worth the extra signature churn in this PR — flagging only so it's a decision rather than an oversight.

---

### Follow-ups filed, not blocking

- #1392 — `finance-ops/membership-payment-plan/page.tsx` drives both routes whose gate changed here but has no own-household predicate at all. Pre-existing, degrades acceptably (the 403 surfaces as a red notification), and adding the mirror needs `householdId` in a registered route's response — so it's a registry PR first, per the boundary-isolation rule. It's just the only queue surface left unaligned after this PR.
- #1393 — `trusted-adult/service.ts` `overrideReview` keeps the structurally identical bypass and the "a sysadmin must" copy. Agreed it's out of scope (different primitive, different approval regime, not covered by the #1390 decision); filed so the codebase doesn't sit indefinitely on two answers to the same question.

### Tests

Right shape. The route-level tests exercise `{isBoardMember: true, isSysadmin: true}` explicitly, which is the combination #1390 was actually about. The `membershipBgNonBlocking` change is the strongest of the five — promoting the applicant's own lead to sysadmin *in the database* mid-test proves the guard reads the household relationship rather than caller-supplied roles. Setting only `isSysadmin` there, to avoid inflating the suite's "email all board members" fallback sets, is a real trap correctly avoided.

One gap: nothing asserts the new 403 message text, which is the drift class item 1 belongs to. Low priority; the copy isn't load-bearing.

The lockout consequence raised in the description isn't a concern in practice — the board isn't closely enough related for a case to have no eligible decider.

dkaygithub:
All three addressed, plus a merge with `main` that needed a decision.

**1. Stale copy — fixed** (8acaafd2). Both strings in `membership-ops/applications/page.tsx` now read "someone outside your household must." Your point about why the sweep missed them is the useful part: the grep was for `"a sysadmin must"` and these were `"(or a sysadmin) must"`. Re-swept on the substring `sysadmin` alone across `src` — the only survivor is `trusted-adult/service.ts:490`, which is #1393.

**2. Arity test — replaced** (ecc4b3d5). You're right that it was a no-op: `Function.length` stops at the first defaulted parameter, so `opts: {...} = {}` — the likelier way someone re-adds it — sails past. Dropped it for the assertion with teeth:

```js
it("still conflicts when the actor holds every privileged role", async () => {
    const privileged = { isSysadmin: true, isBoardMember: true, isBackgroundCheckReviewer: true, isKeyholder: true };
    expect(await hasHouseholdConflict(db(3, privileged), 1, 3)).toBe(true);
});
```

The actor-row flags are the load-bearing part — it fails if anyone reintroduces a role read inside the primitive, which the arity check couldn't see at all. Compile-time enforcement of the deleted parameter stays where you said it belongs: `tsc`, at every call site.

**3. Single-field opts bag — keeping it.** Calling it as a decision rather than an oversight: `certifyPaymentPlan` has four non-test call sites, three of which pass `{ reason }`, and a bare positional `reason?: string` reads worse at those sites than the named field. The bag also has somewhere to go — the break-glass/audit mechanism this PR defers is the kind of thing that lands as a second field. Not worth the churn here.

**Merge with `main`** — one conflict, in `overrideBlocked`. #1412 landed on the same signature: it widened `reset` to reach an in-progress review as well as a BLOCKED one, added the doc comment explaining why `approve` stays BLOCKED-only, and its side of the merge still carried `opts?: { isSysadmin?: boolean }`. Took #1412's contract and comment, dropped the parameter. Verified after the merge that no caller passes a 4th argument and no `isSysadmin` read survives near any conflict guard.

`tsc --noEmit` clean. `--testPathPatterns 'conflictOfInterest|membership'`: 40 suites, 373 tests, all passing.

thpr (approved):
Both blocking items are resolved at `ecc4b3d5`. Approving.

**Item 2 (the arity test) — fixed, and better than what I suggested.** Widening the `db` helper to stamp arbitrary fields on the actor row, then asserting against `{isSysadmin, isBoardMember, isBackgroundCheckReviewer, isKeyholder}`, tests the property that actually matters: the guard reads the household relationship and nothing else on the row. The commit message naming the precise hole — `Function.length` stopping at the first defaulted parameter, so a reintroduced `opts = {}` sails through — is the right thing to leave behind for whoever reads that test next.

**Item 1 (stale copy) — resolved, though by the merge rather than by this branch.** `8acaafd2` / #1412 brought the two `membership-ops/applications/page.tsx` strings to "— someone outside your household must." Worth knowing they were fixed elsewhere concurrently rather than here, in case that copy comes up again.

**Item 3 (single-field opts bag)** — left as is, which was the suggestion.

Re-verified the conflict-of-interest surface against the post-merge head: six `hasHouseholdConflict` call sites, all three-argument; no `isSysadmin` reintroduced in `conflictOfInterest.ts`, `payment.ts`, or `trusted-adult/service.ts`; no `"(or a sysadmin) must"` anywhere in `src/`.

One correction to my earlier review. I flagged `trusted-adult/service.ts` as still carrying the bypass and filed #1393 for it. That was read off this branch while it sat 7 commits behind `main` — #1410 had already landed the same fix on `main`, and it closes #1393. So the codebase is not sitting on two answers; `overrideReview` lost its `opts` parameter and carries the matching copy. #1392 (the membership payment-plan queue's missing client-side mirror) I re-checked against current `main` and it does still stand.

PR #1395, opened 2026-07-31
security(boundary): register PATCH/DELETE /api/attendance/manual/[id] (registry-first)

jee7s:
Registry entries for the two self-correction routes #1357 adds, landed registry-first per the boundary isolation rule. Boundary files only — no app code.

**Depends on #1394.** Until that merges, `check-route-coverage --strict` hard-errors `orphan-registry` on these entries (the route file does not exist on `main` yet), so this PR's own unit job stays red. #1394 makes that case a warning while keeping a hard error for the genuinely stale case. Merge order: **#1394 → this → #1357**.

## The entries

Both are self-scoped writes on the caller's own `Visit`. Ownership is the admission gate — the route resolves the visit by id and 404s anything that is not the caller's own live row, with no existence oracle — and `their_own` (bound to `Visit.personId`) is the per-row backstop.

- **`PATCH`** edits the times and returns the visit: `envelope: 'visit'`, `returns: ['Visit']`.
- **`DELETE`** tombstones the row and returns `{ success, flagged }` — no model data, so no `returns`.

The grant stops at `their_own:personal`. `arrivedAt`/`departedAt` are the only `personal`-tier columns and the member authored them; `deletedAt`/`deletedById` are `internal` and stay stripped.

Both routes keep `withAuth` and are **not** migrated to `handler()`. Registered-but-unmigrated is an established state here — a dozen existing entries are absent from `migrated-routes.txt` — and it matches the sibling `POST /api/attendance/manual`.

Each field was checked against `security/core.ts` rather than assumed: `Envelope` is `string | null` with no closed vocabulary, so `'visit'` is correct because it matches the actual JSON key the route returns; `returns` is optional and `validateRouteGrants` skips an empty one; the `their_own:personal` token is statically typed, so `tsc` enforces it rather than a runtime parse. `returns: ['Visit']` on PATCH is load-bearing, not decoration — it is what makes `their_own` resolvable, and dropping it would silently remove this entry from `delivery.test.ts` coverage.

`DELETE`'s grant is inert today (it returns no model data). Kept for symmetry with `PATCH` so the two entries don't disagree for a future reader — say the word if you'd rather it carry an empty token list.

## Test change (`tests/security/registryAuthz.integration.test.ts`)

This suite generates its cases from the registry and dynamically imports each route's module, so **without this change the PR would go red on integration** — the module does not exist yet in the register-first state.

Worse, it would not have self-healed when #1357 landed. These are the **first non-GET entries in the registry** (all 14 existing are GET), and the "allowed caller" sanity case drives the endpoint with an empty `Request` against a seeded program id: `PATCH` would hit `req.json()` → 400, `DELETE` would hit the ownership check → 404. Both would fail `expect(res.status).toBeLessThan(400)`. That is a trap for every future write route, not just these two.

Two narrow changes:

- **Route-file-absent** — cases that drive the handler use `it.skip` until the file lands. `call()` is untouched, so a file that exists and then fails to import still throws loudly. The static `has a handled authorize gate` case still runs unconditionally, so a register-first entry with an unhandled `authorize` is still caught.
- **The 2xx sanity, for non-GET only** — asserts the response is **not 401/403** rather than `< 400`. A 400 from validation or a 404 from an ownership check both prove the caller cleared authz, which is what that assertion exists to show. Skipping it outright would have dropped the only "an allowed caller actually gets in" signal for every future write route. GET keeps `< 400` unchanged.

Plus a 3-line meta-guard asserting at least one plan resolves on disk — without it, a wrong `__dirname` offset would skip every case and the whole file would pass green.

## Verification

`tsc --noEmit --pretty false` clean (only the 2 pre-existing `@aws-sdk/client-lambda` errors), `npm run lint` exit 0, `npm test -- --testPathPatterns security` exit 0 (17 suites / 619 tests, including this entry's `delivery` coverage).

**Not run: `registryAuthz.integration.test.ts` itself** — no Postgres available in my environment. It compiles and lints, but neither the skip behaviour nor the reworked assertion has been observed executing; CI's `integration-tests` job is the first real exercise. The path arithmetic *was* verified directly (a throwaway unit test from the same directory resolved `/api/profile` → true and `/api/attendance/manual/[id]` → false).

### DIFF
--- a/checkin-app/src/security/registry.ts
+++ b/checkin-app/src/security/registry.ts
@@ -33,13 +33,15 @@ defineRoute({
 });
 
 // Self-correction of the caller's OWN visit: PATCH edits the times, DELETE
-// tombstones the row. Ownership is the admission gate — the route 404s any id
-// that is not the caller's own live row — and their_own (Visit.personId) is the
-// per-row backstop. The grant stops at 'personal' (arrivedAt/departedAt); the
-// 'internal' tombstone fields stay stripped.
+// tombstones the row. `[id]` is a Visit id, NOT a person id, so 'self' cannot
+// express the gate — it compares the id param against auth.user.id. Ownership
+// is enforced in the route body, which 404s any id that is not the caller's own
+// live row; their_own (Visit.personId) is the per-row field backstop. The grant
+// stops at 'personal' (arrivedAt/departedAt); 'internal' tombstone fields stay
+// stripped.
 defineRoute({
     endpoint: 'PATCH /api/attendance/manual/[id]',
-    authorize: 'self',
+    authorize: 'authenticated',
     envelope: 'visit',
     // Bag: { Visit }.
     returns: ['Visit'],
@@ -50,7 +52,7 @@ defineRoute({
 
 defineRoute({
     endpoint: 'DELETE /api/attendance/manual/[id]',
-    authorize: 'self',
+    authorize: 'authenticated',
     // No bag — the response is { success, flagged }, no model data.
     envelope: null,
     orderedView: [
--- a/checkin-app/src/security/registry.ts
+++ b/checkin-app/src/security/registry.ts
@@ -32,6 +32,32 @@ defineRoute({
     ],
 });
 
+// Self-correction of the caller's OWN visit: PATCH edits the times, DELETE
+// tombstones the row. Ownership is the admission gate — the route 404s any id
+// that is not the caller's own live row — and their_own (Visit.personId) is the
+// per-row backstop. The grant stops at 'personal' (arrivedAt/departedAt); the
+// 'internal' tombstone fields stay stripped.
+defineRoute({
+    endpoint: 'PATCH /api/attendance/manual/[id]',
+    authorize: 'self',
+    envelope: 'visit',
+    // Bag: { Visit }.
+    returns: ['Visit'],
+    orderedView: [
+        ['authenticated', ['their_own:personal', 'member', 'public']],
+    ],
+});
+
+defineRoute({
+    endpoint: 'DELETE /api/attendance/manual/[id]',
+    authorize: 'self',
+    // No bag — the response is { success, flagged }, no model data.
+    envelope: null,
+    orderedView: [
+        ['authenticated', ['their_own:personal', 'member', 'public']],
+    ],
+});
+
 defineRoute({
     endpoint: 'GET /api/programs/[id]',
     authorize: 'public',
--- a/checkin-app/tests/security/registryAuthz.integration.test.ts
+++ b/checkin-app/tests/security/registryAuthz.integration.test.ts
@@ -23,6 +23,8 @@
 // this is an integration test that needs the real client to seed personas.
 jest.unmock('@/lib/prisma');
 
+import * as fs from 'fs';
+import * as path from 'path';
 import { allRoutes, type Authorize } from '@/security/core';
 import type { BusinessRole } from '@/types/auth';
 import type { SessionUser } from '@/types/participant';
@@ -72,6 +74,14 @@ const PLANS: RoutePlan[] = Array.from(allRoutes()).map(([endpoint, spec]) => {
 
 function importPathFor(routePath: string): string {
     return `@/app${routePath}/route`;
+}
+
+/** A registry entry may legally precede its route file (register-first: an
+ *  unused defineRoute is inert). Cases that drive the handler are skipped until
+ *  the file lands; a file that exists and then fails to import still throws. */
+function routeFileExists(routePath: string): boolean {
+    const base = path.resolve(__dirname, `../../src/app${routePath}/route`);
+    return ['.ts', '.tsx'].some(ext => fs.existsSync(base + ext));
 }
 
 describe('Registry route admission gates', () => {
@@ -154,21 +164,31 @@ describe('Registry route admission gates', () => {
         return verb(new Request(url, { method: plan.method }), ctx);
     }
 
+    // Guards the skip below: a wrong path would resolve every route as absent
+    // and skip the whole file green.
+    it('resolves route files on disk', () => {
+        expect(PLANS.filter(p => routeFileExists(p.routePath)).length).toBeGreaterThan(0);
+    });
+
     for (const plan of PLANS) {
         describe(plan.endpoint, () => {
+            // Every case using this drives the real handler, so it waits for the
+            // route file. The gate check below is static and always runs.
+            const itServed = routeFileExists(plan.routePath) ? it : it.skip;
+
             it('has a handled authorize gate (route not silently skipped)', () => {
                 expect(plan.gate).not.toBe('unhandled');
             });
 
             // ── unauthenticated ───────────────────────────────────────────────
             if (plan.gate === 'public') {
-                it('admits an unauthenticated caller (public route)', async () => {
+                itServed('admits an unauthenticated caller (public route)', async () => {
                     mockSession.mockResolvedValue(null);
                     const res = await call(plan);
                     expect(res.status).toBeLessThan(400);
                 });
             } else {
-                it('rejects an unauthenticated caller with 401', async () => {
+                itServed('rejects an unauthenticated caller with 401', async () => {
                     mockSession.mockResolvedValue(null);
                     expect((await call(plan)).status).toBe(401);
                 });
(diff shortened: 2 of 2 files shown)


dkaygithub (approved):

thpr:
## Review

The test-side work here is careful and the reasoning about `returns`/`envelope` checks out against `security/core.ts`. One thing in the registry entry is wrong, though, and one framing claim doesn't hold.

### Blocking: `authorize: 'self'` is the wrong gate for this endpoint

`resolveAccess` binds `self` to the resource id param and compares it to the **caller's Person id** (`src/security/access-resolvers.ts:246`):

```ts
case 'self': {
    if (auth.type !== 'session') return { allowed: false };
    const target = params.id ?? params.participantId;
    if (target === undefined) return { allowed: true };
    const targetId = parseInt(target, 10);
    return { allowed: !isNaN(targetId) && targetId === auth.user.id };
}
```

`[id]` on this route is a **Visit** id. So the declared gate compares a visit id against a person id: it admits only on numeric coincidence and rejects every genuine owner.

The precedent this looks like — `GET /api/profile`, the only other `self` entry — has **no** id param, so it lands in the `target === undefined` branch where `self` degenerates to "authenticated, handler scopes to `auth.user.id` itself". That's the opposite half of the branch from this route.

It's inert today because the route stays on `withAuth`, but the registry entry *is* the written policy, and it becomes a live 403-everything bug the moment anyone migrates these verbs to `handler()`. `authorize: 'authenticated'` is the accurate declaration, with the comment saying ownership is enforced in the route body by the 404-on-not-yours lookup — which is exactly what this PR's description already argues.

### The sibling comparison doesn't hold

> it matches the sibling `POST /api/attendance/manual`

`POST /api/attendance/manual` is in `scripts/legacy-authz-routes.txt`, not in the registry. It's grandfathered, not registered-but-unmigrated. The "a dozen existing entries are absent from `migrated-routes.txt`" part is accurate (14 registered, 3 migrated), but every one of those is a **pre-existing** route being brought into the registry.

That distinction matters, because registering without migrating satisfies the `new-route-old-authz` ratchet — whose stated rule is *"New routes must use `defineRoute()` + `handler()`"* — with only half the pair. The stripper never runs, so the `their_own:personal` grant enforces nothing at runtime; and `direct-json` only fires on fully-migrated routes, so the handler can `NextResponse.json` freely. For #1357's brand-new surface that's a real gap, not the established state. It's a joint call with #1357 rather than something to resolve here, but it should be a deliberate one.

### `--strict` isn't armed, so the dependency note is inaccurate

> Until that merges, `check-route-coverage --strict` hard-errors `orphan-registry` on these entries, so this PR's own unit job stays red.

It's green. Root `package.json:21` is `npm -w checkin-app run check-route-coverage`, so CI's `-- --strict` expands to `npm -w checkin-app run check-route-coverage --strict` and the inner `npm run` swallows `--strict` as an npm config. From this PR's own unit job (`91052822115`):

```
npm warn Unknown cli config "--strict". This will stop working in the next major version of npm.
✗ [orphan-registry] … PATCH /api/attendance/manual/[id] has no corresponding route file
✗ [orphan-registry] … DELETE /api/attendance/manual/[id] has no corresponding route file
2 error(s), 174 warning(s)
(advisory mode — exiting 0; pass --strict to fail the build)
```

So #1394 isn't a mechanical prerequisite for this PR. It's still worth landing first — as the prerequisite for actually arming `--strict` (details in #1394).

### Verified, no action

- `envelope`/`returns`: `validateRouteGrants` (`src/security/scopes.ts:245`) does `if (returns.length === 0) continue`, so `DELETE`'s `their_own:personal` grant with no `returns` is skipped, not errored. As described.
- `Visit` tiers: `arrivedAt`/`departedAt` are `personal`, everything else on the model is `public`. `deletedAt`/`deletedById` don't exist on `main` yet — #1357 adds them as `internal`, so the comment will be true once that lands. Nothing here pins it, which is fine for a forward reference; just worth knowing it's unverified until #1357.
- The reworked write assertion holds against the real route: #1357's `PATCH` does `await req.json().catch(() => null)` inside a `try`, so an empty-body drive really does return 400 rather than throwing an unhandled rejection out of `call()`. Good catch on the trap for future write routes — the `< 400` assertion would genuinely have failed there.
- The `resolves route files on disk` meta-guard is the right instinct for a skip driven by a path computation.

thpr:
Follow-up to my note above, on sequencing — this PR matters more than the body claims, and #1394 matters less.

#1357's unit job (`90475399733`) fails on the *other* rule:

```
✗ [new-route-old-authz] …/api/attendance/manual/[id]/route.ts — PATCH … is not in the security registry and not in the frozen legacy baseline
✗ [new-route-old-authz] …/api/attendance/manual/[id]/route.ts — DELETE … same
(2 new-route-old-authz error(s) block even in advisory mode)
##[error]Process completed with exit code 1.
```

`new-route-old-authz` blocks unconditionally by design; `orphan-registry` is swallowed by advisory mode. So the deadlock the two PRs describe is asymmetric:

- route file, no registry entry → blocks today → **#1357**
- registry entry, no route file → advisory → **this PR, green**

This PR is therefore the whole unblocker for #1357 — merge order is **#1395 → #1357**, with #1394 orthogonal (worth landing to arm `--strict`, but not a gate on either). Suggest dropping the "Depends on #1394" framing.

That makes the `authorize: 'self'` issue in my earlier comment the only thing standing between this and merge.

thpr (changes requested):
Requesting changes on one item: `authorize: 'self'` is the wrong admission gate for this endpoint.

`resolveAccess` binds `self` to `params.id ?? params.participantId` and compares it to `auth.user.id` (`src/security/access-resolvers.ts:246`). `[id]` on this route is a **Visit** id, so the declared gate compares a visit id against a person id — it admits only on numeric coincidence and rejects every genuine owner. The precedent it resembles, `GET /api/profile`, has no id param and lands in the `target === undefined` branch where `self` degenerates to "authenticated"; that's the opposite half of the branch from this route.

Inert today because the verbs stay on `withAuth`, but the registry entry is the written policy and it becomes a live 403-everything bug the moment anyone migrates these to `handler()`. `authorize: 'authenticated'` is the accurate declaration, with the comment noting ownership is enforced in the route body by the 404-on-not-yours lookup — which is what the PR description already argues.

Everything else I'd take as-is. Detail on the other points (the legacy-baseline sibling, registered-but-unmigrated vs. the ratchet, and the sequencing correction — this PR unblocks #1357 on its own, #1394 is orthogonal) is in my two comments above.

jee7s:
Fixed in `94be0ee2` — you were right on all three, and the `self` one was a real bug I would have shipped as written.

**The gate.** Both entries are now `authorize: 'authenticated'`, with the comment stating why: `[id]` is a Visit id, so `self` compares it against `auth.user.id` and admits only on numeric coincidence. I verified `access-resolvers.ts:246` matches your reading exactly, including the `target === undefined` branch that makes `GET /api/profile` work — which is precisely the half this route does not land in. Ownership stays where the PR body already argued it lives: the route body's 404-on-not-yours lookup, with `their_own` as the per-row field backstop.

**The sibling claim was wrong** and I have dropped it. `POST /api/attendance/manual` is in `legacy-authz-routes.txt` — grandfathered, not registered-but-unmigrated — and you're right that every registered-but-unmigrated entry is a pre-existing route being brought in. I had conflated the two.

**On `--strict`:** confirmed, and my dependency note was wrong. Root `package.json` re-enters `npm run`, so `-- --strict` is swallowed as an npm config. But the reason #1357 is red is narrower than "advisory mode": `new-route-old-authz` blocks *even in advisory mode* by design (`check-route-coverage.ts` header rule 6, and the log line `2 new-route-old-authz error(s) block even in advisory mode`), while `orphan-registry` does not. So your conclusion holds and is now verified from both sides — **this PR unblocks #1357 on its own, and #1394 is orthogonal.** I have corrected #1394's description accordingly; it stands on its own merits as the prerequisite for arming `--strict`.

**On the register-without-migrate gap** — agreed it is a joint call with #1357 and should be deliberate rather than inherited. My read: the ratchet's intent is `defineRoute()` + `handler()`, and taking only half means the stripper never runs, so `their_own:personal` documents a policy nothing enforces at runtime. For a brand-new PII-bearing surface that is worth closing rather than deferring. I'd rather migrate these two verbs to `handler()` in #1357 than leave a grant that reads as enforcement and is not — but that is more diff in an already-approved PR, so I will follow your call. Raising it there.

jee7s:
Two CI notes on the latest push (`e0a34df7`):

**`isolation` failure was the stale-branch artifact, now cleared.** It named `checkin-app/src/app/programs/[id]/page.tsx` as a boundary-violating companion — a file this PR has never touched. Main had moved one commit ahead (#1401 edits that file) and the check's changed-file diff picked up the drift. Merged `origin/main` in; the file list is back to `registry.ts` + `tests/security/`.

**The integration failure is not from this PR — `main` itself is red.** `programAnnounceRecipients.integration.test.ts` hardcodes program end dates against an Aug 1 membership-year boundary, so it inverted when the UTC clock crossed 2026-08-01 overnight. It fails identically on #1357, which shares no code with this branch. Fix is up as **#1413**; once that lands this PR should be green on its own.

thpr (approved):
Fix confirmed in `94be0ee2`. Both entries are `authorize: 'authenticated'` and the comment states the reason precisely. I checked the knock-on effects and there are none: `planAuthorize` maps `'authenticated'` and `'self'` to the same `session` gate, and `authorizeNeeds` returns `{}` for both — so the registryAuthz case table and the derived `ctxNeeds` are unchanged. Admission is now any authenticated caller with ownership enforced by the route body's 404-on-not-yours lookup, which is what #1357 actually does.

Dropping the `POST /api/attendance/manual` sibling claim was the right call too.

On CI: the only red job is `Run Tests (integration)`, failing on `programAnnounceRecipients.integration.test.ts` › "expired-at-midyear boundary". That is the Aug 1 date-literal breakage on `main`, not this branch — #1413 touches exactly that one file. Not a reason to hold this.

Approving. On the register-without-migrate question you raised: I agree with your instinct and it is worth closing rather than deferring — a `their_own:personal` grant that reads as enforcement while no stripper runs is the kind of thing that gets cited as coverage later. But that belongs in #1357, not here; this PR's job is the registry entries and it does that correctly. Raise it there and I will weigh in.

PR #1401, opened 2026-07-31
fix(programs): never pre-check a participant on the enrollment page

dkaygithub:
## The defect

The enrollment member-select pre-selected a participant on load:

```ts
const me = members.find((p) => p.id === currentUserId);
const def = me && enrollBlock(me).reason === null
  ? me.id
  : members.find((m) => enrollBlock(m).reason === null)?.id;
```

Default to me if I'm eligible, else the first eligible member. On a program with **no upper age limit** an adult clears the age check, so a household lead who opened the page to pay for a child arrived with themselves already ticked. One click on **Complete Enrollment** created a second `PENDING` row — holding a capacity seat, against a payment nobody meant to owe.

## The change

Nobody is pre-checked. Enrolling creates a `PENDING` row and a payment obligation, so each participant has to be an explicit click — including the account holder, the easiest person to enroll by accident.

The empty selection is now a real state the page handles rather than bounces out of:

- Both submit handlers `return` on an empty selection instead of `router.push('/')`. Losing the page you were on is a strange answer to "you haven't picked anyone."
- The scholarship button carries the same `disabled` guard as the enroll button — it didn't before, so it was reachable with nothing selected.
- A hint line ("Check the box next to each person you want to enroll.") keeps the disabled state from being a dead end.
- The `catch` fallback no longer selects the current user either.

### Behavior change worth calling out

This also drops the **one-click resume** for members with a payment still owed. Those rows stay selectable and labelled `(Payment pending — select to finish payment)`, but finishing that checkout is now a deliberate click like any other enrollment. That follows from the same rule — no `PENDING` row without a deliberate click — but it wasn't the reported defect, so it's a deliberate call rather than a side effect. Happy to carve out an exception for `reason === 'pending'` if the one-click resume is worth keeping.

## Tests

Every enrolling test now picks its participants the way a household does, through a `selectMember` helper. Two tests replace the obsolete "falls back to the first eligible member" case:

- nothing is pre-checked and **Complete Enrollment** starts disabled;
- the signed-in adult is not pre-checked on a program with `maxAge: null` — the specific shape that produced the incident.

## Verification

- `tsc --noEmit` — clean.
- The page's unit suite, narrowed with `--testPathPatterns` — **35 passed**, 0 failed.
- `eslint` on both changed files — clean.

Not run: the integration and flow tiers (no local DB / dev server). This change is client-only and no other suite renders this page.

### Unrelated pre-existing failure

`src/app/program-ops/programs/[id]/__tests__/page.test.tsx` fails 2 tests with `Exceeded timeout of 5000 ms`. Confirmed pre-existing — reproduced on a clean tree at 8dbe14c1 with these changes stashed, identical `2 failed, 20 passed`. Same mega-test-vs-5s-timeout class as #1388 and #1389. Left for its own PR.

### DIFF
--- a/checkin-app/src/app/programs/[id]/__tests__/page.test.tsx
+++ b/checkin-app/src/app/programs/[id]/__tests__/page.test.tsx
@@ -31,6 +31,12 @@ function renderPage() {
     return renderWithProviders(<ProgramEnrollmentPage params={params} />);
 }
 
+// Nobody is ever pre-checked, so every enrolling test picks its participants the
+// way a household does. findByLabelText waits out the in-flight household fetch.
+async function selectMember(name: string) {
+    fireEvent.click(await screen.findByLabelText(name));
+}
+
 const household = {
     household: {
         householdMembers: [
@@ -80,6 +86,7 @@ describe("ProgramEnrollmentPage", () => {
         expect(screen.getByText("(Enrolled)")).toBeInTheDocument();
         expect(screen.getByText("(Too young)")).toBeInTheDocument();
 
+        await selectMember("Kid One");
         fireEvent.click(screen.getByRole("button", { name: "Complete Enrollment" }));
         await waitFor(() =>
             expect(notifications.show).toHaveBeenCalledWith(expect.objectContaining({ message: "Successfully enrolled!" })),
@@ -135,9 +142,11 @@ describe("ProgramEnrollmentPage", () => {
         fireEvent.change(screen.getByLabelText(/Emergency contact phone/), { target: { value: "5125550000" } });
         fireEvent.click(screen.getByRole("button", { name: "Save & continue to enroll" }));
 
-        // Back to member-select with the now-enrollable child preselected.
+        // Back to member-select with the now-enrollable child selectable — but
+        // still unchecked, so enrolling them stays a deliberate act.
         await screen.findByText("Which of your household wants to enroll?");
-        expect(await screen.findByLabelText("New Kid")).toBeChecked();
+        expect(await screen.findByLabelText("New Kid")).not.toBeChecked();
+        await selectMember("New Kid");
         fireEvent.click(screen.getByRole("button", { name: "Complete Enrollment" }));
         await waitFor(() =>
             expect(notifications.show).toHaveBeenCalledWith(expect.objectContaining({ message: "Successfully enrolled!" })),
@@ -197,10 +206,12 @@ describe("ProgramEnrollmentPage", () => {
         expect(screen.getByText("(Adult)")).toBeInTheDocument();
         expect(screen.queryByText("(DOB missing)")).not.toBeInTheDocument();
         // PENDING enrollment is a RESUMABLE state, not a dead end: labeled as
-        // payment-pending, preselected, with the enroll button available — NOT
-        // the misleading household-setup affordance.
+        // payment-pending and selectable, with the enroll button available — NOT
+        // the misleading household-setup affordance. Resuming payment is still an
+        // explicit click, so the row is not pre-checked.
         expect(screen.getByText("(Payment pending — select to finish payment)")).toBeInTheDocument();
-        expect(screen.getByLabelText("Only Kid")).toBeChecked();
+        expect(screen.getByLabelText("Only Kid")).not.toBeChecked();
+        expect(screen.getByLabelText("Only Kid")).toBeEnabled();
         expect(screen.queryByRole("button", { name: "Finish setting up your household to enroll" })).not.toBeInTheDocument();
         expect(screen.getByRole("button", { name: "Complete Enrollment" })).toBeInTheDocument();
     });
@@ -219,6 +230,7 @@ describe("ProgramEnrollmentPage", () => {
         fireEvent.click(screen.getByRole("button", { name: "Enroll" }));
         await screen.findByText("Which of your household wants to enroll?");
 
+        await selectMember("Kid One");
         fireEvent.click(screen.getByRole("button", { name: "Pay on Shopify" }));
         // No chargeable variant → persistent error, and NO enrollment is created.
         await waitFor(() =>
@@ -284,6 +296,7 @@ describe("ProgramEnrollmentPage", () => {
         fireEvent.click(screen.getByRole("button", { name: "Enroll" }));
         await screen.findByText("Which of your household wants to enroll?");
 
+        await selectMember("Kid One");
         fireEvent.click(screen.getByRole("button", { name: /request a scholarship or payment plan/i }));
         await waitFor(() =>
             expect(notifications.show).toHaveBeenCalledWith(expect.objectContaining({ message: expect.stringMatching(/Requested! Please check your email/) })),
@@ -310,6 +323,7 @@ describe("ProgramEnrollmentPage", () => {
         fireEvent.click(screen.getByRole("button", { name: "Enroll" }));
         await screen.findByText("Which of your household wants to enroll?");
 
+        await selectMember("Kid One");
         fireEvent.click(screen.getByRole("button", { name: /request a scholarship or payment plan/i }));
         expect(await screen.findByText("Already pending.")).toBeInTheDocument();
     });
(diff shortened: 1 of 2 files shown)


thpr (changes requested):
## Overview

Root cause is correctly identified. `enrollBlock` only checks age/DOB/enrollment, so on a program with `maxAge: null` an adult passes, and the "default to me if eligible" line silently pre-checked the household lead. One click then created a real `PENDING` row holding a capacity seat against an unowed payment. Deleting the defaulting and making empty selection a handled state is the right size of fix — not a patch on the age check.

Two files, client-only. No API, schema, or security-boundary surface touched. CI green across unit, integration, flow, coverage gate, and isolation.

## What's good

- **`return` instead of `router.push('/')`** on empty selection (`page.tsx:185`, `:238`). Bouncing to `/` for "you picked nobody" was a bug in its own right.
- **Scholarship button gets the same `disabled` guard** as the enroll button. It was genuinely reachable with an empty selection before.
- **Comments follow AGENTS.md** — present tense, state the constraint, no history. The `enrollBlock` PENDING comment stays accurate since selectability is unchanged.
- **Tests were converted, not weakened.** `selectMember` drives the real checkbox instead of asserting on internal state, and the two replacement tests pin the exact incident shape (`maxAge: null` + signed-in adult).

## Blocking

### 1. `Force Enroll (Override)` kept no empty-selection guard — `page.tsx:590`

`requiresOverride` is set after a failed enroll attempt, the alert renders as a sibling of the button stack, and the checkbox group above it stays interactive. So: uncheck everything, click **Force Enroll** → `handleEnroll(true)` hits the new `return` at line 238 and does nothing, with no feedback. Pre-PR it at least navigated somewhere. This PR added the guard to the scholarship button for exactly this reason and skipped the third button:

```tsx
disabled={enrolling || selectedParticipantIds.length === 0}
```

### 2. The `catch` fallback is now a dead end — `page.tsx:160-164`

On a `/api/household` network throw, members becomes `[{ Myself, dateOfBirth: null }]` and the selection is now empty. That branch never calls `setNeedsSetup`, so it stays `false`. On any age-gated program `enrollBlock(Myself)` returns `'dob'` → checkbox disabled → enroll button disabled → and the new hint tells the user to "Check the box next to each person" when there is no checkable box.

Adding `setNeedsSetup(true)` to that catch routes them to the household-setup affordance instead. Error path, low frequency, but this PR is what turned it into a dead end.

## Non-blocking

### 3. On the open question — I'd restore the `pending` carve-out

The stated rule is "no `PENDING` row without a deliberate click," but pre-selecting a pending row creates nothing: the row already exists, the POST 409s, and `aggregateEnrollOutcomes` folds it into the checkout set. The rationale doesn't reach that case, and the removal has a copy cost — the top CTA reads "Continue enrollment" (`page.tsx:491`), then the panel says "Which of your household wants to enroll?" with the hint "Check the box next to each person you want to enroll," shown to someone who is finishing a payment, not enrolling.

Worth noting the old code's pending-preselect was already narrow — it only fired when `def == null`, i.e. nobody new to enroll. Restoring just that branch is one line and leaves the incident fix fully intact, since the defect was `me` being preselected as a *new* enrollment:

```ts
setSelectedParticipantIds(members.filter((m) => enrollBlock(m).reason === 'pending').map((m) => m.id));
```

If you keep the removal instead, make the hint conditional on whether any row is pending.

### 4. Test coverage gap

Nothing pins the scholarship button's new `disabled` state. The tests click it after `selectMember`, so the enabled path is covered, but deleting the `disabled` prop this PR just added wouldn't fail anything. One assertion in the "pre-checks nobody" test, on a `hasPrice` program, closes it.

Minor: `selectMember` is a toggle named as a setter. Fine at current call sites (each name selected once), a trap if anyone reuses it.

## Risk

Low. Client-only, no schema/route/registry surface, no security-boundary implications, no perf impact. The behavior change worth flagging to whoever owns the enrollment UX is #3 — a household with an unpaid balance now needs an extra click to resume checkout.

dkaygithub:
Both blocking items were real. Fixed in a54f30de.

## 1. `Force Enroll (Override)` — fixed as suggested

Correct that the checkbox group stays interactive while the override alert is up, so the button was reachable with an empty selection and did nothing. It now carries the same guard as its two siblings — one predicate across all three buttons:

```tsx
disabled={enrolling || selectedParticipantIds.length === 0 || loadingHousehold}
```

I kept `|| loadingHousehold` rather than the two-term version in your snippet, purely so the three buttons read identically. `loadingHousehold` can't be true when the override alert is showing, so it's inert here.

New test: `disables the override button once the selection is emptied` — drives a 403 `requiresOverride` response, asserts the override button is enabled with a member checked, then unchecks and asserts it goes disabled.

## 2. Solo-fallback dead end — fixed, but not with an unconditional `setNeedsSetup(true)`

The diagnosis is right and this PR is what created the dead end. I implemented it differently than suggested, because `setNeedsSetup(true)` in that catch is too broad in one direction and breaks an existing test:

The existing case `falls back to a solo household entry when /api/household throws` runs against `baseProgram({ minAge: null, maxAge: null })`. With no age gate, `enrollBlock(Myself)` returns `null` — the solo entry is genuinely enrollable, the checkbox is live, and there's nothing to route away from. An unconditional `setNeedsSetup(true)` would hide a working selection behind a setup affordance the user doesn't need, and would fail that test.

The dead end is specifically "the fallback entry is *blocked*," so the catch now runs the same predicate the success path uses at line 142:

```ts
const solo = { id: currentUserId, name: "Myself", dateOfBirth: null };
setHouseholdMembers([solo]);
const reason = enrollBlock(solo).reason;
setNeedsSetup(!(reason === null || reason === 'pending'));
```

Age-gated program → `'dob'` → setup affordance. Ungated → live checkbox, existing test still green.

New test: `offers household setup when the solo fallback can't clear an age-gated program`. Worth noting for anyone reading it — the checkbox group renders *outside* the `needsSetup` ternary, so the disabled `Myself` row is still in the DOM; only the button stack swaps. The test asserts that shape (row present but disabled, no hint text, no enroll button) rather than the row disappearing. I had that wrong on the first pass and the test caught it.

## 3. The `pending` carve-out — not changed here, and your argument is better than mine

You're right that the rationale doesn't reach that case: the row already exists, the POST 409s, and `aggregateEnrollOutcomes` folds it into the checkout set, so pre-selecting it creates nothing. The copy mismatch is real too — "Continue enrollment" → "Which of your household wants to enroll?" → "Check the box next to each person you want to enroll" is the wrong script for someone finishing a payment.

I've left it alone in this push because the call belongs to whoever owns enrollment UX, and this round was scoped to unblocking. It's a one-line restore of the `def == null` branch whenever you want it — say the word and I'll add it plus a test, or make the hint conditional on `pending` if the removal stands.

## 4. Test coverage gap — acknowledged, not addressed

Correct that nothing pins the scholarship button's new `disabled` prop; deleting it would still pass. Not fixed in this push (out of the blocking scope I was given), but it's one assertion on a `hasPrice` program in the "pre-checks nobody" test and I'm happy to add it.

Same for the `selectMember` naming — it is a toggle, and it's safe only because each name is selected once per test. Renaming to `toggleMember` is a mechanical follow-up if you'd rather not leave the trap.

## Verification

- `tsc --noEmit` — clean.
- The page's unit suite — **37 passed**, 0 failed (up from 35; both new tests fail against the pre-fix code).
- `eslint` on both changed files — clean.

thpr (approved):
Both blocking items fixed in a54f30de, and #2 is fixed better than I suggested.

**1. Override button guard** — same predicate as its two siblings. Keeping `|| loadingHousehold` for symmetry is the right call; it's inert while the override alert is up, and one predicate across three buttons reads cheaper than a third variant. The test drives a 403 `requiresOverride`, asserts enabled with a member checked, then unchecks and asserts disabled — both directions, and it fails against pre-fix code.

**2. Solo-fallback dead end** — my `setNeedsSetup(true)` was wrong. It would have broken the existing "renders the solo Myself fallback" case, where an ungated program leaves a genuinely checkable row. Reusing the success path's own enrollable predicate is correct: age-gated → setup affordance, ungated → live checkbox.

Confirmed the closure is sound — the panel only renders past a non-null `program`, so `enrollBlock`'s `if (!program)` early-return can't fire in that catch and silently mark the row enrollable. And the test asserts the real shape (disabled `Myself` row still in the DOM, no hint, no enroll button) rather than the row disappearing, which matches where the checkbox group sits relative to the `needsSetup` ternary.

**3 and 4** — both filed as follow-ups rather than held against this PR: the `pending` carve-out is an enrollment-UX call, and the scholarship-button assertion plus the `selectMember` rename are mechanical.

One observation, not a defect: in the catch branch an already-`ACTIVE` current user resolves to `reason === 'enrolled'` → "Finish setting up your household to enroll," which is odd copy. The success path does the same for an all-enrolled household, so this change is consistent with existing behavior rather than introducing anything new.

Unit, flow, deploy build, isolation, s-read, and kiosk all green. Integration was still running when I approved.

Issue #1400, opened 2026-07-31
Leader view of Roster does not show parent phone #


dkaygithub:
It shows emergency contact number, but it should show parent(s) phone numbers first

jee7s:
PROMOTE — thin as written, but it clears the bug bar: actual (emergency contact number is shown), expected (parents' phone numbers first) and the surface (the leader roster view) are all unambiguous, and there is no product decision open — the ordering question, parents above emergency contacts, is answered in the report itself.

The work is built and stacked in two open PRs. #1402 is the security-boundary half, granting program leads `their_program_households:pii` on `Person` household-lead rows — a parent is not a program participant, so `their_program_participants` never reaches them and `Person.phone` strips to `undefined` regardless of what the route selects. It ships alone and ahead per the registry-first flow, and its own body flags that it widens live traffic on merge. #1403 is stacked on it and closes this issue.

So what remains is not implementation, it is the policy review on #1402 — and that review should be taken together with #1409, which widens what the same lead sees on the same roster tab. Two independent registry grants to the same role in the same week is how a boundary drifts.

---
_Posted by Claude._

PR #1402, opened 2026-07-31
feat(security): let program leads reach the parents' contact details

dkaygithub:
## What & why

The Roster tab of `/program-ops/programs/[id]` shows a child's emergency
contacts but not their parents' phone numbers — #1400. A parent is not a
program participant, so `their_program_participants` never reaches them, and
`Person` had no household-scoped binding at all: a lead holds only `everyones`
on a parent's row, and `Person.phone` (`pii`) strips to `undefined` no matter
what the route selects. Delivering it is a policy widening, so it ships alone
and ahead of the feature, per the registry-first flow.

**This PR widens live traffic on merge** — see the FINDING below. The follow-up
adds the household-members select + the roster UI, but the widening does not
wait for it.

## The grant

`Person.their_program_households`, bound as a conjunction:

```ts
their_program_households: {
    all: [
        { field: 'householdId', inCtx: 'householdIdsInScopePrograms' },
        { field: 'isHouseholdLead', isTrue: true },
    ],
},
```

granted at `pii` on the `programLeadMentor` / `programCoreVolunteer` views of
`GET /api/programs/[id]`.

The `isHouseholdLead` conjunct **is** the policy: only the (at most two) adults
modeled as the family's leads are reachable — siblings and other members of an
in-scope household hold nothing. `isTrue` is a new `Match` kind in the scope
grammar, strict `=== true`, so a row that failed to select the column fails
closed. It carries a `field:`, so `validateBindings`' field-existence check
covers it like every other matcher.

## FINDING for the CODEOWNERS reviewer — what reaches a lead the moment this merges

`GET /api/programs/[id]` already returns **full-row** `Person` selects outside
the participant bag:

- `checkin-app/src/app/api/programs/[id]/route.ts:56` — `volunteers: { where: { person: LIVE_PERSON }, include: { person: true } }`
- `checkin-app/src/app/api/programs/[id]/route.ts:78` — `leadMentor: true`

`stripValue` recurses both as model `Person` and resolves scopes per row, so for
**any program volunteer or lead mentor who is also a household lead of a
household with a child in one of the caller's programs** — a parent volunteer,
the common case — a lead / core-vol now receives:

- `email`, `phone` — `pii`, the grant this PR adds;
- `dateOfBirth`, `allergies`, `notificationSettings`, `emailSuppressed` —
  `personal`, via the **pre-existing** `their_program_households:personal`
  token, which now resolves on those same rows because `Person` is bound to the
  scope for the first time.

Those rows resolved `everyones` only before, i.e. public tier. Nothing gates
this behind the follow-up.

Two ways to take it back to inert, both deliberately **not** taken here:

1. A route PR first narrowing `volunteers.include.person` / `leadMentor` to the
   fields the roster renders. That is a route change, not a boundary one, so it
   cannot ship in this PR under the isolation rule.
2. A purpose-built `their_program_household_leads` scope granted only at `pii`.
   That removes the personal-tier collateral, but parent-volunteer email/phone
   still ships on merge — it shrinks the finding, it does not eliminate it — at
   the cost of a new token in the grammar. Reusing the existing scope name was
   the chosen trade-off.

**Cross-program bleed.** `householdIdsInScopePrograms` is the union across *all*
programs the caller leads/core-vols, while the row set is "any Person in this
program's bag". On program A's page, a lead of both A and B resolves `pii` on a
volunteer of A who leads a household with a child in **B**. Pre-existing scope
semantics, but previously unreachable on this route — EC rows in this bag only
ever came from this program's own households. Noted in the binding comment.

## What this deliberately does NOT widen

`fieldVisible` (core.ts) matches tiers **exactly** — there is no lattice — so a
`pii` token reaches `pii` fields only, on rows where the scope resolves:

- **`Household.intakeNotes` (pii)** — unreachable. `Household` binds no scope
  but `their_households`, so a `their_program_households` token resolves to
  nothing on a Household row. The schema GUARD comment previously said "no
  household-scoped pii token may ever be granted to a lead-facing view that
  returns Household rows"; this PR rewrites it to state the invariant that is
  actually load-bearing — *never bind Household to a program- or
  keyholder-scoped scope* — since the old blanket wording would forbid a change
  that provably does not leak.
- **`Household.line1..postalCode`** — unreachable. `internal`, and no view here
  holds an internal token. Comment updated to stay accurate now that leads also
  get parent contacts.
- **`GET /api/trusted-adults/operational`**, the only other route granting this
  scope, selects `household: { select: { id, name } }` and no Person rows —
  its views are unchanged. `GET /api/safety/trusted-adults` returns Person but
  grants `everyones:*` only.

## Tests

- `src/security/__tests__/household-lead-program-scope.test.ts` (new) — scope
  resolution (parent yes; non-lead same household no; lead of an out-of-scope
  household no; missing `isHouseholdLead` or `householdId` fails closed) and
  field-stripping against the **live registry tokens**, including a positive
  delivery assertion (lead actually receives name/email/phone) and a negative
  that the pii grant does not reach `intakeNotes` or the address.
  Also covers the **`ProgramVolunteer.person` row shape the route already
  ships** — the FINDING above is pinned by a test, not just described: one
  assertion for the pii delivery, one for the personal-tier collateral
  (`dateOfBirth`/`allergies`/`emailSuppressed`), one negative for a volunteer
  who is not a household lead. Plus
  `Object.keys(SCOPE_BINDINGS.Household) === ['their_households']`, which makes
  the rewritten `intakeNotes` GUARD enforceable rather than prose-only.
- `tests/security/scopeBindingsEquivalence.test.ts` — the new binding is
  encoded as an explicit, documented delta in `expectedFromSnapshot` (the
  third), never by editing the frozen oracle. Fixture rows gained
  `isHouseholdLead` so the binding is actually exercised: one positive plus
  both negatives.

`registryAuthz.integration.test.ts` still has no volunteer-row coverage; the
unit-level pins above stand in for it.

`npx tsc --noEmit` clean; the security-scoped jest run is 634/634 green
(18 suites). The full unit suite has 8 failures across 4 page suites, all
5s-timeout flakes under load — 3 pass in isolation and the 4th
(`ProgramDetailsPage > edits general-tab fields and saves settings
successfully`) fails identically on clean `main` at 8dbe14c1.

Design for #1400.

### DIFF
--- a/checkin-app/prisma/schema.prisma
+++ b/checkin-app/prisma/schema.prisma
@@ -265,9 +265,10 @@ model Household {
   /// @sensitivity:public
   name    String
   /// Home address is strict-band intimate data. 'internal' (not 'personal')
-  /// so it sits ABOVE the *:personal grants that deliver emergency contacts to
-  /// program leads (their_program_households:personal) and keyholders
-  /// (keyholders:personal): they get EC, never the family's address. Only
+  /// so it sits ABOVE every household-scoped grant that reaches program leads
+  /// (their_program_households:pii/:personal — emergency contacts plus the
+  /// parents' own contact details) and keyholders (keyholders:personal): they
+  /// get people to call, never the family's address. Only
   /// board/sysadmin (everyones:internal) see it on registered routes; the
   /// household's own view is hand-shaped by the self-household routes.
   /// @sensitivity:internal
@@ -285,10 +286,13 @@ model Household {
   /// volunteer-only household types the note that tells a background-check
   /// reviewer to mark them volunteer. Stays 'pii' (NOT internal like the other
   /// narrative fields) because reviewers hold exactly everyones:pii on
-  /// GET /api/membership/reviews and must read it. GUARD: no household-scoped
-  /// pii token may ever be granted to a lead/keyholder-facing view that
-  /// returns Household rows, or this narrative leaks — leads/keyholders get
-  /// household data via *:personal (EC) only.
+  /// GET /api/membership/reviews and must read it. GUARD: what keeps it away
+  /// from leads/keyholders is that Household binds NO scope but their_households
+  /// (scopeBindings.ts) — a household-scoped token like
+  /// their_program_households:pii resolves to nothing on a Household row, which
+  /// is why GET /api/programs/[id] can grant one for the parents' Person rows
+  /// without exposing this. Never bind Household to a program- or
+  /// keyholder-scoped scope, or the narrative leaks with it.
   /// @sensitivity:pii
   intakeNotes String?
 
--- a/checkin-app/src/security/__tests__/household-lead-program-scope.test.ts
+++ b/checkin-app/src/security/__tests__/household-lead-program-scope.test.ts
@@ -22,6 +22,7 @@
 import { stripValue } from '@/security/stripper';
 import { scopesHeld, type CallerContext } from '@/security/access-resolvers';
 import { getRoute, type Role, type Token } from '@/security/core';
+import { SCOPE_BINDINGS } from '@/security/scopeBindings';
 import '@/security/registry';
 
 function ctx(opts: Partial<CallerContext> = {}): CallerContext {
@@ -137,6 +138,12 @@ describe('Parent field-stripping on GET /api/programs/[id]', () => {
         expect(out.phone).toBeUndefined();
     });
 
+    it('the Household model stays bound to their_households only', () => {
+        // The intakeNotes GUARD in schema.prisma rests on this: a program- or
+        // keyholder-scoped token must never resolve on a Household row.
+        expect(Object.keys(SCOPE_BINDINGS.Household)).toEqual(['their_households']);
+    });
+
     it('the pii grant does NOT reach Household.intakeNotes or the address', () => {
         const tokens = tokensFor(ENDPOINT, 'programLeadMentor');
         const household = { id: 42, name: 'Smith', intakeNotes: 'please call mom first', line1: '1 Main St', city: 'Austin' };
@@ -145,5 +152,59 @@ describe('Parent field-stripping on GET /api/programs/[id]', () => {
         expect(out.intakeNotes).toBeUndefined();
         expect(out.line1).toBeUndefined();
         expect(out.city).toBeUndefined();
+    });
+});
+
+// The route ships `volunteers: { include: { person: true } }` and `leadMentor: true`
+// — full-row Person selects outside the participant bag. A volunteer who is also a
+// household lead of an in-scope household (a parent volunteer) therefore resolves the
+// new binding on merge, with no follow-up. These pin what that delivers.
+describe('Parent volunteers, the shape the route already returns', () => {
+    const parentVolunteerRow = {
+        programId: 100,
+        personId: 500,
+        isCore: false,
+        person: {
+            ...parentRow,
+            dateOfBirth: new Date('1985-04-02'),
+            allergies: 'penicillin',
+            emailSuppressed: false,
+        },
+    };
+
+    it('delivers the parent volunteer contact details to a lead', () => {
+        const tokens = tokensFor(ENDPOINT, 'programLeadMentor');
+        const out = stripValue('ProgramVolunteer', parentVolunteerRow, tokens, lead) as {
+            person: Record<string, unknown>;
+        };
+        expect(out.person.email).toBe('sam@example.com');
+        expect(out.person.phone).toBe('5125551234');
+    });
+
+    it('also delivers their personal tier — the collateral of reusing the scope', () => {
+        // their_program_households:personal is pre-existing on this view; binding
+        // Person to the scope is what makes it resolve here. Narrowing the route's
+        // select is what takes these back off the wire.
+        const tokens = tokensFor(ENDPOINT, 'programLeadMentor');
+        const out = stripValue('ProgramVolunteer', parentVolunteerRow, tokens, lead) as {
+            person: Record<string, unknown>;
+        };
+        expect(out.person.dateOfBirth).toEqual(new Date('1985-04-02'));
+        expect(out.person.allergies).toBe('penicillin');
+        expect(out.person.emailSuppressed).toBe(false);
+    });
+
+    it('a volunteer who is not a household lead keeps public tier only', () => {
+        const tokens = tokensFor(ENDPOINT, 'programLeadMentor');
+        const row = {
+            ...parentVolunteerRow,
+            person: { ...parentVolunteerRow.person, isHouseholdLead: false },
+        };
+        const out = stripValue('ProgramVolunteer', row, tokens, lead) as {
+            person: Record<string, unknown>;
+        };
+        expect(out.person.name).toBe('Sam Smith');
+        expect(out.person.email).toBeUndefined();
+        expect(out.person.allergies).toBeUndefined();
     });
 });
--- a/checkin-app/src/security/__tests__/household-lead-program-scope.test.ts
+++ b/checkin-app/src/security/__tests__/household-lead-program-scope.test.ts
@@ -37,6 +37,7 @@ function ctx(opts: Partial<CallerContext> = {}): CallerContext {
         householdIdsInScopePrograms: new Set(),
         eventIdsInScopePrograms: new Set(),
         activeVisitorIds: new Set(),
+        ledHouseholdMemberIds: new Set(),
         ...opts,
     };
 }
--- a/checkin-app/src/security/__tests__/household-lead-program-scope.test.ts
+++ b/checkin-app/src/security/__tests__/household-lead-program-scope.test.ts
@@ -84,13 +84,13 @@ describe('Person.their_program_households scope resolution', () => {
         expect(scopesHeld('Person', outsiderParentRow, lead).has('their_program_households')).toBe(false);
     });
 
-    it('fails closed when the row omits isHouseholdLead', () => {
-        const { isHouseholdLead: _omitted, ...noFlag } = parentRow;
+    it('fails closed when the select omitted isHouseholdLead', () => {
+        const noFlag = { id: 500, householdId: 42, name: 'Sam Smith', phone: '5125551234' };
         expect(scopesHeld('Person', noFlag, lead).has('their_program_households')).toBe(false);
     });
 
-    it('fails closed when the row omits householdId', () => {
-        const { householdId: _omitted, ...noHousehold } = parentRow;
+    it('fails closed when the select omitted householdId', () => {
+        const noHousehold = { id: 500, isHouseholdLead: true, name: 'Sam Smith', phone: '5125551234' };
         expect(scopesHeld('Person', noHousehold, lead).has('their_program_households')).toBe(false);
     });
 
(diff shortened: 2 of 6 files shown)


thpr (changes requested):
## Review

The grant itself is well-built; the problem is the inertness claim. The registry/scope diff is boundary-only and complies with the isolation rule.

## Blocking: the PR is not inert

`GET /api/programs/[id]` already returns Person rows with **full-row selects**, beyond the participants:

- `checkin-app/src/app/api/programs/[id]/route.ts:56` — `volunteers: { where: { person: LIVE_PERSON }, include: { person: true } }`
- `checkin-app/src/app/api/programs/[id]/route.ts:78` — `leadMentor: true`

`stripValue` recurses both as model `Person` and resolves scopes per row (`checkin-app/src/security/stripper.ts:84`; relation map `ProgramVolunteer.person` in `generated/classifications.ts:489`). So on merge, with no follow-up, a lead / core-vol viewing the program receives — for **any program volunteer or lead mentor who is a household lead of a household with a child in one of the caller's programs** (a parent volunteer, the common case):

- `email`, `phone` (pii — the new grant)
- `dateOfBirth`, `allergies`, `notificationSettings`, `emailSuppressed` (personal — the pre-existing `their_program_households:personal` token, which now resolves on these rows)

Today those rows resolve `everyones` only, so the caller gets public tier. This is exactly the collateral the FINDING says "the route's select is what keeps off the wire" — but that sentence describes the *follow-up's* select for the new household-members bag, not the selects already in the route. Nothing gates it behind the follow-up: the widening is live the moment this merges.

Options:

1. Land a route PR first narrowing `volunteers.include.person` / `leadMentor` to the fields the roster UI renders. Then this PR is genuinely inert and the body's claim holds.
2. Take the alternative you considered and rejected — a purpose-built `their_program_household_leads` scope granted only at `pii`. Removes the personal-tier collateral entirely; parent-volunteer email/phone still ships on merge, so the FINDING needs rewording, not deleting.
3. Keep as-is and rewrite the FINDING to state what actually reaches leads today. Reviewer's call — but it must not read "inert".

## Secondary

- **Cross-program bleed.** `householdIdsInScopePrograms` is the union across *all* programs the caller leads/core-vols (`access-resolvers.ts:110-118`), while the row set is "any Person in this program's bag". On program A's page, a lead of A and B now resolves `pii` on a volunteer of A who leads a household with a kid in **B**. Pre-existing scope semantics, but previously unreachable — EC rows in this bag only ever came from this program's own households. Worth a line in the binding comment.
- **Test gap.** The new suite covers parent / sibling / outsider rows and the Household negative well, but nothing exercises a `ProgramVolunteer.person` or `leadMentor` row — the two shapes the route ships today. That is why the suite stayed green; `registryAuthz.integration.test.ts` has no volunteer-row coverage either. Add both (parent-volunteer positive, plus an assertion pinning whether `dateOfBirth`/`allergies` are delivered).
- **The rewritten `intakeNotes` GUARD is prose-only.** "Never bind Household to a program- or keyholder-scoped scope" is now the load-bearing invariant. One line in the security suite makes it enforceable: assert `Object.keys(SCOPE_BINDINGS.Household)` ⊆ `['their_households']`.

## What's good

- `isTrue` is the right minimal grammar addition: strict `=== true` so an unselected column fails closed, carries `field:` so the `allMatches` recursion (`scopes.ts:132`) puts it under `validateBindings`' field-existence check, and the header doc block was updated alongside.
- Equivalence oracle handled correctly — new binding encoded as a documented delta in `expectedFromSnapshot`, frozen snapshot untouched, fixture rows gained the flag so one positive and both negatives actually run.
- `Person.isHouseholdLead` is `public` tier, so using it as the scope discriminator leaks nothing itself.
- `GET /api/trusted-adults/operational` verified clear: `returns: ['TrustedAdult', 'Household', 'TrustedAdultReview']`, no Person.

Findings above come from reading the route, stripper, and resolvers — tests were not run in the review checkout.

dkaygithub:
Confirmed — you're right, and I took **option 3**: the grant stands, the inertness claim is gone.

**Blocking — "the PR is not inert"** (a06cd0a0)

- Title dropped `(inert)`.
- PR body: the old "**This PR is inert**" line is replaced with "**This PR widens live traffic on merge**", and the FINDING section is rewritten around `volunteers.include.person` / `leadMentor` — naming both full-row selects, who the affected rows are (a parent volunteer), and the two tiers that land (`email`/`phone` from the new pii grant; `dateOfBirth`/`allergies`/`notificationSettings`/`emailSuppressed` from the pre-existing `:personal` token that now resolves there). It records that those rows held `everyones` only before, and lists your options 1 and 2 as deliberately-not-taken, with why.
- Same correction in `registry.ts` — the old comment claimed "the route's select is what keeps them off the wire", which described the follow-up's select, not the ones already in the route.

**Secondary**

- **Cross-program bleed** — noted in the binding comment in `scopeBindings.ts`, and called out as its own paragraph in the body.
- **Test gap** — fixed (c25eb8c9). `household-lead-program-scope.test.ts` gained a `Parent volunteers, the shape the route already returns` block over a `ProgramVolunteer` row with a nested full-row `person`: pii delivery, the personal-tier collateral pinned explicitly (`dateOfBirth`/`allergies`/`emailSuppressed`), and a non-lead volunteer negative. The FINDING is now enforced by a test rather than only described — a later narrowing of the route's select has to come update it. `registryAuthz.integration.test.ts` still has no volunteer-row coverage.
- **`intakeNotes` GUARD prose-only** — added your assertion: `expect(Object.keys(SCOPE_BINDINGS.Household)).toEqual(['their_households'])`.

`npx tsc --noEmit` clean; the security-scoped run is 634/634 across 18 suites.

The route-narrowing (your option 1) is still the right end state, but it's a route change and can't ship in this PR under the isolation rule.

thpr (changes requested):
The rewrite is accurate now — thanks, the FINDING, the bleed note, and the volunteer-row pins all land. But option 3 alone isn't enough to merge on: documenting a leak doesn't stop it. Still requesting changes.

To be clear about what's blocking: `dateOfBirth`, `allergies`, `notificationSettings`, `emailSuppressed` for adult volunteers and lead mentors reach program leads on merge, and nothing in either PR takes them back off. I checked #1403 — its route change is purely additive (`household.householdMembers`); `volunteers.include.person` and `leadMentor` are untouched there. So the end state after both land still leaks, with no PR scheduled to fix it.

What I need before approving: a **definitive, merge-ordered plan**, not a note that narrowing is "the right end state". Any of these works.

**A. Narrowing PR first, this one rebases onto it.** Route-only diff, no UI change — the frontend types already tell you the full surface:

- `program-ops/programs/[id]/page.tsx:41` — `volunteers: { personId, isCore, person: { name, email } }[]`
- `program-ops/programs/[id]/page.tsx:43` and `programs/[id]/page.tsx:23` — `leadMentor: { name, email } | null`

so:

```ts
volunteers: { where: { person: LIVE_PERSON }, include: { person: { select: { id: true, name: true, email: true } } } },
leadMentor: { select: { id: true, name: true, email: true } },
```

(keep `id` — it is the `their_own` scope key for the `authenticated` view.) That reduces the FINDING to "a lead also sees a parent-volunteer's email", which is a defensible widening rather than a medical/DOB one.

**B. The purpose-built `their_program_household_leads` scope**, granted only at `pii`. Boundary-only, so it ships inside this PR under the isolation rule, and it removes the personal-tier collateral outright. Email/phone on volunteer/mentor rows still widen, so pair it with A eventually — but the sharp edge is gone on merge.

**C. Block the merge on A.** If you'd rather keep the scope reuse, open the narrowing PR now, land it, then merge this. A "do not merge until #NNNN" marker plus the linked PR is a plan; a paragraph saying it can't ship here is not.

My preference is B in this PR plus A as its own route PR before #1403 — that ordering leaves no window where the personal tier is live.

One correction while you're in here: the claim in your reply that "a later narrowing of the route's select has to come update it" doesn't hold. `household-lead-program-scope.test.ts` builds its own `parentVolunteerRow` literal, so narrowing the Prisma select leaves that test green — it pins the boundary policy, not the route's wire shape. Nothing in the suite asserts the select. Worth saying so in the test comment, otherwise it reads as a guard it isn't.

Minor, non-blocking: `leadMentor` has no test case despite being the same exposure path, and the pinned row omits `notificationSettings` while the FINDING lists it.

dkaygithub:
The 4 pieces of data listed as blocking I do not think are worthy of blocking 

1. Volunteer dob: we've changed to have Boolean isAdult, so not a real leak
2. Allergies: appropriate exposure, if volunteer has allergic reaction, program leader should have been warned
3. Notification settings: who cares
4. Email notification settings: who cares

thpr:
My concern is this is a latent issue waiting to pounce.

Two concerns: (1) An audit of checkin for this form of broad select turned up two sources both leaking PII in a way that is not desirable, see #1426 #1425 .

(2) This creates a mechanism that ignores our declared security surface: Any Person column someone later tags pii or personal lands on the program roster automatically — no route change, no registry change, no CODEOWNERS review of this route, and no test breaks. The schema PR author is making a tiering decision, not a "should program leads see this" decision, and **nothing in the diff tells them the two are the same decision here**. 

An explicit select inverts that — a new column is invisible until someone edits the route, so the exposure decision happens at the route with the security context in view. That's the real reason this isn't just cleanup in my opinion.

thpr (approved):
Clearing my changes-request: the rebase resolved it, though not by anything in this PR.

The new base (`8a420411`) contains #1425, which narrowed `GET /api/programs/[id]` to an explicit select. At your base the route now reads:

- `volunteers.person: { select: { id, name, email } }`
- `leadMentor: { select: { id, name, email } }`
- `participants.person: { select: { id, name, email, phone, householdId, household: {...} } }`

None of those carries `isHouseholdLead`. The binding is a conjunction, so it resolves on **no row this route returns**, and `GET /api/programs/[id]` remains the only Person-returning registered route granting the scope — `GET /api/events/[id]` doesn't grant it, `GET /api/trusted-adults/operational` returns no Person, and `GET /api/safety/trusted-adults` selects household members without the flag or `householdId`. The volunteer/lead-mentor exposure I blocked on cannot happen. Approving.

**One thing to fix before merge, docs-only but load-bearing.** The FINDING is now false, and it is the prose a CODEOWNERS reviewer reads as the policy record. `registry.ts` still says:

> this widens live traffic. The route already returns full-row Person selects outside the participant bag — `volunteers.include.person` and `leadMentor`

That shape does not exist at this PR's own base. Same for the test block titled `Parent volunteers, the shape the route already returns` — it no longer is.

Please restate both around the real reason this is inert, which is a better guard rail than either version we've had:

> No Person select on a route granting `their_program_households` carries `isHouseholdLead`, so the binding resolves nowhere today. Adding that flag to any such select turns BOTH the `:pii` and the pre-existing `:personal` token on for that row — so a select that adds it must stay narrow, or `dateOfBirth`/`allergies`/`notificationSettings`/`emailSuppressed` ride along.

That is exactly what #1403 then does, deliberately, with a six-field select — worth naming it there as the reason the select is narrow, not just that it is.

Keep the volunteer test block; it is a useful pin on the policy ("if a lead row ever carries these columns, they ship"). Just retitle it so it does not claim to mirror the route's wire shape, and note in the comment that it constructs its own row — no route change can break it.

Standing observation, not for this PR: the reason this stopped being a leak is that #1425 replaced a whole-row `include` with an explicit select. Everywhere that pattern survives, a column added and tagged `pii`/`personal` later ships automatically to whoever holds a resolving token — no route diff, no registry diff, no failing test. I have an audit of the remaining instances running separately.

Minor, still open from the last round: `leadMentor` has no test case despite being the same path, and the pinned row omits `notificationSettings` while the FINDING lists it.

PR #1403, opened 2026-07-31
feat(programs): show the parents' phone numbers on the leader roster

### LINKED ISSUE #1400: Leader view of Roster does not show parent phone #

dkaygithub:
## What & why

The Roster tab of `/program-ops/programs/[id]` showed a child's **emergency
contacts** but never their **parents**, so a lead reaching for a phone number
found the backup before the family. Parents now render directly above the
emergency contacts:

```
Email: kid@example.com          Phone: 512-555-9999
Parent/Guardian: Sam Smith - 512-555-1111, sam@example.com; Jamie Smith - 512-555-2222
Emergency Contact: Aunt Ada - 512-555-3333
```

Fixes #1400.

> [!IMPORTANT]
> ## Disclosure note — needs the approving reviewer's explicit sign-off
>
> **After this PR, any member of any enrolled household can see the _names_ of
> the other enrolled families' adults.**
>
> `Person.name` is tier `public`, so the per-field stripper cannot hide it. The
> only thing between these new rows and a caller is the route's association
> gate, and that gate admits any member of any enrolled household — not just
> staff. Phone and email are **not** affected: they are `pii`, and the
> integration test verifies they strip for a non-staff enrolled peer while
> surviving for the lead mentor.
>
> This is a **new class of disclosure**: previously the roster exposed enrolled
> *children*; it now also exposes the *non-participant adults* of those
> households. #1402's audit covered the pii/personal collateral of the grant but
> not this public-tier row widening.
>
> It may well be the intended trade — the roster already discloses the children,
> and a program lead's own workflow is the point of #1400. But it is a policy
> call, not an implementation detail, and it should be an explicit yes rather
> than a side effect. **If it is not wanted**, the fix is to drop
> `household.householdMembers` in the non-privileged branch of the route
> (`src/app/api/programs/[id]/route.ts`), not to lean on tiers — tiers cannot
> hide a `public` field.
>
> Raised by @thpr in review; carried here at his request. Please state the
> decision explicitly on this PR before approving.

## Changes

**`GET /api/programs/[id]`** — the `household` include gains `householdMembers`
filtered to leads:

```ts
householdMembers: {
    where: { isHouseholdLead: true, ...LIVE_PERSON },
    orderBy: { id: "asc" },
    select: { id: true, householdId: true, isHouseholdLead: true, name: true, email: true, phone: true },
},
```

`householdId` and `isHouseholdLead` are **both** scope keys for the `Person`
`their_program_households` binding — drop either and the row resolves no scope,
so the contact fields strip away for everyone. Same failure mode the
`emergencyContacts` select guards against with its `householdId`.

The select stays deliberately narrow. That view also grants `:personal` on
these rows (see the FINDING in #1402), so anything wider would put a parent's
`dateOfBirth` and `allergies` on the wire.

**`ProgramRosterTab`** — a `Parent/Guardian:` line above `Emergency Contact(s):`
on each active-participant card. Pending participants are unchanged; they have
no emergency-contact block either.

## Judgment calls

- **A participant who leads their own household is filtered out of their own
  parent list.** An enrolled adult is not their own parent, so `parentsOf()`
  drops `m.id === p.personId`.
- **Client-side contact fields are optional.** The stripper omits them for a
  viewer without the grant, so a parent can legitimately render as a bare name
  rather than `undefined - undefined`.
- **The label is `Parent/Guardian:`, not `Parents:`.** An `isHouseholdLead` can
  be a grandparent or a legal guardian; the old label claimed more than the data
  supports. Not pluralized — "Parent/Guardians" reads worse than the constant
  label.

## Tests

- `program-ops/programs/[id]/__tests__/page.test.tsx` — a new focused test (kept
  separate rather than grown onto the existing edge-case test, per the recent
  5s-timeout splits): asserts the formatted parent line, asserts parents precede
  emergency contacts via `compareDocumentPosition`, and asserts the
  self-leading adult's card reads `Parent/Guardian: N/A`.
- `programsIdAPI.integration.test.ts` — runs the REAL stack (handler → registry →
  `scopesHeld` → `stripBag`), so it proves the grant **delivers** rather than
  just that the select asked for it. A lead receives the parent's
  name/phone/email; a non-lead sibling in the same household is absent; an
  enrolled non-staff caller gets the parent's name (public) with phone/email
  stripped; and the anonymous / non-enrolled leak tests now canary `PARENT_NAME`
  and `PARENT_PHONE` alongside `ENROLLED_NAME`.
- **New regression pin for #1425's narrowing.** The fixture now also makes the
  parent a core volunteer — the row that resolves `their_program_households` and
  so carries `:pii` + `:personal`. Only the route's select keeps `dateOfBirth` /
  `allergies` / `googleId` off `volunteers[].person` and `leadMentor`, and
  nothing was asserting that. The test names any key beyond `{id, name, email}`
  on either, so a widened select fails loudly instead of silently.

## Review status

- **Merge-base refreshed.** The branch was last rebased before #1402 merged
  (Aug 8), so its three-dot diff still carried all 6 of that PR's files as if
  they were ours — `security-boundary-isolation.yml` would have reported 3
  non-companion violations, and only the `label` check ever ran. `origin/main`
  is now merged in (clean, no conflicts — the branch's copies of `registry.ts` /
  `scopeBindings.ts` / `scopes.ts` were byte-identical to what #1402 landed).
  The three-dot diff is down to the 5 files that are actually this feature, and
  CI runs for real.
- **@thpr's 🔴 blocker** (the `volunteers` / `leadMentor` over-select) was fixed
  on `main` by #1425. This PR adds the regression test that pins it.
- **@thpr's 🟢 items** addressed: the `Parent/Guardian` copy, the anonymous
  `PARENT_NAME` canary, `parentsOf(p)` hoisted to one call per participant per
  render, and the volunteer/lead-mentor regression test.
- **@thpr's 🟡** is the Disclosure note above — open, and deliberately not
  decided here.

`npx tsc --noEmit` clean, `eslint --max-warnings 0` clean on the touched files,
roster page suite 27/27. Integration tests are CI-only (shared dev DB).


thpr (changes requested):
Reviewed the branch against its base (#1402), the registry, `scopeBindings.ts`, and the generated classifications.

## What it does

Adds a `householdMembers` include (household leads only) to `GET /api/programs/[id]`, plumbs it through `ProgramDetail`, and renders a `Parents:` line above `Emergency Contact(s):` on each active roster card.

Mechanically the new code is right. Both scope keys (`householdId`, `isHouseholdLead`) are selected, `isTrue` is strict so an unselected column fails closed, `LIVE_PERSON` is applied, `householdMembers` already resolves to `Person` in `classifications.ts` so the stripper walks it correctly, and `programs/[id]` is already on `EDGE_INCLUDE_ALLOWLIST`. The integration test running the real handler → registry → `scopesHeld` → `stripBag` path is the right level for this and is the strongest part of the PR.

---

## 🔴 The stated mitigation isn't actually implemented

Both PRs in this stack rest on the same claim. #1402's FINDING: the pre-existing `their_program_households:personal` token now resolves on parent rows, and *"the route's select is what keeps them off the wire"*. This PR: *"The select stays deliberately narrow… anything wider would put a parent's `dateOfBirth` and `allergies` on the wire."*

True of the new `householdMembers` select. **Not** true of the two other `Person` includes on the same route, both untouched:

- `checkin-app/src/app/api/programs/[id]/route.ts:55` — `volunteers: { where: { person: LIVE_PERSON }, include: { person: true } }`
- `checkin-app/src/app/api/programs/[id]/route.ts:87` — `leadMentor: true`

`include: { person: true }` returns every `Person` scalar. Before #1402 a volunteer's `Person` row resolved **no** scope on this route — `their_program_participants` is `id ∈ participantIdsInScopePrograms`, and a volunteer isn't a participant — so everything above `public` stripped. Once #1402 binds `Person` to `their_program_households`, any volunteer or lead mentor who is a household lead of a household with a kid enrolled (the parent-volunteer, i.e. the common case) resolves the scope, and the `programLeadMentor` / `programCoreVolunteer` views hold both `:pii` and `:personal`.

Newly delivered to lead mentors and core volunteers on those rows:

| field | tier |
|---|---|
| `dateOfBirth` | personal |
| `allergies` | personal |
| `notificationSettings` | personal |
| `emailSuppressed` | personal |
| `googleId` | pii |

Nothing consumes any of it — `ProgramDetail` declares `person: { name, email }` for volunteers and `leadMentor: { name, email }`. So the fix is free:

```ts
volunteers: {
    where: { person: LIVE_PERSON },
    include: { person: { select: { id: true, householdId: true, isHouseholdLead: true, name: true, email: true } } },
},
```

and the same shape for `leadMentor`. Keep `householdId` / `isHouseholdLead` — they're the scope keys, and dropping them strips `email` too.

This lands the moment #1402 merges, independent of this PR. But this is the PR that makes the stack live, and the one asserting the select is narrow, so flagging it here. Fix in either.

Related, lower priority: `participants[].person` is also a bare `include`, so `dateOfBirth` / `allergies` ride the wire for enrolled kids. Pre-existing (`their_program_participants:personal` was already granted), not a regression — but narrowing it in the same pass would make the "narrow select" claim hold for the whole route.

## 🟡 Parents' names now reach every enrolled peer family

`Person.name` is `public`, so the association gate — not the stripper — is the only thing between the new rows and a caller, and that gate admits any member of any enrolled household. The PR's own integration test confirms it:

```ts
expect(parents[0].name).toBe(PARENT_NAME); // public tier survives
```

So a parent in household A can now enumerate the *adults* of every other family in the program, not just the enrolled children. New class of disclosure (non-participant adults); #1402's audit covers the pii/personal collateral but not this public-tier row widening.

It may well be fine — the roster already discloses the children — but it's a policy call, not an implementation detail, and it should be stated explicitly for the CODEOWNERS reviewer rather than arriving as a side effect. If it isn't wanted, the fix is to drop `household.householdMembers` in the non-privileged branch, not to lean on tiers.

## 🟢 Smaller things

- **"Parents" is the wrong word for `isHouseholdLead`.** A household lead can be a grandparent or a legal guardian; the schema comment calls them leads, and this repo is careful about people-vocabulary. `Parent/Guardian:` says the same thing without the false claim.
- **No anonymous canary for the parent.** The file's own convention is `expect(JSON.stringify(data)).not.toContain(ENROLLED_NAME)` on the anonymous path. `PARENT_NAME` deserves the same line — the metadata branch covers it structurally, but the canary is what makes the guarantee legible.
- **No test for the volunteer / lead-mentor rows.** Follows from the first finding: asserting a volunteer-parent's `dateOfBirth` is absent from the lead-mentor response would have caught it, and is the regression guard once fixed.
- `parentsOf(p)` is called three times per participant per render (`ProgramRosterTab.tsx:251-253`). Hoist to a `const` in the map body.
- Zero parents renders singular `Parent: N/A`. Consistent with the existing emergency-contact line, so fine — just noting the page test leans on it (`getByText("Parent:")`) as the proxy for "the self-leading adult was excluded", which is indirect.
- `formatParent(m: ReturnType<typeof parentsOf>[number])` round-trips through `parentsOf` to name a type that could be extracted once directly.
- `householdMembers?` is optional in `ProgramDetail` while `emergencyContacts` is required, though both come from the same include and the stripper drops fields, never arrays. The `?? []` makes it harmless.
- Pending participants get no parent line. Deliberate and consistent with the existing EC treatment, but #1400's motivation ("a lead reaching for a phone number") arguably applies there too — worth a sentence on the issue if it's being deferred.

## Merge order

The base/CI warnings in the description are accurate and important: no workflow triggers on a non-`main` base, so the empty check list is not a green one. #1402 first, then retarget and let CI actually run — particularly the integration suite, which is where the delivery assertions live and which wasn't run locally.

---

**Verdict:** feature code is correct and the integration test is the right shape. Blocking on the volunteer / `leadMentor` over-select — two-line fix, and it's the exact leak the stack's own FINDING claims is already prevented. The parent-name disclosure to peer families needs an explicit yes/no from the security reviewer before merge.

<sub>Separately, on #1402's `f2521445`: replacing the derived `{ ...parentRow } minus one key` fixtures with hand-copied literals weakens those negative tests — they'll keep passing while silently drifting from `parentRow` (they already dropped `email`). And the lint premise looks off: the same rest-destructure pattern ships on `main` at `src/app/api/membership-ops/households/route.ts:31` and passes `--max-warnings 0`. `const noFlag: Record<string, unknown> = { ...parentRow }; delete noFlag.isHouseholdLead;` keeps the derivation with no unused var. The test-name change to "when the select omitted…" is a genuine improvement — keep that either way.</sub>

jee7s:
@thpr — revised. Taking your [review](https://github.com/innovationtreehouse/checkin/pull/1403#pullrequestreview-4832634215) point by point.

### 🔴 The volunteer / `leadMentor` over-select

Fixed on `main` by #1425, which landed the exact narrowing you wrote out:

```ts
person: { select: { id: true, name: true, email: true } }   // volunteers
leadMentor: { select: { id: true, name: true, email: true } }
```

Nothing was pinning it, though, so it could widen back silently — which is the
half of your finding that survived the fix. Added the regression guard you
asked for: the integration fixture now makes the parent a **core volunteer**,
so the assertion runs against the row that actually resolves
`their_program_households` (the parent-volunteer common case), not an inert
one. It fails on any key beyond `{id, name, email}` and names the offender:

```ts
const SELECTED = ['id', 'name', 'email'];
const vol = data.volunteers.find(v => v.personId === parentId);
expect(Object.keys(vol.person).filter(k => !SELECTED.includes(k))).toEqual([]);
expect(Object.keys(data.leadMentor).filter(k => !SELECTED.includes(k))).toEqual([]);
```

I left `participants[].person` alone — pre-existing, not a regression, and
widening the blast radius of this PR to cover it seemed worse than a follow-up.
Say the word if you want it in the same pass.

### 🟡 Parents' names reaching every enrolled peer family

**Not decided here — deliberately.** It is now a
**Disclosure note**
at the top of the PR description, stated as plainly as I could manage: any
member of any enrolled household can see the *names* of other enrolled
families' adults, because `Person.name` is `public` and the association gate —
not the stripper — is the only thing in front of it. Phone and email are
unaffected; the pii band strips correctly for peers and the integration test
proves it.

The note says it needs the approving reviewer's explicit sign-off before merge,
and records your alternative (drop `household.householdMembers` in the
non-privileged branch, don't lean on tiers) in case the answer is no. That's a
call for you and CODEOWNERS, not for this PR to make quietly.

### 🟢 Smaller things

| Your ask | Done |
|---|---|
| "Parents" is wrong for `isHouseholdLead` | `Parent/Guardian:`. Dropped the pluralization with it — "Parent/Guardians" reads worse than the constant label. |
| No anonymous canary for the parent | The anonymous leak test now canaries `PARENT_NAME` **and** `PARENT_PHONE`; the non-enrolled test canaries `PARENT_NAME`. |
| `parentsOf(p)` called 3× per participant per render | Hoisted to one `const parents` in the map body. |
| The `getByText("Parent:")` proxy is indirect | Agreed — it now asserts that card's line reads `Parent/Guardian: N/A` directly. |

Left as-is, with reasoning rather than silence: `formatParent`'s
`ReturnType<typeof parentsOf>[number]` round-trip, the optional
`householdMembers?` vs required `emergencyContacts`, and pending participants
having no parent line. All three are cosmetic, all three are consistent with
what the file already does, and none change behaviour — happy to take any of
them if you'd rather they were tidied.

### Merge order / CI

You were right that the empty check list wasn't a green one, and the cause
outlived the retarget: the branch was last rebased **before** #1402 merged on
Aug 8, so its three-dot diff still carried all 6 of that PR's files as if they
were ours. `security-boundary-isolation.yml` would have reported 3
non-companion violations, and only the `label` check ever ran.

`origin/main` is now merged in (merge commit, no rebase — clean, no conflicts;
the branch's copies of `registry.ts` / `scopeBindings.ts` / `scopes.ts` were
byte-identical to what #1402 landed). The three-dot diff is down from 11 files
to the 5 that are actually this feature, with no `src/security/` or
`prisma/` files left in it:

```
checkin-app/src/app/__tests__/programsIdAPI.integration.test.ts
checkin-app/src/app/api/programs/[id]/route.ts
checkin-app/src/app/program-ops/programs/[id]/ProgramRosterTab.tsx
checkin-app/src/app/program-ops/programs/[id]/__tests__/page.test.tsx
checkin-app/src/app/program-ops/programs/[id]/page.tsx
```

Also confirmed nothing silently reverted in the merge: #1613's `useOrgTime`
context in `ProgramRosterTab` and #1444's `Fee`/`FeePayment` drop both survive,
and every `src/security/` and `prisma/` file is byte-identical to `main`.

Local: `tsc --noEmit` clean, `eslint --max-warnings 0` clean on the touched
files, roster page suite 27/27 (including the previously flaky general-tab
test). Integration suite is CI-only — it runs for real now.

Re-requested your review.

jee7s:
@thpr — ready for your re-review, with one decision only you (or CODEOWNERS) can make. Your review points were addressed point-by-point in the 2026-08-19 revision: the over-select fix that landed on `main` via #1425 now has the regression guard you asked for (asserting against a parent who is a core volunteer), the smaller asks are done (`Parent/Guardian:` label, anonymous canaries, hoisted `parentsOf`, direct `N/A` assertion), and the stale three-dot isolation problem is resolved — the diff is down to the five feature files. CI green, conflict-free.

Two things gate the merge: clearing the stale changes-requested, and the **Disclosure note** at the top of the PR description — peer families seeing other enrolled families' adults' *names* (phone/email strip correctly) needs your explicit yes/no, with your alternative (drop `household.householdMembers` in the non-privileged branch) recorded there if the answer is no.

thpr (approved):
Approving. Every point from the previous review is addressed or properly escalated.

The regression pin is better than what I asked for. Making the parent a **core volunteer** puts the assertion on the row that actually resolves `their_program_households` — the parent-volunteer case — rather than an inert one, and

```ts
expect(Object.keys(vol.person).filter(k => !SELECTED.includes(k))).toEqual([])
```

fails with the offending keys named, so a re-widened select can't go quiet again.

## The disclosure note — explicit yes

Recording the decision here as asked. Peer families seeing other enrolled families' adults' *names* is the intended trade, not a side effect:

- the roster already discloses the enrolled children by name, and `Person.name` is `public` for that reason;
- a program lead reaching the family is the whole point of #1400;
- phone and email are `pii` and strip correctly for a non-staff enrolled peer — the integration test proves that half, and it is the half that would actually matter.

No change needed. Proceed.

## Non-blocking

The `beforeAll` leaked-state cleanup deletes the program and the people without first clearing the enrollments joining them — neither `ProgramParticipant` nor `ProgramVolunteer` declares `onDelete`, so both are Prisma-default `Restrict`, and a run that dies between `create` and `afterAll` leaves rows that make the next run's cleanup throw on the FK. This PR adds a second row type with that shape and fixes only `afterAll`.

Pre-existing and not yours to carry: filed as #1692, which sweeps all six affected suites. Two lines here if you want it out of the way, otherwise it goes with the sweep:

```ts
await prisma.programParticipant.deleteMany({ where: { personId: { in: existingUserIds } } });
await prisma.programVolunteer.deleteMany({ where: { personId: { in: existingUserIds } } });
```

<sub>🤖 Review by [Claude Code](https://claude.com/claude-code)</sub>

Issue #1407, opened 2026-07-31
POST /api/programs/[id]/participants has no orgMemberOnly gate


thpr:
`POST /api/programs/[id]/participants` — the enrollment write path — never checks `Program.orgMemberOnly`. Members-only enrollment is enforced only by the *read* routes hiding the program.

## Where the gate does and doesn't exist

The three places that gate `orgMemberOnly` are all reads:

- `api/programs/route.ts` — `canSeeOrgMemberOnly` filters the catalog list
- `api/programs/[id]/route.ts` — 403 `Forbidden: Member-Only Program` on program detail
- `api/programs/[id]/eligible-participants/route.ts` — filters the staff candidate list

`api/programs/[id]/participants/route.ts` POST checks, in order: the program exists, the caller may enroll this person (self / household lead / sysadmin+board), the external-admin comp confirmation, then `enrollmentStatus === 'CLOSED'`, age, and capacity. There is no `orgMemberOnly` check anywhere in it.

So a non-member who knows (or guesses) a members-only program id can `POST` straight to it and be enrolled. The program id is a small integer, and a program can flip to `orgMemberOnly` after someone has already seen it, so the id is not a secret.

## Impact

The enrollment lands as a normal row. For a paid program the participant still has to check out, so this is not free access to a paid program — but it does seat them on the roster of a program the org decided is members-only, and a free members-only program is enrolled outright (`initialStatus = 'ACTIVE'` when both price fields are null).

## Suggested fix

Add the same gate the sibling read routes use, inside the `enforceLimits` block so a deliberate board override can still comp someone in (consistent with how closed-enrollment, age, and capacity behave):

```ts
if (currentProgram.orgMemberOnly && !(await isDuesSettled(participantId))) {
    return NextResponse.json({ error: "This program is for Treehouse Members only.", requiresOverride: true }, { status: 400 });
}
```

Note the predicate is checked on `participantId`, not the caller — a household lead enrolling a dependent should be judged on the dependent's household, which is the same household. `isDuesSettled` (not `isActiveOrgMember`) is the current rule after #1397: a household whose dues are paid and whose background check is still with the board counts.

## Notes

Pre-existing; found while implementing #1397, which widened the three read gates but deliberately did not add a new gate to the write path. Untested by the current suite — `programsParticipantsAPI.integration.test.ts` covers auth, age, capacity, and comp overrides, but has no members-only case.

PR #1421, opened 2026-08-01
feat(programs): member pricing and members-only access once dues are paid

### LINKED ISSUE #1397: Use member pricing for programs when a membership application is pending only the background check
A household that has **paid its membership dues** but is still waiting on background check clearance currently gets **non-member** program pricing. Proposal: treat that state as member-priced.

## Current behaviour

The state in question is `PENDING_BG_CLEARANCE` — it means exactly "payment satisfied, background check outstanding". `lib/membership/payment.ts:208`:

```ts
const activating = !!process.bgClearedAt;
… status: activating ? "ACTIVE" : "PENDING_BG_CLEARANCE"
```

`OrgMembership.status` is only flipped to `ACTIVE` on the `activating` branch, so in `PENDING_BG_CLEARANCE` the membership row is not ACTIVE. Program pricing keys on exactly that:

- `personRecordIsActiveOrgMember` → `household.orgMembership.status === "ACTIVE"` (`lib/orgMembership.ts:49`)
- `isActiveOrgMember` / `isActiveOrgMemberThrough` → `ACTIVE_ORG_MEMBER_PERSON_WHERE`

Consumed for pricing at `api/programs/[id]/route.ts:41-42,94` and `api/programs/[id]/discount-code/route.ts:31,37`.

So the family has paid, and is charged the non-member rate for a program while the board finishes reviewing.

## The main design constraint

**Do not widen the shared "active member" predicate to fix this.** `isActiveOrgMember` / `ACTIVE_ORG_MEMBER_PERSON_WHERE` are used well beyond pricing — the program-announce email audience (`lib/notifications.ts`), outreach recipients (`lib/outreach/recipients.ts`), the shop org-members list, people search, payment-plan eligibility. Broadening the predicate would silently hand a household that has not cleared its background check every other member benefit and put it in member-only mailings.

This wants a **pricing-specific** predicate (something like "dues settled for the covered term") layered next to the existing one, leaving `ACTIVE` as the answer to "is this a member" everywhere else. The coverage-window rule still applies — `isActiveOrgMemberThrough` + `programCoverageDate` — so whatever predicate is added has to answer the same "covered through the program's run" question, not just "paid".

## Questions to settle before building

- **What if the background check is later rejected?** The household can land in `BLOCKED`/`DENIED` after already enrolling at the member rate. Is the discount honoured, reversed, or is the enrollment unwound? This is the decision that most shapes the implementation, because Shopify has already taken the money at the discounted price by then.
- **Does this extend to enrollment eligibility, or pricing only?** Members-only programs (`orgMemberOnly`) are a separate gate on the same predicate. The request as written is about price; letting a not-yet-cleared household into a members-only program is a bigger step, and background check clearance is a safety control.
- **Does the reconciliation path accept it?** Member-priced program checkouts mint a discount code (`api/programs/[id]/discount-code`). Worth confirming the finance reconciliation treats a member-priced order from a non-`ACTIVE` household as expected rather than raising an exception.
- **Renewals too, or initial applications only?** A renewing member is `ACTIVE` throughout, so this only bites initial applications — worth confirming that matches intent.

## Scope note

Pricing only, per the request. No change to the membership state machine itself: `PENDING_BG_CLEARANCE → ACTIVE` still requires clearance, and background check clearance remains a precondition for membership.

cc @thpr @dkaygithub

thpr:
Fixes #1397

## The problem

A household whose membership application is **paid** but still waiting on background-check clearance sits at `PENDING_BG_CLEARANCE`, and its `OrgMembership` is not yet `ACTIVE`. Program pricing and the members-only gates key on `ACTIVE`, so a family that had paid in full was charged the **non-member** rate and locked out of members-only programs for however long the board's review took.

## Approach

The issue's main design constraint was: do not widen the shared "is an active member" predicate. `isActiveOrgMember` / `ACTIVE_ORG_MEMBER_PERSON_WHERE` are consumed well beyond programs — the announce-email audience (`lib/notifications.ts`), outreach recipients, the shop org-members list, people search, payment-plan eligibility — and broadening them would hand an uncleared household every other member benefit and put it in member-only mailings.

So this adds a **separate, program-scoped predicate** next to the existing one. `ACTIVE` remains the answer to "is this a member" everywhere else, including the login gate and facility access.

- **`duesSettledAwaitingBg`** (`lib/membership/lifecycle.ts`) — a `StateSet` for `PENDING_BG_CLEARANCE` with `paidAt` set. It lives in the lifecycle definition module so `has`/`where` derive from one spec and the status-literal drift guard stays satisfied. `paidAt` is redundant with the status today (`activate()` is the only writer and always stamps it) and is asserted anyway — a money decision shouldn't rest on that invariant holding for a row some future path writes differently.
- **`DUES_SETTLED_PERSON_WHERE` / `isDuesSettled` / `isDuesSettledThrough`** (`lib/orgMembership.ts`) — `ACTIVE`, or a paid application awaiting clearance. `membershipValidThrough` now yields the same coverage horizon for those households; `isActiveOrgMemberThrough` gates on `ACTIVE` *before* asking for a horizon, so its behaviour is unchanged. The duration comparison is factored into one private `coversThrough` shared by both predicates.

## What moved onto the new predicate

**Pricing** (the original request):
- `api/programs/[id]/discount-code` — the minted member discount code
- `api/programs/[id]` — `viewerMemberPricingEligible`. `viewerIsMember` stays `ACTIVE`-only; it answers "is this a member", and the UI's "your membership ends before this program finishes" alert depends on that meaning.

**Members-only access** (per @dkaygithub's comment on the issue — *"this same state should allow paid households to see and enroll in member only programs"*):
- `api/programs` — the catalog `canSeeOrgMemberOnly` filter
- `api/programs/[id]` — the members-only 403 on program detail
- `api/programs/[id]/eligible-participants` — the candidate filter

The access gates stay status-only, exactly as the `isActiveOrgMember` calls they replace were — no coverage-window check was added where there wasn't one. Pricing keeps its `…Through` variant.

## Decisions a reviewer should weigh in on

The issue listed four open questions. How I settled them:

1. **Background check later rejected.** The discount is honoured; nothing is unwound. The board's existing paid-while-blocked refund alert covers the money case, and by then Shopify has taken the payment anyway. No reversal code.
2. **Enrollment eligibility as well as pricing.** Yes — see above. The issue body flagged this as the bigger step; @dkaygithub asked for it explicitly.
3. **Reconciliation.** Untouched and unaffected — `lib/finance/reconcile.ts` and `matchAudit.ts` sweep *membership* orders (and already include `PENDING_BG_CLEARANCE`), not program orders.
4. **Renewals.** Unaffected. A renewing household is `ACTIVE` throughout, so this only ever bites initial applications — which matches the issue's stated intent.

Also worth noting: #1370 (unified signup, design PR) decides "paid but not yet cleared → member pricing" *inside* its bundled checkout. This lands the same policy for the standalone enrollment path, so the two shouldn't diverge; whichever merges second should point at the single predicate rather than re-answer the question.

## No state-machine change

`PENDING_BG_CLEARANCE → ACTIVE` still requires two reviewers, and clearance remains a precondition for membership itself. No new status, no new transition — `docs/generated/lifecycle/membership.md` is byte-identical (verified by `artifactsDrift.test.ts`).

## Testing

- `npx tsc --noEmit` clean, `npm run lint` clean
- `npm run test:ci` — 2416 pass / 257 suites
- `npm run test:integration` — 1259 pass / 128 suites, against a real Postgres

New coverage:
- Unit (`lib/__tests__/orgMembership.test.ts`): the two `where` shapes asserted against each other, the paid-pending coverage horizon, and the boundary comparison.
- Integration (`programsIdAPI`, `programsAPI`): a paid-pending household gets `viewerMemberPricingEligible: true` with `viewerIsMember: false`, 200s on a members-only program, and sees it in the catalog — then, once its process is moved back to unpaid `PENDING_PAYMENT`, loses pricing and gets a 403.

## Out of scope (filed separately)

While tracing the enrollment path I found that `POST /api/programs/[id]/participants` has **no** `orgMemberOnly` check at all — members-only enrollment is enforced only by the read routes hiding the program. Pre-existing and independent of this change, filed as #1407 rather than widened into this PR.

### DIFF
--- a/checkin-app/src/app/__tests__/programsAPI.integration.test.ts
+++ b/checkin-app/src/app/__tests__/programsAPI.integration.test.ts
@@ -136,6 +136,45 @@ describe('Programs API Integration Tests', () => {
              expect(orgMemberOnly).toBeDefined(); // Revealed because admin
         });
 
+        // #1397: dues paid, background check still with the board — the catalog
+        // shows them members-only programs, same as an ACTIVE household.
+        it('reveals member-only programs to a household that has paid but is awaiting background clearance', async () => {
+             const paidPending = await prisma.person.create({
+                 data: {
+                     email: 'paid-pending-programs-api-test@example.com',
+                     name: 'Paid Pending',
+                     household: {
+                         create: {
+                             name: 'Test HH',
+                             orgMembership: {
+                                 create: {
+                                     status: 'NONE',
+                                     processes: { create: { kind: 'INITIAL', status: 'PENDING_BG_CLEARANCE', paidAt: new Date() } },
+                                 },
+                             },
+                         },
+                     },
+                 },
+                 select: { id: true, householdId: true },
+             });
+
+             try {
+                 (getServerSession as jest.Mock).mockResolvedValue({ user: { id: paidPending.id } });
+
+                 const req = new Request('http://localhost:4000/api/programs', { method: 'GET' });
+                 const res = await GET(req as unknown as import("next/server").NextRequest);
+                 expect(res.status).toBe(200);
+
+                 const programs = await res.json();
+                 const orgMemberOnly = programs.find((p: { name?: string }) => p.name === 'Member Only API Test Program');
+                 expect(orgMemberOnly).toBeDefined();
+             } finally {
+                 await prisma.orgMembershipProcess.deleteMany({ where: { orgMembership: { householdId: paidPending.householdId } } });
+                 await prisma.orgMembership.deleteMany({ where: { householdId: paidPending.householdId } });
+                 await prisma.person.delete({ where: { id: paidPending.id } });
+             }
+        });
+
         // The catalog is deliberately anonymous-readable, so the projection — not a
         // route gate — is what keeps `personal`-tier data off the wire. Both the
         // cached anonymous path and the live session path must ship public columns only.
--- a/checkin-app/src/app/__tests__/programsIdAPI.integration.test.ts
+++ b/checkin-app/src/app/__tests__/programsIdAPI.integration.test.ts
@@ -415,6 +415,92 @@ describe('Individual Program API Integration Tests', () => {
             expect(data.viewerIsMember).toBeUndefined();
             expect(data.viewerMemberPricingEligible).toBeUndefined();
         });
+
+        // #1397: dues paid, background check still with the board. Not a member
+        // (nothing else in the app treats them as one), but priced as one and
+        // admitted to members-only programs.
+        describe('a household that has paid but is awaiting background clearance', () => {
+            let paidPendingId: number;
+            let paidPendingHouseholdId: number;
+
+            beforeAll(async () => {
+                const person = await prisma.person.create({
+                    data: {
+                        email: 'paid-pending-prog-id-api-test@example.com',
+                        name: 'Paid Pending',
+                        household: {
+                            create: {
+                                name: 'Test HH',
+                                orgMembership: {
+                                    create: {
+                                        status: 'NONE',
+                                        processes: { create: { kind: 'INITIAL', status: 'PENDING_BG_CLEARANCE', paidAt: new Date() } },
+                                    },
+                                },
+                            },
+                        },
+                    },
+                    select: { id: true, householdId: true },
+                });
+                paidPendingId = person.id;
+                paidPendingHouseholdId = person.householdId;
+            });
+
+            afterAll(async () => {
+                await prisma.orgMembershipProcess.deleteMany({ where: { orgMembership: { householdId: paidPendingHouseholdId } } });
+                await prisma.orgMembership.deleteMany({ where: { householdId: paidPendingHouseholdId } });
+                await prisma.person.deleteMany({ where: { id: paidPendingId } });
+            });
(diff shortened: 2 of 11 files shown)


dkaygithub (approved):

thpr:
Rebased onto current `main`, which now includes #1411 (the enrollment-POST `orgMemberOnly` gate, filed from this work as #1407).

Added a second commit: #1411 was written against a base without `isDuesSettled`, so it gates on `isActiveOrgMember`. Combined with this branch that left the write path stricter than the reads — a paid, not-yet-cleared household would see a members-only program in the catalog, open its detail page, be quoted the member price, then be refused at enrollment. Swapped that one call so all four gates agree.

The gate stays inside `enforceLimits`, so a confirmed board/sysadmin comp still seats a non-member. Added the matching integration case (a `PENDING_BG_CLEARANCE` household with `paidAt` set enrolls successfully); #1411's non-member, member, and admin-comp cases are untouched and still pass.

Re-verified after the rebase: tsc clean, lint clean, unit suite 2426 pass, integration suite 1264 pass.

PR #1447, opened 2026-08-02
feat(membership-audit): add the "students 18+ as of the member-year start" roster

### LINKED ISSUE #1230: "Students ≥18 as of Sept 1" report
**Backlog item M7** — from the backlog baseline (`docs/backlog/INDEX.md`, PR #1192).

"Students ≥18 as of Sept 1" report — current AND next member-year

thpr:
Fixes #1230 (backlog item M7).

## Why

The board needs to know which household members are legally adults **for a given member year** — the age-18 individual-agreement obligation and the person-level background-check trigger both key on it — and no surface answered that. Age was only ever judged as-of today, which is the wrong question when the obligation attaches at the year boundary.

## What

A board-only Membership Audit view (`/membership-audit/turning-18`, "Students 18+") listing non-lead household members who are 18 or older as of the **current** member-year start and as of the **next** one, so both the standing adults and the ones aging in on the coming boundary are visible in one table.

- **Household leads are excluded.** This is the "child of the signer turns 18" list, and a lead is the signer.
- **Not scoped to program enrollees.** Program students can sit in non-member households, so scoping the query would drop exactly the rows worth seeing. Instead every row carries the programs the person is enrolled in (any program not yet `FINISHED`), with a "Hide those not in a program" checkbox that filters the already-fetched list client-side.
- **No DOB → a count, not a blank row.** People whose age can't be judged come back as `unknownDobCount` and render as a hygiene note. `isDeclaredAdult` people are excluded from that count — a lead marked them 25+, there is nothing to chase.
- **The boundary is the configured `BoardSettings.orgMembershipYearBoundary`**, not a hardcoded Sept 1. With none configured the page says so and renders nothing to act on.

The age filter runs in the query (`dateOfBirth <= birthCutoff(next)`) rather than over every loaded person. `programYear.test.ts` pins `birthCutoff` against `calculateAge` on both sides of the boundary so the SQL-side bound and the displayed age can't drift apart.

## Reviewer notes

Three drift guards fired during the build and each is registered rather than worked around:

- **`livePersonDriftGuard`** rejected `...LIVE_PERSON` spread through a shared local const — it text-scans the query's own argument object. Inlined at both query sites.
- **`authzRegistry`** required negative-authz coverage: a row in `authzRoleRejection.integration.test.ts` (401 anon / 403 plain user) plus the `AUTHZ_TESTED` entry.
- **`routeAuthDrift` rule 3** flagged the `ProgramParticipant` edge include (all-public-tier, so row *existence* is the sensitive thing). The handler is `withAuth roles [sysadmin, board]`, so it gets a keyed `EDGE_INCLUDE_ALLOWLIST` entry with that justification. That file is `checkin-app/tests/security/*`, an allowed companion under the boundary-isolation workflow — not a boundary file — so this does not need to ship as its own PR.

## Verification

- `tsc --noEmit` clean (only pre-existing missing-`xlsx` errors in this worktree).
- Unit suite: **2428 passed / 257 suites**.
- Integration suite against a throwaway Postgres: **1260 passed / 128 suites**.
- Live query smoke against real Postgres: a synthetic member born 2008-08-15 returned `ageAtCurrent 17 / ageAtNext 18` with its program badge, boundaries `2025-09-01` / `2026-09-01`. Confirms the query shape executes, not just type-checks.
- `npm run lint` clean.

Deliberately skipped: CSV export and column sorting. The list is small and the board's next step is opening a household, not exporting. Worth adding when it outgrows a screen.

### DIFF
--- a/checkin-app/src/__tests__/authzRegistry.test.ts
+++ b/checkin-app/src/__tests__/authzRegistry.test.ts
@@ -90,7 +90,6 @@ const AUTHZ_TESTED = new Set<string>([
     'membership-audit/person-bg',
     'membership-audit/households-missing-contact',
     'membership-audit/unclaimed-households',
-    'membership-audit/turning-18',
     'membership-ops/applications/archive',
     'membership-ops/applications/unarchive',
     'membership-ops/applications/certify-payment',
--- a/checkin-app/src/__tests__/authzRegistry.test.ts
+++ b/checkin-app/src/__tests__/authzRegistry.test.ts
@@ -90,6 +90,7 @@ const AUTHZ_TESTED = new Set<string>([
     'membership-audit/person-bg',
     'membership-audit/households-missing-contact',
     'membership-audit/unclaimed-households',
+    'membership-audit/turning-18',
     'membership-ops/applications/archive',
     'membership-ops/applications/unarchive',
     'membership-ops/applications/certify-payment',
--- a/checkin-app/src/app/__tests__/authzRoleRejection.integration.test.ts
+++ b/checkin-app/src/app/__tests__/authzRoleRejection.integration.test.ts
@@ -47,6 +47,7 @@ import { GET as BADGES_GET } from '@/app/api/facility/badges/route';
 import { GET as FAC_VISITS_GET, PATCH as FAC_VISITS_PATCH } from '@/app/api/facility/visits/route';
 import { GET as MISSING_CONTACT_GET } from '@/app/api/membership-audit/households-missing-contact/route';
 import { GET as UNCLAIMED_GET } from '@/app/api/membership-audit/unclaimed-households/route';
+import { GET as TURNING_18_GET } from '@/app/api/membership-audit/turning-18/route';
 import { POST as CERTIFY_PAYMENT_POST } from '@/app/api/membership-ops/applications/certify-payment/route';
 import { POST as APP_EXTERNAL_POST } from '@/app/api/membership-ops/applications/external/route';
 import { POST as REVIEW_OVERRIDE_POST } from '@/app/api/membership-ops/applications/review-override/route';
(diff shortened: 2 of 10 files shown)


dkaygithub:
Just read through this.
Tempted to approve as is.

My gut says that there are a plethora of bugs with the conflation of {minor/child/student/kid/under 18/dependant} language we've given Claude in various contexts.

thpr:
Doc Standard Proposal in pr #1458  - not that's modifying a proposal which is still in progress and being updated (see #1428 #1445 )

thpr:
## Restructured after CI

`check-route-coverage` rejected the route:

```
✗ [new-route-old-authz] api/membership-audit/turning-18/route.ts — GET /api/membership-audit/turning-18
  is not in the security registry and not in the frozen legacy baseline
  (scripts/legacy-authz-routes.txt). New routes must use defineRoute() + handler() from @/security.
```

I had read the sibling Membership Audit routes as evidence that registration was optional here. They're **grandfathered** in `scripts/legacy-authz-routes.txt`, not exempt. New routes must register.

That makes the registry entry a boundary change, so it ships separately: **#1490**. This PR is now based on that branch and should be retargeted to `main` once it merges. ("Coverage gate" was a cascade from the unit job, not an independent failure — `No merged coverage-summary.json`.)

### What the migration changed beyond the gate

`stripBag` walks each bag key as a classified model and drops every field the schema doesn't classify. The computed row this PR originally returned — `ageAtCurrent`, `ageAtNext`, `programs`, `householdName` — would have been **stripped to nothing on the wire, with a 200**. So the route now returns the classified inputs `{ Person, BoardSettings }` and the page derives both as-of ages with the same pure helpers. Same pattern `GET /api/finance-ops/payment-plans` uses for its next-year flag.

Two knock-ons:

- **The unknown-DOB count is gone.** Ad-hoc scalar, stripped; recovering it client-side would mean shipping the whole directory's DOBs to the browser. Membership Audit → Compliance already surfaces people with no recorded age.
- **Dates render through `formatDateOnly`** instead of a hand-rolled UTC-pinned `toLocaleDateString`, matching the calendar-date convention from #1424, which landed while this was open.

### New test

`turningEighteenAPI.integration.test.ts`. The stripper fails *silently* — wrong bag key logs a warning and returns an empty 200 — so this asserts the shape survives: `dateOfBirth` present, household and program nested, leads and under-18s excluded, 403 for a non-board caller.

### Re-verified locally

- `check-route-coverage` → `0 error(s)`
- `tsc --noEmit` clean
- Unit: **2491 passed / 262 suites**
- Integration against a throwaway Postgres: **1303 passed / 131 suites**
- `npm run lint` clean

jee7s (approved):

Issue #1396, opened 2026-07-31
Participant merge transfers the background check date but leaves the application at background-check review


jee7s:
Found in production testing (@jee7s). Merging a participant transfers the tombstoned participant's background check date to the survivor, but the in-flight membership application stays parked at background-check review. The date says the household is covered; the process state disagrees.

**Assessment: not intended.** The merge deliberately moves the newer date on the stated rationale that it is the same human, and it re-points the process row to the survivor — but it never re-evaluates the derived state that date feeds. That reads as a gap in the merge's cross-entity sweep, not a deliberate carve-out.

## Why the two disagree

Two separate fields carry "background check" state, and only one of them moves:

- `Person.lastBackgroundCheck` — the compliance date. The merge takes the newer of the two (`membership-ops/participants/merge/route.ts:94-100`, *"Same human — an older check is never the right survivor"*).
- `OrgMembershipProcess.bgClearedAt` — per its schema doc, *"The single source of truth for 'BG cleared' … the membership only goes ACTIVE once this is set AND dues are paid."*

`bgClearedAt` is stamped from a still-valid prior check at exactly two moments, both via `householdBgIsFresh`:

- `lib/membership/intake.ts:390` — at intake
- `lib/membership/renewal.ts:191` — when a member begins renewal

A merge is neither. The route re-points the process row (`merge/route.ts:326`, `orgMembershipProcess.updateMany` → `subjectPersonId: keepId`) and updates the date, but nothing re-runs the freshness predicate. So a household that becomes fresh *because of* the merge never gets `bgClearedAt` stamped.

## Impact

An application waiting on background-check clearance stays blocked even though the household now holds a valid background check. The board has to run a 2-of-N review for a background check that already exists and is still in its recheck window.

There is a second-order effect worth checking during the fix: `lib/membership/external.ts:112` computes `holdForNote = !process.bgClearedAt && !!household.intakeNotes?.trim()`, so an unstamped `bgClearedAt` also keeps the intake-note payment hold in place.

## Reproduction

1. Household with an in-flight application sitting at background-check review (`bgClearedAt` null).
2. A duplicate participant record carries a `lastBackgroundCheck` inside `BoardSettings.bgRecheckMonths` of the membership-year boundary.
3. Merge the duplicate into the surviving participant.
4. Survivor's `lastBackgroundCheck` updates to the newer date. Process `bgClearedAt` stays null and the application stays at review.

## Suggested direction (not prescriptive)

[...]

PR #1450, opened 2026-08-02
fix(household): keep merge tombstones out of the "needs a lead" surfaces

thpr:
## What

A merged-away Person is a **tombstone**, not a deleted row (`LIVE_PERSON = { mergedIntoId: null }`, `checkin-app/src/lib/person/filters.ts:12`): it keeps its `householdId`, and the participant merge clears its `isHouseholdLead`. `LIVE_PERSON` was applied inconsistently across household reads, so a household whose only remaining member is a tombstone ("live-empty") leaked into board surfaces.

`BROKEN_HOUSEHOLD_WHERE` matched such a household **by construction** — clearing `isHouseholdLead` is exactly what the merge does — so every merge permanently added a board to-do that nobody could clear, since the only name offered to promote was a tombstone.

Affected before this PR:

- `GET /api/admin/broken-households` — listed it, and its `householdMembers` include had no live filter either, so it rendered the merged-away person's name.
- `GET /api/membership-audit/unclaimed-households` — listed it via the BROKEN arm of `UNCLAIMED_OR_BROKEN_HOUSEHOLD_WHERE`.
- `GET /api/nav/todo-counts` — two board badges incremented off those same predicates.
- `GET /api/safety/emergency-contacts` — worse in kind: `household.findMany` with **no `where` at all** and no live filter on the include, so the keyholder-facing front-desk sheet carried the household and the tombstone's name.

## How

Both shared predicates in `checkin-app/src/lib/household/filters.ts` now count live people only, with the two halves of the intent split apart:

```ts
export const BROKEN_HOUSEHOLD_WHERE: Prisma.HouseholdWhereInput = {
    householdMembers: { some: LIVE_PERSON, none: { isHouseholdLead: true, ...LIVE_PERSON } },
};
```

- `none: { isHouseholdLead: true, ...LIVE_PERSON }` — a household whose only lead is a tombstone **is** broken and stays listed if it has other live members.
- `some: LIVE_PERSON` — a household with no live members is excluded: there is nobody to promote.

The unclaimed arm's two lead clauses get `...LIVE_PERSON` for the same reason.

`householdMembers` includes filtered to live people on the broken-households, emergency-contacts, and households-missing-contact routes; emergency-contacts also gained `where: { householdMembers: { some: LIVE_PERSON } }` so a live-empty household is off the sheet entirely.

Follows the approach already taken by `/api/membership-ops/households` and `/api/membership-audit/compliance`.

## Reviewer notes

**Behavior change — empty households drop off the broken list and its badge.** That falls straight out of the "no live members ⇒ exclude" rule, and it takes the pre-existing genuinely-empty case with it. The old route comment said empty households were shown "even though there's no one to promote"; that inclusion was unactionable by its own admission. `brokenHouseholdsAPI.integration.test.ts` asserted it, so the assertion and the stale route/badge comments are updated to the new intent. `No members to promote.` in `membership-audit/broken/page.tsx` is now an unreachable fallback — left in place, harmless.

**`findHouseholdsMissingValidContact()` is deliberately unchanged.** Its scope is membership *state* (ACTIVE / in-flight process), not roster, and `service.integration.test.ts` asserts on purpose that an ACTIVE membership with zero members counts. The tombstone leak on that surface was only the lead include, which is fixed.

**The `LIVE_PERSON` drift guard cannot see this class of bug.** It scans `person:` keys, the Person delegate, and join delegates — `household.findMany({ include: { householdMembers } })` matches none of them, which is why all four surfaces drifted unnoticed. Widening the guard to `householdMembers:` is a bigger, separate change and is not attempted here.

## Testing

New `checkin-app/src/lib/household/__tests__/tombstoneHouseholds.integration.test.ts`: a live-empty household is absent from all three list routes; the tombstone's name is absent from the broken household's members and from the contact sheet; and each nav badge is asserted equal to its own list's length — the list and its count share the predicate precisely so they can't diverge, and `filters.ts` says so. Verified it fails against pre-fix source (3 of 4 cases).

- `npm run test:ci` — 2421 passed, 257 suites
- `npm run test:integration -- --runInBand` — 1262 passed, 129 suites
- `npx tsc --noEmit` — clean

Found while analysing #1396; independent of it, and #1396 is not implemented here.

### DIFF
--- a/checkin-app/src/__tests__/livePersonDriftGuard.test.ts
+++ b/checkin-app/src/__tests__/livePersonDriftGuard.test.ts
@@ -190,9 +190,6 @@ const ALLOWLIST: Record<string, string> = {
     'app/api/auth/dev-personas/route.ts': 'Dev-only persona picker (config.isDevInstance() gated), scoped to `email: { endsWith: "@example.com" }` — a merge tombstone\'s email is always rewritten to merged-*@deleted.invalid (merge/route.ts step 1 CAS), so it can never match this domain filter.',
     'app/api/dev/shopify/orders-paid/route.ts': 'Dev-only mock-webhook firer (config.isDevInstance() gated). Both join reads pin `personId: { in: participantIds }` from the request body and are echoed back to the dev UI — no production or user-facing effect.',
     'lib/dev/seed-helpers.ts': 'Dev seed helper: picks an arbitrary sample of existing persons for local macro/demo flows. No production or user-facing effect.',
-
-    // ── unresolved: listed so the guard stays green, NOT because it is correct ──
-    'app/api/safety/emergency-contacts/route.ts': 'NEEDS REVIEW — deliberately left unfiltered pending a product call. This is the evacuation roster: every member carries its open `visits` and `isPresent` is derived from them, and a merge leaves a concurrently-open visit on the TOMBSTONE (see lib/scan-service.ts). Filtering here would therefore trade a duplicate ghost row for the chance of reporting a household absent while someone is still in the building. Which way a safety surface should fail is not a call to make from a drift guard.',
 };
 
 // ── scanner (walk / captureObject copied from lifecycleStatusLiteralAllowlist.test.ts) ──
--- a/checkin-app/src/app/__tests__/brokenHouseholdsAPI.integration.test.ts
+++ b/checkin-app/src/app/__tests__/brokenHouseholdsAPI.integration.test.ts
@@ -3,8 +3,9 @@
  */
 /**
  * Integration tests for the Broken Households tab.
- *  - GET /api/admin/broken-households lists households with zero leads (incl. empty
- *    ones) and excludes households that already have a lead.
+ *  - GET /api/admin/broken-households lists households whose live members include no
+ *    lead, and excludes both households that already have a lead and households with
+ *    no live member to promote.
  *  - POST /api/household/lead lets a BOARD MEMBER assign a lead to a household they
  *    are not a member of, which un-breaks it.
  */
@@ -54,7 +55,7 @@ describe('Broken Households API Integration Tests', () => {
             data: { email: 'youth-broken-api-test@example.com', name: 'Broken Youth', householdId: brokenHouseholdId, dateOfBirth: new Date('2015-01-01') }
         });
 
-        // 2. Leadless household with no participants -> still broken (included).
+        // 2. Leadless household with no members -> nobody to promote, so NOT listed.
         const empty = await prisma.household.create({ data: { name: 'Broken API Test HH Empty' } });
         emptyHouseholdId = empty.id;
 
@@ -69,7 +70,7 @@ describe('Broken Households API Integration Tests', () => {
 
     afterAll(cleanup);
 
-    it('GET lists leadless households (including empty) and excludes led ones', async () => {
+    it('GET lists leadless households and excludes led and empty ones', async () => {
         (getServerSession as jest.Mock).mockResolvedValue({
             user: { id: boardId, isSysadmin: false, isBoardMember: true }
         });
@@ -81,7 +82,7 @@ describe('Broken Households API Integration Tests', () => {
         const data = await res.json();
         const ids = data.households.map((h: { id: number }) => h.id);
         expect(ids).toContain(brokenHouseholdId);
-        expect(ids).toContain(emptyHouseholdId);
+        expect(ids).not.toContain(emptyHouseholdId);
         expect(ids).not.toContain(ledHouseholdId);
 
         const broken = data.households.find((h: { id: number }) => h.id === brokenHouseholdId);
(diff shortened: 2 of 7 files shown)


jee7s:
The diagnosis is right and the predicate rewrite in `lib/household/filters.ts` is the correct shape. `BROKEN_HOUSEHOLD_WHERE` did match a live-empty household by construction once the merge cleared `isHouseholdLead`, and splitting it into `some: LIVE_PERSON` plus `none: { isHouseholdLead: true, ...LIVE_PERSON }` fixes that cleanly.

One surface in the sweep is missed, and the PR body reads as though it is covered.

`GET /api/membership-audit/unclaimed-households` still has an unfiltered include:

```ts
include: { householdMembers: true }
```

and the page renders every one of them:

```tsx
<Text key={m.id} size="sm" c="dimmed">{m.name || 'Unnamed'} • {m.email}</Text>
```

with the household title also falling back to `householdMembers.find(p => p.name)?.name`. So a genuinely unclaimed household that also holds a merge tombstone still shows the merged-away person's name and email as someone to chase, which is the same leak this PR exists to close. The route is not in the diff.

Worth noting the existing fixture already exercises it: `brokenHh` has one live member plus `Ghost2`, and the new assertion for that route checks ids only, so the case passes without covering it.

Fix looks like `where: LIVE_PERSON` on that include, plus extending the unclaimed assertion to member names rather than ids.

Separately, I checked the flagged consequence that genuinely empty households drop off the broken list, and it is safe to accept: `/api/membership-ops/households` applies no member-count filter, so those rows stay visible to the board there and do not disappear org-wide.

One coordination note. #1455 allowlists `api/safety/emergency-contacts/route.ts` as deliberately unfiltered "pending a product call", and this PR filters it. The two merge cleanly, so git will not stop it; the result is that #1455's stale-entry assertion fires on whichever lands second, and the product question gets answered silently. The question is whether a tombstone's still-open visit should keep its household on the evacuation roster. `merge/route.ts` leaves that visit in place when the keeper already has one open, so the case is reachable.

---
_Posted by Claude._

thpr:
Good catch on `unclaimed-households` — fixed in d851bf68. `include: { householdMembers: { where: LIVE_PERSON } }`, and you're right that the assertion was the reason it slid past: `brokenHh` already carried `Ghost2` alongside its live member, so the case was exercised and only the ids were compared. The assertion now pins the member list itself:

```ts
expect(broken.members).toEqual([
    { id: expect.any(Number), name: `Live ${TAG}`, email: `live-${TAG}@example.com` },
]);
```

Verified it fails against the pre-fix route. The PR body overclaimed that surface; the fix makes the claim true rather than the other way round.

## On the #1455 overlap

Taking the product call rather than leaving it to land silently, because the reachable case resolves cleanly in one direction.

`merge/route.ts:247` leaves the tombstone's open visit in place **only when the keeper already has one open**:

```ts
const keeperHasOpenVisit = await tx.visit.findFirst({ where: { personId: keepId, departedAt: null }, ... });
moved.visits = (await tx.visit.updateMany({
    where: { personId: mergeId, ...(keeperHasOpenVisit ? { departedAt: { not: null } } : {}) },
    data: { personId: keepId }
})).count;
```

So a tombstone with an open visit implies a live keeper with an open visit — the merge's own comment says it, "same human". The tombstone's row is a duplicate of a presence the roster already reports under the keeper, never the sole record of someone in the building. Filtering it cannot make the sheet read emptier than the building is; it removes a second row for one person.

The household-level exclusion holds under the same reasoning. If the tombstone's household still has live members, it stays on the sheet with their real presence. If it has none, the only presence it could ever have reported was that duplicate — and the human is on the roster under the keeper's household. A live-empty household on an evacuation sheet is a name with nobody attached to it and no contact who can act.

That said, the two do have to be sequenced, and the follow-on is #1455's to make either way: its `safety/emergency-contacts` allowlist entry needs to come out (its own stale-entry assertion will say so), and its `broken-households` / `households-missing-contact` / `unclaimed-households` entries overlap this branch line-for-line. Happy to rebase onto #1455 instead if you'd rather that one own the whole sweep — say which and I'll reshape this.

Also noted your check on empty households staying visible via `/api/membership-ops/households` — that's the reassurance I wanted and didn't verify myself.

Re-ran after the change: `test:ci` 2421 passed / 257 suites, `test:integration --runInBand` 1262 passed / 129 suites, `tsc --noEmit` clean.

jee7s:
**Area of concern: a red required check, from one line in a file this PR does not touch.**

`Run Tests (unit)` fails on `livePersonDriftGuard.test.ts` — the stale-allowlist assertion returns `["app/api/safety/emergency-contacts/route.ts"]`. Merged #1455 listed that route as `NEEDS REVIEW — deliberately left unfiltered pending a product call`, and this PR filters it. Deleting the allowlist entry answers the question and turns the check green.

The safety argument checks out against main: `merge/route.ts:310-314` leaves a tombstone's open visit in place only when the keeper already has one, so a tombstone's open visit always implies a live keeper already reporting presence.

Also worth noting the scope has shrunk — `unclaimed-households` and `households-missing-contact` are no longer in the diff and `admin/broken-households` is comment-only, because #1455 shipped identical filters. The `BROKEN_HOUSEHOLD_WHERE` rewrite in `lib/household/filters.ts` is the remaining novel content, and it is the actual fix.

---
_Posted by Claude._

jee7s (approved):

PR #1451, opened 2026-08-02
fix(merge): refuse a person merge across mismatched household membership

thpr:
## What

`POST /api/membership-ops/participants/merge` never compared the two subjects' household membership state. `OrgMembership` is 1:1 with `Household` (`householdId Int @unique`) and the merge never writes it — the only membership statement in the route is the `PERSON_BG` `subjectPersonId` re-point. The tombstone keeps its `householdId`; the keeper stays in the keeper's household. Nothing reconciled the two.

This adds a pre-transaction guard that refuses the merge with a 400, and mirrors the check in the analyze route so the picker warns before the operator commits.

Found while analysing #1396. Independent of it — this is a separate, pre-existing bug and this PR does not implement #1396.

## Why — two failure modes

**1. Value stranded.** The merged-away record is the last live member of a household whose membership is `ACTIVE`. Post-merge the paid membership sits on a household with zero live members, while the surviving human is in a different household. The existing lead guard (`isLead && householdOthersCount > 0`) does not fire for a sole member, so nothing caught this.

**2. Restriction laundered.** `DENIED`/`REVOKED` are negative state, and the login block is derived **live** from the person's current household:

```ts
const denied = orgMembershipStatusBlocksLogin(p.household?.orgMembership?.status);
```

(`src/lib/authClaims.ts:34` — `denied` then clears every role flag, `householdLead`, `toolStatuses` and `programsLed`.) Merge a duplicate sitting in a `DENIED` household into a record in a clean household and the surviving human is simply **not denied** on the next token refresh. This does not require the household to be emptied, so no existing guard fired at all.

## The rule

Both harms turn on **which side is merged away**, so the guard is asymmetric. It refuses when either holds:

- **Restriction differential** — the two statuses differ and either is `DENIED` or `REVOKED`. Access is read from the survivor's current household, so the merge would change which restriction applies to this person.
- **Stranded membership** — the merged-away household is `ACTIVE` and would be left with no live member. Only `ACTIVE` carries value worth stranding: `NONE` grants nothing and `DENIED`/`REVOKED` are negative state, so a live-empty household in those states strands nothing.

The live-member count ignores tombstones — a household whose remaining members are all merged away has nobody left to use its membership.

| kept / merged-away | merged-away is its household's last live member | outcome |
| --- | --- | --- |
| NONE / NONE | either | merge |
| ACTIVE / NONE — the ordinary dedupe | either | merge |
| ACTIVE / ACTIVE | no | merge |
| ACTIVE / ACTIVE | yes | refuse — membership stranded |
| NONE / ACTIVE | yes | refuse — membership stranded |
| DENIED or REVOKED on one side, anything different on the other | either | refuse — restriction changes hands |
| DENIED / DENIED, REVOKED / REVOKED | either | merge |

The ordinary dedupe row matters: `createParticipantWithHousehold` gives every fresh Google sign-in its own household with no `OrgMembership` row and makes that person its lead, so a duplicate account reads `NONE` and is its own sole live member. A member's family creating a second account has to stay mergeable.

## What changed

- **New** `src/app/api/membership-ops/participants/merge/membershipGuard.ts` — `householdMembershipStatus()` (a household with no `OrgMembership` row reads as `NONE`) and `membershipMergeBlock(keep, merge)` implementing the rule above.
- **`merge/route.ts`** — both `findUnique`s now include `household.orgMembership`; the guard runs pre-transaction alongside the existing `mergedIntoId` / lead / stranded-login-identity guards, counting live (un-merged) others in the merged-away household.
- **`merge/analyze/route.ts`** — selects `orgMembership.status` (public tier) and `mergedIntoId` on household members, and returns `membershipBlock: { aAsKeeper, bAsKeeper }`. Both directions, because the rule is asymmetric and the keeper isn't fixed until the UI scores and the operator swaps.
- **`membership-ops/participants/merge/page.tsx`** — applies the direction matching the current keeper, renders the reason, disables **Proceed to Preview**, and adds "Swapping which record is kept clears this" when only one direction is blocked.

Error copy names remedies that exist in the app: assign the person to another household from the Participants page, or set the household's membership on the Households page.

## Out of scope, deliberately

The guard does **not** move or merge the two `OrgMembership` rows. That drags in dues, Shopify orders and process history. Refusing and letting the board resolve it by hand is the intended scope.

## Tests

Ten in `merge/__tests__/route.integration.test.ts`:

- Merged-away record is the last live member of an ACTIVE household — refuse. It is the sole member **and** the lead, so the test proves the existing lead guard cannot see this shape.
- ACTIVE keeper vs ACTIVE merged-away, merged-away side is its household's last live member — refuse. Equal status, still a stranding.
- An ACTIVE household whose other members are all tombstones — refuse. Pins the live-only count.
- ACTIVE household with another live member remaining — allow.
- DENIED merged-away / NONE keeper — refuse. Asserts through `orgMembershipStatusBlocksLogin` that the merged-away side is denied and the keeper is not, i.e. exactly the state the merge would have laundered.
- REVOKED keeper / NONE merged-away — refuse. The restriction on the keeper's side counts too.
- DENIED / DENIED — allow.
- ACTIVE keeper / duplicate alone in its own membership-less household — allow. The archetypal dedupe.
- Explicit `NONE` row vs. no `OrgMembership` row at all — allow.
- Analyze reports per direction and reports both null once the two share a household.

The fixture gained `makeHousehold()` plus `extraHouseholdIds` teardown.

Runs (all foreground):

- The merge integration suite → **42/42** (was 32 before this PR).
- Guard replaced with the earlier symmetric equality check → **8 fail**, including both directions of the bug: the ordinary-dedupe allow case and the ACTIVE/ACTIVE stranding refusal.
- Full unit run (`test:ci`) → **2421/2421**, 257 suites.
- `authzRoleRejection` + `auditLog` integration (the other suites touching this route), together with the merge suite → **168/168**.
- `tsc --noEmit` and `eslint` clean.

No flow test: the journey needs board-side membership setup that HTTP alone can't reach cheaply, and the integration tests cover the guard directly. No unit test for the page Alert.

## Live-data audit — NOT RUN, needs someone with prod read access

Both queries below are read-only. They were syntax-checked against a throwaway Postgres and confirmed to **detect** on a synthetic tombstone (ACTIVE household, sole member, merged into a NONE household): 1 row each on that fixture, 0 rows on a clean DB. They have **not** been run against live data.

**A. Households that are live-empty (every member has `mergedIntoId` set) with a non-NONE membership:**

```sql
SELECT h.id AS household_id, h.name, m.status,
       (SELECT count(*) FROM "Person" p WHERE p."householdId" = h.id) AS people_rows
FROM "Household" h
JOIN "OrgMembership" m ON m."householdId" = h.id
WHERE m.status <> 'NONE'
  AND NOT EXISTS (SELECT 1 FROM "Person" p WHERE p."householdId" = h.id AND p."mergedIntoId" IS NULL)
ORDER BY h.id;
```

**B. People whose current household membership status differs from that of a household they were merged out of:**

```sql
WITH RECURSIVE chain AS (
    SELECT id AS tombstone_id, "mergedIntoId" AS survivor_id, 1 AS depth
    FROM "Person" WHERE "mergedIntoId" IS NOT NULL
    UNION ALL
    SELECT c.tombstone_id, p."mergedIntoId", c.depth + 1
    FROM chain c JOIN "Person" p ON p.id = c.survivor_id
    WHERE p."mergedIntoId" IS NOT NULL
),
terminal AS (
    SELECT DISTINCT ON (tombstone_id) tombstone_id, survivor_id
    FROM chain ORDER BY tombstone_id, depth DESC
)
SELECT t.tombstone_id,
       tp."householdId" AS merged_away_household,
       COALESCE(tm.status, 'NONE') AS merged_away_status,
       t.survivor_id,
       sp."householdId" AS surviving_household,
       COALESCE(sm.status, 'NONE') AS surviving_status
FROM terminal t
JOIN "Person" tp ON tp.id = t.tombstone_id
JOIN "Person" sp ON sp.id = t.survivor_id
LEFT JOIN "OrgMembership" tm ON tm."householdId" = tp."householdId"
LEFT JOIN "OrgMembership" sm ON sm."householdId" = sp."householdId"
WHERE COALESCE(tm.status, 'NONE') IS DISTINCT FROM COALESCE(sm.status, 'NONE')
ORDER BY t.tombstone_id;
```

Query B follows merge chains (A→B→C) to the terminal survivor, so a mid-chain tombstone isn't mistaken for the survivor.

Reading the results: rows in query A with status `ACTIVE` are stranded value. Rows in query B where `merged_away_status` is `DENIED`/`REVOKED` against a weaker survivor are escaped restrictions — those want a login-access check on the surviving person, not just a membership fix. Query B also surfaces `NONE` → `ACTIVE` pairs, which the new rule permits and which are not harm; read it as a superset.

### DIFF
--- a/checkin-app/src/app/api/membership-ops/participants/merge/__tests__/route.integration.test.ts
+++ b/checkin-app/src/app/api/membership-ops/participants/merge/__tests__/route.integration.test.ts
@@ -356,26 +356,60 @@ describe("Merge Participants API", () => {
     });
 
     // OrgMembership is 1:1 with Household and the merge never moves it: the tombstone
-    // keeps its householdId, the keeper stays in its own household. So the two sides'
-    // membership status must already agree, or the merge either strands the membership
-    // or drops a restriction.
+    // keeps its householdId, the keeper stays in its own household. Both resulting
+    // harms depend on which side is merged away, so the guard is asymmetric.
     describe("household membership guard", () => {
-        it("400s when the merged-away record holds an ACTIVE membership and the keeper's household is NONE (value stranded)", async () => {
-            // Tombstone is the SOLE live member of an ACTIVE household — the lead guard
-            // (lead + others > 0) can't see this shape.
+        it("400s when the merged-away record is the last live member of an ACTIVE household (value stranded)", async () => {
+            // Sole live member AND lead — the existing lead guard (lead + others > 0)
+            // cannot see this shape.
             const memberHh = await makeHousehold("Paid Member Household", "ACTIVE");
             await prisma.person.update({ where: { id: pMergeId }, data: { householdId: memberHh, isHouseholdLead: true } });
 
             const res = await POST(mergeReq(pKeepId, pMergeId));
             expect(res.status).toBe(400);
             const data = await res.json();
-            expect(data.error).toContain("different Treehouse membership status");
+            expect(data.error).toContain("last live member");
             expect(data.error).toContain("ACTIVE");
 
             // Refused before the transaction: the ACTIVE membership still has a live
             // member, and the surviving human isn't sitting in the NONE household.
             expect((await prisma.person.findUnique({ where: { id: pMergeId } }))?.mergedIntoId).toBeNull();
             expect(await prisma.person.count({ where: { householdId: memberHh, mergedIntoId: null } })).toBe(1);
+        });
+
+        it("400s on ACTIVE keeper vs ACTIVE merged-away when the merged-away side is its household's last live member", async () => {
+            // Equal status on both sides, and still a stranding: the equality alone
+            // says nothing about who is left to use the merged-away membership.
+            await prisma.orgMembership.create({ data: { householdId, status: "ACTIVE" } });
+            const otherActive = await makeHousehold("Other Active Household", "ACTIVE");
+            await prisma.person.update({ where: { id: pMergeId }, data: { householdId: otherActive } });
+
+            const res = await POST(mergeReq(pKeepId, pMergeId));
+            expect(res.status).toBe(400);
+            expect((await res.json()).error).toContain("last live member");
+            expect((await prisma.person.findUnique({ where: { id: pMergeId } }))?.mergedIntoId).toBeNull();
+        });
+
+        it("counts only LIVE others: an ACTIVE household whose remaining members are all tombstones still blocks", async () => {
+            const memberHh = await makeHousehold("Hollow Member Household", "ACTIVE");
+            const alreadyMerged = await prisma.person.create({ data: { name: "Prior Tombstone", householdId: memberHh, mergedIntoId: pKeepId } });
+            extraPersonIds.push(alreadyMerged.id);
+            await prisma.person.update({ where: { id: pMergeId }, data: { householdId: memberHh } });
+
+            const res = await POST(mergeReq(pKeepId, pMergeId));
+            expect(res.status).toBe(400);
+            expect((await res.json()).error).toContain("last live member");
+        });
+
+        it("allows an ACTIVE household merge when another live member remains", async () => {
+            const memberHh = await makeHousehold("Shared Member Household", "ACTIVE");
+            const sibling = await prisma.person.create({ data: { name: "Sibling", householdId: memberHh } });
+            extraPersonIds.push(sibling.id);
+            await prisma.person.update({ where: { id: pMergeId }, data: { householdId: memberHh } });
+
+            const res = await POST(mergeReq(pKeepId, pMergeId));
+            expect(res.status).toBe(200);
+            expect((await prisma.person.findUnique({ where: { id: pMergeId } }))?.mergedIntoId).toBe(pKeepId);
         });
 
         it("400s when the merged-away record's household is DENIED and the keeper's is NONE (restriction laundered)", async () => {
@@ -385,7 +419,7 @@ describe("Merge Participants API", () => {
             const res = await POST(mergeReq(pKeepId, pMergeId));
             expect(res.status).toBe(400);
             const data = await res.json();
-            expect(data.error).toContain("different Treehouse membership status");
+            expect(data.error).toContain("which restriction applies");
             expect(data.error).toContain("DENIED");
 
             // The login block is derived live from the person's household (authClaims),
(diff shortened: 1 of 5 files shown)


jee7s:
Both failure modes are real and I verified them. `OrgMembership.householdId` is `@unique` and the merge never rewrites it, so value genuinely can strand; and `authClaims.ts:34` derives the login block from the person's *current* household, so merging out of a DENIED household does launder the restriction on the next token refresh. Worth fixing.

The guard as written is symmetric, though, and both harms are not:

```ts
if (keepStatus === mergeStatus) return null;
```

That refuses on any difference, while every harm in the PR body requires the **merged-away** side to be the non-NONE one. Two consequences.

**It blocks the most common dedupe.** `createParticipantWithHousehold` gives every fresh Google sign-in its own household with no `OrgMembership` row, and makes that person its lead. So the duplicate reads NONE and is its own sole member, which clears the existing `isLead && householdOthersCount > 0` guard and then hits this new 400. Keeper ACTIVE against duplicate NONE is the archetypal case (a member's family creates a second account), nothing is stranded because a NONE membership is worth nothing, and no restriction is escaped. The UI disables Proceed to Preview, so there is no way through the tool at all.

**It still permits the case it was written to stop.** ACTIVE keeper against ACTIVE merged-away passes the equality check, and if that tombstone was its household's sole live member, the merge strands a paid membership. That is failure mode 1, allowed.

The matrix in the PR body labels NONE/ACTIVE as refuse-because-stranded, which holds in one direction only, while the code is direction-agnostic even though POST knows which id is the keeper.

An asymmetric rule keyed on the merged-away side covers both: refuse when the merged-away household holds a non-NONE membership that would be left with no live member, and refuse when either side is DENIED or REVOKED and the two differ. That keeps the DENIED-laundering block and the stranding block while letting the ordinary duplicate through.

One smaller thing: the error text says to "resolve the membership on these households first", and there is no UI for moving an `OrgMembership` between households, so the remedy it offers cannot be carried out.

---
_Posted by Claude._

thpr:
Both right, and the second one is the worse of the two — the guard permitted the exact failure it was written to stop. Fixed in 9e9a035a; the rule is now keyed on the merged-away side.

**Over-blocking.** Confirmed the mechanism: `createParticipantWithHousehold` (`src/lib/auth-options.ts:58`) creates the household, then the person, then `addHouseholdLead`, and never touches `OrgMembership` — so a fresh Google sign-in is `NONE`, sole live member, and its own lead. `isLead && householdOthersCount > 0` is false, and the equality check then took it. ACTIVE keeper against a NONE duplicate now merges.

**Under-blocking.** ACTIVE/ACTIVE with the merged-away record as its household's last live member now refuses. Equality never said anything about who is left to use the membership.

The two clauses:

- statuses differ **and** either is `DENIED`/`REVOKED` → refuse (restriction changes hands)
- merged-away household is `ACTIVE` **and** would be left with no live member → refuse (stranding)

One deviation from your wording. You said refuse when the merged-away household holds a **non-NONE** membership left with no live member; I narrowed that clause to `ACTIVE`. Taken literally, non-NONE would refuse `DENIED`/`REVOKED` on both sides — which clause 1 lets through as equal — and a live-empty `DENIED` household strands nothing of value. `DENIED`/`DENIED` and `REVOKED`/`REVOKED` now merge; say the word if you want them blocked instead.

Also worth flagging: the live-member count ignores tombstones. `householdOthersCount` in the existing lead guard does not, so a household whose other members are all merged away reads as populated there. I left that guard alone — different question, and changing it changes when the lead guard fires.

**Analyze.** The rule is asymmetric and the keeper isn't fixed until the UI scores and the operator swaps, so analyze now returns `membershipBlock: { aAsKeeper, bAsKeeper }`. The page applies the direction matching the current keeper and appends "Swapping which record is kept clears this" when the other direction is clean — which is the way through for the case you flagged as having none.

**Error copy.** You're right that there is no UI for moving an `OrgMembership`. It now names remedies that exist: assign the person to another household from the Participants page (`POST /api/membership-ops/participants/[id]/household`, the "Assign Household" modal), or set the household's membership on the Households page. The stranding message leads with "swap which record is kept", since that is usually the actual fix.

**Tests** — 10 now, including the ordinary-dedupe allow, ACTIVE/ACTIVE stranding refuse, all-tombstone household refuse, live-sibling-remains allow, and `DENIED`/`DENIED` allow. Dropping the old symmetric equality check back in fails 8 of them, covering both directions of your review. Full merge suite 42/42, `test:ci` 2421/2421, `authzRoleRejection` + `auditLog` + merge 168/168, `tsc` and `eslint` clean.

The PR body's matrix is updated to match.

jee7s:
**Area of concern: rebasing this against main will silently produce a type error if done by taking one side.**

DIRTY against both #1448 and #1449. #1448 already added `where: LIVE_PERSON` to `householdMembers` in the merge POST and analyze routes, which makes the `.filter(m => m.mergedIntoId === null)` here redundant — and keeping that filter while taking main's `select` (which no longer carries `mergedIntoId`) is a tsc failure, not a silent bug.

Two consequences for the body: the claim that `householdOthersCount` in the existing lead guard ignores tombstones is now stale, since #1448 fixed exactly that.

The current head has never run CI — the last run was against an earlier commit, before the asymmetric-rule rewrite.

---
_Posted by Claude._

thpr:
Rebased onto `origin/main` (48 commits) and force-pushed — `b45d8370` + `debc6079`. The rule is unchanged; what changed is how the live-member count is obtained.

**Correction to my last comment.** I said `householdOthersCount` in the lead guard counts tombstones. That was true of the base I wrote against and is not true of current `main`: #1448 put `where: LIVE_PERSON` on the `householdMembers` include in both the merge and analyze routes, for the same reason — it cites a legitimate merge falsely blocked by a tombstone inflating that count. Nothing to flag there any more.

**What the rebase changed here.** #1448 and #1449 (PERSON_BG dedupe) touch the same two routes and the same test file, so all three conflicted. Resolution kept both sides everywhere. Because the include is now `LIVE_PERSON`-filtered upstream, my own tombstone filtering was doing the same job twice, so I dropped mine rather than leave two mechanisms:

- the analyze select no longer pulls `mergedIntoId` on household members;
- both routes count live others straight off the filtered include, which in the POST route collapses to reusing the existing `householdOthersCount`.

The `counts only LIVE others` test still pins the behaviour — an ACTIVE household whose remaining members are all tombstones still blocks — it is just now enforced by the include rather than by a filter in the guard's caller.

Two things worth knowing about the merge itself: the 3-way merge kept **both** copies of a `let extraHouseholdIds` declaration that each side had added independently — a duplicate binding in one scope, sitting outside any conflict marker — and the test file's two new tombstone cases from #1449 land immediately before this PR's `describe` block. Both handled.

Re-verified on the new base, all foreground: `tsc --noEmit` and `eslint` clean; merge + `authzRoleRejection` + `auditLog` integration **174/174** (up from 168 — #1449 added merge cases); full `test:ci` **2515/2515** across 265 suites, including the widened `livePersonDriftGuard` from #1455 that now scans nested Person relation pulls like the ones this PR touches.

The open question from my last comment stands: I narrowed the stranding clause to `ACTIVE` rather than non-NONE, so `DENIED`/`DENIED` and `REVOKED`/`REVOKED` merge. Say if you want those blocked instead.

jee7s (approved):

Issue #1260, opened 2026-07-24
🔴 BG clearance stamps every household lead, not the person checked (SA1 — safety-critical)


thpr:
**Backlog item SA1** — `docs/backlog/INDEX.md`.

🔴 **CRITICAL SAFETY DEFECT.** `clearBackgroundCheck`'s household branch (`src/lib/membership/review.ts:305`) blanket-`updateMany`s the BG date onto **every lead** in the household. Clearing ONE adult therefore makes an **un-checked adult read as background-checked** — one person's check satisfies another's, so an uncleared adult supervising youth shows as cleared.

**Not moot.** The SA2 re-import fixed old *data*; the blanket stamp itself never stopped, so new pollution accrues on every household clearance. Design (proposed, verified vs main): `checkin-app/docs/designs/1260_BG_PER_ADULT_SUBJECT.md` — fix is on the write side; household rollups (membership needs ONE checked lead) are correct and stay. Unblocks SA3 (per-adult predicate).

_(This issue was briefly mis-scoped to 'enforcement/grace' during a decomposition — corrected. Enforcement/grace is now SA18; self-attest widening is SA17/#1452; consent is #961; supplier-affirmation needs clarification.)_

thpr:
**Backlog reconciliation (2026-08-01)** — PR #961 shipped the escalating student-nudge slice. **BLOCKER surfaced (interview):** the nudge's `/membership` self-attest (#875) keys off the household process — widening it to accept a **bare PERSON_BG** obligation is required, not deferred (silence-on-attest = core CS). Also #1412 landed a partial (confirm BG attestations + a way back). Source: INDEX SA1.

jee7s:
**Status: READY, PENDING #1470.** Moving to Iteration 2 / **In progress** — the implementation exists and is the right shape. What remains is one CI defect, three unanswered questions, and a scoping call on the remediation half.

**Verified against `origin/main`:** the defect is live. `checkin-app/src/lib/membership/review.ts`'s household branch still runs an unfiltered `tx.person.updateMany({ where: { householdId, isHouseholdLead: true } })`, while the `PERSON_BG` branch directly above stamps only its subject. The suppression loop is real: `personBgVerdict` returns `FRESH` for a falsely-stamped adult, `openPersonBg`'s guard (b) in `personBgTriggers.ts` then skips them, and nothing ever closes a stale stamp.

**Two corrections to the body:**
[...]

PR #1454, opened 2026-08-02
docs: design for per-adult background-check subjects (#1260)

thpr:
Design doc for #1260 — proposed, for review. No app code.

## The defect

`clearBackgroundCheck`'s household branch (`src/lib/membership/review.ts:305`):

```ts
await tx.person.updateMany({ where: { householdId, isHouseholdLead: true }, data: { lastBackgroundCheck: now } });
```

Two reviewers approve one household application and **every** household lead gets stamped — including an adult who never consented and whose PDF does not exist at Averity. They then read `FRESH` to every consumer, so they can volunteer with youth indefinitely while every surface reports them cleared.

It compounds: the false stamp makes `personBgVerdict` return `FRESH`, so `openPersonBg` skips them (dedup guard b) and they never reach `peopleNeedingBgCheck`. The machinery that would catch this is disarmed by the thing it was meant to catch. Each household cycle re-stamps them.

Known legacy — the policy block at `personBgCheck.ts:18-21` names it and defers it. The `PERSON_BG` branch six lines above stamps only its subject and is correct.

## Root cause

Nothing in the household path ever records **who** the check was for. `bgConsentAt` is a bare process-level timestamp with no person FK; the Averity link is one static org-wide URL with no per-person token. So there is no subject to narrow the `updateMany` to — it has to be captured somewhere first.

## Approach

Capture it where it is actually known: **the reviewer is holding the PDF, which names the person.**

- One nullable FK, `BackgroundCheckAttestation.subjectPersonId`. No new table.
- A checkbox group on the reviewer card — still one Approve click, selection is the only addition. Approval counts go per-subject.
- **The applicant's consent flow is unchanged.** No new controls, no new copy.

An earlier revision captured the subject at consent time via per-lead checkboxes and a join table. Superseded — a family's honour-system claim cannot catch "we submitted one form and assumed both were covered", whereas the reviewer-side capture does: the second adult simply never reaches two approvals. Both alternatives are recorded in the doc with why they lost.

## What deliberately does not change

Household freshness rollups stay exactly as written. **Membership requires ONE checked adult lead**; a second adult's check is a volunteer obligation, not a membership gate. An earlier revision of this design proposed flipping `householdBgIsFresh` to "every lead fresh" — that would have tightened the membership requirement beyond policy, and is called out in the doc as wrong.

With the stamp fixed, no new nag surface is needed: a volunteering second lead is already a `ProgramVolunteer`, so they are already in the population `openPersonBg` and `peopleNeedingBgCheck` scan. Removing the false stamp is sufficient, not merely necessary.

## Remediation — in-app, no script

The instance is in AWS and awkward to reach directly, and deciding which of two adults actually had a check is a judgement call against the PDFs that board members hold, not whoever has DB access.

Almost all of it already exists: `PUT /api/membership-ops/participants/[id]` accepts `lastBackgroundCheck: null`, is board-gated, and already writes a proper audit row. Only the **worklist** is new — one section on the existing compliance dashboard, reusing the `PersonSection` component it already renders twice.

Identification is exact, not heuristic: `clearBackgroundCheck` writes a single `const now` to both every lead's `lastBackgroundCheck` and the process's `bgClearedAt`, so equality holds to the millisecond. Likely-subject evidence comes from the `markBgConsent` audit actor. The dashboard labels; a board member clicks.

## Notes for review

- **Scope check:** SA2 was retired on "no polluted data exists" with a warning to re-verify when SA1 was scoped (`docs/backlog/INDEX.md:303`). Verified here — **it does not moot this.** SA2 was pre-import *data*; the blanket stamp itself never stopped.
- **Unique-constraint trap** flagged in the doc: `@@unique([processId, reviewerId, subjectPersonId])` with a nullable subject weakens duplicate protection, since Postgres treats NULLs as distinct. PG 15 is confirmed in both `deploy/docker-compose.yml` and CI, so `UNIQUE NULLS NOT DISTINCT` applies — but Prisma cannot express it, so it needs raw SQL plus a `compare-schema-dumps.sh` reconcile.
- **One deliberate departure** wanting sign-off: the compliance dashboard documents itself PULL-ONLY and would gain its first mutation. I read that property as being about *automation* ("the system never auto-revokes; a human must follow up"), which a per-row button does not violate — but that is my reading of someone else's design comment.
- **Three open questions** at the foot of the doc: the 2026-07-01 cutoff, how hard to work the "cannot tell" rows, and whether a reviewer may name a non-lead adult.
- No test tier currently covers this: both integration tests asserting the stamp use single-lead households, so nothing locks in the blanket behaviour and nothing would catch its removal. Test plan leads with the regression case.

Referenced without a closing keyword — #1260 stays open for the implementation PR.

### DIFF
--- a/checkin-app/docs/designs/1260_BG_PER_ADULT_SUBJECT.md
+++ b/checkin-app/docs/designs/1260_BG_PER_ADULT_SUBJECT.md
@@ -1,80 +1,71 @@
 # Per-adult background-check subjects (household path)
 
-Status: **PROPOSED — for review**
-Issue: [#1260](https://github.com/innovationtreehouse/checkin/issues/1260) · Backlog item **SA1**
-(ENHANCE · NEEDS-DESIGN · size L)
-
-**#1260 is now scoped to exactly this defect** (it was decomposed from the original five-slice SA1
-bundle). This design is the whole of it — an implementation PR should carry `Fixes #1260`; a
-design-doc PR references it plainly, no closing keyword.
-
-The other former SA1 slices, none of them dependencies:
-
-| Former slice | Now |
-|---|---|
-| cron cohort-open | **already shipped** — `api/cron/person-bg-annual/route.ts` calls `runPersonBgAnnualSweep` |
-| consent link/email | [#961](https://github.com/innovationtreehouse/checkin/issues/961) 18+ student nudges — PERSON_BG has no in-app consent orchestration; the only path is the board's manual `submitPersonBgForReview` → `markBgConsent` (`personBgSubmit.ts:14`) |
-| self-attest widening | [#1452](https://github.com/innovationtreehouse/checkin/issues/1452) (SA17) — self-attest must accept a bare PERSON_BG obligation; blocks #961 |
-| enforcement/grace blocking | SA18 — warn-only is by design (`backlog/CUJS.md:235`), board-gated by SA5. Explicitly **not** proposed here |
-| supplier affirmation | unresolved — no code, no backlog context beyond the INDEX line; needs clarification before anyone scopes it |
-
-**Resolves a flagged backlog TODO.** `docs/backlog/INDEX.md:303` retired SA2 with a warning: *"⚠
-likely also moots SA1's 'blanket-stamp→per-adult migration' sub-part — verify when SA1 is scoped."*
-Verified while scoping this: **it does not moot it.** SA2 was about pre-import pollution; the
-blanket `updateMany` never stopped running, so new pollution has accrued from every household
-clearance since. That line should be corrected, and INDEX's SA1 row now needs to match #1260's
-reduced scope.
-
-Related issues:
-- [#1429](https://github.com/innovationtreehouse/checkin/issues/1429) (SA11) **volunteer-onset BG
-  trigger** — the same second-adult case, fired when an adult *starts volunteering* rather than at
-  the annual sweep. Complementary: this design fixes the stamp so the second adult becomes visible;
-  #1429 makes them visible *sooner*. See [Read side](#read-side--unchanged-and-why). Its "when a
-  SECOND adult volunteers, BOTH adults need a check" wording needs reconciling against domain rule 1
-  — it is a volunteer trigger, not a membership requirement.
-- [#1396](https://github.com/innovationtreehouse/checkin/issues/1396) participant merge transfers the
-  BG date — adjacent to the merge-propagation note in [Remediation](#remediation).
-- [#1224](https://github.com/innovationtreehouse/checkin/issues/1224) `PERSON_AGREEMENT` — **design
-  only, never implemented** (the enum has no such kind). The agreement side of the same
-  "individuals, not households" shift. Not a dependency.
-
-Unblocks **SA3** (household-composition sweeps), which needs a per-adult predicate.
-Also see: `personBgCheck.ts` policy block (the rule this violates) · `HOUSEHOLD_LEAD_MODEL.md`.
-
-## Domain rules
-
-> **Canonical home is the rules register**, `docs/rules/membership.md` § *Background checks* and
-> § *Background checks for program-attached adults* ([#1445](https://github.com/innovationtreehouse/checkin/pull/1445),
-> open at time of writing). It carries the policy citations this restatement lacks. **Once #1445
-> merges, collapse this section to a pointer** rather than maintaining two copies — the register is
-> the single source, and a design doc restating rules is exactly the drift the register exists to end.
->
-> Restated here only because the design below is unreadable without them. The register sources them:
-> rules 1 and 3 to *Membership Policy, Art. VI §VI.1* ("At least one adult in each family is checked;
-> so is any other family member 18 or over who will be present regularly"), and rule 2 to *Volunteer
-> Policy, Art. IV* ("Every volunteer 18 or older is checked. The obligation attaches to the role, not
-> only to the household") — which is this design's thesis, in the policy's own words.
->
-> **One reconciliation point.** The register says "present regularly"; this design's population is
-> `ProgramVolunteer`-driven (rule 4). Those coincide only because there is no informal volunteering
-> here — confirmed, and recorded as rule 4. If that ever stops being true, the code's population is
-> narrower than the policy's and rule 4 is where it breaks.
-
-1. **Membership requires ONE background-checked adult lead.** Not all of them. A household with a
-   checked lead is compliant for membership purposes, full stop.
-2. **A second adult's check is a VOLUNTEER obligation, never a membership gate.** If the other lead
-   volunteers, they need their own check — that changes *how we nag them*, not whether the family
-   is a member. Posture is warn-only per the `personBgCheck.ts` policy block.
-3. **Checks are per-adult.** One person's check must never satisfy another's. This is the rule the
-   current code breaks.
-4. **There is no informal volunteering.** This is a youth organisation; every volunteering adult is
-   recorded as a `ProgramVolunteer`. The volunteer roster is therefore *authoritative*, not
-   best-effort — which is what lets the nag population below be treated as complete.
-
-Rules 1 and 2 mean the household-level rollups are **correct as written** and stay untouched — see
-[Read side](#read-side--unchanged-and-why). The defect is entirely on the write side.
+Issue: [#1260](https://github.com/innovationtreehouse/checkin/issues/1260) · Status: **PROPOSED — for review**
+· Scope/related work in the [appendix](#appendix--issue-scope-and-related-work).
 
 ## Problem
+
+When two reviewers approve a household's background check, the system marks **every adult lead in
+that household** as background-checked — not just the one whose check was actually reviewed.
+
+So an adult who has never been checked, whose report does not exist at the vendor, shows up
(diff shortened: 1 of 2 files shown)


thpr:
Both gaps confirmed against the code and folded in — `1c5175c6`.

**1. `overrideBlocked` approve — confirmed, and worse than the doc's legacy tail.**

Verified the mechanism: `approve` ([`:397`](https://github.com/innovationtreehouse/checkin/blob/main/checkin-app/src/lib/membership/review.ts#L397)) does not delete attestations, unlike `reset` ([`:385`](https://github.com/innovationtreehouse/checkin/blob/main/checkin-app/src/lib/membership/review.ts#L385)), so a BLOCKED process carries at most one APPROVE per subject and `subjectsWithTwoApprovals` can never reach 2. Empty set, every time.

The consequence is worth naming precisely, because it's an inversion rather than a shortfall: the board force-approves, `bgClearedAt` sets, the membership activates, and **no adult is stamped at all** — the household drops straight to `STALE_BG` holding a `bgClearedAt` with no person behind it. Today's blanket stamp at least stamps someone.

Resolved by having the override **assert** its subjects rather than count them — it already bypasses the two-reviewer rule for `bgClearedAt`, so it does the same for the stamp, via a `subjectOverride` parameter on `clearBackgroundCheck`. Same lead checkbox group on the `review-override` surface, same non-empty-or-400 requirement as `attest`. `reset` is untouched. The existing `hasHouseholdConflict` gate on `overrideBlocked` ([`:364`](https://github.com/innovationtreehouse/checkin/blob/main/checkin-app/src/lib/membership/review.ts#L364)) isn't widened by this. Three test cases added, asserting the *stamp* rather than the status — asserting status alone is exactly what would have missed this.

**2. Merge — correct, and the doc mis-filed it.**

`resolveKeeperUpdate` takes the newer date unconditionally with no provenance, which is precisely what this design says must not exist, and the write fix doesn't touch it. Filing it under one-time remediation with a "delete the section when empty" step was wrong: that sub-list refills on every future merge. Now scoped **permanent**, with step 6 narrowed to the blanket-stamp list only, and closing the hole explicitly deferred to #1396.

Cross-referenced the COI point there too — verified merge is `withAuth({ roles: ['isSysadmin','isBoardMember'] })` with no `hasHouseholdConflict`, so it's an unguarded route for a board member to place a BG date on someone in their own household. That belongs to #1396's design, not this one, but it shouldn't get lost.

**Minors**

- **`LIVE_PERSON`** — you're right that the present tense oversized it. Verified `merge/route.ts:210` is the only non-null `mergedIntoId` writer and sets `isHouseholdLead: false` in the same CAS, so tombstone-leads can only be residue from merges predating that. Reworded to "narrower than it looks and closed going forward… noted for completeness, not as live risk." Interested in the opposite-direction bug you chipped — if the lead guard counts tombstones, that's the same inconsistency picking the wrong side twice.
- **Split-subject stall** — promoted to a stated operational mode. Named as a failure that *cannot happen today*, accepted rather than designed away (binding reviewer 2 to reviewer 1's selection defeats the independence the two-reviewer rule exists for), with the mitigations spelled out: the `subjects[]` payload shows reviewer 2 that Alex sits at 1/2 before they choose, and force-approve resolves a stall outright. Plus a note that reviewers should name every adult whose PDF they read.
- **`PERSON_BG` denormalisation** — recorded as considered-and-not-taken, with your reasoning (doesn't remove the raw-SQL requirement). Worth having in the doc since it reasons explicitly about which rows carry null.

Nothing here changes the core model or the applicant-facing surface.

thpr:
Checked this design against the in-flight PRs — `a88359bb`. Nothing invalidates the model; three things needed reconciling.

**#1445 rules register — this doc should defer to it.**

`docs/rules/membership.md` § *Background checks* carries the rules with policy citations, and sources rule 2 almost verbatim: *"Every volunteer 18 or older is checked. The obligation attaches to the role, not only to the household"* (Volunteer Policy, Art. IV). That's this design's thesis in the policy's own words. Rules 1 and 3 come from *Membership Policy, Art. VI §VI.1*.

My hand-written Domain rules section now says so and is **marked to collapse to a pointer once #1445 merges** — a design doc restating rules is the drift the register exists to end. Kept for now only because #1445 is still open and the design is unreadable without them.

One reconciliation worth a second opinion: the register says checks cover *"any other family member 18 or over who will be **present regularly**"*, where this design's population is `ProgramVolunteer`-driven. Those coincide only because there's no informal volunteering here — now recorded as rule 4. If that ever stops holding, the code's population is narrower than the policy's, and rule 4 is the seam.

**#1448 / #1450 — same omission as `review.ts:305`, opposite consequences.**

I'd located the merge lead-count guard bug independently (`householdMembers: true` at `merge/route.ts:130`, consumed at `:158`, counting tombstones and *falsely blocking* legitimate merges) before seeing #1448 already fixes it with `{ where: LIVE_PERSON }`. With #1450 that's three lead-scoped queries missing the same filter, fixed three separate times, failing in opposite directions — over-blocking in merge, over-writing here.

Recorded as an open question rather than a change: should `householdLeadsWhere` bundle `LIVE_PERSON` outright, or should `livePersonDriftGuard.test.ts` cover lead-scoped queries? Three point fixes for one class suggests the predicate is the right place. Out of scope for this PR either way.

**#1449 — no impact, cited anyway.**

It moves `openPersonBg`'s dedup guard to a shared `personBgOpen.where` StateSet, same statuses. This design leans on that guard for the second-lead handoff to Triggers A/C, so I've noted the change; semantics are unchanged and the handoff holds.

**#1396 deferral still stands.** Verified none of the four in-flight merge branches (#1448/#1449/#1450/#1451) touch `resolveKeeperUpdate`'s newer-wins date rule — zero changed lines matching `lastBackgroundCheck` on any of them. So the merge remains an ongoing source of unearned dates, the remediation sub-list stays permanent, and closing it stays with #1396.

thpr:
Correcting my #1448 note above — `b20d2abb`. The mechanism was right; the framing and the proposed remedy were not.

**Mechanism (stands).** The guard blocks on `householdOthersCount > 0`, counting `householdMembers` loaded bare at `merge/route.ts:130`. Tombstones keep their `householdId` (the CAS nulls identity fields and clears `isHouseholdLead`, never moves the household), so they were counted. Over-counting a blocking guard over-blocks: merging a lead out of a household whose only other members are tombstones got a false 400 naming members that don't exist.

**Correction 1 — severity was flattened.** "Same omission, opposite consequences" implied equivalence. #1448 counted household *members*, and every merge creates one, so it was live and compounding. `review.ts:305` filters `isHouseholdLead: true`, which the CAS sets false, so it only ever reached pre-#1105 residue. Different severity classes.

**Correction 2 — the remedy I proposed was wrong.** Bundling `LIVE_PERSON` into `householdLeadsWhere` wouldn't have caught #1448: that's a `householdMembers` include and doesn't use the helper. Neither does `review.ts:305`.

**What's actually generalisable.** `livePersonDriftGuard.test.ts` exists for this class and matches:

```
/\b(?:prisma|tx|db)\.person\.(findMany|findFirst|count|aggregate|groupBy)\s*\(/g
```

plus nested pulls keyed on `person:`. Neither bug matches — `review.ts:305` is `tx.person.updateMany` (**no write verb is in the pattern**), and `merge/route.ts:130` is a `findUnique` (explicitly excluded) pulling a Person-typed relation named `householdMembers`. Neither was allowlisted because neither was ever seen.

So the class fix is extending the guard along those two axes: **write verbs**, and **Person-typed relations under names other than `person`**. That would have caught both, and is worth its own issue — the allowlist's value depends on the matcher seeing every site, and right now a whole verb category is dark.

dkaygithub (changes requested):
Can design docs get some coherent scoping?

Like : problem / objective / exec summary/ then design, then detailed design

I feel like Claude often jumps into appendix or almost alternatives considered without sharing what the problem is that we're solving in plain English

thpr:
Restructured — `8741a952`. The critique was right: the doc opened with issue-decomposition tables and a rules blockquote, so the actual problem sat ~90 lines down and a reader hit provenance and alternatives before learning what was being solved.

New order is **problem → objective → executive summary → design → detail**, with everything else moved to an appendix.

**Problem** is now first and in plain English — no code, no file paths, no line numbers:

> When two reviewers approve a household's background check, the system marks **every adult lead in that household** as background-checked — not just the one whose check was actually reviewed. So an adult who has never been checked, whose report does not exist at the vendor, shows up everywhere as cleared.

**Objective** — a background-check date on a person means that person was checked; nothing else should ever set it. Plus the two constraints: membership behaviour must not move, and applying families must not have to do anything new.

**Executive summary** is a table of outcomes rather than mechanism — what changes for reviewers (one checkbox group), for applicants (nothing), in the data, for membership (unchanged), how existing bad data gets fixed (board worklist, no script), and what it costs.

**Moved to an appendix:** the SA1 slice decomposition, the backlog TODO resolution, and the related-issue list. That's provenance — useful to keep, not context a reader needs before the problem statement.

The mechanism-level narrative is intact under "How it breaks today", and the design sections keep their detail.

Applied the same treatment to the sibling doc on #1456 (`7c7e890b`) so both read consistently.

Worth folding into the docs standard if this shape is what we want generally — `docs/DOCUMENTATION_STANDARD.md` is in flight under #1428, and "problem / objective / exec summary / design / detailed design" is a cheap section ordering to state there once rather than re-review per doc.

dkaygithub (approved):

PR #1455, opened 2026-08-02
fix(test): teach the LIVE_PERSON guard to see writes and renamed Person relations

thpr:
## What

The `LIVE_PERSON` drift guard (`checkin-app/src/__tests__/livePersonDriftGuard.test.ts`) could only see half the ways this codebase touches `Person`. Its matcher was **read-only** and **name-bound**:

```js
const PERSON_CALL_RE = /\b(?:prisma|tx|db)\.person\.(findMany|findFirst|count|aggregate|groupBy)\s*\(/g;
```

…plus nested pulls keyed on the literal `person:`. Two blind spots, each of which had already let a real bug through:

1. **No write verbs.** `lib/membership/review.ts` stamped `lastBackgroundCheck` on household leads via an unfiltered `tx.person.updateMany` — never matched, so never reviewed.
2. **Person-typed relations under other names.** The merge route pulled `householdMembers: true` (a `Person[]` relation), counted tombstones into its household-lead guard, and falsely blocked legitimate merges. Invisible to a matcher looking for `person:`.

## Widening

- **Class 1** gains `updateMany` / `deleteMany`. `update` / `delete` / `upsert` stay out for the same reason `findUnique` does — they take a *unique key*, so they touch the one row the caller already resolved and can neither sweep in a tombstone nor inflate a count.
- **Class 2** now matches every relation field typed `Person` / `Person[]`, with the name list **derived from `prisma/schema.prisma` at test time** rather than hand-written. A relation under an unobvious name is exactly the blind spot, so a hand-kept list would reopen it on the next schema addition. Today that resolves to 13 names (`person`, `householdMembers`, `subjectPerson`, `reviewer`, `leadMentor`, `user`, …); a new one is in scope the day it lands. A third assertion pins the parse so it can't silently go empty.

The header gains a **write rule**, because the read rule does not transfer: a read that sees a tombstone is a ghost in a list, but a write that hits one is silent corruption of a row nobody will look at again. So mass writes filter by default, and are allowlisted only when they are *tombstone machinery* (the merge CAS, which targets the tombstone on purpose) or *structural* (a write that must touch every row pointing at something, liveness irrelevant).

## Triage

Widening flagged 19 files. Every one was reviewed individually; nothing was blanket-allowlisted.

**13 fixed** — all `where` narrowings, no behaviour change for live rows:

| Site | Why it needed the filter |
|---|---|
| `/household`, `admin/broken-households`, `membership-audit/{compliance,households-missing-contact,unclaimed-households}`, `membership-ops/{applications,contacts,households}`, `membership/reviews`, `safety/trusted-adults`, `finance-ops/payments` | Human-facing household rosters and lead-contact lists — the sweep-vs-display rule's display side. A tombstone's rewritten `merged-*@deleted.invalid` address showed up as a person to contact; on `finance-ops/payments` the pull is `take: 1`, so a tombstone could shadow the real lead entirely. |
| `membership/request-payment-plan` | The lead set is an **authorization** check (`some(p => p.id === currentUserId)`); a merged-away identity must not satisfy it. |
| `webhooks/resend` | Stamped/cleared `emailUndeliverableAt` on tombstones (the ids came from a helper that already filters — now self-guarding at the write). |
| `lib/membership/review.ts:305` | The household-lead background-check stamp. Minimal `...LIVE_PERSON` only — its logic is being redesigned in #1454, deliberately not restructured here. |

**6 allowlisted**, with reasons:

- `membership-ops/participants/import` — **structural** write: `person.updateMany({ where: { householdId: source } })` reparents every row so the following `household.delete` can run. Skipping tombstones leaves the RESTRICT FK unsatisfiable and rolls the whole merge back. (The lead-cap reads in the same file *do* filter.)
- `membership/reviews`, `lib/membership/review.ts` — FK-pinned **to-one** relations (`subjectPerson`, `reviewer`): one row chosen by the FK, the same boundary as `findUnique`-by-id, and Prisma takes no `where` on a to-one include anyway. Whether a `PERSON_BG` process whose subject was merged away should still sit in the queue is a lifecycle question about the *process*, flagged `NEEDS REVIEW` rather than guessed at.
- `membership-ops/participants/merge{,/analyze}` — untouched here; already fixed by the in-flight #1448. Listed so this guard can land independently; **the two entries come out when #1448 merges** (the stale-entry assertion will say so).
- `safety/emergency-contacts` — `NEEDS REVIEW`, deliberately left unfiltered. It's the evacuation roster: `isPresent` is derived from each member's open `visits`, and a merge leaves a concurrently-open visit on the *tombstone*. Filtering would trade a duplicate ghost row for the chance of reporting a household absent while someone is still in the building. Which way a safety surface should fail is a product call, not a drift guard's.

## Testing

- `npm run test:ci` — 2422 passed, 257 suites. One oracle updated: `participantPiiMinimization.test.ts` pins `/api/household`'s exact query shape, which now carries the `where`.
- `npm run test:integration` — 1258 passed, 128 suites, against a fresh seeded Postgres.
- `npx tsc --noEmit` clean.

Independent of #1260 and #1448 — this fixes the guard, not the bugs it should have caught.

### DIFF
--- a/checkin-app/src/__tests__/livePersonDriftGuard.test.ts
+++ b/checkin-app/src/__tests__/livePersonDriftGuard.test.ts
@@ -193,10 +193,6 @@ const ALLOWLIST: Record<string, string> = {
 
     // ── unresolved: listed so the guard stays green, NOT because it is correct ──
     'app/api/safety/emergency-contacts/route.ts': 'NEEDS REVIEW — deliberately left unfiltered pending a product call. This is the evacuation roster: every member carries its open `visits` and `isPresent` is derived from them, and a merge leaves a concurrently-open visit on the TOMBSTONE (see lib/scan-service.ts). Filtering here would therefore trade a duplicate ghost row for the chance of reporting a household absent while someone is still in the building. Which way a safety surface should fail is not a call to make from a drift guard.',
-
-    // ── fixed in flight elsewhere (drop these when the PR lands) ───────────────
-    'app/api/membership-ops/participants/merge/route.ts': 'The `householdMembers: true` pull feeding the household-lead guard counts tombstones, so a legitimate merge can be falsely blocked — a real bug, already fixed by the in-flight PR #1448 (`where: LIVE_PERSON`). Listed only so this widened guard can land independently of it; remove the entry when #1448 merges.',
-    'app/api/membership-ops/participants/merge/analyze/route.ts': 'Same in-flight fix (PR #1448) for the analyze view\'s `householdMembers` pull, which must agree with the POST\'s guard; remove the entry when #1448 merges.',
 };
 
 // ── scanner (walk / captureObject copied from lifecycleStatusLiteralAllowlist.test.ts) ──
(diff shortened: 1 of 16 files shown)


jee7s (approved):

thpr:
Rebased onto main now that #1448 has merged.

- Dropped the two `merge{,/analyze}` allowlist entries — both routes filter `LIVE_PERSON` on main, so the entries were stale and the merge queue's stale-entry assertion (correctly) failed.
- That exposed a scanner false positive the widened name list made reachable: `merge/analyze/route.ts` documents its explicit select with "a plain `householdMembers: true` returns full Person rows one level down" — a relation name **inside a comment**, read as a site. The reverse would be just as wrong: a `LIVE_PERSON` mentioned in a comment reading as a query's filter. The scanner now masks comment bodies to spaces (same length, so every index still lines up) before running all three classes.

No other site changed status. Re-verified on the rebased tree: `npm run test:ci` 2473 passed / 262 suites, `npx tsc --noEmit` clean.

PR #1461, opened 2026-08-02
docs: design for merge-time background-check carryover (#1396)

thpr:
Design doc for #1396 — proposed, for review. No app code.

## The defect

Participant merge moves `Person.lastBackgroundCheck` to the survivor (`merge/route.ts:93`, newer-wins) but never re-evaluates the derived `OrgMembershipProcess.bgClearedAt`. The date says the household is covered; the process state disagrees, and an application can stay parked at background-check review.

## Why a doc and not a ~100-line PR

The issue's suggested shape — extract `clearBackgroundCheck`'s household block, call it inside the merge transaction — is right, and it is small. What sits underneath it is not.

**The intake-note clause is a policy question the code cannot answer.** `intakeNotes` is a bare `String?`; nothing records whether a human has ever read one. Two readings are defensible — a gate on the pre-payment review (what the code implements) or a standing "a human touches this household first" flag — and they disagree at exactly the two states holding the fix's entire value. The issue's "reuse the full predicate" advice silently picks one. Left open for the board, with a `noteAcknowledgedAt` field noted as the thing that would make it answerable in code.

**Provenance breaks in a way it doesn't for intake or renewal.** Those reuse a prior check from the *same* household, so the approving reviewers stay one query away. Merge pulls a date from the tombstone's household, which the survivor's history does not contain. Copying attestations is ruled out — they mean "reviewer X attested on THIS application" and a board reset deletes them by `processId`. The replacement is left open.

**It turns an ungated route into a privileged one.** The merge has no conflict-of-interest check, which is fine while deduplicating people is not a membership decision. The carryover makes it one, including a path to `ACTIVE`, while every sibling clearance path is COI-gated (`attest`, `overrideBlocked`, and #1391's rule that no role bypasses it). That is a correctness requirement, not a preference.

## What the analysis settled

- **The two-sided state space collapses.** Both merge subjects can hold in-flight processes; the cross product reduces to "the keeper's household process, always" — but only via the existing pre-transaction lead guard, and the collapse is the result of the analysis rather than a safe starting assumption.
- **The state the issue names is the one state the fix must never touch.** For a household `INITIAL`, `PENDING_BG_REVIEW` is reachable only when held for a note. The fix's value lives in the parallel-track states the issue does not mention.
- **The lifecycle-machine surface is not a driver.** It reduces to a single new edge. Recorded because the first read suggested otherwise.

## Four live bugs found while scoping, all unrelated to #1396

Each is tracked separately; none are introduced by this design.

1. **Merge never compares the two households' membership states** — more serious than #1396 itself. `OrgMembership` is 1:1 with `Household` and the merge never writes it. A paid `ACTIVE` membership can be stranded on a household with no live members, and a `DENIED`/`REVOKED` household can be escaped by merging out of it, since the login block is derived live from the person's household (`authClaims.ts:34`).
2. **`LIVE_PERSON` is applied inconsistently**, so a household left with only tombstoned members matches `BROKEN_HOUSEHOLD_WHERE` by construction — a permanent, unactionable board to-do that also reaches the keyholder-facing emergency sheet.
3. **The merge lead guard counts tombstones**, spuriously refusing valid merges with a 400.
4. **A merge can leave one person with two concurrent `PERSON_BG` reviews** — the re-point is blind, and no unique index covers `subjectPersonId`.

## Notes for review

- **Three open questions** at the foot of the doc: what an intake note obliges (the one that decides most of the fix), provenance, and emails. All are board calls; the doc gives recommendations for two and deliberately none for the first.
- **Interaction with #1454** (per-adult BG subjects): both rewrite `clearBackgroundCheck`'s household branch, so a textual conflict is certain. They compose semantically — #1454 makes it stamp named subjects, this carryover deliberately stamps nobody since it reuses an existing date rather than minting `now()`. #1454 should land first and this rebase onto it.
- **Emails are the sharpest reversible decision.** From `PENDING_BG_CLEARANCE` the carryover reaches `ACTIVE`, which no existing shortcut path can — so a dedupe cleanup could mail a family "Welcome to the Treehouse". Recommendation is to fire it rather than activate silently, but it wants explicit sign-off.

Referenced without a closing keyword — #1396 stays open for the implementation PR.

### DIFF
--- a/checkin-app/docs/designs/1396_MERGE_BG_CARRYOVER.md
+++ b/checkin-app/docs/designs/1396_MERGE_BG_CARRYOVER.md
@@ -1,0 +1,691 @@
+# Merge-time background-check carryover
+
+**Issue:** [#1396](https://github.com/innovationtreehouse/checkin/issues/1396) — participant merge
+transfers `Person.lastBackgroundCheck` to the survivor but never re-evaluates the derived
+`OrgMembershipProcess.bgClearedAt`, so an application stays parked at background-check review.
+
+**Status:** design. No production code accompanies this doc.
+
+## Why a doc and not a PR
+
+The issue's suggested shape ("extract `clearBackgroundCheck`'s household block, call it inside
+the merge transaction") is right and is ~100 lines. Five things sit underneath it that a PR
+description can't hold:
+
+0. **A merge has two sides, and the issue reasons about one.** Both people can sit in households
+   with in-flight processes, and both can hold an open `PERSON_BG`. That cross product does
+   collapse — to "the keeper's household process, always" — but only via a pre-transaction lead
+   guard nobody would think to look for, and the collapse is the *result* of the analysis, not an
+   assumption you may start from (see [Two sides, not one](#two-sides-not-one)). Working it
+   through also turns up four live bugs unrelated to #1396 — one of them (the merge never compares
+   the two households' membership states) more serious than #1396 itself.
+1. **The #900/#907 note clause is a policy question the code cannot answer.** Eight non-terminal
+   statuses can carry `bgClearedAt = null`; the carryover acts on three. Whether the note clause
+   applies at each depends on what the board *means* by a note — a gate on the pre-payment review,
+   or a standing "human touches this first" flag — and nothing in the schema records whether a
+   human ever read one. The two readings disagree at exactly the two states holding the fix's
+   entire value (see [The predicate](#the-predicate)). The issue's "reuse the full predicate"
+   advice silently picks one, and picks it wrong at the state it names.
+2. **Provenance breaks in a way it doesn't for intake/renewal.** The merge is the first
+   `bgClearedAt` stamp whose originating approval is not in the stamped household's own history.
+   Open question, deliberately unresolved below.
+3. **It turns an ungated route into a privileged one.** The merge has no conflict-of-interest
+   check because deduplicating people isn't a membership decision. The carryover makes it one —
+   including a path to `ACTIVE` — while every sibling clearance path is COI-gated
+   (see [Conflict of interest](#conflict-of-interest)). That is a correctness requirement the issue doesn't
+   mention, and it is the reason this can't be a quiet 100-line patch.
+4. **The emails and the audit contract are decisions, not defaults.** The merge is the first path
+   that can flip a household to `ACTIVE` as a side effect of an operator's data cleanup.
+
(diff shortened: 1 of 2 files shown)


jee7s:
Content-wise this is the kind of design doc I want to see — four bugs found, three decisions escalated rather than guessed, and a clear account of why #1396 is a design rather than a 100-line PR.

One thing blocks it, and it is a `git mv`.

The file lands at `checkin-app/docs/designs/1396_MERGE_BG_CARRYOVER.md`, but `docs/DOCUMENTATION_STANDARD.md` §4 says working docs live at `docs/in-design/<name>.md`, and `docs/designs/` is a directory the migration plan deletes. Your own sibling #1457 was moved to `docs/in-design/TOMBSTONE_REMOVAL.md` during review, so this is just the one that slipped through.

Move it to `docs/in-design/` and drop the issue-number prefix to match, and I have nothing else.

---
_Posted by Claude._

thpr:
Moved to `docs/in-design/MERGE_BG_CARRYOVER.md`, prefix dropped. Rebased onto current `main` while I was in there — the branch was 18 behind.

It turned out to be more than the `git mv`, because §4.1 of the standard you cited also specifies how a working doc opens, and `BG_PER_ADULT_SUBJECT.md` follows it. So the doc now does too:

- **Problem** rewritten in plain English. It previously opened on `Person.lastBackgroundCheck` and `OrgMembershipProcess.bgClearedAt` — which §4.1 calls out precisely: that describes the symptom in the source rather than the problem.
- **Objective** and **Executive summary** added.
- **Rules this design obeys** added, naming the four it relies on, per §5.
- **Provenance moved to an appendix** — the "why a doc and not a PR" argument, the live bugs found while scoping, and related work. All of it justifies the design to a reviewer without explaining it to a reader, and above the problem it displaced it.

48 relative links repointed for the new depth and root; anchors and paths verified to resolve. The related-work entry for #1260 now points at its merged doc rather than the issue.

Two notes while I was adjacent to it:

- `LIFECYCLE.md` is still at `checkin-app/docs/designs/LIFECYCLE.md` on `main`, so references to it are linked there rather than assumed migrated.
- The appendix records the sequencing against #1260: both rewrite the same household clearance branch, so that one lands first and this rebases onto it. They compose semantically — that design makes clearance stamp named subjects, this carryover deliberately stamps nobody.

jee7s (approved):

PR #1470, opened 2026-08-03
fix(membership): stamp only the adult a background check actually covered

### LINKED ISSUE #1260: 🔴 BG clearance stamps every household lead, not the person checked (SA1 — safety-critical)

thpr:
Fixes #1260. Implements `docs/in-design/BG_PER_ADULT_SUBJECT.md`.

## The defect

Clearing a household's background check stamped **every** household lead:

```ts
await tx.person.updateMany({ where: { householdId, isHouseholdLead: true }, data: { lastBackgroundCheck: now } });
```

So an adult who had never been checked — whose report does not exist at the vendor — read as cleared everywhere. The membership consequence is nil (the household really does have one checked adult, which is all membership requires). The safety consequence is the whole defect: that adult can volunteer with youth indefinitely while every surface reports them cleared.

Worse, the false stamp hid them from the machinery meant to catch it. `personBgVerdict` returned `FRESH`, so `openPersonBg` skipped them, so neither the annual sweep nor new-member activation ever opened a `PERSON_BG`, so they never reached `peopleNeedingBgCheck` — and the next household clearance refreshed the lie. The loop re-armed every cycle.

## Why it wasn't a one-line fix

Nothing in the household path had ever recorded **who** a check was for, at any step. One consent checkbox per household, one timestamp with no person attached, and a single org-wide vendor URL with no per-person identity. There was no recorded subject to narrow the `updateMany` to.

The subject now comes from the one place it already exists: **the reviewer is holding the vendor's report, and it names the person.** Recording that is one extra click on a screen only board members see — and costs applying families nothing. Applicants see **no change at all**.

## What changed

**Model.** `BackgroundCheckAttestation` gains a nullable `subjectPersonId` FK. No new table. Null on `PERSON_BG` (the process already names its subject), on any `REJECT` (whole-process), and on legacy rows.

**Reviewer surface.** The card gains one checkbox group listing every live household lead with that lead's own approval count. Still one Approve click; one attestation row per named subject, so a reviewer holding two reports for one family records both at once.

**Clearance.** An adult clears at two approvals; clearance stamps exactly those adults. `bgClearedAt` still fires on the **first** adult to reach two, so activation timing, `applyVolunteerStatus`, the `FOR UPDATE` lock, and the household freshness rollups are all untouched. Membership requires one checked adult, not all of them — this must not tighten that, and a test pins it.

**Naming is mandatory.** Approving a household without naming anyone is a 400, not a silent no-op — that is the blanket-stamp bug in human form.

**Board force-approve names its own subjects.** This is not a legacy tail. A `BLOCKED` process carries at most one `APPROVE` per subject (a second would already have cleared it), so counting attestations there yields the empty set *every time*. Left alone, the override would set `bgClearedAt` with no adult behind it — strictly worse than the behaviour it replaces, since today's override at least stamps someone. So the override asserts its subjects directly, and an unnamed one is the same 400.

**Eligibility goes per-subject.** A reviewer who named one lead still has the other to name, so the application stays in their queue. `approvedAwaitingSecond` now excludes what is still actionable — otherwise a half-named row would be counted both green and grey.

## Accepted new failure mode

If reviewer 1 names only Alex and reviewer 2 names only Sam, neither reaches 2/2 and the review stalls awaiting a third reviewer. This **cannot happen today**.

Not designed away on purpose: the alternative is letting one reviewer's subject choice bind the other's, which defeats the independence the two-reviewer rule exists for. Two things keep it from being silent — the card renders per-adult counts, so reviewer 2 sees Alex at 1 of 2 before choosing, and the board's force-approve resolves any stall outright. There is a test asserting the stall so it stays deliberate rather than becoming a surprise.

## Remediation (the data already in the database)

The write fix stops new pollution. It cannot undo existing false stamps, and **no script can** — nothing in the database records who was actually checked, so only a person holding the vendor reports can tell a real stamp from a fabricated one.

So the board gets a worklist on the compliance dashboard they already use.

**The join key is exact, not a heuristic.** A single `const now = new Date()` wrote both `bgClearedAt` and every lead's `lastBackgroundCheck`, so equality to the millisecond identifies a blanket stamp. Three classes fall out on their own and need no special-casing: single-lead households (never two matching leads), the fresh-check intake/renewal shortcut (stamps `bgClearedAt` without touching any `Person`), and `PERSON_BG`. A clearance made under the new per-adult rules is excluded by carrying a subject-named attestation.

**Evidence is labelled, never acted on.** `markBgConsent` writes one audit row per process whose actor is whoever recorded consent — the applicant themselves on the primary path, a board member on the backstop. A lead who attested their own consent is the best available signal of who went to the vendor, so the row says so. Nothing is pre-selected, nothing is bulk, nothing is automatic.

**Merge-inherited dates get their own permanent section.** `resolveKeeperUpdate` takes the later of two dates unconditionally with no subject provenance, so every future merge can mint a date that traces to nobody. That list does not empty and this PR deliberately does not redesign the merge rule — the hole belongs to #1396.

**Cutoff is a control, not a constant.** Defaults to the board's confirmed 2026-07-01 and stays editable, so the real distribution of clearance dates can widen it without a redeploy.

### Deliberate departure worth signing off

The compliance route documents itself as PULL-ONLY, and this adds the page's second action button. Confirmed with the requester before building. It is not a new capability: it calls the pre-existing board/sysadmin-gated `PUT /api/membership-ops/participants/[id]` with `lastBackgroundCheck: null` — the same thing the participants edit modal already does, same permission, same audit row. One person at a time; no process and no `bgClearedAt` is ever touched.

Blast radius of a wrong click is small and shapes how hard the board should work each row: clearing the wrong lead cannot cost a membership (the household keeps its other checked adult), so the worst case is a redundant re-check request.

## Migration

Additive: nullable FK, no backfill, no `NOT NULL` on an existing column. Old code ignores it through the whole rolling-deploy drain window; new code reads subject-less rows as legacy.

**The unique index is hand-written SQL and must stay that way.** Replacing `@@unique([processId, reviewerId])` with the three-column version weakens duplicate protection for legacy and `PERSON_BG` rows, because Postgres treats `NULL`s as distinct — two `(process, reviewer, NULL)` rows would both be allowed, letting a `PERSON_BG` reviewer attest twice. `NULLS NOT DISTINCT` restores it, Prisma cannot express the attribute, and `migrate diff` cannot see it — so a regenerated migration would silently drop it. Called out in the migration's header comment so the next `coalesce-migrations` run does not quietly regenerate a plain unique.

Verified against a real Postgres 16: the attribute survives `migrate deploy`, and `prisma migrate diff --from-config-datasource --to-schema --exit-code` reports **no difference**, so the CI drift gate stays clean.

The in-transaction `already_attested` check under the existing `FOR UPDATE` process lock is what actually serializes concurrent attestations; the index is defence-in-depth. Please do not let it quietly become less than that.

Index name is `map:`-pinned — the generated name would be 67 characters, past Postgres's 63-character identifier limit.

## Deviations from the design doc

**The reviewer payload shape.** The doc specifies a derived `subjects: [{ personId, name, approvals, isFresh }]`. That cannot work: the security stripper builds each response row only from fields present in the generated classification map, so an invented key is dropped silently. Instead the route selects `attestations: { subjectPersonId }` and the card counts per lead. The doc's actual goals hold — per-subject counts are visible, and reviewer identity never leaves the server. Cost is the `returns` entry in #1469. `isFresh` dropped; nothing renders it.

**One behaviour change the doc doesn't mention.** A household with zero live leads now drops out of every reviewer's queue, because there is nobody to name. It stays visible to the board on the applications view. Broken data either way — the alternative was leaving an un-actionable card in the queue forever.

## Tests

The bug was invisible to the suite: both integration tests asserting the stamp use **single-lead** households, so nothing locked in the blanket behaviour and nothing would have caught its removal. Every new case uses a two-lead household.

New file `bgPerAdultSubject.integration.test.ts` covers:

- **The regression itself** — both reviewers name only Alex ⇒ Alex stamped, **Sam's `lastBackgroundCheck` stays null**. This is the test that would have caught the report.
- Membership unaffected — same scenario still reaches `PENDING_PAYMENT`, `bgClearedAt` set, `householdBgIsFresh` still true.
- The nag turns on — Sam now reads `NEEDED` where they previously read `FRESH`.
- Both named ⇒ both stamped; split subjects ⇒ nobody stamped and the process stays queued.
- Empty selection ⇒ 400 (including a junk-typed body); naming a non-lead ⇒ 400; a rejection names nobody.
- Per-subject eligibility, including that naming the same adult twice is still refused.
- A subject-less attestation counts toward nobody.
- Force-approve stamps its named subjects and no others; unnamed ⇒ 400; `reset` unchanged.
- Remediation: inclusion rules (single-lead, shortcut, and per-adult clearances all excluded), evidence labelling, the cutoff filter, and merge-inherited detection.

**Results.** Full unit suite 2440 passed / 257 suites. Full integration suite 1286 passed against a local Postgres 16.

One integration failure, **pre-existing and unrelated**: `Program Age Bounds › should block a participant who only turns minAge tomorrow`. Confirmed it fails identically on a stashed clean tree at the same base commit — not touched by this PR.


jee7s:
The write-side fix is the right one and I would want it. `clearBackgroundCheck`'s household branch blanket-stamps `lastBackgroundCheck` onto every lead, so an unchecked adult reads as cleared everywhere and is simultaneously hidden from `openPersonBg` and the annual sweep. That is a safety hole that re-arms itself, and it matches backlog SA1 / #1260.

Two things before it merges.

**The integration failure is a real defect in this PR, not flake or main breakage.** It fails `programAnnounceRecipients.integration.test.ts` › "with no date boundary configured" with `Received array: []`. Root cause:

`membershipValidThrough` returns `null` when `BoardSettings.orgMembershipYearBoundary` is unset, and `coversThrough` treats null as covered. That suite's top-level `beforeAll` never sets the field its own name promises; only the nested describe manages it. So the test inherits whatever the previous suite left in the shared `--runInBand` database. Several boundary-writing suites restore with `if (saved) update(...)`, which never deletes a row they created, and `personBgTriggers.integration.test.ts` leaks a Sept-1 seed forward. With that leaked value and today's clock, `nextBoundary(Sep 1) = 2026-09-01`, which is earlier than the fixture's `endAt` of 2026-12-01, so every covered household fails the duration check and the recipient list comes back empty.

This PR trips it because it changes the byte size of six integration suites and adds a seventh, and jest's default sequencer orders by file size. A re-run of the same commit will not clear it. The smallest correct fix is to have that top-level `beforeAll` set the boundary it claims (null) rather than inherit it, which also closes the hole for every future PR that shifts suite order.

Note the PR body attributes the red run to a different pre-existing failure, so this one has probably not been seen yet.

**The remediation half may be unnecessary.** `docs/backlog/INDEX.md:303` records SA2 as dropped because "no polluted data exists; everyone was re-imported per-adult after the DB change" (2026-07-21), and explicitly warns that this "likely also moots SA1's blanket-stamp to per-adult migration sub-part, verify when SA1 is scoped." This ships that worklist with a cutoff defaulting to 2026-07-01, which predates the re-import, plus a second action button on a route the PR itself notes is documented PULL-ONLY. If the re-import did clean everything, the only real pollution is the couple of weeks since #1260 was filed. Worth confirming before carrying it on an already 1,325-line diff.

Also worth an explicit sign-off rather than leaving it implied: the new failure mode where two reviewers naming different subjects stall a review until a third reviewer arrives.

---
_Posted by Claude._

jee7s:
**Area of concern: nothing technical — two decisions.**

The integration failure is fixed: `2696ab10` sets the boundary that `programAnnounceRecipients` never set, and the current head is fully green including both migration jobs.

What remains is scope. `docs/backlog/INDEX.md:320` carries an 2026-08-01 correction confirming the "SA2 moots SA1" speculation was wrong for the code defect — so the write fix stands. But the same line still records that the 2026-07-21 re-import cleared pre-existing pollution, which makes the worklist's `2026-07-01` default cutoff wider than the real window (post-re-import clearances only). A count settles whether the dashboard section is warranted at all.

Still outstanding: the sign-off you asked for on the split-subject stall, where two reviewers naming different subjects leave a process waiting on a third.

---
_Posted by Claude._

thpr:
"The remediation half may be unnecessary. " I don't intend to directly write SQL against the live DB and set Claude with that constraint.  If someone else wants to do that, we can change the direction and scope here.  We just need an owner for that.

jee7s (approved):

thpr:
## One register file to update before this merges

This PR closes a divergence the domain register still records, and its diff does
not touch `docs/rules/`.

`docs/rules/membership.md` ends with `## Policy requirements not yet enforced
here`, whose first entry is:

> **A household clearance stamps every lead, not the person who was checked.**
> The household path carries one consent and one clearance for the whole
> application, then writes that date onto each of its leads, with nothing tying
> it to whoever submitted. This answers the household question correctly — policy
> wants one adult checked — but the date is a per-person fact, and the
> program-attached surface reads it per person. A lead who never took a check
> reads there as covered rather than needed.

That is exactly what this PR fixes, so **the entry should be deleted in this
PR** — the register is meant to describe what is true now, and leaving it says
the blanket stamp is still live to anyone who reads the file after you merge.

Nothing needs adding in its place. The Procedure rule this contradicted is
already in the file and becomes true for the first time when you land:

> A check covers the adult who took it. One person's clearance never satisfies
> another's.  `[Decision]`

Two things worth checking against your diff while you are in there, both stated
as rules in the same file and both easy to break by accident:

- *"A clearance is dated when the second reviewer clears it, and the validity
  window runs from that date."* Your description says clearance still fires on
  the first adult to reach two approvals, so this should be unaffected — worth a
  glance since you changed what "clears" means per person.
- *"A household that has settled its dues but is still waiting on its background
  check is a member for program pricing and for reaching members-only programs,
  and for nothing else."* Your note that membership requires one checked adult
  and this must not tighten it is the same invariant from the other side.

Context: found by the domain-register audit on #1528. The appendix sections are
being retired — where an issue owns a gap, the issue is the record, and where a
PR closes one, the PR deletes the entry. This is the second kind.

Issue #1462, opened 2026-08-02
Record program withdrawal with history (delete-vs-withdraw exit state on ProgramParticipant)


thpr:
## Problem

Withdrawal from a program **hard-deletes** the `ProgramParticipant` row. `withdrawAndReleaseHold` ([capacity.ts:66](https://github.com/innovationtreehouse/checkin/blob/main/checkin-app/src/lib/program/capacity.ts#L66)) is the single exit path for manual withdrawal, the non-payment kick, and grace-period auto-withdrawal, and all three `delete()`. So a person who left is indistinguishable from one who was never there — there is no history to read.

That in turn feeds false positives downstream: `PROGRAM_ATTACHED_WHERE` ([personBgTriggers.ts:19](https://github.com/innovationtreehouse/checkin/blob/main/checkin-app/src/lib/membership/personBgTriggers.ts#L19)) can't exclude people who withdrew, so `PERSON_BG` obligations and the board compliance chase list ([compliance/route.ts:94](https://github.com/innovationtreehouse/checkin/blob/main/checkin-app/src/app/api/membership-audit/compliance/route.ts#L94)) keep chasing them, and we can't write honest copy explaining *why* someone was flagged.

## Scope (agreed)

This issue is **the withdraw exit state only** — recording a real withdrawal with history, on the participant row. It is deliberately narrow. Removal is **not** always a withdrawal:

- **Delete (never-there)** — program not started yet, or an early no-show. The person effectively never participated; no history is worth keeping. Keep the existing `delete()` and its exactly-once hold-ledger boundary ([capacity.ts:56](https://github.com/innovationtreehouse/checkin/blob/main/checkin-app/src/lib/program/capacity.ts#L56)) **unchanged** — no new guard needed on this path.
- **Withdraw (was-here-then-left)** — real participation, then exit. Record `WITHDRAWN` / `endedAt` and stop deleting.

### Which path, per caller

The `PENDING | ACTIVE` status already carries the "were they ever really in" signal, and **both automatic crons only ever touch `PENDING` rows:**

- **Pre-start / non-payment kick** (`cron/pending-participants`) → **delete**. Never paid, never active.
- **Grace-expiry** (`cron/scholarship-grace-expiry`) → **delete**. The row sits in `PENDING_HELD_DENIED` the whole time — denied a plan, never paid, timed out. Never participated.
[...]

PR #1477, opened 2026-08-03
feat(membership): individual agreement for adult children (#1224)

### LINKED ISSUE #1224: Age-based individual agreement per adult child (≥18, child-of-signer)

thpr:
Fixes #1224

## Problem

A household signs **one** membership agreement, and a minor child is covered by it. Once a child turns 18 they are a legal adult and can no longer be bound by a parent's signature — they need their own. A spouse does not: a spouse stays on the household agreement.

## What ships

A new `OrgMembershipProcess` kind, `PERSON_AGREEMENT`, mirroring `PERSON_BG`: `orgMembershipId` null, `subjectPersonId` set. **No new columns** — the migration is a single additive `ALTER TYPE`.

**Opened automatically for** a household member who is not a lead, is 18–25 with a date of birth on file, is in a member household, and is attached to a program that is running or ended within the last year. Nightly, idempotent, one per person per membership year.

**The age band is the spouse guard.** No field records spouse-vs-child. But `isDeclaredAdult` means "over 25, no DOB" — set at intake and stamped by the nightly purge as people cross 26. So a non-lead adult **over 25** is a spouse the household never marked as a lead; a non-lead **18–25 with a DOB** is a child who turned 18.

**Signing:** one resolver shared by `getOrCreateContractSigningUrl` and `syncContractStatus`, checking `isHouseholdLead` **first** so a stray individual agreement can never shadow a lead's household process and stall activation. A non-lead signing their own bypasses the household lead gate — the one gate that opens here. `markContractSigned` flips it straight to `ACTIVE` and skips `advanceExternalIfComplete`, which would deref a null `orgMembershipId`.

`syncContractStatus` had to change too, and it is easy to miss: it resolves through `household.orgMembership.processes`, which **can never contain** a `PERSON_AGREEMENT`. Left alone, an adult child who signs and returns completes only if the unreliable Zoho webhook fires; otherwise the card keeps demanding a signature already executed and a re-click mints a second ceremony.

**Surfaces:** a sign card on `/membership` for the subject; outstanding and over-25-not-requested lists plus a manual request button for the board. **Gates nothing.**

## Deltas from the design doc

The design doc (`1224_PERSON_AGREEMENT.md`) is **deleted** in this PR, per the documentation standard — a per-change design doc is a working doc, and at merge its standing rules move into `docs/rules/membership.md` and the doc goes. That means this section is the only record of how the built thing differs from the design that was merged, so it is complete rather than a summary.

| Design as merged | What ships | Why |
|---|---|---|
| Population: program-attached, ≥18, non-lead | Adds **18–25 with a DOB on file** | The design had no way to exclude a spouse and accepted flagging them. The age band is the guard the data can actually express. |
| Age judged **as of the member-year boundary** (copied from `PERSON_BG`) | Age judged **as of now** | Boundary-relative age is only safe for a once-a-year sweep. Run nightly it flags a 17-year-old whose birthday merely falls before the next boundary — asking a minor to sign, the exact failure this prevents. |
| An **annual sweep** at the boundary | **Nightly**, on the existing `api/cron/nightly` | An annual pass misses everyone who starts qualifying after it fires: a 19-year-old added to a program the day after the boundary would wait a year. Also avoids a new cron route, which would ship inert — cron scheduling lives out-of-band. |
| **12-month `Program.startAt`/`endAt` window** + a NULL-date rule | **Kept** (dropped, then restored in review) | I dropped it arguing the age band bounded the damage. It does not: the band bounds *who* is asked, not *how many times*. Attachment rows are never cleared when a program ends, so the unbounded predicate re-asks someone who took one class at 18 every cycle until they turn 26. Restored, NULL-date rule included. |
| Dedup from the boundary | Dedup from **boundary − renewal lead window** | Mirrors `settledThisCycleWhere`. Without it, someone who starts qualifying two weeks before a boundary signs and is asked again days later. |
| "Build both manual buttons" | Only the **agreement** button is new | The BG one already exists on the same page (`Record external check & submit`). |
| Manual path unspecified | Drops the program-attachment requirement **and** the age ceiling; keeps the lead refusal and requires a known age | That is the button's purpose — reaching someone the automatic rule cannot. The lead refusal is not cosmetic: an open agreement on a lead wedges the household signing flow. |
| Open question: family-facing "you can ignore this" copy | **Dropped as undeliverable** | No class year, no graduation state, no attachment history — the system cannot tell someone who left from someone who stayed. The only honest version is a generic caveat shown to everyone, which trains people to ignore a legal-agreement request. |
| Open question: "mark your spouse as a lead" copy | **Unnecessary** | The age band excludes over-25 non-leads from the automatic population, so the case does not arise there. |
| Response shape unspecified | Route goes through `defineRoute()` + `handler()`, returning `id/kind/status` only | A new route may not use old authz. The first cut returned the whole process row, carrying `subjectPersonId` and both Zoho ids out for no reason. |
| — | **New:** each board surface explains why it disagrees with the other | Added in review. See below. |

## Two board surfaces that disagree on purpose

`/membership-audit` now carries both the 18+ roster (#1495) and these agreement lists, over structurally the same population — both are 18–25, since the DOB is stripped at 26. They judge age differently **on purpose**: the roster as of the member-year start, the agreement lists as of today. A board member reading them back to back would see different people and no explanation.

Each surface now names the other and says why, and the rule and its consequence are recorded in `docs/rules/membership.md`. The reciprocal note on the roster is in **#1495**, which has not landed yet — until it does, the link from Compliance points at a route that does not exist.

## Rules extracted

`docs/rules/membership.md` gains, under Procedure, the standing rules a later change could violate: who signs their own agreement, that it re-signs yearly, the age reference and its deliberate divergence from the background-check rule, the automatic population and the over-25 exclusion, the lead refusal, the manual open and its two refusals, and that it gates nothing.

The Assumptions section gains the one the whole distinction rests on: **both signing adults are marked as household leads**, because the app records no relationship and lead status plus age is all that separates a spouse from an adult child. That replaces the old assumption that adult children are handled entirely by hand, which this work makes untrue.

## Landed prerequisites

All merged; nothing outstanding.

- #1467 — moved `PROGRAM_ATTACHED_WHERE` to `person/filters.ts`. Without it this feature's new module closes a `const` edge across an existing import cycle and dies at module init.
- #1468 — stopped mock implementations leaking in `external.test.ts`. Without it the `not_lead` guard test silently stops guarding the moment the signing path calls `findFirst` — which is what this PR does.
- #1489 — the registry entry, isolated per the boundary rule.

## Migration safety

Checklist run. Additive enum value; not destructive, single statement, no new fields, no `@sensitivity` change. Non-TS readers grepped to zero. Timestamp bumped to `20260802130000` off a collision with #1470 (independent objects, so the interleave was harmless either way).

Drain window: old code never queries the new kind, and these rows carry `orgMembershipId` null so they cannot appear in any household-process query old code makes.

## Verification

- full unit suite: **2537 passed**, 267 suites
- full integration suite against a real Postgres, on a database created from scratch: **1326 passed**, 131 suites
- migrations applied from scratch on an empty database
- type-check and lint clean

New coverage: the age band and the cycle floor (including "signs Aug 20, boundary rolls Sept 1, must not be re-asked"), the route's deny paths and refusals, the response carrying only `id/kind/status`, the population predicate and dedup against a real DB, and two in `external.test.ts` pinning that a non-lead cannot sign the household agreement but can sign their own.

One fix came out of the merge with main: the awaiting-agreement list now filters `subjectPerson` through `LIVE_PERSON`. The sharpened drift guard caught it — an obligation whose subject was merged away would have stayed on the board's chase list.

**Correction to an earlier note on this PR:** I reported `programAgeBounds.integration.test.ts` as a pre-existing, apparently clock-dependent failure. It is not clock-dependent — it passes on a freshly created database. It was failing on a scratch DB I had been reusing across runs, so the cause is accumulated state, not this branch and not the clock.

## Known limits, stated deliberately

- The compliance list only surfaces **program-attached** over-25 candidates for the manual button. The route accepts any `personId`, so reaching someone in no program needs a person picker that does not exist yet.
- The one-year window is only as good as the program data. **Nothing ends a program**: `endAt` is optional and `phase` defaults to `PLANNING`, moving only on a manual settings edit. A program left undated reads as running forever and its members stay in the population. Filed as #1512; the change here is monotonic — never worse than attached-ever, better wherever a program carries an end date.
- #1462 tracks the "no exit signal for program attachment" gap, and depends on #1512: a sweep keyed on `endAt` does nothing where `endAt` is null.
- `src/security/registry.ts` still carries an issue number in a comment. It is a boundary file, already merged, and editing it here would break boundary isolation.


thpr:
## Update: route now goes through the security boundary

CI caught what local runs could not: `check-route-coverage` rejects a **new** route on `withAuth`. A route method in neither the security registry nor the frozen legacy baseline is new surface on old authz — a hard error even in advisory mode, and correctly so.

The route now uses `defineRoute()` + `handler()`:

- **#1489** carries the registry entry alone (boundary-isolation rule).
- This PR points the route at it and merges that branch, so the entry is present in-tree. **Isolation is expected red here until #1489 lands on main**; once it does, merging main in drops `registry.ts` from this diff and both checks go green.

Two things improved on the way through:

- The response is now a declared bag — `{ OrgMembershipProcess: { id, kind, status } }` behind the `process` envelope. The old version returned the whole process row, which carried `subjectPersonId` and both Zoho ids back out for no reason. A new test pins the shape so a later widening of the select can't quietly ship the subject's identity.
- The `AUTHZ_TESTED` entry is dropped — that guard scans for the `roles:` withAuth gate, which this route no longer has. Deny paths are still covered by the route's own tests, now exercised through `resolveAccess`.

jee7s:
**Area of concern: it carries a registry entry inline, which makes `isolation` unsatisfiable.**

`src/security/registry.ts` trips `is_boundary()`, and once that fires every other changed file must be a companion. Fourteen of the twenty-four are not — the migration, `cron/nightly/route.ts`, both compliance surfaces, `membership/page.tsx` and seven `lib/membership/*` modules. `prisma/migrations/**` is specifically not on the companion list.

The registry hunk here is byte-identical to #1489. Land #1489 first, then rebase so `registry.ts` drops out of this diff — otherwise two `defineRoute` calls for the same endpoint land and the registry throws at module init.

Also DIRTY against #1449 in `personBgTriggers.ts`, `lifecycleStatusLiteralAllowlist.test.ts` and `cron/nightly/route.ts`, and no CI has run on the current head.

---
_Posted by Claude._

thpr:
Review is stale; a boundary PR was shipped earlier

jee7s (approved):

Issue #1476, opened 2026-08-03
Decision: widen the facility/visits gate to isOperations


thpr:
Split out of the attendance/visit correction surface design (`checkin-app/docs/designs/1256_ATTENDANCE_CORRECTION_SURFACE.md`), where it was open question §6.1. Moved here so the design doc can describe only what is built; AT3 (#1254) shipped deliberately **without** this change.

## The decision

`GET/POST/PATCH/DELETE /api/facility/visits` is gated `roles: ['isSysadmin', 'isBoardMember']`. Should `isOperations` be added?

The design recommends **yes** — ops is a facility-operations role and visit management is facility operations. But widening a permission gate is a board call, not a default worth taking silently inside a feature PR, so the status quo shipped.

## Current state

- Route: `checkin-app/src/app/api/facility/visits/route.ts` — `['isSysadmin', 'isBoardMember']` on all four verbs.
- Page: `checkin-app/src/app/facility-ops/visits/page.tsx` — `useRequireRole(['isSysadmin', 'isBoardMember'])`.

The two sets are currently equal, which is the invariant that matters.

## If adopted

Apply the same 3-role set to **both** the route (`withAuth roles`) and the page (`useRequireRole`) in one change. They must stay equal — the two drifting apart is exactly what AT13 (#1259) was, fixed in #1350. In the design's §1 permission matrix, ops moves from 🟡 to ✅ on insert-for-others, edit, and delete.

Worth checking at the same time whether ops should also reach the AT12 correction-review screen (#1258) when that lands, since it is the oversight half of the same surface.

## Scope note

Gate-only. No behavioural change to any handler.

jee7s:
**HOLD** — the implementation is fully specified but the decision itself is unmade, and it is explicitly a board call rather than a default. That is precisely why it should not sit in Planning.

Current state re-verified on `origin/main`: both `checkin-app/src/app/api/facility/visits/route.ts` (all four verbs) and `checkin-app/src/app/facility-ops/visits/page.tsx` still gate exactly `['isSysadmin', 'isBoardMember']`. Accurate, not stale.

To promote, answer:

1. Does the board approve adding `isOperations` to the facility/visits gate?
2. If yes, should ops also reach the AT12 correction-review screen (#1258) when it lands, or is that a separate call?

[...]

Issue #1258, opened 2026-07-24
Admin hour-correction review screen


thpr:
**Backlog item AT12** — from the backlog baseline (`docs/backlog/INDEX.md`, PR #1192).

Admin hour-correction review screen — surface how often self/hour corrections happen + let admins review them, NOT buried in the audit log (companion oversight surface to AT5)

- Origin: CREATE · Readiness: NEEDS-DESIGN · Size: S

_Auto-created from the backlog INDEX; see the doc for bucket context and cross-links._

thpr:
**Backlog reconciliation (2026-08-01)** — AT12 (hour-correction review screen) is covered by the **merged correction-surface design #1352** as the oversight half of AT3/AT5. Open sub-decision teed up: **who sees raised corrections** (ops/board/lead) — verify whether #1352 settled it. Readiness IN-DESIGN. Source: INDEX AT12.

jee7s:
**Status: NOT READY.** Moving to Iteration 2 / **Ready** — nothing is in flight (no open PR implements AT12; #1478 excludes it by name), and the design's two load-bearing premises about the audit trail are no longer true on `main`.

`checkin-app/docs/designs/1256_ATTENDANCE_CORRECTION_SURFACE.md` §4 is detailed enough to build from, but both of the screen's axes depend on premises that the AT5/AT3 work invalidated after the design merged.

**What changed under the design.** §4 tells the implementer to split self from staff via `actorId === secondaryAffectedEntity` (§6.6, which #1478's doc rewrite marks "adopted — every visit audit write now sets it to the subject person"). That claim is wrong for two paths: `checkin-app/src/app/api/events/[id]/attendance/route.ts:123,153` and `checkin-app/src/app/api/my-programs/conflicts/resolve/route.ts:69` both store the **event id** there. Both columns are bare `Int`, so the comparison silently returns "self" on collision rather than failing. The `newData.type` fallback §4 offers instead is also no longer a discriminator: since #1478 a household-lead proxy correction writes `type: "self_correction"` with `actorId ≠` subject.

And §4's "the same significance function runs at read time" is not uniformly possible. The roster-mark rows carry no times to recompute from (`newData: { participantId, associatedEventId, synthetic }`, no `oldData`); `facility/visits` EDIT/DELETE store whole-row JSON but no `significance`; only the manual self/household paths persist `newData.significance`.

**Clarifications needed before implementation:**

[...]

PR #1478, opened 2026-08-03
feat(attendance): household-lead, program-lead and staff visit management (AT3)

### LINKED ISSUE #1254: Visit management for staff
**Backlog item AT3** — from the backlog baseline (`docs/backlog/INDEX.md`, PR #1192).

Visit management for staff. (a) A user can insert their own past visit but CANNOT edit it once inserted — no edit path for an existing visit (self or others), insert-only. (b) Whether ops / program-lead can ADD a visit for others is unverified — needs a code check.

- Origin: ENHANCE · Readiness: NEEDS-DESIGN + VERIFY(b) · Size: S

thpr:
Fixes #1254

> **Blocked on #1492** — the registry-first half. `POST /api/facility/visits/insert` is role-gated, so `check-route-coverage --strict` fails here until that entry is on `main`. Verified locally that route + entry together take the lint to exit 0.
>
> Its own endpoint rather than a `POST` on `/api/facility/visits`: adding a verb to an existing legacy route file cannot satisfy the lints in any PR ordering — registry-first trips `orphan-registry`, route-first trips `new-route-old-authz`, both together trip boundary isolation. Filed as #1491.

#1357 (AT5) has merged; this branch is **rebased onto `main`** and the diff is now the 6 AT3 commits alone (27 files). Full unit suite, the visit/attendance/event/cron integration suites, and the migration chain were all re-run green after the rebase.

AT3 is the non-`self` half of the correction-surface design (`checkin-app/docs/designs/1256_ATTENDANCE_CORRECTION_SURFACE.md`), which #1357 built the `self` half of. Four slices plus a doc consolidation.

## What this adds

**Household-lead correction of a member's visits.** The only path by which a minor's visit can be corrected at all — they cannot self-serve. A lead may insert, edit, and delete visits for anyone in their own household, on exactly the same trust-first terms as their own. Scope resolution is one server-side function, `visitSubject` (`src/lib/visit/scope.ts`): self always, plus your own household's members if you lead it. The household comes from the actor's row, so a `personId` in the body is only ever a target checked against it — a non-lead naming a household peer is denied, and so is a lead reaching into another household. Leadership, not membership, is the grant.

**Staff insert-for-others at an arbitrary past time.** `POST /api/facility/visits/insert`, the walk-in path neither the kiosk (live only) nor the event roster mark (program-scoped, event window) could record. Closed visits only — an open one would put someone on the live in-the-building roster on staff say-so.

**Program-lead correction brought in line.** `manualEditAttendance` already existed and already had the right scope; it was the one human visit-write path that hard-deleted and wrote *no audit row at all*. Its Absent branch now tombstones, and both branches log.

**The `VisitSource` split.** `SYSTEM` collapsed three unrelated events into one value. No checkout path audits a `Visit`, so `departedVia` is the only record of how a visit closed — and the two machine closers have opposite accuracy. A facility close is stamped when the building actually closed; the cron stamps at its own run time, so the real departure may be hours earlier. That distinction is what lets the significance rule suppress corrections to machine stamps by *source* rather than magnitude. Before this, `significance.ts` had to infer the closer at runtime because the enum could not tell them apart.

## Two decisions worth reviewing

**The open-visit facility guard now follows the subject, not the actor.** An open backfill asserts the subject is in the building now, so it obeys the same keyholder-first rule as `/api/scan`. Left on the actor, a keyholder parent could backfill an open visit for their non-keyholder child into an empty building — recording a child present, alone, with no keyholder on site. Self keeps reading the session claim (unchanged behaviour); the proxy case reads the subject's row.

**Proxy significance is a ×2 multiplier, not a floor.** A floor would flag every proxy edit including a two-minute typo fix — the "required justification reads as distrust" failure the design rejects. Multiplicative matters: a machine-close weight of 0 stays 0 when doubled, so a lead fixing their child's `AUTO_CLOSE` departure still never flags. An additive weight would make that happy-path fix the loudest alert in the system.

## Migration safety

Expand only, in two migrations. `SYSTEM` **stays in the enum**: during a rolling deploy the previous release serves traffic against the fully-migrated schema and still writes it from all three paths, so dropping the value would fail every one of them for the whole drain window. The drop is a follow-up release.

- `..._visit_source_split_add` — three `ALTER TYPE … ADD VALUE`, deliberately *not* wrapped in a transaction, because Postgres forbids using a value in the transaction that added it. Additive and idempotent, so a partial apply is harmless.
- `..._visit_source_split_backfill` — the row mapping, wrapped, since its two updates must land together.

Legacy departures are fused history and cannot be fully recovered. One case can: the roster mark is the only writer that puts `SYSTEM` on `arrivedVia` and writes both fields together, so a `SYSTEM`/`SYSTEM` pair maps to `LEAD_MARKED`. Everything else maps to `AUTO_CLOSE`, the conservative "do not trust this time" reading.

**Not a security-boundary change.** `classifications.ts` tiers *fields*, not enum values, so `arrivedVia`/`departedVia` stay `public` and the generated file comes out byte-identical — verified by regenerating and diffing to empty. The isolation workflow fires only on the boundary source dirs (excluding generated output), the generator script, or a genuine re-tier. The design doc's own Security-boundary note claimed otherwise and contradicted its §3; corrected here.

## Domain rules register

`docs/rules/attendance-checkin.md` is updated for what this ships. Two entries there were made false and would not have prompted anyone otherwise:

- **The household-lead entry** sat under *Policy requirements not yet enforced here*, where a stale claim is indistinguishable from a deliberate known gap. Removed, and replaced by a rule in *The visit record* stating what a lead may now do — insert, edit and remove for household members, lead-only — placed next to the self-correction rules, since the decided semantics are that a lead's correction works exactly as correcting one's own does.
- **"Adding one for someone else is scoped to an event roster or a live scan, never an arbitrary past time"** — the staff insert here makes that untrue. Split into the capability and the limit that survives it: a visit recorded for someone else is always a closed one, because appearing on the list of who is in the building follows from a badge at the kiosk, not from staff saying so.

**Deliberately not touched:** the self-correction half of the first entry ("A member cannot correct their own visit"). It is also stale — that shipped in #1357 — but it belongs to the design-doc retirement work on `claude/retire-shipped-design-docs`, and fixing it here would collide.

**No program-lead rule written.** This PR changes how a roster removal is *stored* (tombstone rather than row removal), not who may do it or what it means, and the register carries no claim that contradicts.

## Verification

- Full migration chain applies to a fresh database (`migrate deploy`, exit 0).
- Backfill run verbatim against seeded legacy rows: `SYSTEM`/`SYSTEM` → `LEAD_MARKED`/`LEAD_MARKED`, `SCANNER`/`SYSTEM` → `SCANNER`/`AUTO_CLOSE`, `WEB` untouched, zero `SYSTEM` rows remaining.
- Unit suite: 2492 tests / 262 suites green.
- Integration: 131 tests across the 19 visit/attendance/event/cron suites, against a real Postgres.
- `tsc --noEmit` clean.

## Deliberately not in scope

- **`isOperations` gate widen** — split out to #1476. The gate stays `['isSysadmin', 'isBoardMember']` on both route and page; widening a permission gate is a board call, not a silent default inside a feature PR.
- **Dropping the `SYSTEM` enum value** — the contract stage, a follow-up release.
- **AT12** (#1258) — the correction-review screen, unbuilt.
- **The advisory-lock gap** on `facility/visits` `PATCH`/`DELETE` and `manualEditAttendance` — pre-existing, documented as §7 Parallel work, being repro'd and fixed separately against `main`.

## Doc

The design doc is rewritten in the present tense to describe what actually runs — `Validated current state` (a snapshot of pre-implementation `main`) is replaced by `The write surface`, an inventory of every route that can touch a `Visit`. Reviewers comparing the doc to the code should now find them the same thing.


jee7s:
**Area of concern: the route is not registered, and the code comment says it is.**

`Run Tests (unit)` fails with `✗ [new-route-old-authz] … POST /api/facility/visits/insert is not in the security registry and not in the frozen legacy baseline`. That rule blocks even in advisory mode, and the `Coverage gate` red is downstream of the same job.

The new route's header comment claims "Declared in the security registry" — `registry.ts` on this branch has no `visits/insert` entry; the branch only adds it to `AUTHZ_TESTED`.

#1492 is the registry-first entry for this endpoint and is green. Land that, then rebase.

---
_Posted by Claude._

jee7s (changes requested):
Raising these here rather than in the AT12 design doc (#1497) — per @thpr, defects in an open PR belong as a review on that PR, not buried in a doc. All verified against this PR's current head `6d128500`.

This PR's own rewrite of `1256_ATTENDANCE_CORRECTION_SURFACE.md` states the invariant plainly: *every visit audit write sets `secondaryAffectedEntity` to the subject person*, and the audit-coverage inventory says each row carries "`secondaryAffectedEntity` = whose visit it is, so acting-for-another reads off the inequality without a join."

On this head that is **nine of twelve writers**, so the invariant is false the day it merges. Three writes still store a non-person id.

## 1. Three writers store an event id in the subject column

| Site | Stores | Should store |
|---|---|---|
| `api/events/[id]/attendance/route.ts:123` | `eventId` | `pId` |
| `api/events/[id]/attendance/route.ts:153` | `eventId` | `pId` |
| `api/my-programs/conflicts/resolve/route.ts:69` | `visit.associatedEventId` (nullable) | `visit.personId` |

Both audit id columns are bare `Int` with no foreign key, so this does not fail loudly — `actorId === secondaryAffectedEntity` silently returns "self" whenever a person id and an event id collide, and both sequences start at 1.

**No information is lost by the change.** At the roster-mark sites the event id is already in `newData.associatedEventId`; at `conflicts/resolve` it is already in `oldData.associatedEventId`. The subject is in scope at all three (`pId`, `visit.personId`).

Worth noting the near-miss: `events/[id]/route.ts` (the `manualEditAttendance` action this PR adds) stores `targetId` correctly at all three of its writes. It is `events/[id]/**attendance**/route.ts` — one path segment over — that does not. Easy to read as already done.

## 2. The audit-coverage inventory lists a writer that does not exist

It names `membership-ops/.../merge` as a `Visit` audit writer. It writes none — the merge re-parents visits with `tx.visit.updateMany` and logs `tableName: "OrgMembershipProcess"` and `"Person"` only. That behaviour is right and should stay; a merge changes no attendance fact. The claim should come out of the inventory.

## 3. Significance is persisted on one file out of five write paths

`newData.significance` is written only by `attendance/manual/[id]/route.ts`. Zero occurrences in `events/[id]/route.ts`, `facility/visits/route.ts`, `my-programs/conflicts/resolve/route.ts`, or `events/[id]/attendance/route.ts`.

This PR introduces `PROXY_MULTIPLIER = 2` and a `byProxy` option specifically to weight an adult changing another person's record — and then adds three writers in `events/[id]/route.ts` where a program lead corrects a member's attendance, none of which score. One file reaches the multiplier.

Suggested call sites, roughly four lines each: `facility/visits` PATCH and DELETE, `conflicts/resolve` DELETE, and `manualEditAttendance` EDIT and DELETE. All already carry the `oldData` those calls need — `facility/visits` because this PR just added it, `manualEditAttendance` because it was written that way.

Not blocking on its own, but a flagged view that silently omits the paths capable of the largest corrections reads as "nothing significant happened this month" when nobody looked.

## Why fix it here rather than after

Doing it later means a follow-up PR **plus a permanent read-side fallback** for every row written between the two merges: those rows keep an event id in the column and nothing distinguishes them from correct rows except their age. The fallback never goes away, because a reader cannot tell which convention an old row followed.

Item 1 is three lines in two files. Items 2 and 3 are small and in files already in this diff. None of them touches a boundary file, so none disturbs this PR's isolation.

---
_Posted by Claude._

thpr:
Verified all three against `6d128500` before touching anything. Items 1 and 2 were correct and are fixed in `c7d6ad51`. Item 3 I'm declining, with reasoning — and the review's own framing is what led me to a fourth defect it didn't catch.

## 1. Subject column — fixed, and thank you for the collision argument

Confirmed at all three sites. Now `pId` at both `events/[id]/attendance` writes and `visit.personId` at `conflicts/resolve`. Nothing lost, exactly as you said: the event was already in `newData.associatedEventId` and `oldData.associatedEventId` respectively.

The argument that moved me was the permanent read-side fallback, not the collision itself. A collision needs an id coincidence; the fallback is unconditional and forever, because nothing on the row says which convention it followed.

Worth recording for the next reader: the old convention was **deliberate**, not an oversight. `eventAttendanceAPI.integration.test.ts` carried the comment *"Per-Visit audit rows carry the event in secondaryAffectedEntity; testEventId is fresh per run, so this is leak-proof"* and used it as its cleanup key. Four assertions there, plus one in `auditLog.integration.test.ts` and one in `myProgramsConflictsResolveAPI`, all pinned it. Those are repointed to the subject; the suite-scoping key is now the participants, which is equally leak-proof.

I checked the read side before changing it: no consumer interprets the column semantically. `AuditLogPanel.tsx` renders it as a bare `` +${id} ``.

## 2. Merge in the inventory — fixed

Confirmed: `membership-ops/.../merge` writes `tableName: "OrgMembershipProcess"` and `"Person"` only, no `Visit` row. The doc claim is gone, and I added the reason it's right that way — a merge changes who owns an attendance fact, not the fact, so it's an identity operation and AT12 shouldn't surface it as a correction.

## 3. Significance on five paths — declining, and the doc was the actual defect

Two reasons the blanket wiring would make things worse:

**The delete floor turns it into noise.** `deleteSignificance` always returns `flagged: true` — deliberately, so a member erasing their own visit is always seen. Wire that into `manualEditAttendance`'s Absent branch and every routine roster correction emails the board. Marking someone Absent is the lead's ordinary weekly workflow, not an anomaly. Same for `conflicts/resolve`: resolving a duplicate is cleanup. A flag that fires on the happy path is the failure mode §2 spends a paragraph on for `AUTO_CLOSE` — it just arrives through a different door.

**On `facility/visits` the actor is the recipient.** That route is gated `['isSysadmin', 'isBoardMember']`, so the flag would be the board notifying itself.

**And AT12 doesn't need it.** §"No new model" is explicit that significance is *recomputed* from the audit row's old/new values — no persisted field, no request table. I checked that every path you list writes the source fields (`arrivedVia`/`departedVia`) into `oldData`, so the correction-review screen can score all of them. `newData.significance` in `attendance/manual/[id]` is a filtering convenience, not the mechanism.

Where you were right is that the doc left this implicit and so read as universal. Added a **"Which paths raise the live flag"** section stating which paths fire, both reasons the others don't, and the lever if the board later wants leads' corrections in the feed — the recipient set (§6.3) plus a per-path threshold, not the delete floor.

## 4. Found while checking item 1: `conflicts/resolve` was hard-deleting

`await tx.visit.delete({ where: { id: visit.id } })` — a real row removal, contradicting the same doc's *"every delete path is a soft-delete update"* and destroying a record the tombstone design promises is reversible. Its test asserted `expect(gone).toBeNull()`, so it was pinned in place.

The `LIVE_VISIT` drift guard can't catch this: it deliberately exempts single-row `delete`/`update`, so the one call shaped like this slipped the net.

Now tombstones with `deletedById`, and the test asserts the tombstone instead. Resolving a conflict is a judgement about which of two overlapping records is real — backing that judgement out has to be possible.

## Verification

Unit 2520/265 suites, integration **full run** 1323 passed / 6 skipped / 0 failed across 130 suites, `check-route-coverage --strict` exit 0, `tsc` clean.

The four repointed assertions are the useful signal here — every one of them failed on the first run and had to be understood rather than mechanically updated.

jee7s (approved):

jee7s (approved):

PR #111, opened 2026-04-04
🛡️ Sentinel: [CRITICAL] Fix Timing Attack in Cron Endpoints

dkaygithub:
🚨 Severity: CRITICAL
💡 Vulnerability: The cron API endpoints (`nightly`, `pending-participants`, `reminders`, `post-event`) were using the standard string comparison operator (`!==`) to compare the provided `authorization` header with the expected `CRON_SECRET`. Standard string comparisons fail fast (as soon as a character mismatch occurs), which leaks timing information to an attacker. This timing leak theoretically allows an attacker to guess the correct secret character-by-character and bypass authorization, allowing them to illegally trigger jobs like kicking participants, issuing mass emails, or auto-closing the facility.
🎯 Impact: Unauthorized execution of automated administrative jobs, potential data corruption, unauthorized notifications, and DoS.
🔧 Fix: Updated all four cron endpoints to import the `crypto` module and utilize `crypto.timingSafeEqual()` for the comparison. Both the provided header and the expected token are converted to `Buffer` objects before checking. A strict `.length` equality check is performed first (which is safe and standard practice before a constant-time check).
✅ Verification: Ran `npm run lint` and `npm test` locally to verify changes. Code review passed.

---
*PR created automatically by Jules for task [16111397985925022488](https://jules.google.com/task/16111397985925022488) started by @dkaygithub*

PR #1479, opened 2026-08-03
🛡️ Sentinel: [HIGH] Fix XSS vulnerability in dev email viewer

dkaygithub:
🚨 **Severity:** HIGH
💡 **Vulnerability:** Used `dangerouslySetInnerHTML` to render captured email bodies in the dev sent-mail viewer (`checkin-app/src/app/dev/sent-mail/page.tsx`). Even though this is a developer tool, rendering arbitrary HTML can lead to XSS if a malicious payload is captured in the email content.
🎯 **Impact:** Potential execution of malicious scripts within the context of the dev dashboard, potentially leading to unauthorized access or actions within the development environment.
🔧 **Fix:** Replaced `dangerouslySetInnerHTML` with a securely sandboxed `iframe` using the `srcDoc` attribute. The `sandbox` attribute is configured with `allow-popups allow-popups-to-escape-sandbox` to prevent script execution while preserving the ability to click embedded links or tokens safely.
✅ **Verification:** Replaced the vulnerable code, applied styling to ensure the iframe is responsive and visible, and updated the sentinel log with the finding. Note that standard tests and linters are currently failing in the sandbox environment due to missing dependencies.

---
*PR created automatically by Jules for task [4655278643892921787](https://jules.google.com/task/4655278643892921787) started by @dkaygithub*


thpr:
## Review

Direction is right — a sandboxed iframe is the correct pattern for rendering captured mail. But the chosen sandbox flags break the page's one reason to exist, and the stated severity doesn't match the actual exposure.

### Blocking

**1. Links no longer complete token flows — the page's entire purpose.**

The file header states it plainly: *"this page renders it so link/token flows can be completed without a real key."* Grepped every email template — **none** use `target="_blank"`; they emit plain `<a href>` (e.g. `unsubscribeFooter` in `src/lib/outreach/render.ts`, the CTA in `src/lib/email-templates/post-event.ts:26`).

With `sandbox="allow-popups allow-popups-to-escape-sandbox"`:

- No `allow-top-navigation*` → a plain link navigates **the iframe itself**, so the app loads inside a 300px box instead of the tab.
- No `allow-same-origin` → the frame gets an opaque origin, so the navigation is treated as cross-site. NextAuth's session cookie is SameSite=Lax by default (`src/lib/auth-options.ts` sets no `cookies` override) → **not sent**. Any authenticated dev link lands logged-out.
- Both popup flags are dead weight — nothing opens a popup without `_blank` or scripts.

One-word fix that keeps the XSS win (still no `allow-scripts`) and restores click-through as a real top-level navigation:

```
sandbox="allow-popups allow-popups-to-escape-sandbox allow-top-navigation-by-user-activation"
```

**2. Unverified.** The PR body notes *"standard tests and linters are currently failing in the sandbox environment due to missing dependencies."* So the page was never rendered and no link was clicked in it. Given finding 1, that shows.

### Non-blocking

- **Stale comment** (`checkin-app/src/app/dev/sent-mail/page.tsx:69-70`). "Dev-only rendering of our own captured template HTML… Never runs in prod (page 404s); content is what we ourselves sent" was the justification for `dangerouslySetInnerHTML`. It now dangles, explaining nothing. Per AGENTS.md: delete stale comments when you touch the code. Replace it with the real non-obvious why — which sandbox tokens are present and what each one buys.
- **Fixed `minHeight: 300px`** — auto-sizing is impossible without `allow-scripts`, so every email becomes a scrollbox regardless of length. A reasonable tradeoff, but it is a UX regression from natural inline height and isn't acknowledged anywhere.
- **`color: "#111"` is inert** — the `srcDoc` document doesn't inherit parent styles. Dead prop; drop it.
- **Duplicate `title="Captured Email"`** on every row. Use `title={email.subject}` — better a11y, one character more.
- **`.jules/sentinel.md` heading says "React CSS/HTML Injection"**, body says XSS. Pick one.

### Severity assessment

HIGH looks overstated. Every template routes user input through `escapeHtml` (`src/lib/email-templates/base.ts`), and `substituteTokens` escapes `{{name}}` explicitly (`src/lib/outreach/tokens.ts:37`). The only unescaped HTML is the board-authored outreach body — intentionally rich, trusted-admin authored. The page 404s outside dev instances (`config.devToolsActive()`) and reads a dev DB. The realistic path is a *future* template forgetting `escapeHtml`, which makes this reasonable defense-in-depth rather than a live HIGH.

### Other checks

- No CSP / `frame-ancestors` set anywhere, so framing itself won't be blocked.
- No test scrapes this page; `flow-tests/membership-activation-fanout.flow.test.ts:39` explicitly says so. No test breakage.
- Not a security-boundary file, so the boundary-isolation CI rule doesn't apply.

### Verdict

Add `allow-top-navigation-by-user-activation` and fix the stale comment, then this is a clean defense-in-depth improvement.

jee7s:
**Area of concern: the sandbox drops the session cookie, so the page stops doing the one thing it exists for.**

With no `allow-same-origin` the iframe has an opaque origin, so clicking a captured magic link is a cross-site navigation and the `SameSite=Lax` NextAuth cookie is not sent. Combined with the absent `allow-top-navigation-by-user-activation`, the link also navigates inside the ~300px frame rather than the tab. Every dev magic link lands logged out.

Adding `allow-top-navigation-by-user-activation` restores click-through without granting scripts. The comment above the block still says links are clickable, which the change makes false.

Separately: #1486 is the same fix on the same lines and should close in favour of this one, but its per-email `title={...email.subject}` is better than the static `title="Captured Email"` here and is worth porting.

---
_Posted by Claude._

jee7s (dismissed):

thpr:
Wait, we approved this with blocking changes?

jee7s:
I've cleared my review. I got my tabs backwards and approved the wrong PR. This wasn't my agent doing an erroneous approval

thpr (changes requested):
Requesting changes — see the detailed review in https://github.com/innovationtreehouse/checkin/pull/1479#issuecomment-5170965116.

Blocking item: the `sandbox` value drops link click-through, which is the page's whole purpose. Plain `<a href>` links (no template uses `target="_blank"`) navigate the iframe itself rather than the tab, and the opaque origin means the SameSite=Lax NextAuth session cookie isn't sent, so authenticated dev links land logged-out. Adding `allow-top-navigation-by-user-activation` fixes both and keeps the XSS win, since `allow-scripts` stays off.

Also please refresh the now-stale comment above the iframe — it still justifies the removed `dangerouslySetInnerHTML`.

thpr:
Closing as a duplicate of #1617 — same finding, same fix, from an earlier Sentinel run.

Two details here were better than #1617 and I have carried both over as review comments on it: the iframe `title`, and sizing with `minHeight` rather than a fixed `height`.

Both versions share one defect, tracked on #1617: the sandbox as configured does not preserve link clicking. A sandbox without `allow-same-origin` or `allow-top-navigation` still lets the frame navigate itself, so a plain left-click loads the app inside the frame under an opaque origin — no `SameSite=Lax` session cookie, no scripts — which is the token-completion flow this page exists for. Fix is `srcDoc={`<base target="_blank">${email.html}`}`.

---

PR #1482, opened 2026-08-03
security(isolation): certify whole-entity decommissions so model drops ship as one PR

dkaygithub:
Rebased onto current `main` and consolidated. This branch now carries the whole decommission-exception effort — @dkaygithub's original five commits with their authorship intact, plus three that close what review found. **#1526 and #1524 are absorbed; they can be closed once this lands.**

## Why it was rebased

The branch predated #1501, which moved the workflow off the checked-out `HEAD` because on a `pull_request` run that ref is `refs/pull/N/merge` — the base merged with the current tip of main. Merging the old side would have reverted that fix (`grep -c HEAD_SHA`: main 5, old branch 0). Both conflicts are resolved keeping both sides.

## The reason this exists now works

**Running the certifier against #1444's actual head certified it:**

```
Certified whole-entity decommission:
  - model FeePayment (dropped from schema in this PR)
```

It did not before, and that was not a bug in #1444. Two edits were being refused that a whole-entity drop *forces*:

- **`registry.ts` had to remove `'Fee'` from a surviving route's `returns:` array.** `returns` is typed `readonly Models[]` and `Models` is keyed off the generated classifications, so dropping the model stops the entry compiling. The edit is compelled by the type system, and removing an element only narrows what the route may return.
- **`scopeBindings.ts` rewrote comments naming Fee.** Those comments describe a model that will not exist.

So FR7 — and #354 behind it — was blocked on a rule with no way to say yes.

Both sides now pass through one normaliser. It forgives **comments** and **a `returns:` element naming a model dropped from `schema.prisma` in the same PR**, and nothing else. Twelve cases pin the split: a role change, `authenticated`→`public`, a widened tier, an endpoint change, an *addition* to `returns`, and dropping a model that *survives* all still fail byte-strict.

Two details where this could have gone wrong, both handled:

- **String contents are preserved.** The existing `stripStringsAndComments` erases them; normalising with it would have made `authorize: 'public'` compare equal to `authorize: 'authenticated'`.
- **A file whose only change is forgiven now yields zero removals rather than an error**, so `certifyDecommission`'s global "nothing was decommissioned" guard is what refuses an empty PR. The guard moved; it did not disappear.

## What the review found, and what it cost

| Fixed | Was |
|---|---|
| Certifier audited the wrong tree | Used checked-out `HEAD` (= the merge ref) after #1501 moved the workflow off it, so main's advance could decide this PR's verdict. `--head` is now required. |
| Any deletion was admitted | A file nothing imports can still be a security control. Now only route files of removed registry entries. |
| Derived route paths were trusted | Absence at head read as "endpoint dead" — but a path that never existed is also absent. Base must now export the verb first. |
| Renaming a boundary file bypassed the gate entirely | `git diff --name-only` reports only the destination, so renaming `middleware.ts` kept CI green while disabling the site-wide org-login gate. `--no-renames` fixes it. Verified in a scratch repo. |
| `middleware.ts` had no CODEOWNERS entry | Every other `is_boundary` path had one. |
| The failure message never named the boundary file | It listed the *feature* files; a boundary file is a companion, so a developer read "must ship it ALONE" over a list not containing the file that tripped it. |

Plus `process.exitCode` over `process.exit()`, which can truncate buffered stdout on a pipe — how the workflow runs it.

## Verification

- **36/36** in `scripts/__tests__/boundary-decommission.test.ts`.
- The certifier executed against #1444's real head: exit 0, certified.
- `--head` proven to move the audited tree: same argv, two `--head` values, two different verdicts on the same file.
- Workflow YAML parses; the run block passes `bash -n`.

## Known fail-closed corner

Removing an **already-inert** route entry fails certification, because the base file must export the verb before its absence at head means anything. Only reachable on a PR that also carries a violation, since a boundary-only PR never invokes the certifier. Documented in the workflow header; split such a removal into its own PR.


thpr:
## Review

**What it does:** adds a fail-closed certifier (`checkin-app/scripts/lib/boundary-decommission.js` + CLI) that lets the boundary-isolation workflow admit *one* extra PR shape — a boundary diff that is nothing but whole top-level entries spliced out byte-for-byte, whose subjects (Prisma model / route verb) also leave in the same PR. Certified PRs may then carry `prisma/migrations/**` + file deletions. Also fixes boundary scripts flagging themselves, plus docs.

**Verification run:** no `node_modules` in the review worktree, so no jest — instead the lib was loaded on bare node and driven against the real `scopeBindings.ts` / `registry.ts` / `schema.prisma` at PR head.

- Segmentation round-trips both real files, names all 26 bindings entries and all 18 registry calls. The parser is sound on today's shapes.
- Real dynamic segments (`[id]`) map fine; no route groups and no `route.tsx` in the repo — `endpointToRouteFile` is the exact inverse of `check-route-coverage.ts`'s `fileToEndpointPath`.
- The design is genuinely fail-closed on the boundary side, and the `all:`-widening rejection is the right instinct.

Three things worth fixing before merge.

---

### 1. Blanket deletion latitude admits deleting an auth gate — Medium/High

`certifyDecommission` admits *any* file with status `D`. Confirmed to pass:

```
changed:    [scopeBindings.ts M, schema.prisma M, <drop migration> A, checkin-app/src/middleware.ts D]
violations: [<migration>, checkin-app/src/middleware.ts]
=> ok: true, removedModels: ["Visit"]
```

`src/middleware.ts` is the site-wide org-login gate (dev + ops-stg) and the `token?.denied` → `/access-denied` redirect. It is imported by nothing, so deleting it breaks neither tsc nor the build; it is not CODEOWNERS-gated. A decommission PR could carry that deletion through isolation unreviewed.

This is exactly the argument the PR body makes for *not* trusting textual removals in boundary files ("a deleted line can be a deleted check") — just not applied to app-file deletions.

Cheapest fix: treat `checkin-app/src/middleware.ts` as a boundary file in both `is_boundary` (workflow) and `isBoundaryPath` — it is one. Tighter fix: admit only deletions the decommission implies (route files of `removedEndpoints`).

### 2. Route-kill check fails open when the endpoint→file mapping misses — Medium

`boundary-decommission.js` (the `removedEndpoints` loop) treats `readHead(file) == null` as proof the verb stopped being served. Nothing validates that the derived path is the real one. Confirmed:

```
remove defineRoute 'PATCH /api/attendance/manual/[id]' + 'DELETE /api/attendance/manual/[id]', delete nothing
=> ok: true
```

Those two are legitimately inert today (register-first, no route file), so this is not a live hole — but the check cannot distinguish "inert" from "the string concat guessed wrong". A route group `(admin)`, a catch-all, a `route.tsx`, or a rewrite turns a silent `null` into a certified kill of a still-served endpoint — the precise #1394 scenario the exception is meant to make safe.

Suggested fix — require the base file to resolve and export the verb before accepting its absence at head:

```js
const base = readBase(file);
if (base == null || !exportsVerb(base, verb)) {
    reasons.push(`${endpoint}: ${file} does not serve ${verb} at base — mapping unverified`);
} else if (readHead(file) != null && exportsVerb(readHead(file), verb)) { … }
```

Removing a genuinely inert entry is a boundary-only PR with no violations, so the certifier never runs on it — nothing legitimate regresses.

### 3. `isBoundaryPath` is a hand-copy of the shell `is_boundary` — Medium

Two independent definitions of "what is a boundary file", kept in sync by a test that is itself a hand-written mirror (`boundary-decommission.test.ts`, the `isBoundaryPath` describe). Divergence is asymmetric: if the shell grows an entry the JS does not have, `boundaryChanged` misses that file and the "engine files never qualify" rejection silently does not fire — fail-open.

The workflow already computes the boundary set. Pass it in (`--boundary f1 f2 … -- violations…`) and delete `isBoundaryPath` from the lib. Same for `exportsVerb`, which duplicates `VERB_EXPORT_RE` from `check-route-coverage.ts:46` verbatim — two copies of one security-relevant regex.

---

### Smaller

- `check-boundary-decommission.js` — `process.exit(main(...))` can truncate buffered stdout when stdout is a pipe (GitHub Actions). Use `process.exitCode = main(...)`.
- `check-boundary-decommission.js` — the comment on the `--name-status` parse says R/C rows "surface each for the path checks", but the code keeps only `rest[rest.length - 1]` (the destination). Behaviour is still fail-closed; the comment is just wrong. Per `AGENTS.md`, a comment that does not match the code is worse than none.
- `boundary-decommission.js` — `stripStringsAndComments` handles single quotes and `//` only. `checkRoundTrip` guarantees the line *partition*, not correct nesting, so a stray `{` in a double-quoted string could let one entity segment swallow the next; a head that removes both would certify with only the first model schema-checked. Impact is bounded (whole-entry removals narrow, they do not widen) but the weld check is what gets skipped. Either bail on `"` / backtick / `/*` inside a container, or say so in the comment.
- `.github/CODEOWNERS`: the two new lines break the column alignment of the block they join.
- Docs/renumber are clean — no other file references "rule 7"/"rule 8".

### Tests

Genuinely adversarial and the right shapes (`all:` widening, reorder, addition, no-op, outbound, engine file, surviving model, modified app code). Gaps that map to the findings above: nothing covers the unresolvable-route-path path, nothing pins the deletion latitude as a deliberate decision, and the CLI wrapper (arg parsing, `merge-base`, `gitShow` null-on-missing) is untested. `scripts/__tests__/` is picked up by the default jest run (`testMatch` reaches it, no ignore pattern excludes it); it was not executed in this review.

### Non-blocking

Head is `168c8dd3` on `3f0b90db`, so it predates #1481 (`b446b43d`), which edited the same workflow's trigger/skip steps. The rebase is trivial as the body notes, but the merged result is worth re-reading once — both hunks land in one file.

jee7s:
**Area of concern: the certifier admits any deletion, including a security control.**

`certifyDecommission` accepts any path with git status `D`. thpr demonstrated `checkin-app/src/middleware.ts` passing certification — the site-wide org-login gate and the `token?.denied` redirect, imported by nothing, so deleting it breaks neither tsc nor the build. That is this PR's own "a deleted line can be a deleted check" argument, not applied to app files.

Findings 2 (route-kill fails open when the endpoint-to-file mapping misses) and 3 (`isBoundaryPath` hand-copies the shell `is_boundary` and can diverge fail-open) are also unaddressed at the current head.

Also DIRTY: merged #1481 rewrote the same header block in `security-boundary-isolation.yml`. That part is trivial.

Flagging that this is the highest-leverage item in the queue — #1444 cannot merge until it does.

---
_Posted by Claude._

dkaygithub:
@thpr — thanks for the review. F1 fixed in c4ddfb4c (middleware.ts is now a boundary file in `is_boundary`/`is_companion`/`isBoundaryPath`). F3 split out as #1526 (single-source boundary set; base retargets to main after this merges).

On **F2** I'd like to push back. The route-kill check *can* fail open on a bad path guess — but the failure has no security consequence, because removing a route's registry entry fails **closed** at runtime:

```js
// handler.ts
const spec = getRoute(endpoint);
if (!spec) return apiError('Internal Server Error', 500);
```

An un-registered live route 500s — it serves nothing. So a wrongly-certified "kill" of a still-served endpoint is at worst an **availability** regression (a route goes down, loud, caught in staging/tests), never an **exposure widening** — which is the only invariant isolation exists to protect.

Contrast F1: deleting `middleware.ts` fails *open* (the login gate just disappears). Same "checker waives review" shape, opposite blast radius — which is exactly why F1 was a real hole and F2 isn't.

Happy to add the base-resolves guard anyway if you want the certified reason to be strictly truthful, but I don't think it's blocking.

thpr:
Re-reviewed at `9ddcd53d`. The rebase is clean — both scripts, the test, the deploy doc and CODEOWNERS are byte-identical to `c4ddfb4c`, the only AGENTS.md delta is main's own doc churn, and #1501's `"$BASE_SHA"..."$HEAD_SHA"` handling survived intact. Re-ran the parser and the certification probes against the four boundary changes main landed since the old base (#1502, #1492, #1489/#1490, #1559): 26 bindings entities, 22 registry entities, no parse error; a clean `Visit` decommission certifies; the `middleware.ts` deletion is rejected. **Finding 1 is fixed and holds after the rebase.**

Also **withdrawing finding 2** (route-kill "fails open" on a bad path guess) — reasoning posted on #1526, short version: removing a whole `defineRoute` never widens on either path, so a falsely certified route-kill unlocks exactly what a correct one unlocks and doesn't expand the blast radius past a legitimate decommission.

Two findings below, and both belong **here rather than on #1526** — flagging that explicitly since #1526 is stacked on this branch and it's ambiguous which PR owns them. Both are functional gaps in what this PR ships, not in the refactor; #1526 can just rebase over them. (I raised both on #1526 first — apologies for the routing noise.)

## 1. Security test files are boundary files, so the decommission PR blocks itself

`is_boundary` matches `checkin-app/src/security/*` with only `generated/` carved out. Shell `case` globs match `/`, so all ten files under `checkin-app/src/security/__tests__/` are classified as boundary files and reach `certifyDecommission`, which rejects anything outside `DECLARATIVE_FILES`. Verified on this branch:

```
boundaryChanged: [scopeBindings.ts]                          => ok: true

boundaryChanged: [scopeBindings.ts,
                  src/security/__tests__/visit-household-lead-scope.test.ts]
  => ok: false
     "…visit-household-lead-scope.test.ts: non-declarative boundary file —
       direction undecidable, never decommission-eligible"
```

Not a contrived path: `visit-household-lead-scope.test.ts` is a per-model scope oracle for `Visit` — exactly what a `Visit` decommission has to update or delete — and rule 7 as written in this PR names "plus test-oracle updates" as part of the single decommission PR. Deleting the oracle rather than editing it doesn't help; `is_boundary` classifies by path regardless of git status, so a `D` row poisons the set identically.

So the exception refuses to fire for the shape it was built for, and blames a test file while doing it. Worth confirming against the real #1444 FR7 diff before merge — the end-to-end verification in the PR description presumably didn't carry a `src/security/__tests__` change.

Cheapest fix is one arm above the existing one, mirroring how `generated/` is already carved out:

```bash
checkin-app/src/security/__tests__/*) return 1 ;;
```

`is_companion` already admits those paths, so they stay legal companions. One side effect worth naming so it's a decision and not a surprise: a PR touching *only* `src/security/__tests__/` would stop tripping isolation at all. Given `is_boundary`'s own comment scopes it to "policy/runtime files — the security boundary itself", a test file in that set reads as misclassification rather than intent — but it does change the base rule, not just the exception.

## 2. The CLI's `HEAD` is the merge ref

#1501 landed on main while this branch was out, and changed what `HEAD` means in this workflow: on a `pull_request` event checkout resolves `refs/pull/N/merge`, so `HEAD` carries main's advance. That fix replaced every `...HEAD` in the YAML with an explicit `$HEAD_SHA`. Post-rebase the workflow and the script it invokes now disagree, inside one file — the step computes `VIOLATIONS` from `"$BASE_SHA"..."$HEAD_SHA"`, then calls:

```yaml
node checkin-app/scripts/check-boundary-decommission.js --base "$BASE_SHA" -- "${VIOLATIONS[@]}"
```

and the script does `merge-base(baseSha, 'HEAD')`, `diff ${mergeBase}..HEAD`, and `gitShow('HEAD', p)`.

The reason this survived the rebase is that it isn't a conflict — the YAML hunk is, `check-boundary-decommission.js` has no markers at all, so nothing in the resolution surfaces it.

What it does when main drifts from `base.sha`:

- `changed` picks up main's files, so a boundary file **main** touched lands in `boundaryChanged` and certification is refused, naming a file this PR never touched.
- `readHead` returns the *merged* `scopeBindings.ts` / `registry.ts`, so an entry main added independently reads as `entry added: …` and the diff is refused.

Both fail closed — a flaky-refusal bug, not a hole. One narrow reversal: `violations` comes from the workflow's correct SHAs while `deleted` is built from the CLI's `changed`, so if main deletes a file this PR *modified*, `deleted.has(v)` is true and modified app code is admitted as "a deletion". Contrived, but a genuine flip.

Fix: pass `--head "$HEAD_SHA"`, use it for both the diff and `gitShow`, drop the internal `merge-base` — `"$BASE_SHA"..."$HEAD_SHA"` is already merge-base→head, and matching the workflow's own two SHAs exactly is the point.

## Smaller, unchanged since the first review

- `middleware.ts` is now a boundary file, so **any** PR touching it must ship alone — a new constraint on ordinary work that nothing records. The AGENTS.md bullet, rule 7, and the workflow's own "What this enforces, mechanically" header bullet are all silent (that last one is now stale on three counts: `middleware.ts` plus the two certifier scripts).
- `middleware.ts` isn't CODEOWNERS-gated, while every other boundary file is — a solo-PR requirement with no required reviewer attached.
- Rule 7 is worth one sentence saying that `ROW_SCOPE_KEY` removal **widens**: `makeScopesHeld` returns the empty set for a row missing its scope key, but with the entry gone `scopeKey === undefined` and the same row gets `everyones`. For that container the schema-exit check is the only guard, unlike `SCOPE_BINDINGS`/`registry.ts`, which narrow on their own. Without that noted, a later reader relaxes it on the reasonable-sounding belief that removals narrow.
- Still open from the first review, all minor: `process.exit(main(...))` stdout truncation, the `--name-status` comment claiming R/C rows surface both paths when only the destination is kept, and `stripStringsAndComments` handling single quotes and `//` only.

thpr:
Re-reviewed at `e9eb0191`.

## Fixed, verified

- **`--head "$HEAD_SHA"`** — the CLI now audits the PR's own tree; `merge-base`, the `--name-status` diff and `readHead` all use it.
- **Deletion latitude narrowed** to `impliedDeletions` (the route files of removed registry entries). This closes the *class* that finding 1 only patched one instance of — confirmed a `src/lib/fees.ts` deletion is now refused with the new message, not just `middleware.ts`.
- **`readBase` must export the verb at base**, with the resulting inert-entry corner documented in the workflow header.
- `process.exitCode`, the corrected R/C comment, and `middleware.ts` added to CODEOWNERS.
- **The `--no-renames` catch is a good one.** Rename detection reporting only the destination meant renaming `src/middleware.ts` kept its name out of `CHANGED` entirely — and since nothing imports it, tsc and the build stay green while the gate stops running.

I probed the new `normalizeEntry` relaxation adversarially and it holds: removing a `returns:` element for a model this PR drops is forgiven; removing a *surviving* model's element is refused; adding one is refused; an `authorize:` edit inside a surviving entry is refused. All `returns:` arrays in `registry.ts` are single-line, so the documented multi-line fail-closed corner doesn't bite the real file — and line 111 carries `'Fee'`, so the FR7 case works.

## Still open

### 1. `src/security/__tests__/*` is still classified as a boundary file

No `__tests__` arm was added to `is_boundary`, and it still reproduces:

```
boundaryChanged: [scopeBindings.ts,
                  src/security/__tests__/visit-household-lead-scope.test.ts]
  => "…visit-household-lead-scope.test.ts: non-declarative boundary file —
      direction undecidable, never decommission-eligible"
```

Scoping this more precisely than I did last time: **#1444 itself will not trip it** — no file under `src/security/__tests__/` mentions `Fee`/`FeePayment`. It bites the next decommission. `visit-household-lead-scope.test.ts` is the per-model scope oracle for `Visit`, so a `Visit` drop has to delete it, and deleting it is refused identically — `is_boundary` classifies by path, not by git status. A latent trap rather than an immediate blocker, but rule 7 promises "plus test-oracle updates" ships in the same PR.

The fix is still one arm above the existing one, mirroring the `generated/` carve-out:

```bash
checkin-app/src/security/__tests__/*) return 1 ;;
```

`is_companion` already admits those paths. The side effect to decide on deliberately: a PR touching *only* `src/security/__tests__/` would stop tripping isolation at all.

### 2. Rule 7 and the AGENTS.md bullet now describe behaviour the code no longer has

The workflow header was updated in this round; the docs were not.

- Both still say the certifier admits "file deletions". It now admits only the route files of removed registry entries. A PR packaged to rule 7's letter — model drop plus an unrelated deletion — gets refused.
- Rule 7 still calls the exception "byte-strict". Comment edits and `returns:` element removals for dropped models are now normalised away. That relaxation is deliberate and well argued in the code; the doc asserts the opposite of it.
- Neither carries the new inert-entry corner (workflow header only).

While that section is open, it is still worth the sentence that `ROW_SCOPE_KEY` removal **widens**: `makeScopesHeld` returns the empty set for a row missing its scope key, but with the entry gone `scopeKey === undefined` and the same row gets `everyones`. For that container the schema-exit check is the only guard, unlike `SCOPE_BINDINGS`/`registry.ts`, which narrow on their own. Without it recorded, a later reader relaxes the rule on the reasonable-sounding belief that removals narrow.

### 3. Minor: comment forgiveness is file-wide

`stripComments` runs over the whole file, not just entries naming a dropped subject. Verified: rewriting an unrelated rationale comment on a *surviving* `defineRoute` still certifies. It cannot widen exposure, and the header argues that honestly — noting it only because the registry's comments are the audit trail the isolation rule exists to protect, and scoping the forgiveness to comments that name a dropped subject would be a small tightening.

thpr:
Superseded by #1590, merged as `72d98303`. That PR carries all eleven commits from this stack with their original authorship, rebased onto `main`, plus the two remaining review items: the `src/security/__tests__/` carve-out in `is_boundary` (security tests were classified as boundary files, so a decommission blocked itself on its own scope oracle) and rule 7 / AGENTS.md rewritten to describe the certifier that actually shipped. Verified on `main`.

Closing rather than merging because the stack was rooted on `boundary-decommission-exception` — merging a stacked PR into a branch nothing will ever merge reports `MERGED` while the code never reaches `main` (#1447).

Issue #1165, opened 2026-07-20
Delete DoB for all adults


thpr:

thpr:
Needs to be deep delete - meaning the old value can't be captured in audit logs - we want no record of it

thpr:
Design + implementation shipped in #1326 (merged, commit 4e17af5).

No separate design doc — the design is captured in the PR description and inline code comments. Summary:

- **Threshold 26+, not 18+.** No age-gated program targets anyone over 25 (`MAX_PROGRAM_AGE`), so a DoB for a 26+ person is dead weight; 18-25 keep theirs because youth program caps (`maxAge` up to 25) still need exact age.
- **Policy seam:** `checkin-app/src/lib/person/adultDob.ts` — `normalizeAdultDob()` strips DoB + sets `isDeclaredAdult` for 26+.
- **Write-time strip** in the 4 interactive DoB-write routes (`profile`, `household`, `household/member`, `membership-ops` participant create).
- **One-time backfill:** `checkin-app/prisma/migrations/20260724120000_purge_adult_dob`.
- **Nightly self-heal:** `cron/nightly` re-runs the purge so members crossing 26 are cleaned within a day.

Deferred: bulk CSV import leans on the nightly sweep rather than a write-time strip (import is slated for rewrite).

PR #1495, opened 2026-08-05
feat(membership-audit): add the "students 18+ as of the member-year start" roster

### LINKED ISSUE #1230: "Students ≥18 as of Sept 1" report

thpr:
Fixes #1230 (backlog item M7).

Re-lands #1447, which merged into `claude/registry-turning-18` rather than `main`. That branch's own PR (#1490, the registry declaration) had already merged to `main` 29 minutes earlier, so #1447's app code was left stranded on the branch tip and never reached `main`. #1447 is closed and cannot be retargeted, so both of its commits are cherry-picked onto current `main` here — same nine files, no conflicts.

The registry entry this route depends on is already on `main` from #1490, so the handler wiring resolves without any boundary-file change in this PR.

## Why

The board needs to know which household members are legally adults **for a given member year** — the age-18 individual-agreement obligation and the person-level background-check trigger both key on it — and no surface answered that. Age was only ever judged as-of today, which is the wrong question when the obligation attaches at the year boundary.

## What

A board-only Membership Audit view (`/membership-audit/turning-18`, "Students 18+") listing non-lead household members who are 18 or older as of the **current** member-year start and as of the **next** one, so both the standing adults and the ones aging in on the coming boundary are visible in one table.

- **Household leads are excluded.** This is the "child of the signer turns 18" list, and a lead is the signer.
- **Not scoped to program enrollees.** Program students can sit in non-member households, so scoping the query would drop exactly the rows worth seeing. Instead every row carries the programs the person is enrolled in (any program not yet `FINISHED`), with a "Hide those not in a program" checkbox that filters the already-fetched list client-side.
- **No DOB → a count, not a blank row.** People whose age can't be judged come back as `unknownDobCount` and render as a hygiene note. `isDeclaredAdult` people are excluded from that count — a lead marked them 25+, there is nothing to chase.
- **The boundary is the configured `BoardSettings.orgMembershipYearBoundary`**, not a hardcoded Sept 1. With none configured the page says so and renders nothing to act on.

The age filter runs in the query (`dateOfBirth <= birthCutoff(next)`) rather than over every loaded person. `programYear.test.ts` pins `birthCutoff` against `calculateAge` on both sides of the boundary so the SQL-side bound and the displayed age can't drift apart.

## Reviewer notes

Three drift guards fired during the original build and each is registered rather than worked around:

- **`livePersonDriftGuard`** rejected `...LIVE_PERSON` spread through a shared local const — it text-scans the query's own argument object. Inlined at both query sites.
- **`authzRegistry`** required negative-authz coverage: a row in `authzRoleRejection.integration.test.ts` (401 anon / 403 plain user) plus the `AUTHZ_TESTED` entry.
- **`routeAuthDrift` rule 3** flagged the `ProgramParticipant` edge include (all-public-tier, so row *existence* is the sensitive thing). The handler is `withAuth roles [sysadmin, board]`, so it gets a keyed `EDGE_INCLUDE_ALLOWLIST` entry with that justification. That file is `checkin-app/tests/security/*`, an allowed companion under the boundary-isolation workflow — not a boundary file — so this does not need to ship as its own PR.

## Verification

Re-run against current `main` after the cherry-pick:

- `tsc --noEmit` clean.
- Unit tests covering the touched files (`programYear`, `routeAuthDrift`, `pageRegistry`): 3 suites, 458 passed.
- The two integration suites (`turningEighteenAPI`, `authzRoleRejection`) were **not** re-run locally — leaving those to CI.

From the original #1447 run: full unit suite 2428 passed / 257 suites; full integration suite 1260 passed / 128 suites; live query smoke against real Postgres returned `ageAtCurrent 17 / ageAtNext 18` for a synthetic member born 2008-08-15 with boundaries `2025-09-01` / `2026-09-01`; `npm run lint` clean.

Deliberately skipped: CSV export and column sorting. The list is small and the board's next step is opening a household, not exporting. Worth adding when it outgrows a screen.

---

**Update — rebased onto `main` @ `a00bed98`.** The branch carries a third commit, `feat(membership-audit): explain why this roster differs from the agreement lists`, which adds a note to the page explaining why this roster (age as of the member-year start) lists different people than the individual-agreement lists on Compliance (age as of today). #1477 carries the reciprocal note.

Re-verified after the rebase — which pulled in the UTC-midnight DOB changes (#1494, #1496) that this route's `dateOfBirth <= birthCutoff` filter depends on: `tsc --noEmit` clean; `programYear` / `routeAuthDrift` / `pageRegistry` suites 458 passed.

The earlier `isolation` failure on this PR was a false positive, not a boundary change. That job diffs the PR **merge ref** against `pull_request.base.sha`, which is pinned at PR-open. `main` advanced five commits afterwards — two of them registry PRs (#1492, #1489) touching `checkin-app/src/security/` — so the job attributed those upstream boundary files to this PR and listed 18 unrelated files as violations. This PR touches no file under `checkin-app/src/security/`.

### DIFF
--- a/checkin-app/src/__tests__/authzRegistry.test.ts
+++ b/checkin-app/src/__tests__/authzRegistry.test.ts
@@ -91,7 +91,6 @@ const AUTHZ_TESTED = new Set<string>([
     'membership-audit/person-bg',
     'membership-audit/households-missing-contact',
     'membership-audit/unclaimed-households',
-    'membership-audit/turning-18',
     'membership-ops/applications/archive',
     'membership-ops/applications/unarchive',
     'membership-ops/applications/certify-payment',
--- a/checkin-app/src/__tests__/authzRegistry.test.ts
+++ b/checkin-app/src/__tests__/authzRegistry.test.ts
@@ -91,6 +91,7 @@ const AUTHZ_TESTED = new Set<string>([
     'membership-audit/person-bg',
     'membership-audit/households-missing-contact',
     'membership-audit/unclaimed-households',
+    'membership-audit/turning-18',
     'membership-ops/applications/archive',
     'membership-ops/applications/unarchive',
     'membership-ops/applications/certify-payment',
--- a/checkin-app/src/app/__tests__/authzRoleRejection.integration.test.ts
+++ b/checkin-app/src/app/__tests__/authzRoleRejection.integration.test.ts
@@ -48,6 +48,7 @@ import { GET as FAC_VISITS_GET, PATCH as FAC_VISITS_PATCH } from '@/app/api/faci
 import { POST as FAC_VISITS_INSERT_POST } from '@/app/api/facility/visits/insert/route';
 import { GET as MISSING_CONTACT_GET } from '@/app/api/membership-audit/households-missing-contact/route';
 import { GET as UNCLAIMED_GET } from '@/app/api/membership-audit/unclaimed-households/route';
+import { GET as TURNING_18_GET } from '@/app/api/membership-audit/turning-18/route';
 import { POST as CERTIFY_PAYMENT_POST } from '@/app/api/membership-ops/applications/certify-payment/route';
 import { POST as APP_EXTERNAL_POST } from '@/app/api/membership-ops/applications/external/route';
 import { POST as REVIEW_OVERRIDE_POST } from '@/app/api/membership-ops/applications/review-override/route';
@@ -192,6 +193,7 @@ describe('Protected-route role rejection', () => {
         { name: 'POST /api/facility/visits/insert', invoke: () => FAC_VISITS_INSERT_POST(nreq('http://localhost/api/facility/visits/insert', 'POST', {})) },
         { name: 'GET /api/membership-audit/households-missing-contact', invoke: () => MISSING_CONTACT_GET(nreq('http://localhost/api/membership-audit/households-missing-contact')) },
         { name: 'GET /api/membership-audit/unclaimed-households', invoke: () => UNCLAIMED_GET(nreq('http://localhost/api/membership-audit/unclaimed-households')) },
+        { name: 'GET /api/membership-audit/turning-18', invoke: () => TURNING_18_GET(nreq('http://localhost/api/membership-audit/turning-18')) },
         { name: 'POST /api/membership-ops/applications/certify-payment', invoke: () => CERTIFY_PAYMENT_POST(nreq('http://localhost/api/membership-ops/applications/certify-payment', 'POST', {})) },
         { name: 'POST /api/membership-ops/applications/external', invoke: () => APP_EXTERNAL_POST(nreq('http://localhost/api/membership-ops/applications/external', 'POST', {})) },
         { name: 'POST /api/membership-ops/applications/review-override', invoke: () => REVIEW_OVERRIDE_POST(nreq('http://localhost/api/membership-ops/applications/review-override', 'POST', {})) },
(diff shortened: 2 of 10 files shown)


thpr:
Added the reciprocal cross-reference note to the roster page (9bc56a33), pairing with the one on Compliance in #1477.

Both surfaces sit under `/membership-audit` over structurally the same population (both 18–25, since #1165 strips the DOB at 26) but judge age differently on purpose — this roster as of the member-year start, the agreement lists as of today. A board member reading them back to back would otherwise see different people and no explanation. Each now names the other and says why they disagree: anyone with a birthday between today and the next member-year start appears on the agreement lists and not here.

Pushed a plain commit onto this branch — it was clean and level with origin at the time. Type-check, lint, and the membership-audit suites are green.

jee7s:
The age math holds up — I tried to break it and couldn't. `birthCutoff` is the exact SQL inverse of `calculateAge`, both UTC; ages are derived at the boundary rather than at today; `LIVE_PERSON` is at the query site; the registry grant is deliberately narrow. One gap worth a look, and it is narrower than it first appears.

## Null DOB never matches the filter

`route.ts:37` filters `dateOfBirth: { lte: birthCutoff(next) }`. `NULL <= x` is NULL in SQL, so any row without a date of birth is silently absent — not excluded by a rule, just never considered.

**For the population this roster is about, that is mostly the right answer**, and worth stating so the fix doesn't overshoot. Program participation caps at `MAX_PROGRAM_AGE = 25` (`programAge.ts:78`, enforced by `validateProgramAgeBounds`), and the two paths that null a DOB only fire above that line: the nightly purge at `cron/nightly/route.ts:83` (`>= 26 years`) and `normalizeAdultDob` (`age > MAX_PROGRAM_AGE`), both of which set `isDeclaredAdult = true` on the way out. So a purged row is out of program range by construction. The roster is not losing students, and the check only really applies to participants decidedly under 26.

## The residual that does matter: `DOB_MISSING`

There is a third way to hold a null DOB, and it is not an adult. `normalizeAdultDob(null)` returns `{ dateOfBirth: null }` and leaves `isDeclaredAdult` to the caller (`adultDob.ts:32`), so a household member added without a birthdate and without the over-25 tick sits at `{ dateOfBirth: null, isDeclaredAdult: false }` — **age unknown, not age ≥ 26**.

That person could be 17 and about to turn 18, which is precisely who this list exists to surface. They are absent, and nothing on the page says so.

The codebase already treats this as a distinct case rather than folding it into "adult": `personBgVerdict` returns `DOB_MISSING` with the comment *"age unknown, so we can neither..."* (`personBgCheck.ts:36,63`), and the sibling compliance view routes it as its own verdict for exactly this reason.

## Suggestion

Rather than widening to `isDeclaredAdult` — which would pull in the 26+ population the purge deliberately removed — surface the unknowns as their own small count, the way the compliance view does:

```ts
// alongside the main query
const unknownAge = await prisma.person.count({
    where: { ...LIVE_PERSON, isHouseholdLead: false, dateOfBirth: null, isDeclaredAdult: false },
});
```

Filtering on `isDeclaredAdult` needs no `select`, so this raises no `@sensitivity` question.

Two smaller things that follow from the same point:

- **The empty state asserts a negative the query cannot support.** `page.tsx:120` says *"Nobody turns 18 by the next member year."* With unknowns excluded, the honest sentence is that nobody *with a recorded birthdate* does.
- **The heading is "18+" but the set is effectively 18–25**, since 26+ have no DOB to match on. Worth a word in the description so a reader doesn't take the absence of older members as data.

## Sequencing

#1477 links to this route at `membership-audit/compliance/page.tsx:259` and that page is already in its diff, so #1477 ships a dead link until this lands. Worth merging this first, or dropping the anchor there.

---
_Posted by Claude._

jee7s (approved):

PR #1497, opened 2026-08-05
docs(attendance): design proposal for the correction-review screen (AT12)

jee7s:
Design proposal for #1258 (backlog AT12, the admin hour-correction review screen). Doc only — no code, no schema change, nothing built.

#1258 was assessed NOT READY on 2026-07-26 with six clarifying questions. This closes five of them from the code and narrows the sixth. What remains open is two product judgements that code cannot settle, both with a recommendation.

## Gated on #1523

Per @thpr, the defects this design depends on were moved out of the document and onto the PR that owned them, as a CHANGES_REQUESTED review on #1478. That has now played out:

1. **`secondaryAffectedEntity` holds the subject person on every visit audit write — satisfied.** #1478 closed the three broken writers and merged; I re-verified all twelve writes on `main` at `d4b38c5b`.
2. **Every edit and delete path persists `newData.significance` — still open, now #1523.** It was raised on #1478 as non-blocking, and #1478 merged without it.

So requirement 2 outlived its host and is filed as **#1523** rather than left in a merged PR thread. §7 gates on that, and records the cost this document predicted: deferring the fix turned four lines inside an existing PR into a change of its own.

## What the design settles

- **The actor axis is a comparison, not a join** (§2), given the invariant above. The design's contribution is the argument for *not* special-casing the exceptions at read time: a read-side workaround for rows written before the fix can never be retired, because nothing on a row records which convention it followed.
- **Read the persisted flag, do not recompute it** (§3). Recomputing cannot be filtered or paginated in the database, so the default view would load and score every row in the range before cutting the first page. The drill-down recomputes, since it has one row loaded, and shows both values when they disagree — which is also what makes threshold recalibration observable.
- **Ship on the gate Facility Ops already enforces** (§4), sysadmin and board. The governing design says ops too, but `facility-ops/layout.tsx:10` locks the whole section to sysadmin+board. Granting ops the route while the layout keeps them out recreates AT13, the defect this design family exists to close. #1476 is the open decision; this widens with everything else when that lands.
- **Reuse `getPeriodStart` from `facility/trends`** (§4). #1423 made it org-timezone-correct; extract it rather than writing a third bucketing implementation.
- **The participant merge is a documentation error** in the governing design's audit-coverage inventory, not a gap (§1). It writes no `Visit` audit row and should not start. One consequence survives and is handled: a subject id on an older row may point at a person since merged away, which `Person.mergedIntoId` resolves.

## On #1502

§2 now says explicitly why `led_households` cannot serve as the actor axis, since it is the obvious thing to reach for. It is resolved for the *viewer* at request time from current household leadership, never recorded on the audit row, so it cannot say what relationship held when the edit happened. The appendix cites #1502 alongside #1478 as the registry-first precedent this design follows in §4.

## Still open

Two, both with a recommendation, both in §6: whether a roster-mark insert deserves a fabricated significance score (recommend no), and whether v1 ships the threshold-calibration control (recommend read-only first).

## Placement

`docs/in-design/` per `DOCUMENTATION_STANDARD.md` §4, with no `Status:` header — §6.7 is explicit that the folder carries that meaning. The doc is deleted when the work merges, per the lifecycle table in §2 of that standard.

## Verification

Code claims checked against `origin/main` at `0441d65c` and re-verified against #1478's current head `6d128500` before the review was filed. No code, so no tests to run and no `tsc` to clear.

Design PR, so it references #1258 without a closing keyword — the issue stays open for implementation pickup.

---
_Posted by Claude._

### DIFF
--- a/docs/in-design/ATTENDANCE_CORRECTION_REVIEW.md
+++ b/docs/in-design/ATTENDANCE_CORRECTION_REVIEW.md
@@ -3,8 +3,8 @@
 Issue: [#1258](https://github.com/innovationtreehouse/checkin/issues/1258) (backlog AT12)
 
 Governing design: `checkin-app/docs/designs/1256_ATTENDANCE_CORRECTION_SURFACE.md` §4.
-This is a proposal; nothing here is built, and §1 and §2 correct three of its
-claims.
+This is a proposal; nothing here is built. §1 and §2 disprove three of its §4
+claims, and §7 puts correcting them in the implementation PR's scope.
 
 ---
 
@@ -68,8 +68,7 @@ says why for each.
 ## 1. The substrate: what the audit trail actually records
 
 The screen has three axes: kind, actor class, and time. Kind and time come off
-the audit row directly. Actor class does not, and this is where the governing
-design's §4 rests on a premise that is not true.
+the audit row directly. Actor class does not.
 
 `AuditLog` carries `actorId`, `action`, `tableName`, `affectedEntityId`,
 `secondaryAffectedEntity`, `oldData` and `newData` (`checkin-app/prisma/schema.prisma`).
@@ -77,57 +76,36 @@ Every field is `@sensitivity:internal`. Both id columns are bare `Int` with no
 foreign key and no discriminator; nothing in the schema says what
 `secondaryAffectedEntity` points at, and nothing stops two meanings sharing it.
 
-Every writer that logs a `Visit` row on `origin/main` at `0441d65c`:
+**The actor axis is sound on `main` today.** An earlier draft of this section
+catalogued five writers that stored a non-person id in `secondaryAffectedEntity`;
+#1478 closed the last three, and all twelve visit audit writes now store the
+subject. That history is not repeated here — §2 states the invariant the screen
+depends on, and §7 records that it is met.
 
-| # | Writer | Action | Subject identity | Before/after times | Significance |
-|---|---|---|---|---|---|
-| 1 | `attendance/manual/route.ts:143-151` | CREATE | `Number(userId)`, the member, who is also the actor | after only; a create has no before | no |
-| 2 | `attendance/manual/[id]/route.ts:127-136` | EDIT | `visit.personId` | both, with the old sources | **yes** |
-| 3 | `attendance/manual/[id]/route.ts:176-185` | DELETE | `visit.personId` | before, with sources | **yes** |
-| 4 | `events/[id]/attendance/route.ts:117-125` | EDIT | **the event id**; the person is in `newData.participantId` | none; sets `associatedEventId`, touches no time | no |
-| 5 | `events/[id]/attendance/route.ts:147-155` | CREATE | **the event id**; the person is in `newData.participantId` | none | no |
-| 6 | `facility/visits/route.ts:102-109` | EDIT | **not set** | **after only; no `oldData` at all** | no |
-| 7 | `facility/visits/route.ts:159-166` | DELETE | **not set**; person is inside a whole-row `oldData` | before, with sources | no |
-| 8 | `my-programs/conflicts/resolve/route.ts:63-79` | DELETE | **`visit.associatedEventId`**, which is nullable; person is in `oldData.personId` | before, with sources | no |
+What survives is narrower, and §3 rests on it. Two shapes of audit row carry no
+diffable content, and no fix changes that because there is nothing to record:
 
-Paths under `checkin-app/src/app/api/`. #1478 has since added further writers and
-repaired the broken ones; §7 states what this design still needs. Three
-consequences follow, and they are the whole of this design's difficulty.
+| Writer | Why there is nothing to diff |
+|---|---|
+| `events/[id]/attendance/route.ts` — roster mark (CREATE and EDIT) | Stores no times at all. A create has no earlier value, and the edit sets `associatedEventId` without moving a clock. |
+| `facility/visits/route.ts` — EDIT, rows written before #1478 | Stored no `oldData`. #1478 added it, so only the pre-fix tail is affected. |
 
-**The self-versus-staff test does not work.** Governing design §6.6 offers
-`actorId === secondaryAffectedEntity` as the cheap proof of a self-correction.
-Rows 4, 5 and 8 put an event id in that column and rows 6 and 7 leave it null.
-Both columns are `Int`, so the comparison does not fail on a type; it returns
-"self" whenever a person id and an event id collide, which they will, because
-both sequences start at 1. A screen whose primary axis is "did this person edit
-their own record or someone else's" cannot rest on a test that answers wrong
-without saying so.
+**Significance therefore cannot be recomputed uniformly**, which is the whole of
+§3's argument. `editSignificance` needs the old times and the old
+`arrivedVia`/`departedVia` to weight them
+(`checkin-app/src/lib/visit/significance.ts`); neither shape above supplies them.
 
-**The fallback does not work either.** §4 offers `newData.type` as the
-discriminator: `manual_entry` and `self_correction` mean self, absence means
-staff. #1478 makes a household lead's proxy correction write
-`type: "self_correction"` with an actor who is not the subject, which is exactly
-the case the axis exists to separate. A marker naming the *route* cannot answer
-a question about the *relationship between two people*.
-
-**Significance cannot be recomputed uniformly.** §4 says "the same significance
-function runs at read time". `editSignificance` needs the old times and the old
-`arrivedVia`/`departedVia` to weight them (`checkin-app/src/lib/visit/significance.ts`).
-Row 6 stores no `oldData`, so there is nothing to diff. Rows 4 and 5 store no
-times at all, because a roster mark has no earlier value and an association
-change moves no clock.
-
-**The participant merge is a documentation error, not a gap.** The governing
-design's audit-coverage inventory lists it as a `Visit` audit writer. It is not
-one: the merge re-parents visits with
+**The participant merge is a documentation error in the governing design, not a
+gap in the data.** Its audit-coverage inventory lists the merge as a `Visit`
+audit writer. It is not one: the merge re-parents visits with
 `tx.visit.updateMany` and logs `tableName: "OrgMembershipProcess"` and
 `tableName: "Person"` only. That behaviour is right and should stay — a merge
 moves rows between two records of the same human and changes no attendance fact,
-so auditing it as a correction would fill the feed with non-corrections. The
-claim should come out of that inventory. One consequence does survive: a subject id in an
-older audit row may belong to a person since merged away. Merges delete nothing
-and `Person.mergedIntoId` records the survivor, so the drill-down resolves the
-name and can follow the pointer.
+so auditing it as a correction would fill the feed with non-corrections. One
+consequence does survive into this screen: a subject id on an older row may
+belong to a person since merged away. Merges delete nothing and
+`Person.mergedIntoId` records the survivor, so the drill-down resolves the name
+and can follow the pointer.
 
 ## 2. The actor axis, and what it requires of the writers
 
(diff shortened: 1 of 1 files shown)


thpr (changes requested):
This hides issues with another PR in a design doc.  Those items belong as a "request changes" on the PR and this PR should gate on those being done correctly

thpr:
see also #1502

jee7s:
@thpr agreed, and fixed. The findings were doing no work where they were — #1478's author had no reason to read a design doc for AT12, and nothing blocked on them.

**Moved to #1478 as a CHANGES_REQUESTED review**, with file:line and the exact change for each:
1. Three writers store an event id in `secondaryAffectedEntity` (`events/[id]/attendance/route.ts:123,153` and `my-programs/conflicts/resolve/route.ts:69`).
2. `newData.significance` is persisted by one file out of five write paths.
3. The audit-coverage inventory lists the participant merge as a `Visit` audit writer; it writes none.

I re-verified all three against #1478's current head `6d128500` before filing — it had moved from `882b72c6` since I first looked, and I did not want to request changes against a stale ref.

**This PR now gates on that.** New §7 states both requirements and the sequence (#1478 resolves them → this merges → registry entry alone → route and page). §2 and §3 keep only the part that is actually design: the argument for why the screen must *not* absorb the gap at read time. A read-side workaround for rows written before the fix can never be retired, because nothing on a row records which convention it followed — so the invariant has to hold at the writers first. Net −27 lines.

**On #1502** — thanks, that changed §2. `led_households` is the obvious thing to reach for as the actor axis, so the doc now says explicitly why it does not work: it is resolved for the *viewer* at request time from current household leadership, never recorded on the audit row, so it cannot say what relationship held when the edit happened. I also cited #1502 as the registry-first precedent §4 follows, since it is a cleaner example than the one I had.

Separately, reviewing #1502 turned up a pre-existing gap in a file it touches — `activeVisitorIds` is built without a `deletedAt` or `LIVE_PERSON` filter, so a tombstoned open visit reads as active forever. Filed as #1503 rather than raised here, for the same reason as above. It is latent today (no route grants `all_current_visitors`), so it is not urgent, but #1502 is a natural host if it has not merged.

---
_Posted by Claude._

jee7s:
@thpr the gate you asked for has now resolved, and half of it went the other way — updated in `32575f73`.

**Requirement 1 — satisfied.** #1478 closed the three writers and merged. I re-verified every visit audit write on `main` at `d4b38c5b`: all twelve store the subject person (`pId`, `visit.personId`, `targetId`, `subjectId`, `existing.personId`, `removed.personId`). §2's actor axis rests on that, and it now holds.

**Requirement 2 — orphaned, so filed as #1523.** Persisting `newData.significance` on every edit and delete path was raised on #1478 as non-blocking, and #1478 merged without it. One file of seven scores today. It cannot be recovered at read time — some of those rows carry no `oldData` and some carry no times at all — so it is only closable at the write.

That is exactly the cost this document argues against for deferred fixes: four lines inside an existing PR became a change of its own. §7 now says so plainly rather than quietly re-pointing the gate.

This PR now gates on **#1523**, not #1478. Your original point still stands as the reason the doc is shaped this way: the findings are on the PRs and the issue where someone can act on them, and the design only records what it depends on.

---
_Posted by Claude._

thpr (changes requested):
The design is sound and the §7 gating is right. Two things block implementation, three should be fixed before merge.

## Blocking — §4 does not work as written

**The response shape can't pass through `handler()`.** `stripBag` keys the bag by model name and drops anything else (`src/security/stripper.ts:30`); `stripValue` keeps only fields present in `classifications[model]`, fail-closed (`src/security/stripper.ts:63`). The proposed body `{ buckets, rows, total, page, pageSize }` has no model keys, and `kind`, `actorClass`, `actorName`, `subjectName`, `flagged`, `score`, `before`, `after`, `visitId` are all derived names, not `AuditLog` fields. Under the `defineRoute` entry the doc prints, this route returns `{}`.

None of the 19 registered routes returns an aggregate, so there is no precedent to copy — this design is the first, and the shape has to be resolved here. It can't be deferred to the route PR, because the registry entry ships **alone and inert** ahead of it: get it wrong and you land a boundary PR that describes a response that cannot exist.

**Related:** `returns` omits `Visit`. Whatever shape survives, `before`/`after` carry `arrivedAt`/`departedAt`/`arrivedVia`/`departedVia`, which are `Visit` fields.

## Before merge

**§1 contradicts §2 and §7.** §1 is present-tense about defects that no longer exist — the actor-axis test works on `main`, all twelve writers store the subject, and the merge inventory error is already corrected in 1256. A reader hits an 8-row defect table, then learns two sections later that none of it holds. Keep only what §3 still depends on: rows 4 and 5 carry no times, row 6's pre-fix rows carry no `oldData`. Past-tense the rest and the table shrinks to two or three rows. The appendix's "corrected in §1 and §2 above" goes with it.

**`period=day` doesn't exist.** §4's query params offer `day|week|month`; `PeriodType` is `"week" | "month" | "quarter" | "year"` (`src/app/api/facility/trends/route.ts:9`) and the function's final `else` buckets to **year**. Reuse-as-is turns `period=day` into year buckets. Drop it or widen the extracted helper, and say which.

**Name the `docs/rules` extraction target.** `DOCUMENTATION_STANDARD.md` §4 is *extract, then delete* — the appendix only has the delete. One rule here passes §3.9 cold: *every visit audit write stores the subject person in `secondaryAffectedEntity`.* It currently lives in three code comments and merged design prose; `docs/rules/attendance-checkin.md` has no mention of it. §7 marking it "Satisfied" is precisely the phrasing that stops anyone from extracting it. This design's actor axis is a bare integer comparison against that column — if a future writer breaks it, the screen returns wrong answers silently.

**1256 §4 stays wrong.** `newData.type` (line 559), read-time recompute (line 564), `sysadmin + board + ops` (line 575) — all three disproved here, none corrected there. This doc deletes itself at merge, so the corrections go with it. Add the §4 edit to the implementation PR's scope.

## Housekeeping

Branch is three commits behind `main` and still pins `0441d65c`. The appendix cites #1478 at `6d128500`, which predates the two commits that actually satisfied requirement 1 — the PR body's `d4b38c5b` is the correct pin.

## Sequence

Resolve the §4 shape → merge this → #1523 → registry entry alone → route and page. The doc's own sequence puts #1523 first, which is fine, but the §4 shape is independent of it and is the thing that will bite.

---
_Posted by Claude._

jee7s:
@thpr all six addressed in `7b75fbea`, rebased onto `8a420411`.

**The §4 blocker is real and I verified the mechanism rather than taking it.** `stripBag` (`stripper.ts:29-32`) drops any bag key that is not a model name; `stripValue` (`:63-64`) copies only fields present in `classifications[model]`. Under the entry as printed the route returns `{}` — every one of `buckets`, `rows`, `total`, `page`, `pageSize`, `kind`, `actorClass`, `flagged`, `before`, `after` is dropped.

I also went looking for an escape hatch and there isn't one: `handler()` always strips, all 76 classified entries are real Prisma models with no synthetic-view support, and `_count` is special-cased only for *relation* counts, not arbitrary groupings. Nor can a new route sit this out on `withAuth` — `new-route-old-authz` blocks an unregistered new route even in advisory mode.

So §4 no longer specifies a shape. It states the blocker, gives the three options with their real costs, and recommends one:

1. **Model rows, client derives** — passes today, but `oldData`/`newData` are `internal` and the view grants `everyones:internal`, so they ship whole. That is the `system-status/audit-log` shape this section was written to avoid, and it breaks §3's pagination argument because bucket counts from one page describe only that page.
2. **Two requests** — no help; a count is still not a model field.
3. **Extend the boundary** — the right answer, and #1518's per-caller-view `select` is exactly that mechanism.

Recommendation is (1) for v1 with the tight-select principle **explicitly suspended and the reason recorded**, revisited under #1518. The thing that must not happen is (1) shipped silently as though it were intended, because then §4's own argument against the forensic-log shape applies to this screen and nothing says why.

`returns` gains `Visit`. Good catch — `before`/`after` are `Visit` fields and the entry omitted it.

**`period=day`** — confirmed, `PeriodType` is `week|month|quarter|year` and `getPeriodStart`'s final `else` falls through to **year**, so an unrecognised value silently buckets to year rather than erroring. Dropped; the design takes the four that exist and says so.

**§1** was arguing with §7, exactly as you said. It now keeps only what §3 actually depends on — the two row shapes that carry nothing to diff (the roster mark, and pre-#1478 `facility/visits` edits) — and the eight-row table is gone. The appendix's "corrected in §1 and §2 above" went with it.

**The extraction target is named**, and I verified it is absent: `docs/rules/attendance-checkin.md` does not mention `secondaryAffectedEntity` anywhere. §7 now carries the rule to extract, plus your point that "Satisfied" is the phrasing that stops anyone extracting it — a satisfied requirement reads as finished rather than as a rule to keep.

**1256 §4's three claims** are now in the implementation PR's scope with line numbers (559, 564, 575), since this file's deletion would otherwise take the corrections with it.

**Sequence** updated to put the §4 shape first: it is independent of #1523 and it is the thing that blocks the registry entry, which ships alone and inert ahead of the handler.

Still gated on **#1523** for the significance requirement.

---
_Posted by Claude._

thpr (changes requested):
All five findings from the previous review are addressed. §4 now states the boundary blocker with its verified mechanism and three costed options, §1 drops the stale defect table, `returns` gains `Visit`, `period` drops `day`, and §7 gains the rules extraction plus the 1256 §4 corrections.

Re-verified the new claims against `main` at `8a420411`: all 76 `classifications` entries are real models; `new-route-old-authz` blocks even in advisory mode (`scripts/check-route-coverage.ts:367`); the `stripper.ts` line references are correct; #1518 is open and is the per-caller-view `select` work; and #1478 did add `oldData` to the `facility/visits` PATCH, so §1's "pre-fix tail" is accurate.

Three things remain.

## Blocking — the `orderedView` omits `everyones:personal`

`Visit.arrivedAt` and `departedAt` are **`personal`**, not public:

```
Visit: { id: 'public', personId: 'public',
         arrivedAt: 'personal', departedAt: 'personal',
         arrivedVia: 'public', departedVia: 'public', … }
```

The entry printed in §4 grants `['everyones:internal', 'public']`. So adding `Visit` to `returns` — which this commit correctly does — yields `id`, `personId`, `arrivedVia`, `departedVia` and `associatedEventId`, and **not the times**. The screen's entire before/after column strips to null.

The fix is one token: `['everyones:personal', 'everyones:internal', 'public']`, matching what `POST /api/facility/visits/insert` already grants both roles. This has to be right in the registry-only PR, since that is the one that ships alone and inert.

## Option 1 is a tier bypass, not just a tight-select concession

§4 frames option 1's cost as "the `system-status/audit-log` shape". The sharper problem is that `oldData`/`newData` are `internal` JSON blobs which *contain* `personal` Visit times. Field tiers do not reach inside JSON, so shipping the blobs whole under an `everyones:internal` grant delivers `personal` data to a view that was never granted `personal`.

That is not relaxing a principle, it is routing around the classification — worth saying plainly, because once the grant above is corrected the honest version costs nothing extra.

## A fourth option §4 does not list

`stripValue` copies only fields *present on the object* (`stripper.ts:64` — `if (!(field in obj)) continue`). It never requires them to have come from the database. So the handler can synthesize bag entries: `{ AuditLog: rows with the blobs deleted, Person: [...], Visit: [...] }`, where the `Visit` entries carry `arrivedAt`/`departedAt`/`arrivedVia`/`departedVia` extracted from the blobs. Those are real `Visit` field names, so they classify as `Visit` data under an honest `everyones:personal` grant, while the raw blobs never leave the handler.

Honest limits. The `after` state is a real `Visit` row fetchable by id, but `before` exists only in `oldData`, and one bag cannot hold two `Visit` arrays keyed to the same id. Derived scalars still have nowhere to live: `action`/`actorId`/`secondaryAffectedEntity` give `kind` and `actorClass` by client-side comparison, but `flagged`/`score` need `newData` retained.

So it halves the exposure rather than eliminating it — ship `newData` (the small blob), keep `oldData` server-side. Worth costing against option 1 before accepting the suspension, since it needs nothing from #1518.

## Structural — §7's gate is circular

The sequence now opens with "resolve §4's response shape → #1523 lands → this design merges". But §4's own argument is that the shape question "lands here rather than in the route PR". As written, the doc gates its own merge on a question it is the place to answer, so it can never merge.

§4 does answer it — blocker stated, three options costed, one recommended. Say so: the shape is **decided here**, and #1523 is the only gate. Otherwise the recommendation reads as a placeholder.

---
_Posted by Claude._

jee7s:
@thpr all three addressed in `b6b431d1`. I verified each against `main` at `8a420411` before acting — all three hold.

**The grant was wrong and you're right that it had to be caught here.** `classifications.ts:281-282` — `arrivedAt` and `departedAt` are `personal`, everything else the screen needs from `Visit` is `public`. Under `['everyones:internal', 'public']` the entry returns the ids and the via-provenance and strips both timestamps, so the before/after column renders empty. Now `['everyones:personal', 'everyones:internal', 'public']`, matching `POST /api/facility/visits/insert`.

Two deliberate departures from that precedent, since the wave-1 review flagged unearned grants as their own problem: **`pii` stays ungranted** — only `Person.name` is needed and that is `public`, so email cannot ship — and **`member` stays ungranted**, because no field on `AuditLog`, `Person` or `Visit` carries that tier. The sibling route grants `member`; copying it would have been cargo-cult.

**Your reframing of option 1 is the more important correction, and I've taken the wording.** I had it as a tight-select concession. It is not: `oldData`/`newData` are `internal` blobs *containing* `personal` times, field tiers do not reach inside JSON, and so shipping them whole under an `everyones:internal` grant delivers `personal` data to a view never granted `personal`. Routing around the classification, not relaxing a principle. Stated that way it stops being a defensible v1 trade-off.

**Option 4 is now the recommendation.** I confirmed the mechanism — `stripper.ts:64` is `if (!(field in obj)) continue` followed by a plain copy, with no provenance check anywhere — so synthesized `Visit` entries classify as `Visit` data and are governed by the honest grant while the raw blobs stay in the handler. It needs nothing from #1518, which is what makes it strictly better than (1).

I wrote the limits into §4 rather than leaving them for the route PR to discover: `after` is a real `Visit` row fetchable by id, `before` exists only inside `oldData` and one bag cannot hold two `Visit` arrays keyed to the same id — so `before` needs a separate key or a stated convention, and that is the one piece the route PR still designs. `kind` and `actorClass` fall out client-side from `action`/`actorId`/`secondaryAffectedEntity`, which do ship; `flagged` and `score` live in `newData`. So it halves the exposure rather than eliminating it: ship `newData`, keep `oldData` server-side. That trade is now explicit rather than inherited.

**The circular gate is gone.** §7 listed "resolve §4's shape" as a precondition while §4 argued the question can only be answered there — so the doc gated its own merge on itself, exactly as you describe. The sequence is now **#1523 → this merges → registry entry alone → route and page**, with a sentence in §7 saying the shape is decided in §4 and why it cannot be deferred. **#1523 is the only gate.**

---
_Posted by Claude._

thpr (approved):
All four items from the previous review are addressed, and the fixes check out against `main` at `8a420411`.

Verified: `classifications.ts:281-282` is exactly `arrivedAt`/`departedAt`. The `member`-tier justification is right — no field on `AuditLog`, `Person` or `Visit` carries that tier. The `pii` exclusion is right — `googleId`, `email` and `phone` are the only `pii` fields on `Person`, and `name` is `public`. And option 4's stated limit is real: `relations['AuditLog']` is empty, so nothing nests under an audit row, and the bag is a flat `Record` — one `Visit` key, with no way to distinguish `before` from `after` by `id` since they share it.

Approving. Two note-level items to fold in, neither blocking.

## The `everyones:personal` grant is tier-wide, and the entry's comment implies otherwise

Tier tokens apply across every model in the bag — there is no per-model form in the grammar (`scopes.ts` scopes are per-*row* predicates bound per model, not per-tier narrowings). So `everyones:personal` also opens `Person.dateOfBirth`, `allergies`, `notificationSettings` and `emailSuppressed` at the stripper.

This is not a live leak. Under option 4 the handler synthesizes the bag and selects `{ id, name }`, so those fields are never on the object and `if (!(field in obj)) continue` skips them. But the entry's comment still reads:

```
// Bag: rows derived from AuditLog, with actor/subject names resolved
// through Person (id and name are both public tier).
```

An auditor reading the registry entry alone sees `everyones:personal` plus `Person` in `returns`, and has no way to tell that DOB and allergies are out of scope — the comment actively suggests `Person` needs nothing above `public`. One line fixes it: `everyones:personal` is there for `Visit.arrivedAt`/`departedAt`, and the handler selects only `{ id, name }` from `Person`. Cheap, and it is precisely what the boundary-isolation rule exists to make auditable in a registry-only PR.

## Stale sentence at line 347

> Whatever is chosen, `returns` must include **`Visit`** … and the entry above omits it.

The entry above no longer omits it — this commit added it. The paragraph is also redundant now that option 4 is chosen rather than "whatever is chosen". Cut it, or fold the field list into the `everyones:personal` rationale above.

---

With those folded in the document is in good shape. The merge itself remains gated on #1523 by §7's own terms.

---
_Posted by Claude._

PR #1499, opened 2026-08-05
feat(membership): an intake note no longer holds payment

thpr:
## What changed

An applicant note ("anything else we should know?") used to park the membership application at `PENDING_BG_REVIEW` until two reviewers cleared the background check. A family that wrote anything in that box could not pay until a human had read it. That gate is removed: the note is still surfaced to the reviewers, but the review now runs in parallel with payment like every other application.

**Service layer**
- `advanceExternalIfComplete` always advances to `PENDING_PAYMENT` — the `holdForNote` branch and the `intakeNotes` read are gone.
- `submitIntake` / `beginRenewal` — a note no longer disqualifies the `bgClearedAt` shortcut for a household that already holds a still-valid check.
- `blockedResetStatus` — an unpaid board reset returns to `PENDING_PAYMENT` regardless of a note.

**Lifecycle**
- Removed the `PENDING_EXTERNAL_ACTION → PENDING_BG_REVIEW` edge; `BLOCKED → PENDING_BG_REVIEW` is now tagged `kind: PERSON_BG`, and `PENDING_BG_REVIEW → PENDING_PAYMENT` is marked `legacy`. `docs/generated/lifecycle/membership.md` regenerated via `npm run generate:lifecycle-artifacts`.

**UI**
- Dropped the red *"Your application will pause for human review before you can pay"* line under the intake note field.
- Dropped the "Hang tight — your application is being reviewed" status card.
- The reviewer-facing note alert on `/membership-ops/review` stays — that is the point of the field.

## Reviewer notes

- `PENDING_BG_REVIEW` stays in the enum: `PERSON_BG` still enters there, and any **legacy household row** currently parked at that status still clears to `PENDING_PAYMENT` through the untouched `clearBackgroundCheck` path. Those rows now render the generic "Application in progress" card instead of the removed held-for-review card. No data migration is included — say so if you want one.
- The volunteer-designation allowlist (#874) is unaffected: it is applied at the `PENDING_PAYMENT` transition, which every application now reaches directly. A household that wanted volunteer dues *and* is not on the allowlist will now pay full dues before a reviewer can mark them — that is the behavioural consequence of removing the gate.

## Docs

- `docs/rules/membership.md` — the hold rule rewritten; the "applicants are told about the hold" rule deleted.
- `docs/DOCUMENTATION_STANDARD.md` — the hold was the worked example in two places (the §3.8 format sample and the checkable-sentence example); both swapped for the background-check freshness carve-out, preserving the lesson about stating exemptions.
- `docs/backlog/CUJS.md` (A1 steps 2 and 6) and `docs/backlog/INDEX.md` (M11 triage entry) corrected.
- `docs/in-design/MERGE_BG_CARRYOVER.md` — STALE banner added. Its predicate discussion, `holdForNote` bypass risk, and board decision 0 are moot; the `householdBgIsFresh` half stands. Re-deriving that design is separate work.
- `docs/in-design/DOCUMENTATION_MIGRATION.md` — its evidence example marked as since-removed.

## Testing

- `npx tsc --noEmit` clean.
- Unit suite (`test:ci`) — 2499 tests, 264 suites, all pass.
- Integration, serial, scoped to `src/app/__tests__/(membership|personBg|navTodo)` — 195 tests, 19 suites, all pass against a throwaway Postgres.
- Flow tests **not run** — Docker was unavailable in this session, so the standalone flow harness could not start. `flow-tests/membership-intake-notes.flow.test.ts` asserts the note reaches the reviewer queue; that queue includes `PENDING_PAYMENT` rows with consent recorded, which is covered by the passing `membershipReviewAPI` integration suite, but CI is the real check.

### DIFF
--- a/checkin-app/docs/generated/lifecycle/membership.md
+++ b/checkin-app/docs/generated/lifecycle/membership.md
@@ -21,7 +21,7 @@ stateDiagram-v2
     BLOCKED --> ACTIVE: overrideBlocked approve
     BLOCKED --> ARCHIVED: archiveApplication
     BLOCKED --> PENDING_BG_CLEARANCE: overrideBlocked reset
-    BLOCKED --> PENDING_BG_REVIEW: overrideBlocked reset
+    BLOCKED --> PENDING_BG_REVIEW: overrideBlocked reset · PERSON_BG
     BLOCKED --> PENDING_EXTERNAL_ACTION: overrideBlocked reset · RENEWAL · legacy
     BLOCKED --> PENDING_PAYMENT: overrideBlocked reset
     INTAKE --> ARCHIVED: archiveApplication
@@ -32,10 +32,9 @@ stateDiagram-v2
     PENDING_BG_REVIEW --> ACTIVE: clearBackgroundCheck
     PENDING_BG_REVIEW --> ARCHIVED: archiveApplication
     PENDING_BG_REVIEW --> BLOCKED: attest REJECT
-    PENDING_BG_REVIEW --> PENDING_PAYMENT: clearBackgroundCheck
+    PENDING_BG_REVIEW --> PENDING_PAYMENT: clearBackgroundCheck · legacy
     PENDING_EXTERNAL_ACTION --> ACTIVE: markContractSigned · PERSON_AGREEMENT
     PENDING_EXTERNAL_ACTION --> ARCHIVED: archiveApplication
-    PENDING_EXTERNAL_ACTION --> PENDING_BG_REVIEW: advanceExternalIfComplete
     PENDING_EXTERNAL_ACTION --> PENDING_PAYMENT: advanceExternalIfComplete
     PENDING_PAYMENT --> ACTIVE: activate
     PENDING_PAYMENT --> ACTIVE: grantRenewalPayment · RENEWAL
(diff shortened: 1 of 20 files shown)


jee7s:
The gate removal itself is agreed. This is about the other half of the stated behaviour — *"the note is still surfaced to the reviewers"* — which does not hold for one population.

**For a household with a still-valid background check, the note now reaches nobody.**

Dropping the note from the shortcut condition means `bgFresh` alone stamps `bgClearedAt` (`intake.ts:396`, `renewal.ts:198`). The reviewer queue is `awaitingBgReview.where`, which requires `bgClearedAt: null` — so the row never enters `eligibleReviewProcessIds` and no reviewer ever sees the note. Not "later than before"; never. `intake.ts:411` says as much: *"Fresh-check shortcut ⇒ clearBackgroundCheck never runs this cycle."*

The exposure lands where the field matters most. Renewals are returning households, so a valid check is the normal case, and a returning volunteer-only household writing "please mark us volunteer" is the motivating example in `membership-intake-notes.flow.test.ts`'s own header. The PR body says such a household "will now pay full dues before a reviewer can mark them," which reads as *eventually* — for these rows there is no eventually. The intake form also still asks "anything else we should know?", now without the pause warning, so the applicant has no signal that nobody is obliged to read it.

**Smallest fix that keeps both halves:** leave `external.ts` as-is (the advance always opens payment), but keep the note disqualifying the `bgClearedAt` shortcut in `intake.ts` and `renewal.ts`. The row then lands at `PENDING_PAYMENT` with consent recorded and `bgClearedAt` null, which `PARALLEL_BG_STATES` already puts in the reviewer queue. Payment opens immediately; the note still reaches a human. Roughly two lines back out.

**The cost is real and worth weighing:** a household with a perfectly valid check that writes a note gets asked to consent to a redundant background check, purely to keep the row reviewable. That may be exactly why the coupling was removed. If that trade is unacceptable, the note needs a delivery surface independent of the background-check queue — board-visible "households with unread intake notes" — which is separate work. In that case the PR body should say the note is not guaranteed to be read, rather than that it still reaches reviewers.

Two things I checked that are **fine**: the unrun flow test should still pass (`PARALLEL_BG_STATES` includes `PENDING_PAYMENT`, and that path leaves `bgClearedAt` null with `bgConsentAt` set), and legacy `PENDING_BG_REVIEW` rows still clear through the untouched `clearBackgroundCheck`, correctly tagged `legacy: true`.

---
_Posted by Claude._

thpr:
I'm at "do we even need a note anymore"?  I'm trying to figure out what value that has in our system vs just asking folks to email board@

jee7s (approved):

PR #1500, opened 2026-08-05
fix(people): refuse the over-25 declaration when a DOB is on file

thpr:
## What

`PUT /api/membership-ops/participants/[id]` set `Person.isDeclaredAdult` unconditionally:

```ts
if (body.isDeclaredAdult !== undefined) updateData.isDeclaredAdult = Boolean(body.isDeclaredAdult);
```

The route never reads or writes `dateOfBirth`, so an ops/board editor could flag a person holding a real sub-26 DOB. That breaks the app-wide invariant — recorded in `docs/rules/people-households.md` ("Identity and age") and enforced at write time by `normalizeAdultDob` (`src/lib/person/adultDob.ts`) — that a person has a date of birth **or** an over-25 declaration, never both.

This route was the only unguarded flag-writer. Every sibling already guards on `!dob`: `api/household/member`, `api/household`, `api/profile`, `lib/membership/intake.ts` (×3). The nightly cron and the `20260724120000_purge_adult_dob` migration only ever write the pair consistently.

## Fix

Select `dateOfBirth` on the `prior` fetch the route already makes, and drop the flag when one is present:

```ts
if (updateData.isDeclaredAdult === true && priorDob) {
    updateData.isDeclaredAdult = false;
}
```

**Coerce, not 400.** This matches the rule the guard cites: `normalizeAdultDob` treats a sub-26 DOB as superseding the flag and sets it `false`. Refusing the write would diverge from that, and — more importantly — would make an already-inconsistent person **unsaveable**: the edit modal resubmits the whole form (`membership-ops/participants/page.tsx:181`) with `isDeclaredAdult` seeded from the row, so editing that person's name, phone, or background-check date would submit the flag and take the 400 back. Coercing repairs the pair on the next edit instead.

Such rows are reachable today: `membership-ops/participants/import` writes `dateOfBirth` at lines 188, 223 and 230 without touching the flag and without `normalizeAdultDob` — the bulk-import case that helper's own docstring names as slipping past. Nothing reconciles them afterwards; the nightly cron only handles the 26+ direction.

Nothing is silently lost by the coercion: the edit modal hides the switch entirely when a DOB is on file, so no editor can deliberately submit the flag for such a person. The only submitters are the form echoing a stale row back, and direct API calls.

**The DOB is never cleared to make the flag stick.** A real DOB is the more authoritative value; destroying it to satisfy a checkbox would be the worse outcome.

`dateOfBirth` is destructured out before `oldData`, so the audit row's shape is unchanged — no new sensitive field lands in `AuditLog`.

## Why it matters

Readers that take the flag *before* the DOB, and so mis-classify a flagged youth as an adult:

| Reader | Effect |
|---|---|
| `src/lib/getFullAttendance.ts:80` | counts them as a **supervising adult** in the two-deep safety calc |
| `src/lib/membership/personBgCheck.ts:66` | treats them as ≥18 — a **minor pulled into background-check scope** |
| `src/app/api/people/search/route.ts:49` | `isDeclaredAdult: true` OR-leg — a flagged youth matches an adult search |
| `src/app/my-household/page.tsx:45` | renders an "Adult" badge |

`isKnownAdult` (`src/lib/programAge.ts:15`) reads the flag only on the no-DOB path, so the **program-enrolment gate was never affected** — a real DOB already wins there.

## No UI change needed

`src/app/membership-ops/participants/page.tsx:442` already hides the switch when `editingParticipant?.dateOfBirth` is set (landed with #1008), and `/api/people/search` ships `dateOfBirth` to board/sysadmin — the only roles that see the Details button (`isStaff`). The server-side coercion backstops direct API calls and stale client rows.

## Verification

Three integration cases in `src/app/__tests__/adminParticipants.integration.test.ts`:

- **happy path** — no DOB, flag `true` → 200, flag set (so an over-broad guard can't pass unnoticed)
- **DOB present** — a youth already carrying a stale flag, renamed *and* resubmitting the flag → 200, name saved, flag healed to `false`, DOB intact. This is the exact interaction the edit modal produces.
- still allows clearing the flag on a DOB row

Confirmed the DOB case fails against `main`'s route (the flag lands as `true`).

- `npm run test:ci` — 2503 passed, 264 suites
- `npm run test:integration -- --testPathPatterns "auditLog|authzRoleRejection|adminParticipant"` — 145 passed
- `npx tsc --noEmit`, `npm run lint` — clean

The integration run used a throwaway Postgres provisioned with `prisma migrate deploy` (not `db push`, which drops the partial unique indexes).

## Existing data

Rows may already violate the invariant. Detection query (read-only) — **not run against any real database**:

```sql
SELECT id, name, "dateOfBirth"
FROM "Person"
WHERE "dateOfBirth" IS NOT NULL AND "isDeclaredAdult" = true
ORDER BY "dateOfBirth" DESC;
```

The coercion above heals such a row whenever someone next edits that person, but does not reach rows nobody touches. Neither the `purge_adult_dob` migration nor the nightly cron will find them either: both only fix the 26+ direction (null the DOB, set the flag), so a flagged sub-26 row is invisible to them. A one-off sweep setting `isDeclaredAdult = false` where a DOB exists — never nulling the DOB — is worth running; it is deliberately **not** in this PR.

Follow-up worth filing separately: `membership-ops/participants/import` should route its `dateOfBirth` writes through `normalizeAdultDob`, which is what stops these rows being created in the first place.

### DIFF
--- a/checkin-app/src/app/__tests__/adminParticipants.integration.test.ts
+++ b/checkin-app/src/app/__tests__/adminParticipants.integration.test.ts
@@ -293,34 +293,59 @@ describe('Admin Participants API Integration Tests', () => {
             // Cleanup is handled by afterEach
         });
 
-        it('rejects the over-25 declaration when the person has a date of birth on file', async () => {
+        it('sets the over-25 declaration on a person with no date of birth', async () => {
             (getServerSession as jest.Mock).mockResolvedValue({
                 user: { id: testAdminId, isSysadmin: true, isBoardMember: false }
             });
 
-            // A 12-year-old: a declared-adult flag on this row would make them match
-            // the adults filter in /api/people/search and count as a supervising
-            // adult in the check-in safety calc.
+            const adult = await prisma.person.create({
+                data: { email: 'edit-test-user-nodob@example.com', name: 'Adult No DOB', household: { create: { name: "Test HH" } } },
+            });
+
+            const req = new Request(`http://localhost:4000/api/membership-ops/participants/${adult.id}`, {
+                method: 'PUT',
+                body: JSON.stringify({ isDeclaredAdult: true })
+            });
+
+            const res = await PUT(req as unknown as Parameters<typeof PUT>[0], { params: Promise.resolve({ id: adult.id.toString() }) });
+            expect(res.status).toBe(200);
+
+            const dbCheck = await prisma.person.findUnique({ where: { id: adult.id } });
+            expect(dbCheck?.isDeclaredAdult).toBe(true);
+        });
+
+        it('drops the over-25 declaration when the person has a date of birth on file', async () => {
+            (getServerSession as jest.Mock).mockResolvedValue({
+                user: { id: testAdminId, isSysadmin: true, isBoardMember: false }
+            });
+
+            // A 12-year-old already carrying a stale flag, as a bulk import that writes
+            // a DOB without touching it leaves them. The edit form resubmits the whole
+            // record, so an unrelated name edit must still save AND heal the pair — a
+            // declared-adult flag here would make them match the adults filter in
+            // /api/people/search and count as a supervising adult in the safety calc.
             const youth = await prisma.person.create({
                 data: {
                     email: 'edit-test-user-dob@example.com',
                     name: 'Youth With DOB',
                     dateOfBirth: new Date(Date.UTC(2014, 0, 15)),
+                    isDeclaredAdult: true,
                     household: { create: { name: "Test HH" } },
                 },
             });
 
             const req = new Request(`http://localhost:4000/api/membership-ops/participants/${youth.id}`, {
                 method: 'PUT',
-                body: JSON.stringify({ isDeclaredAdult: true })
+                body: JSON.stringify({ name: 'Youth Renamed', isDeclaredAdult: true })
             });
 
             const res = await PUT(req as unknown as Parameters<typeof PUT>[0], { params: Promise.resolve({ id: youth.id.toString() }) });
-            expect(res.status).toBe(400);
+            expect(res.status).toBe(200);
 
             const dbCheck = await prisma.person.findUnique({ where: { id: youth.id } });
+            expect(dbCheck?.name).toBe('Youth Renamed');
             expect(dbCheck?.isDeclaredAdult).toBe(false);
-            // The DOB is the authoritative value — refusing the flag must not clear it.
+            // The DOB is the authoritative value — dropping the flag must not clear it.
             expect(dbCheck?.dateOfBirth).toEqual(new Date(Date.UTC(2014, 0, 15)));
         });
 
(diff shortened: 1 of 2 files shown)


jee7s:
The guard is in the right place — the edit UI already hides the over-25 switch when a DOB is on file, so enforcing that server-side is the correct backstop. One case it makes worse, though:

**A person with both a DOB and a stale `isDeclaredAdult: true` becomes uneditable.**

`membership-ops/participants/page.tsx:326` initializes the form with `isDeclaredAdult: !!p.isDeclaredAdult`, and line 181 submits the whole `editForm` on every save. So for such a row, editing the *name*, *phone*, or *background-check date* now submits `isDeclaredAdult: true` and gets a 400. Nothing about that person can be changed through the UI.

Those rows are reachable: `membership-ops/participants/import/route.ts` writes `dateOfBirth` at lines 188, 223 and 230 without `normalizeAdultDob` and without touching the flag — the "rows that slip past (e.g. bulk import)" case the helper's own docstring names. Nothing reconciles the pair afterwards; the nightly cron only handles 26+, and it produces a consistent row (`dateOfBirth = NULL, isDeclaredAdult = true`).

Worth noting the guard also diverges from the helper it cites. `normalizeAdultDob`'s documented rule is that a sub-26 DOB "supersedes the flag" — it *sets the flag false*. This route refuses the write instead.

Either fix is small:
- guard only on an actual transition (`updateData.isDeclaredAdult === true && prior.isDeclaredAdult !== true && priorDob`), or
- coerce to `false` when a DOB is present, matching `normalizeAdultDob` — this one also self-heals the inconsistent rows on next edit.

Minor: there's no test for the happy path (no DOB → set the flag true → 200). Both new cases are DOB-present, so an over-refusal wouldn't be caught.

---
_Posted by Claude._

thpr:
Confirmed all three claims and switched to coercion — `88458a05`.

The uneditable-row case is real: `page.tsx:181` sends `JSON.stringify(editForm)`, and `:326` seeds `isDeclaredAdult` from the row, so any edit to a both-values person resubmitted `true` and took the 400. And the source is reachable — `participants/import/route.ts` writes `dateOfBirth` at 188/223/230 with no `normalizeAdultDob` and no flag write.

Went with coercion rather than the transition-only guard, for the reason you gave: it matches `normalizeAdultDob`'s documented "sub-26 DOB supersedes the flag" and self-heals on next edit, where the transition guard would leave those rows inconsistent indefinitely.

The silent-drop argument I used for the 400 doesn't hold up — the modal hides the switch when a DOB is present, so no editor can deliberately submit the flag for such a person. The only submitters are the stale form echo and direct API calls, and coercion is right for both.

Added the happy-path case (no DOB → `true` → 200). The DOB case now renames the person alongside the flag, so it exercises the interaction that regressed rather than a bare flag write. 2503 unit + 145 integration green; the DOB case still fails against `main`'s route.

Noted in the PR body as a separate follow-up: the import route should route its `dateOfBirth` writes through `normalizeAdultDob`, which is what stops these rows being created at all.

jee7s (approved):

PR #1504, opened 2026-08-05
docs(workflow): forward-only issue Status ladder for org project 1

thpr:
Docs only. No app code, no schema, no project-field writes, and `AGENTS.md` is deliberately untouched — the replacement text ships as a working doc for review first.

## The problem

Project 1's Status field can't tell you what stage an issue is at. An issue whose **design PR** is out for review and one whose **implementation PR** is out for review both read "In review", so the board can't be read without opening each item. Getting from the first to the second requires moving the item *backwards* through Ready.

Live proof: **#1258** (design PR #1497 open, nothing implemented) and **#1396** (design merged, three implementation PRs open) are both "In review" today, four rungs apart.

Second version of the same failure: **#975** is half delivered — Releases 0+1 shipped in #1464, the `DROP COLUMN` is gated on Release 1 reaching a published *prod* release — and sits in "Ready", which reads as free to pick up.

And four of the eight Status options appear nowhere in `AGENTS.md`. Two of them hold 171 of the project's 209 items.

## What this proposes

A forward-only ladder — each state entered at most once, never returned to:

```
Ideation/Reporting → Planning → Designing → Design review → Ready → In progress → In review → Done
                                └──── design leg ────┘    hinge   └─── impl leg ───┘
```

`Ready` becomes the hinge meaning "the design question is answered, whichever way". Forward *skips* stay legal, so work needing no design still climbs five rungs and pays nothing for the design leg (#1230 traced in the doc).

Net option change: **+Designing, +Design review, −Backlog** — nine, up from eight.

## Two files, split by lifetime

Per the rule this PR also adds to the standard:

- **`docs/in-design/ISSUE_WORKFLOW_RULES.md`** — byte-identical to the `AGENTS.md` section it replaces. Nothing about where it goes or what it replaces; landing it is a copy.
- **`docs/in-design/ISSUE_WORKFLOW_MIGRATION.md`** — everything that expires at cutover: current counts, eight by-hand UI steps, where the five in-flight issues land, open questions. Step 8 is "apply the rules file, delete both".

`docs/DOCUMENTATION_STANDARD.md` gains **§4.2** requiring that split for any working doc that changes a standing rule, plus a clause on agent-guidance item 7. Writing them as one file states the permanent text twice — once as design, once as the block to paste — so a later edit lands in one copy; and it degrades the extract-at-merge step from a file move into a per-paragraph judgement call made months after the reasoning.

## Design calls a reviewer may want to argue with

**Blocked is a label, not a status.** Blocking toggles while the work stays put, so as a status it forces `Ready → Blocked → Ready` — a re-entry, which is the exact guarantee this is buying. It's also orthogonal to progress: an item can be blocked in Planning, Ready, or In review. Labels already toggle and already filter (`-label:blocked` gives the pick-up-able view of Ready). A dedicated "Blocked by" field carries the same information with more machinery.

**Partly delivered is a decomposition, not a status.** A "Partially shipped" state stops being monotonic at three releases, and it hides what matters — it doesn't say what's left or what gates it. #975's own design already names Release 0, 1, and 2 as distinct deploys with an ordering constraint; the decomposition was written, only the issues weren't.

**Backlog retired on evidence.** Zero items in 209. Whatever it was for, Ideation/Reporting and Planning are holding it. An undocumented option nobody uses is the defect this PR is fixing, so keeping it "just in case" reintroduces it. Counter-reading in open question 1: it may be unused *because* it was never documented.

**One backward move survives, declared rather than engineered away.** Rework — a PR closed unmerged, or implementation proving a merged design wrong — re-enters a state the issue has occupied, and is comment-gated on the issue. Removing it costs either re-filing the issue (orphaning its discussion, parent link, and every PR reference) or four more states for a rare case already obvious from the closed PR.

## Reviewer notes

- **Two open questions block specific steps, not the whole migration.** Q3 (does Designing need an assignee? if designs aren't claimed in advance, Designing and Design review collapse into one option) blocks steps 1–3 — decide it before adding options or you add the wrong ones. Q1 (is Ideation/Reporting one bucket or two?) blocks step 4 only.
- **Independent bug found while verifying:** no open PR closes **#1396**. `closingIssuesReferences` is empty on #1450 and #1451 and points to #1260 on #1470, so nothing will close the issue or drive Status → Done. True regardless of this PR.
- **Standard now contradicts itself on where these rules live.** §3.4 says "how we build" belongs in `docs/conventions.md`, and an issue workflow is how we build — but the section lives in `AGENTS.md` today and that's what the migration targets. Either `AGENTS.md` is an unlisted permanent home §2 should name, or these belong in `conventions.md` with `AGENTS.md` pointing at them. Not resolved here.
- **All eight migration steps are manual GitHub UI actions.** Single-select option edits aren't safely scriptable, and deleting an option that still has items on it silently clears their status.
- Verified against the live project and repo, not the issue bodies: option list, all 209 items' status distribution, and each of the five traced issues' state, open PRs, and closing links.


jee7s:
Opened **#1505** stacked on this branch (base `claude/amazing-tesla-df2a71`) rather than commenting, since the gaps are all text.

The ladder covers Status. Eight rules that govern this board live only as verbal guidance and appear nowhere in the repo — I checked, including in the `AGENTS.md` section this replaces, so none of them are regressions here. But once step 9 pastes the rules file over that section, it becomes the board's only written contract, which makes this the place they land.

The two I would press on:

- **Iteration.** The canonical view is filtered `iteration:@current`, so the ladder currently governs a field the default view does not filter on. **34 items sit at Planning or beyond with no Iteration** against 11 in Iteration 2 — `@current` shows 11 of the 44 actually in flight. #1505 adds a backfill step.
- **The Ideation/Reporting → Planning bar.** The transition table says "Triage accepts it", and **141 of 210 items are queued behind that gate**. #1505 gives it a criterion by request kind.

Also worth flagging directly here: your no-automation constraint is not written down anywhere, and this PR introduces a `blocked` label — syncing a label against a project field is the obvious next thing someone builds. #1505 records the rule.

Separately, and independent of #1505: I verified your #1396 finding. `closingIssuesReferences` is empty on #1450 and #1451 and points to #1260 on #1470, so nothing closes it or drives Status → Done. Real and actionable now.

One gap I would raise on this PR itself: the transition table has *Implementation PR merges → Done (automatic)*, but the built-in workflow fires when the **first** PR carrying a closing link merges. #1396 has three implementation PRs. The multi-release rule handles decomposition across releases, not several PRs inside one release — worth a rule that only the last PR carries the closing keyword.

---
_Posted by Claude._

jee7s (approved):

Issue #1456, opened 2026-08-02
Remove LIVE_PERSON: make tombstone exclusion structural, not per-query


thpr:
`LIVE_PERSON` (`checkin-app/src/lib/person/filters.ts`) is an **opt-out** convention: every Person query must remember `{ mergedIntoId: null }` or it silently reads a merged-away record. **113 usages across 40 files**, plus a **31-entry allowlist** on the drift guard. Forgetting produces no error — just a ghost in a list, or a silent write.

## Evidence this fails in practice

Three bugs from this one class:

- **#1448** — the merge's own lead-count guard counted tombstones, *falsely blocking* legitimate merges with a message naming members that don't exist.
- **#1450** — tombstones leaked into the "needs a lead" household surfaces.
- **`review.ts:305`** — the background-check blanket stamp (being fixed under #1260 / PR #1454).

`livePersonDriftGuard.test.ts` exists to catch exactly this and missed all three: it matches read verbs on the `person` delegate plus nested pulls keyed on `person:`, so it sees neither `tx.person.updateMany` nor a Person-typed relation named `householdMembers`. A chip is widening it now — that's mitigation, not a fix. The guard is a regex over source text; it will always have blind spots.

## Why the tombstone exists at all (and why it needn't)

The merge already **repoints** visits and 8 join tables to the survivor. Rows stay on the tombstone in exactly one situation: a **unique-constraint collision** — the same human has a row on both records for the same program / event / tool / process. `merge/route.ts:257` states the coupling outright: *"leave the colliding row on the tombstone (both survive; §3's LIVE_PERSON filter excludes the tombstone's from every count/roster)."*

So the tombstone is a parking space for collision residue, plus a redirect target for badge scans. Neither requires it to be a `Person`.

## Approach

1. **Merge refuses on ambiguous collisions** rather than parking them — "unenroll one of these from Program X first". Cleanup surfaces already exist (`DELETE /api/programs/[id]/participants`, `/api/events/[id]/rsvp`, `/api/shop/certifications`, `/api/attendance/manual/[id]`). #1451 is the same pattern already landing. Bare join rows with no payload auto-dedupe.
2. No new residue ⇒ **resolve the existing stock** (one-time triage; the widened guard's allowlist scopes it).
[...]

PR #1518, opened 2026-08-06
docs(security): target security architecture — strategy + sequencing

thpr:
## What

Two new documents under `docs/security/`:

- **`TARGET_SECURITY_ARCHITECTURE.md`** — the strategy: a from-first-principles design for data visibility, evaluated against the shipped handler system and #1134 rather than anchored to either. Threat model, model-space survey (per-view DTOs, Zanzibar-style ReBAC, in-process policy engines, Postgres RLS), and the verdict: one declarative policy (field tiers + relationship row scopes, per route) that **compiles into the query** — generated `where`, per-caller-view `select` — with the stripper retained as an independent second evaluator whose disagreement is a tripwire.
- **`TARGET_ARCHITECTURE_SEQUENCING.md`** — build order and migration mechanics only (one hop from `withAuth` to the facade, all 118 routes), kept separate so the strategy stays free of plan detail.

## Key decisions in the design

- Ceiling `rows:` semantics (industry practice), exactness scaled to top-tier routes; leak-critical business rules graduate to scopes — two membership scopes (`org_members` strict, `dues_settled` program-only)
- First-match view selection + subsumption validator distinguishing scope-qualified vs unconditional tokens; conjunction entries
- 404 for invisible rows, byte-identical to absent ids; minimum cell size on aggregates
- Declared projections as the single declassification point, reviewed per input field
- Relationship-edge writes treated as policy authoring: derived edge set, audited + reversible, rendered to the board
- `asSystem()` as a registered CODEOWNERS-gated inventory; facade access logging of highest-tier reads
- Per-request role freshness declared as an invariant (verified: `updateAge` is inert under the jwt strategy — next-auth 4.24.14)
- Outbound governed by a recipient-scope rule; governance report pivots by data subject/tier with per-sentence walker attestation
- Build over adopt, bounded by a constrained include grammar — backed by surveys of all 142 checkin routes and the 8 Inventory apps; portfolio-wide the declarations standardize, enforcement engines are per-app compile targets (RLS always generated, never hand-written)

## Relation to other work

- Incorporates the threat review (`TARGET_ARCH_THREAT_REVIEW.md`, on `claude/security-architecture-threats-541f86`), including its verified role-freshness correction; the two branches cross-reviewed each other
- #1134 — the design absorbs Steps 1–2 and re-sequences them; this doc is design-only, no closing keyword
- Both docs conform to `docs/WRITING_STANDARD.md`

## Open issues (in-doc)

1. RLS scope — decided jointly with the portfolio (`/Volumes/Untitled/Inventory` apps)
2. Second-party confirmation on highest-stakes relationship edges — board call


jee7s:
## Review: 15 findings, ranked most severe first

Method note: 10 finder angles, then every candidate independently verified against merge-base `a00331fa` (the PR's actual base — several early candidates died because they were checked against a stale local branch), then a gap sweep, then an adversarial challenge pass before posting. ~17 candidates were refuted in verification and are not listed. All line numbers refer to the files as added by this PR.

---

**1. `TARGET_SECURITY_ARCHITECTURE.md:1034` — The coverage mandate contradicts the standing DECISION RECORD blocking the withAuth collapse.**
"Every `withAuth` route... migrates once, directly to the facade form" — but `docs/security/auth-consistency-analysis.md:205-211` marks Step 6 "(BLOCKED on the `derive` hook)": `handler()`'s strip path "silently drops computed/derived response fields", and routes returning computed envelopes (`isMember`, `isPresent`, fallback names) "must not be force-migrated". Neither new doc mentions `derive` or computed envelopes. An implementer following the mandate silently loses those fields on every computed route.

**2. `TARGET_SECURITY_ARCHITECTURE.md:187` — Invariant I3 outlaws the thrice-recorded rows-private/count-public decision.**
`pentest_findings_2026-04-21.md` deliberately keeps `_count.{participants,volunteers}` for non-privileged callers so anonymous capacity display works; `auth-consistency-analysis.md` records the opposite position as REJECTED ("it can't say 'rows private, count public'"); `routeAuthDrift` pins the same. Under I3 an anonymous caller's I2-filtered set is empty, so every public program renders 0-capacity — re-breaking the #575 fix. I3 needs a declared carve-out.

**3. `TARGET_SECURITY_ARCHITECTURE.md:535` — "The facade auto-includes the `needs` fields in the select" defeats Prisma's native masking and reintroduces subtraction-after-fetch at the declassification point.**
Verified empirically on Prisma 7.8.0: `needs`-declared inputs absent from the caller's select are fetched internally and stripped by Prisma itself. The doc's mechanism explicitly selects raw `dateOfBirth`, so it sits in the response object with only the JS stripper between it and the wire — exactly the pre-stripper exposure the design forbids at lines 204-208. Fix: select only the projection field; let Prisma add and strip the inputs.

**4. `TARGET_SECURITY_ARCHITECTURE.md:605` — The reviewer-anonymity deferral ("a declared-projection problem, not a tier problem") is unsolvable in the doc's own grammar.**
Projections only add or declassify; the stripper strips only inputs "the view doesn't grant" (line 537). `reviewerId` is public-tier and both views on `membership/reviews` grant `public` (registry.ts:275-280: "Widening that select is a policy decision"), so the step-5 generated select ships it and no projection can withhold it. Identical shape to the one that earned `decisionNote` the `deliberative` tier — but with no re-tier, no deadline, no open issue.

**5. `TARGET_SECURITY_ARCHITECTURE.md:630` — The kiosk claim contradicts `presence-identity-tiering.md`, and `/api/attendance` GET falls outside both the four-surface taxonomy and the coverage mandate.**
That doc recommends the kiosk view hold `all_current_visitors:personal` — a row-scoped grant, the opposite of "no row vocabulary" — and names `/api/attendance` as "the one that would break if migrated naively". The route hand-rolls `verifyKioskSignature` + `getOptionalSessionUser`, so it is neither `withAuth` nor `withKiosk`. Neither new doc names it; following them as written either blanks the presence board or leaves the widest presence surface unmigrated forever.

**6. `TARGET_SECURITY_ARCHITECTURE.md:391` — "Merge freely when the tokens are scope-qualified" silently downgrades routes from fail-closed to alarm-only.**
Lines 282-291: single-scope views fail closed on unexpected strip; mixed-scope views only alarm, with the class "derived from its declaration". Merging a second scope token — the validator's recommended resolution — changes exactly that derivation, and nothing (diff, validator output, report) surfaces the enforcement-class transition. `trusted-adults/operational` (registry.ts:336) is a shipped instance of the pattern.

**7. `TARGET_SECURITY_ARCHITECTURE.md:743` — The session-claim rejection rests on two false premises.**
(a) Cost: the jwt callback already includes `household:{include:{orgMembership:true}}` on every request — byte-identical to `ACTIVE_ORG_MEMBER_INCLUDE` — so an `org_members` claim costs zero extra queries and is exactly as fresh as the role flags the same section declares invariant. (b) P0-C is miscited: it was a denied-household leak in a route that never consulted state (no cache, no lapse window), and its fix put denied-state INTO the session (`token.denied`). Only the predicate-collapse argument (#1397) survives; the paragraph should stand on it alone. (The cost argument does hold for `dues_settled`, whose predicate needs `processes` the callback doesn't load.)

**8. `TARGET_SECURITY_ARCHITECTURE.md:476` — The identical-404 convention + byte-identity walker assertion contradict the pentest-recorded deliberate split on `programs/[id]`.**
`pentest_findings_2026-04-21.md`, verbatim: unauthenticated → 404, "(authenticated non-members still get `403`)" — shipped code matches. The walker assertion (lines 490-492) fails on day one against this route. Either supersede the pentest decision explicitly or define how a recorded deliberate 403 coexists with the convention.

**9. `TARGET_SECURITY_ARCHITECTURE.md:189` — I5 cites "#1456's done-when list" as its prover, but that list is four deletion checkboxes and the issue explicitly refuses both couplings.**
#1456 removes `LIVE_PERSON` entirely (tombstone → `PersonMerge` archive); its out-of-scope section says #1134 row visibility is "deliberately not merged into this" and `archivedAt` is "not folded in here". "One injection site, two ledgers" (line 237) asserts precisely the merge #1456 declines. As written, I5 has no prover.

**10. `TARGET_SECURITY_ARCHITECTURE.md:1064` — The RLS acceptance criterion is never reconciled with system readers, and any app-set GUC system context is available to the compromised process T4 posits.**
The candidate tables (Person, EmergencyContact, TrustedAdult, Visit) are exactly what the nine `withCron` jobs and future `asSystem` paths touch; under "no-context = zero rows" they silently no-op. The escape — an app-set GUC or privileged role — is settable by the subverted app layer, materially weakening "the only layer that survives T4" (line 107). Open issue 1 needs an explicit system-context design before the criterion is adoptable.

**11. `TARGET_SECURITY_ARCHITECTURE.md:693` — Derived-completeness fails its own motivating example, and the fixture-gating proposal deadlocks against the shipped isolation CI.**
Enumerating scope×model pairs from `SCOPE_BINDINGS` exempts the `OPT_OUT_PENDING_ROUTE` models — including `VolunteerDesignation`, so "zero volunteer designations... passes vacuously" stays vacuous under the proposed check. And `seedBaseline` lives in `src/lib/dev/seed-helpers.ts`, which is in neither CODEOWNERS nor `security-boundary-isolation.yml`'s companion list: a scope PR carrying its discriminating rows fails isolation, while splitting leaves the mandatory completeness check red in between.

**12. `TARGET_SECURITY_ARCHITECTURE.md:503` — The cell-size trigger excludes the passage's own motivating example and is near-vacuous under I3.**
Lines 501-502 state the motivating caller is "fully authorized for every row in the aggregate", so "an aggregate over rows the caller cannot individually see" never fires on the scholarship-status dashboard it was invented for — and under I3 that trigger class is empty by construction. The trigger must key on grouping shape (cells below threshold), not row visibility.

**13. `TARGET_SECURITY_ARCHITECTURE.md:617` — The outbound sentence misstates the control state, and no sequencing step owns the self-described largest uncontrolled disclosure channel.**
`defineOutbound` surfaces already exist (registry.ts:447/453) but `outboundCall(` has zero production call sites at `a00331fa` — real mail goes through `lib/email.ts` `sendEmail`, called directly. So both axes are unenforced and the real work is adoption, not "one concept". Sequencing steps 1-8 never mention outbound; step 6's exclusion names only the session wrappers — "not deferred as a unit" defers it indefinitely with no owner.

**14. `TARGET_ARCHITECTURE_SEQUENCING.md:21` — The step-3 report generator cannot satisfy the design's own claims, and step 2's declaration changes escape the ratchet.**
The report needs a data-subject pivot no current data provides, per-sentence walker attestation (fail-on-unasserted) against a walker that is brand-new in the same step, and `asSystem()`/wrapper inventories — `asSystem` occurs nowhere in the repo, so that section cannot be generated and the report fails its own honesty gate on day one. Meanwhile the `deliberative`-tier PR, whose marquee sentence is the report's showcase, lands at step 2 before any report exists to diff (design line 668: "Every PR that changes a declaration carries the report diff").

**15. `TARGET_SECURITY_ARCHITECTURE.md:163` — The route census is wrong, and the batch-lane question rests on a CI constraint that does not exist.**
Corrected numbers at `a00331fa`: **116** route files call `withAuth` (the doc's 118 is a string-mention count including 2 comment-only files; 5 callers use `withAuth<T>(` and evade naive greps); **7** files are dual-wrapper (GET on `handler()`, write verb on `withAuth`) and are counted in both the "14" and the "118"; the distinct session surface is 123 files, and the 7 dual files migrate per-verb, not "once". Sequencing line 49's "118 single-route migrations under the two-PR boundary-isolation rule" misreads `security-boundary-isolation.yml`: `is_companion` has no cardinality check and every `src/security/*` file companions every other, so one boundary PR can already batch arbitrarily many registry entries — the open process question is answered by the shipped CI.

---

### Process note (not a ranked finding)

Both files are `Status: PROPOSED` designs parked in `docs/security/`, but `docs/DOCUMENTATION_STANDARD.md` (in force on main since #1567) reserves `docs/in-design/` for exactly this class ("Deleted at merge"; §6.8: never cite in-design as ground truth), and the migration doc freezes `docs/security/`. One-line fix: move both files to `docs/in-design/`. (The standard was still PROPOSED when this PR was opened — it flipped to in-force the next day — so this is a landing-target correction, not an authoring error.)

Also confirmed but cut by the 15-item cap: the incident citations at lines 32/884 are wrong (#127/#122 are error-response leaks, inherited from #1134's text; #957 is top-level field echoes on writes — a counterexample to the relation-traversal sentence citing it); the OR-block widening mechanism at lines 864-868 is empirically false in Prisma (only key collision widens; wrap-never-spread stays right, and the ZenStack-v3 attribution is invented); the "reads as" table cited at line 640 lacks both `their_program_households` and `keyholders` (stale on the two later-added scopes).

🤖 Posted by Claude Code on behalf of @jee7s

jee7s (approved):

Issue #1346, opened 2026-07-25
Date/time canonical-layer remediation (beyond the #1149 program-date slice)


thpr:
**Backlog item PL17** — surfaced by the design review of PR #1338 / issue #1149 (`checkin-app/docs/designs/1149_DATE_TIME_TZ_DESIGN.md`). Folded from `backlog/sources/design_pr_open_items.md`.

The #1149 work grew from a program-date off-by-one into a whole-class remediation. Backlog **P19 covers only the program-date display bug (Finding 2)**; the rest is uncaptured:
- **F1** — DOB written at 3 incompatible conventions (midnight/noon UTC) across writers; exact-match import dedup silently misses → **duplicate Person** (ties M18 person-merge).
- **F7** — `calculateAge`/`isYouth` read local fields off a UTC-midnight DOB → age off-by-one on the birthday; **can reject an 18-yr-old at a minAge:18 gate** (safety-adjacent).
- **F3/F6** — memberSince + lastBackgroundCheck display off-by-one (F6 compliance-facing).
- **F8/F10** — visit-window + facility-trends bucketing mix UTC parse with local calendar fields.
- Single display-tz source (org tz → client); `APP_TIMEZONE` hardcode = second-region gap (deferred until a 2nd region).

Blocked on a **DECISION**: calendar-date storage model (Model A DateTime-by-convention / Model B `@db.Date` / hybrid). P19 becomes a slice of this.

thpr:
**Backlog reconciliation (2026-08-01)** — PL17 design is **merged (#1338)**. Scope beyond the P19 program-date slice: DOB written at 3 conventions (dedup-miss → duplicate Person), an **age-gate birthday flip that can reject an 18-yr-old** (safety-adjacent), memberSince/lastBackgroundCheck display bugs, visit-window/trends bucketing. Blocked on the calendar-date-storage decision. Readiness IN-DESIGN. Source: INDEX PL17.

jee7s:
HOLD — the design is merged but the decision it exists to tee up is still unmade, and the issue now overstates what is left.

[...]

PR #1521, opened 2026-08-06
refactor(dates): calendar dates become @db.Date

thpr:
Answers the storage question Finding 12 of the date/time design left open, and retires that design doc into the durable register.

## The decision

The design offered three options: keep `DateTime` and hold the calendar-date contract by convention (Model A), migrate to Postgres `date` (Model B), or apply B only where it buys a functional win (hybrid). Decision: **Model B, in full.**

The reason is not tidiness. The point is that the schema enforces the classification, so a future `new Date(str)` + `formatDate` cannot reintroduce the off-by-one class the earlier findings were made of. "Consistency-only, no functional win" argues about which migration is most urgent, not about which classification is true, so no column is left behind. This is the same argument the repo already makes in `docs/conventions.md`: an invariant everyone must remember is not an invariant.

## What changed

Six columns, one migration each so a partial rollback stays possible:

| Column | Migration |
|---|---|
| `BoardSettings.orgMembershipYearBoundary` | `20260806130000` |
| `Program.startAt` | `20260806130100` |
| `Program.endAt` | `20260806130200` |
| `OrgMembership.memberSince` | `20260806130300` |
| `Person.lastBackgroundCheck` | `20260806130400` |
| `Person.dateOfBirth` | `20260806130500` |

`Event.startAt`/`endAt` are deliberately excluded. They are genuine instants, not calendar dates, and stay `DateTime`.

## The cast, and the one that would have corrupted the data

Every migration is:

```sql
ALTER TABLE "X" ALTER COLUMN "col" TYPE date USING ("col"::date);
```

and **not** `USING (col AT TIME ZONE 'UTC')::date`. These columns are `timestamp(3)` *without* time zone (Prisma's default; there is no `@db.Timestamptz` in the schema), so `AT TIME ZONE` reinterprets the naive value as a `timestamptz` instant and the `::date` then resolves it in the **connection's** `TimeZone`. On any session west of UTC that silently decrements every row by a day.

Verified on a Postgres session set to `America/Chicago`:

| stored | `col::date` | `(col AT TIME ZONE 'UTC')::date` |
|---|---|---|
| `2026-08-15 00:00:00` | `2026-08-15` | `2026-08-14` |
| `2026-08-15 12:00:00` | `2026-08-15` | `2026-08-15` |

The midnight row is the common case, which is what makes the wrong form dangerous: it corrupts the majority of rows and leaves the minority looking fine.

The cast self-backfills, so there is no separate backfill step. Each row keeps its own calendar day.

## Migration ordering

`20260806120000_normalize_dob_utc_midnight` (#1496) landed on main while this was in progress. It rewrites pre-#1494 `dateOfBirth` rows to UTC midnight while the column is still `timestamp(3)`. These six were originally stamped a day earlier, so on rebase they sorted ahead of it.

They are renumbered to run **after** it. Two reasons: a database that has already applied the normalize would otherwise get six migrations inserted behind it, and once `dateOfBirth` is a `date` column the normalize's `WHERE "dateOfBirth" <> "dateOfBirth"::date` matches nothing, leaving a migration in the tree that never did anything. The order now reads as it happened: normalize the rows, then make the column a `date` so they cannot drift again.

`memberSince` is the one multi-statement migration and carries an explicit `BEGIN`/`COMMIT`, since `migrate deploy` does not wrap a file in a transaction. Postgres refuses to retype a column whose default has no implicit cast to the new type, so the default is dropped and re-set around the change.

## Deploy safety

Old code keeps serving traffic against the migrated schema for the whole drain window. Both directions are safe: a `date` column reads back as the UTC-midnight `Date` old code already handled, and a write of a timestamp is assignment-cast to `date` by Postgres.

Nothing destructive. No `migrate reset`, no `--accept-data-loss`, no drop-and-recreate. Every type change preserves its rows.

## The functional win

Import dedup (`where: { dateOfBirth: parsedDob }`, in `import/preview/route.ts` and `import/route.ts`) is an exact-match consumer. A DOB stored at noon used to miss a midnight probe and create a duplicate Person. Exercised through the Prisma client on a Chicago-timezone process, writing the same calendar day in four conventions:

```
midnight-UTC   | wrote 1990-01-01T00:00:00.000Z | db 1990-01-01 | read 1990-01-01T00:00:00.000Z
noon-UTC       | wrote 1990-01-01T12:00:00.000Z | db 1990-01-01 | read 1990-01-01T00:00:00.000Z
bare-date-str  | wrote 1990-01-01T00:00:00.000Z | db 1990-01-01 | read 1990-01-01T00:00:00.000Z
late-UTC 23:59 | wrote 1990-01-01T23:59:00.000Z | db 1990-01-01 | read 1990-01-01T00:00:00.000Z

exact-match probe for 1990-01-01T00:00Z matched 4 of 4
```

Before the change that probe matched one of the four.

## Doc retirement

`DOCUMENTATION_STANDARD.md` gives a working doc one ending: extract whatever standing rules it established, then delete it. The date/time design has reached that point, so `checkin-app/docs/designs/1149_DATE_TIME_TZ_DESIGN.md` is gone and its content is distributed:

- **`docs/conventions.md`** gains "A day is not a moment". The design's spine was that a moment and a day are different kinds and the code kept conflating them, which is how we build rather than a domain rule.
- **`docs/rules/`** gains nothing, deliberately. Everything domain-shaped was already written: age measured at a program's start date (`programs.md`), absent date of birth reading as youth and absent data not being a zero (`principles.md`), the date-of-birth retention rule (`people-households.md`). Restating any of it would create a second copy to forget to update.
- **Backlog PL17** absorbs what is left.

References to the deleted file are repointed in the same commit rather than as a follow-up, per the standard's sweep rule. Those comments also described calendar dates as stored at UTC midnight "by convention"; they are `date` columns now, so the wording is corrected.

## A live bug found while doing this, not fixed here

The org timezone is an editable sysadmin setting and the server honours it: event creation converts the entered wall clock through it, and facility-trends buckets and labels against it. But the client instant formatters (`formatDate`, `formatTime`, `formatDateTime`, used across roughly 14 surfaces) render through a compiled-in `APP_TIMEZONE = 'America/Chicago'`.

Changing that setting therefore desyncs every displayed time from every stored one, silently. An event entered as 3pm reads back at 2pm. It needs no second region to bite; one settings edit is enough. The two agree today only because the setting's default equals the constant and nobody has changed it.

The retired design framed this as future readiness for a second region, which understated it. PL17 now records it as a live bug, and it is deliberately out of scope here: fixing it needs a decision on how the resolved zone reaches client components.

## Correction in the history

An earlier commit on this branch rewrote PL17 from the design doc's finding list, which had gone stale, and claimed several findings were still open. Checked against the tree, findings 3 through 11 have all shipped: the calendar-date displays read UTC-pinned, `calculateAge` reads UTC fields, the visit-window filters do UTC arithmetic, and facility-trends buckets against the org zone. Two commits walk that back. Reviewing the final state of `docs/backlog/INDEX.md` rather than the intermediate diffs will be less confusing.

## Testing

- Unit: 265 suites, 2515 tests, all passing.
- Integration: 130 suites, 1311 passing, 9 skipped, none failing.
- `tsc --noEmit` and `lint` clean. Worth saying plainly that **tsc proves nothing here**: the TypeScript type stays `Date` on both sides of this change, so a green typecheck cannot see the read-value shift. Integration against a real Postgres is the guard.
- `prisma migrate diff --from-config-datasource --exit-code` reports no drift.
- Applied to a populated database with `dateOfBirth` rows seeded at midnight, noon and `23:59:59.999`; all three come through both migrations as the same calendar day.

No test fixtures needed changing. The sweep across all three test roots found time-bearing literals on these columns, but unit tests mock Prisma so a column-type change cannot reach them, and the integration writes are either already UTC-midnight or truncation-invariant.

## Boundary

No file under `checkin-app/src/security/` is touched. The security generator emits field-to-tier only, so `@db.Date` leaves `classifications.ts` byte-identical; confirmed by regenerating. The boundary-isolation workflow does not apply.

Refs #1149. The remaining display-timezone work stays with #1346.


jee7s:
The six migrations are right where it matters — `col::date` and not `(col AT TIME ZONE 'UTC')::date`, with the reasoning in each file, and the explicit `BEGIN`/`COMMIT` on `memberSince` because `migrate deploy` doesn't wrap a file in a transaction. That last one is easy to miss and would leave a column mid-retype. The cast is also a no-op on `Person.dateOfBirth` because #1496 already normalised those rows, so the two PRs compose cleanly.

**One defect: the re-set default contradicts the convention this PR adds.**

`20260806130300_org_membership_member_since_date` ends with:

```sql
ALTER TABLE "OrgMembership" ALTER COLUMN "memberSince" SET DEFAULT CURRENT_TIMESTAMP;
```

on a column it just converted to `date`. `CURRENT_TIMESTAMP` is a `timestamptz`, so the implicit cast to `date` resolves in the **connection's** `TimeZone` — the same hazard the comment three lines above warns about, and exactly what the new rule in `docs/conventions.md` forbids:

> **A day is never taken from a moment.** Keeping the date part of "now" answers in whatever zone the process happens to be running in; a day comes from a day.

Concretely: with a session zone west of UTC, any row that falls back to the default between 00:00 and 06:00 UTC gets **yesterday** as its `memberSince`. It's masked if the deployment's session zone is UTC — but that is a configuration coincidence, which is the same argument this PR makes about the compiled-in timezone constant.

The body explains *why* the default has to be dropped and re-set (Postgres won't retype a column whose default has no implicit cast). It doesn't note that the semantics changed on the way through.

Either `SET DEFAULT (now() AT TIME ZONE 'UTC')::date`, or drop the default and let the writer supply the day — the latter matching the rule's "a day comes from a day" more literally.

Minor: the migration comments run ~10 lines each across six files. `AGENTS.md` asks for short comments with the rationale in the PR body, and the body already carries all of it.

---
_Posted by Claude._

thpr:
Both fixed in f0dbc7c4. The default was a real defect and the reasoning is exactly right, including that it was masked by a configuration coincidence.

Confirmed it before changing anything, on one server at one instant:

```
SET TimeZone='America/Chicago';  CURRENT_TIMESTAMP::date  ->  2026-08-05
SET TimeZone='UTC';              CURRENT_TIMESTAMP::date  ->  2026-08-06
```

Same expression, same moment, two different days. And the default is reachable on every path that matters: no writer supplies `memberSince`. All six creation sites (`membership-ops/participants`, `membership-ops/households`, `intake.ts`, the import upsert, seed helpers) omit it, so the database default decides the value every time a membership is created.

The default is now `(now() AT TIME ZONE 'UTC')::date`, which is zone-free. Verified identical at UTC-5, UTC and UTC+14.

One thing worth recording, because it constrains the fix. The default expression has to be declared in `schema.prisma` too, or the CI drift check fails: `@default(now())` renders as `CURRENT_TIMESTAMP`, so the migration and the schema disagree and `migrate diff --exit-code` returns 2. It now reads

```prisma
@default(dbgenerated("((now() AT TIME ZONE 'UTC'::text))::date"))
```

with the string matching Postgres's own normalised form of the expression — the `'UTC'::text` and the doubled parens are not decoration, and without them the diff still reports a change. `migrate diff --exit-code` now reports no difference against a database built from the migrations. `memberSince` also stays optional in the generated client, which those six writers depend on; `tsc` is clean.

I did not take the second option, dropping the default so the writer supplies the day, even though it matches "a day comes from a day" more literally. The column is `NOT NULL` and every writer omits it, so removing the default converts all six into runtime failures. Supplying the day at the writers is the better end state, but "which day" is the org-timezone question, and that is PL17 — the same item this PR flags for the compiled-in display constant. Doing it here would answer it in passing for one column.

Comments trimmed to the constraint that is still true and non-obvious, with the rationale left in the body. They were duplicating it.

Re-verified after the change: unit 265 suites / 2515 tests, integration 130 suites / 1311 passed 9 skipped, `tsc` and `lint` clean, no drift, and `classifications.ts` still byte-identical.

jee7s (approved):

PR #1522, opened 2026-08-06
fix(time): display instants in the org's configured timezone

thpr:
The organisation's timezone is an editable sysadmin setting (`AppSettings.timezone`, edited at `/settings/localization`), and the server already honours it — event creation converts an entered wall clock through it, facility trends buckets and labels against it. The client instant formatters did not: `formatDate` / `formatTime` / `formatDateTime` rendered through a compiled-in `America/Chicago`, across roughly fourteen surfaces.

So changing the setting desyncs every displayed time from every stored one, silently and with no error — an event entered as 3pm reads back at 2pm. It needs no second region to bite; one settings edit is enough. The two agree today only because the setting's default equals the constant.

## What changed

The root layout resolves the configured zone and installs it through a new `TimezoneProvider`; the formatters read it.

```tsx
// app/layout.tsx
const timezone = await resolveDisplayTimezone();
<TimezoneProvider value={timezone}> …whole app… </TimezoneProvider>
```

```ts
// lib/time.ts — was a const used directly by all three formatters
let displayTimezone = APP_TIMEZONE;
export function setDisplayTimezone(tz) { displayTimezone = tz || APP_TIMEZONE }
```

`APP_TIMEZONE` is demoted to what it should always have been: the seed default and the fallback when the settings row cannot be read.

**A provider rather than a server-component prop.** These are plain functions called from leaf client components all over the tree — mostly `"use client"` pages with no server parent to thread a prop from. A prop would mean converting every call site to a hook; the provider leaves all ~14 files untouched. It assigns during render, not in an effect, so the server-rendered first pass already formats in the right zone instead of flashing the fallback.

**The layout's read is cached for a minute** (`resolveDisplayTimezone` in `lib/appSettings.ts`), so a page render doesn't query per request, and it falls back to the last known zone — then the constant — when the settings row can't be read. An asleep database renders the app in the default zone rather than failing every page.

**Server-side callers** run in a separate module instance where no provider ever renders, so the two that format instants — the check-in receipt email and the visit-correction board alert — pass the resolved zone per call. Left alone, they'd have kept the compiled-in zone: the same bug wearing a different hat. `options.timeZone` already won over the default, so this is a three-line seam, not new machinery.

## Held deliberately

- `toDatetimeLocal` / `fromDatetimeLocal` stay browser-zone — the native `datetime-local` contract. Untouched, comment intact.
- `formatDateOnly` stays UTC-pinned. A calendar date has no zone; routing it through the org's would shift the stored day for every reader west of it. There's a test asserting it does not move under two different configured zones.

## Reviewer notes

- **Known trade:** nothing re-renders on assignment, so a timezone change takes effect on the next page load (and up to a minute later, given the cache). Real context providers would fix that at the cost of converting every call site to a hook. Say the word if that's wanted instead.
- **Not fixed, still open:** `calculateAge` with the default `asOf = now` still rolls a birthday at UTC midnight (7pm Chicago). The fix there is a caller passing the org-zone calendar day — a day comes from a day — not routing `calculateAge` through this provider. Its comment is corrected, the behaviour is not.
- This is the **single-display-timezone slice of PL17** (#1346), not the whole item — hence no closing keyword. The rest of that class (calendar-date storage, the age-gate rollover) is untouched and still open; the backlog row is narrowed accordingly rather than dropped.
- Zero migrations, zero schema changes.

## Testing

- Unit: 2526 pass / 267 suites
- Integration against a fresh Postgres: attendance-manual (incl. both concurrency suites), notifications, check-in scan, events, facility trends, localization settings — 39 pass / 9 suites
- tsc clean, eslint clean
- New tests: `TimezoneProvider.test.tsx` (renders an instant at `Asia/Tokyo`, asserts `9/1/2026 11:30` not Chicago's `8/31`), a layout test for the resolve→provider→child wiring, and `appSettings.test.ts` for resolve / cache-once / DB-down fallbacks. `pinTimezone` keeps the process zone off the org zone, since CI runs in UTC where this bug is invisible.
- Mutation-checked: reverting the formatters to `APP_TIMEZONE` fails two of them. Not a vacuous guard.


jee7s:
The change is complete, which I checked rather than assumed: I enumerated every consumer of `formatDate`/`formatTime`/`formatDateTime` at this head. All thirteen are `"use client"` components under the new root provider, and the only two server callers — `lib/notifications.ts` and `api/attendance/manual/[id]/route.ts` — both pass `{ timeZone }` explicitly. So no server component silently keeps the compiled-in zone. `formatDateOnly` stays UTC-pinned and `toDatetimeLocal` stays browser-zone, both correct.

**One caveat, and one thing it leaves open.**

`displayTimezone` is a module-level mutable global, and `TimezoneProvider` assigns it from a render body. That is a side effect in render, which React does not guarantee: Strict Mode double-invokes, and concurrent/streaming rendering may render, discard, or reorder. It works today because there is one organisation, one root provider, and a value that is constant per request — so the caveat is fragility rather than a live bug, and the component's comment argues the tradeoff honestly (avoiding a fallback flash). Worth naming so the next person doesn't discover the constraint by breaking it.

The more practical point: **nothing prevents the next server caller from regressing this.** A server component added later that calls `formatDate()` without `{ timeZone }` gets `APP_TIMEZONE` silently — no error, no failing test, and the rendered time is simply wrong by an offset. That is the failure mode `docs/conventions.md` already names:

> **An invariant everyone must remember is not an invariant.** A rule held by convention holds until the first person who has not read the convention writes a query.

Two cheap options: make the instant formatters require an explicit `timeZone` when no provider has installed one (throw or lint in dev), or add a test that asserts every non-`"use client"` caller passes it — the same shape as the existing `livePersonDriftGuard` scan, which exists for exactly this class of "someone will forget" problem.

Worth noting #1521 edits `docs/conventions.md` too, so if both land the guard has an obvious home next to the rule it enforces.

---
_Posted by Claude._

jee7s (approved):

PR #1524, opened 2026-08-06
security(isolation): audit the PR's own head; stop admitting unrelated deletions

jee7s:
Reworked. The previous revision overlapped #1526 on the boundary-set plumbing; that overlap is **gone** — #1526 owns it entirely. This PR is now two commits and nothing else.

## Rebased stack

The base moved, because one of these fixes cannot land on the old one.

```
main
 └── claude/1482-deconflicted    #1482 rebased onto main (was CONFLICTING)
      └── claude/1526-rebased    #1526's three commits, unchanged, dkay's authorship
           └── this PR           two commits
```

**Why the rebase is a prerequisite, not a tidy-up:** `HEAD_SHA` does not exist on `boundary-decommission-exception` — that branch predates #1501, which is where `HEAD_SHA` was introduced. Passing `--head "$HEAD_SHA"` there dies on the unset variable under `set -euo pipefail`. So the `--head` fix below is only expressible after the rebase. #1482 is CONFLICTING against main today regardless, so the rebase has to happen either way; `claude/1482-deconflicted` is it, both conflicts resolved keeping both sides.

If you would rather not take the rebase through my branches, the two commits here apply to whatever #1482 and #1526 land as — but the `--head` commit needs #1501's variable present.

### 1. The certifier audited a tree the PR does not propose

#1501 moved the workflow off the checked-out `HEAD` because on a `pull_request` run that ref is `refs/pull/N/merge` — the branch merged with the *current* tip of main. The certifier never got the same treatment: it was still calling `git diff ..HEAD` and `git show HEAD:`.

Two ways that goes wrong, both of them main's changes deciding this PR's verdict:

- main lands a deletion; the certifier sees it in `changed`, and a violation that should ship separately reads as admissible.
- main deletes a route file; the route-kill check sees it absent at head and certifies an endpoint this PR never touched.

`--head` is now **required**, and the workflow passes `HEAD_SHA`. Making it required rather than defaulted is the point: there is no argument list that reaches the wrong ref.

### 2. "It is a deletion" was not evidence of anything

The exception admitted *any* deleted file riding along with a boundary drop. But a file nothing imports can still be a security control — `middleware.ts` is the standing proof — and deleting one breaks no build and no test. The class was much wider than the thing it was written for.

It now admits only the route files of the registry entries this PR removed. Everything else ships separately, deleted or not.

### 3. The route-kill check trusted a path it derived itself

`endpointToRouteFile` maps `GET /api/x` → `src/app/api/x/route.ts` by construction, and absence at head was read as "the endpoint is dead". A route group, catch-all, rewrite or `route.tsx` makes that path wrong — and a path that never existed is absent at head too, so a live endpoint certifies. The base file must now resolve *and* export the verb before its disappearance means anything.

Fail-closed corner, now documented in the workflow header: removing an already-inert entry fails certification. Only reachable on a PR that also carries a violation, since a boundary-only PR never invokes the certifier.

### 4. Rename bypass (from review)

With rename detection on, `git diff --name-only` reports only the destination. So renaming `middleware.ts` never put its name in `CHANGED`, `is_boundary` never fired, and the step exited green — disabling the site-wide org-login gate as thoroughly as deleting it, with tsc and the build green too. `--no-renames` surfaces D+A so both names hit `is_boundary`.

Verified in a scratch repo: default reports `a/mw.ts`; `--no-renames` reports `a/middleware.ts a/mw.ts`.

Note the two file lists now disagree about renames deliberately. Detection needs the *source*; the certifier's `--name-status` diff keeps the destination only, which is correct for the deletion rule — a renamed file is not a deletion and must not be admitted as one. Comment updated to say that, since it previously claimed the opposite.

### Smaller

- `process.exitCode` instead of `process.exit()`, which can truncate buffered stdout on a pipe — how the workflow runs it.
- CODEOWNERS gains `/checkin-app/src/middleware.ts`, exactly the line thpr specified.
- The failure message now names the boundary file that tripped the rule. It printed only the *feature* files; a boundary file is a companion, so a developer touching `middleware.ts` plus one other file read "must ship it ALONE" over a list not containing `middleware.ts`. Guarded — a re-tier-only failure has no boundary files.
- Hoisted the repeated `REGISTRY`/`ROUTE_FILE` test constants.

## On the previous review

@thpr — findings 1, 2, 5, 6 and the constants are done above. Two are already closed on this branch rather than by me:

- **3 (the middleware test models an impossible config)** — the test on this base is dkay's `9d89c48b`, and it already passes `boundaryChanged: [BINDINGS, MIDDLEWARE]`, which is the fix you asked for. It is the production configuration now.
- **4 (the fallback in the description would reopen the finding)** — you were right that `is_boundary` and `is_companion` are a package. Moot here: that fallback is withdrawn with the whole F3 overlap, and both lines live in #1526 unmodified.

`/checkin-app/scripts/tsconfig.json` was already at the block's column on this base — the misalignment you saw was mine, on the branch that realigned the block.

## Verification

- `scripts/__tests__/boundary-decommission.test.ts` — **25/25 pass**. New pins: `--head` required; the base-resolves guard rejects a derived path that never served the verb; the tightened deletion class rejects an unrelated deletion and admits the removed entry's own route file.
- `--head` proven to move the audited tree: same argv, two different `--head` values, two different verdicts on the same file (`deleted, but not a route file…` vs `neither a migration nor a deletion…`).
- Workflow YAML parses; the run block passes `bash -n`.
- CI on this PR is **not meaningful** — it targets a non-`main` branch, and `ci.yml`/`security-boundary-isolation.yml` only trigger on PRs into `main` or `rel/**`. Zero checks will dispatch. Everything above was run locally.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -19,6 +19,10 @@
 # Authentication & session machinery.
 /checkin-app/src/lib/auth.ts                      @dkaygithub @jee7s @thpr
 /checkin-app/src/lib/auth-options.ts              @dkaygithub @jee7s @thpr
+# The site-wide org-login/access gate. security-boundary-isolation.yml's
+# is_boundary() already treats this as a boundary file; without the entry here
+# the isolation check and the review requirement disagree about what it is.
+/checkin-app/src/middleware.ts                    @dkaygithub @jee7s @thpr
 
 # Contract tests that enforce the policy in CI.
 /checkin-app/tests/security/                      @dkaygithub @jee7s @thpr
--- a/.github/workflows/security-boundary-isolation.yml
+++ b/.github/workflows/security-boundary-isolation.yml
@@ -170,6 +170,7 @@ jobs:
             if [ -z "$RETIER" ] && [ "${#BOUNDARY_FILES[@]}" -gt 0 ] &&
                node checkin-app/scripts/check-boundary-decommission.js \
                  --base "$BASE_SHA" \
+                 --head "$HEAD_SHA" \
                  --boundary "${BOUNDARY_FILES[@]}" \
                  -- "${VIOLATIONS[@]}"; then
               echo "Decommission exception: boundary diff is whole-entry removals of dropped"
--- a/.github/workflows/security-boundary-isolation.yml
+++ b/.github/workflows/security-boundary-isolation.yml
@@ -36,7 +36,14 @@ name: Security boundary isolation
 # to its migration (#1446). When check-boundary-decommission.js certifies
 # that shape — byte-strict, fail-closed: partial edits, additions, re-tiers,
 # reorders, and engine-file changes never qualify — the PR may also carry
-# prisma/migrations/** and file deletions, nothing else.
+# prisma/migrations/** and the route files of the removed entries, nothing else.
+#
+# Known fail-closed corner: the certifier requires each removed route entry's
+# derived file to export the verb AT BASE (otherwise absence at head proves
+# nothing). Removing an ALREADY-INERT entry therefore fails certification. That
+# is only visible on a PR that also carries a violation — a bindings drop plus
+# its migration, say — since a boundary-only PR never reaches the certifier.
+# Split the inert-entry removal into its own PR.
 on:
   pull_request:
     # rel/** stays in the trigger even though enforcement is main-only:
@@ -86,7 +93,16 @@ jobs:
           # merge ref (= base.sha and head.sha) local, so no fetch is needed;
           # if either SHA is unreachable the diff below fails loud, never a
           # silent unaudited pass.
-          if ! CHANGED_RAW=$(git diff --name-only "$BASE_SHA"..."$HEAD_SHA"); then
+          # --no-renames: with rename detection on, `git diff --name-only`
+          # reports ONLY the destination, so renaming a boundary file never puts
+          # its name in CHANGED, is_boundary never fires, and the step exits
+          # green. Next.js runs only src/middleware.ts, so a rename disables the
+          # site-wide gate exactly as thoroughly as deleting it — and nothing
+          # imports it, so tsc and the build stay green too. D+A surfaces both
+          # names. (The certifier's own --name-status diff deliberately keeps the
+          # destination only; detection needs the source, the deletion rule does
+          # not.)
+          if ! CHANGED_RAW=$(git diff --no-renames --name-only "$BASE_SHA"..."$HEAD_SHA"); then
             echo "::error::Could not diff $BASE_SHA...$HEAD_SHA — a SHA is unreachable. Failing rather than passing this PR unaudited."
             exit 1
           fi
@@ -185,6 +201,12 @@ jobs:
             echo "to a separate PR (registry-first for new routes — an unused"
             echo "defineRoute is inert):"
             printf '  %s\n' "${VIOLATIONS[@]}"
+            # Guarded: a re-tier-only failure has no boundary files, and an
+            # empty array still prints one blank bullet.
+            if [ "${#BOUNDARY_FILES[@]}" -gt 0 ]; then
+              echo "Triggered by these boundary files:"
+              printf '  %s\n' "${BOUNDARY_FILES[@]}"
+            fi
             exit 1
           fi
           echo "Boundary change is properly isolated (companions: tests/docs/schema only)."
--- a/checkin-app/scripts/__tests__/boundary-decommission.test.ts
+++ b/checkin-app/scripts/__tests__/boundary-decommission.test.ts
@@ -116,17 +116,19 @@ const bindingsWithoutFeePayment = dropLines(
 describe('parseArgs', () => {
     // The shell → argv → lib wire is now the sole carrier of the boundary set;
     // pin its slicing so a workflow invocation-line edit can't silently drift.
-    it('splits --base / --boundary / -- into the certifier inputs', () => {
-        expect(parseArgs(['--base', 'X', '--boundary', 'a', 'b', '--', 'v'])).toEqual({
+    it('splits --base / --head / --boundary / -- into the certifier inputs', () => {
+        expect(parseArgs(['--base', 'X', '--head', 'Y', '--boundary', 'a', 'b', '--', 'v'])).toEqual({
             baseSha: 'X',
+            headSha: 'Y',
             boundaryChanged: ['a', 'b'],
             violations: ['v'],
         });
     });
 
     it('yields an empty boundary set when --boundary is omitted', () => {
-        expect(parseArgs(['--base', 'X', '--', 'v'])).toEqual({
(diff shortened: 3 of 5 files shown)


thpr (commented):
Reviewed against the PR head. No `node_modules` in my worktree either, so jest did not run — I drove the lib on bare node for the edge cases and verified the git rename behaviour in a scratch repo instead.

## Overview

The shape of all three fixes is right. Each one moves an inference the certifier was making on its own into something the caller has to prove, and each new failure mode is closed. What follows is what I would want changed before merge.

## 1. `middleware.ts` is declared a boundary file but gets no CODEOWNERS entry — Medium/High

Every other path in `is_boundary` has a matching line in `.github/CODEOWNERS` — `src/security/`, the generator, both certifier files. `checkin-app/src/middleware.ts` does not, and there is no catch-all owner in that file.

Concretely: a PR that deletes `middleware.ts` **and nothing else** sets `BOUNDARY_TOUCHED=1`, produces zero violations (it is now its own companion), and exits 0 on `Boundary change is properly isolated`. No security owner is a required reviewer. Isolation without ownership is half a control — CODEOWNERS' own header calls these paths trust anchors that "cannot merge without review."

One line, in the block this PR already realigned:

```
/checkin-app/src/middleware.ts                      @dkaygithub @jee7s @thpr
```

## 2. Renaming `middleware.ts` evades the rule entirely — Medium

Verified in a scratch repo: `git diff --name-only` reports **only the destination** for a rename.

```
--name-only:    a/mw.ts
--name-status:  R100  a/middleware.ts  a/mw.ts
```

So the workflow's `CHANGED` array (`security-boundary-isolation.yml:68`) never contains `checkin-app/src/middleware.ts` for a rename. `is_boundary` never fires, `BOUNDARY_TOUCHED` stays 0, and the step exits on `No security-boundary changes in this PR`. Next.js executes only `src/middleware.ts`, so a rename disables the site-wide gate exactly as thoroughly as the deletion I demonstrated — and it is the same class: a control nothing imports, so tsc and the build stay green. `src/security/*` is not exposed this way because things import it.

Cheapest fix is `--no-renames` on line 68, so a rename surfaces as D+A and both names hit `is_boundary`. Worth noting that the two file lists then disagree about renames on purpose: `check-boundary-decommission.js:69` keeps the destination only and argues that is safe (it is, for the deletion rule), while the workflow's detection needs the *source*.

## 3. The new middleware test models a configuration production cannot produce — Low/Medium

`boundary-decommission.test.ts:328` passes `boundary: [BINDINGS]` and `violations: [MIGRATION, MIDDLEWARE]`. In the shipped workflow `middleware.ts` is both `is_boundary` and `is_companion`, so it is never a violation and always in the boundary set. The real rejection comes from the `engine/tooling boundary file — direction undecidable` branch, not the deletion branch this test asserts on.

The deletion rule is worth testing and the `fees.ts` case at line 306 already does it. Either retitle this one to what it actually covers, or set `boundary: [BINDINGS, MIDDLEWARE]` and assert the engine-file reason — that is the production path, and nothing exercises it today.

## 4. The fallback offered in the description would reopen the finding

> If that is too strict, the deletion fix alone closes the reported hole and the `is_boundary` line can be dropped.

Dropping only `is_boundary` leaves the `is_companion` line, which makes `middleware.ts` a non-violation the certifier never sees — the original hole, unguarded. The two lines are a package, and `is_companion` is *required*, not a convenience: without it a PR touching only `middleware.ts` has itself as a violation, the certifier rejects it as a non-declarative boundary file, and a correctly isolated boundary PR fails CI. Same reason lines 97-99 exist for the scripts.

For sizing the friction: `middleware.ts` has 3 commits in the last 12 months. The strictness is cheap — keep it.

## 5. The failure message never names the file that triggered isolation — Low, but this PR makes it bite

`printf '  %s\n' "${VIOLATIONS[@]}"` lists the *feature* files. `middleware.ts` is a companion, so it is never printed. A developer whose PR touches `middleware.ts` plus anything else sees "must ship it ALONE" over a list that does not contain `middleware.ts`. `RETIER` gets a dedicated `::error::` line; boundary files get none. Add one:

```bash
echo "Triggered by:"; printf '  %s\n' "${BOUNDARY_FILES[@]}"
```

## 6. Route-kill tightening rejects inert-entry removal when the PR also carries a migration — Low, fail-closed

The argument in the description — removing a genuinely inert entry is boundary-only, so the certifier never runs — holds only when the PR has no violations. A decommission that drops a scopeBindings model (with its migration) *and* removes an already-inert registry entry now fails on `mapping unverified`. Fail-closed and rare, so not a blocker, but a line in the header would save the next person the debug.

## Smaller

- `.github/CODEOWNERS:37` — `/checkin-app/scripts/tsconfig.json` was not realigned with the other six in the same block.
- The test file ends `});\n\n` — trailing blank line, likely a prettier / `no-multiple-empty-lines` hit.
- `const REGISTRY` is now declared in three separate test bodies (267, 292, 352). Hoist it next to `BINDINGS`/`SCHEMA`.
- `flagValues` breaks on any value starting with `--`. Not reachable from the workflow's call; fine as is.
- `exportsVerb` duplicating `VERB_EXPORT_RE` — agree with deferring. A build step to share it is a bigger change than this fix should carry.

## What is right

- `process.exitCode` over `process.exit()`, and the reason given is the real one.
- Passing `--boundary` in rather than re-deriving kills the whole drift class. Best of the three fixes.
- The fail-closed guards check out. I drove both edge cases on bare node: a missing `boundary` gives `no boundary file set supplied by the caller`, and an *empty* array — what an empty shell array expands to — gives `no whole-entry removals found in the boundary diff`. Both reject.
- Requiring the base file to export the verb is the right shape for the mapping fix: it makes the mapping prove itself instead of reading absence as death.
- Reverting the `stripStringsAndComments` bail was the right call, and the documented limit is honest. One suggestion — the comment asserts "a removal can only ever narrow," but what actually holds that line is the byte-strict `kept.join('\n') !== headSource` check in `diffSegmentations`: a mis-segmentation makes head fail to reconstruct rather than certify a widening. Citing that mechanism is stronger than the assertion alone.

## Verdict

1 and 2 are the ones to block on — both leave `middleware.ts`, the file this PR exists to protect, reachable by a path CI stays green on. 1 is one line, 2 is one flag. 3 and 5 are cheap. 4 is a description edit.

---
_Posted by Claude._

jee7s:
Superseded by #1590, which landed all four of these commits with their authorship intact, rebased onto `main`, alongside @dkaygithub's five from #1482/#1526 and two further fixes. The certifier work is on `main`; nothing here is outstanding.

Verified after the merge: running the shipped certifier against #1444's real head certifies it — `model FeePayment (dropped from schema in this PR)`, exit 0 — which is what this stack existed to make possible.

thpr:
Superseded by #1590, merged as `72d98303` — this PR was already closed when the stack came down, so noting it here for the trail. All three fixes from it (`--head` required, deletion latitude narrowed to `impliedDeletions`, the `--no-renames` rename bypass) are on `main` with authorship intact.

PR #1525, opened 2026-08-06
docs(membership): propose intake-note acknowledgement — read it, then proceed

jee7s:
Docs only — one new file in `docs/in-design/`. No code, no schema, no migration, no boundary files. **Ready for review; not blocked on anything.**

> **This description was rewritten from scratch.** The previous one argued the gated design against a pre-#1499 `main` and cited by line number. Both are gone. If you read the old version, discard it — the doc is the source, not this summary.

## The rule this proposes

**If a note is provided, a named person reads it, and the record says who and what.**

Reading is an acknowledgement, not a decision. One reader, not the two-of-N background-check review, and not coupled to it.

## What the design is now

**Surface it, do not gate.** The earlier revision blocked `ACTIVE` on an unacknowledged note. That was designed for a `main` where a note already held the application; #1499 removed the hold and `docs/rules/membership.md` now carries it as a standing rule. The diagnosis was never that something is stuck — it is that something is invisible. So: a board-facing list of applications carrying an unacknowledged note, oldest first, with the note shown and an "acknowledged" action. Nothing is held.

The price is stated in the doc rather than hidden: **an unread note stops nobody.** A household that asked for volunteer dues can be activated at full dues before anyone reads the request. That is the trade #1499 already made deliberately, and the remedies it relies on exist. The list is worked, or the note is not acted on.

## The three decisions settled in this revision

**1. The list is board-facing, and the access rule is relaxed to its policy limit.**

The access rule (`membership.md`, Procedure → Renewal) reads *"An intake note is readable by the household's leads and the reviewers, not by its other members"* and carries `[Decision — Policy: Membership Policy, Art. VI §VI.2; Records Policy, Art. IV]`. Board members are not inside it as written.

That rule turned out to be **stricter than the policies it cites** — both already permit board members to read intake notes. The owner established this by reading the policy documents themselves. That matters, because `DOCUMENTATION_STANDARD.md` §3.6 names this exact move as the risky one: *"Sometimes stricter than the policy requires, in which case say so: the risk is someone relaxing it while believing they are aligning to policy."* The doc therefore records the basis rather than presenting the relaxation as a formality — what was read, by whom, that it was read rather than inferred, and that the policies live on Drive (§3.10) rather than in this tree, so **no future reader can re-verify the claim from the codebase**; they have to go to the policies.

**This PR does not edit `docs/rules/membership.md`.** The surface does not exist yet, and §3.7's reason applies — *"a reader who finds their answer stops reading."* A register describing a screen nobody has built is worse than one that is merely narrow. The doc instead fixes the **exact replacement text for both rules**, tier tag and policy anchors intact, to land with the implementation PR.

**2. Snapshot-on-process stays — and the boundary work is now priced.**

`intakeNoteSnapshot` would be the **first `pii` field on `OrgMembershipProcess`**, whose ceiling today is `personal`, held by `shopifyInvoiceUrl` alone. **Four registered routes return the model in their view bag**, so raising the ceiling means re-clearing all four plus regenerating `classifications.ts` — a boundary change that ships in its own PR per AGENTS.md. Cost missed this entirely; it is now an item.

The doc also records that **the argument keeping the note narrow today does not transfer.** `registry.ts` argues `Household.intakeNotes` is out of reach because *"Household binds no scope beyond their_households"* — an argument about **`Household`'s** bindings. On `OrgMembershipProcess` the conclusion holds for a different reason: that model is **unbound**, annotated in `scopeBindings.ts` as *"board/admin today; a household-facing status route is plausible."* Whoever adds that route will be binding a pii-carrying model with the reasoning attached to a different model. The doc says so, so they find it.

And a correction is recorded rather than quietly dropped: on the boundary axis, the separate `IntakeNoteAcknowledgement` model **would have been cheaper** — no scopes bound, no existing view bag, all four routes untouched. The snapshot is still the right shape (no re-open semantics, no edit-materiality comparison, no join table), but it buys that with boundary surface.

**3. The byte-identical skip is adopted, with its cost stated.**

Skip the snapshot when the text is byte-identical to the most recent acknowledged snapshot on that membership.

The defect it addresses is real and is the design's own enforcement mechanism failing: nothing re-prompts the family at renewal, `intakeNotes` lives on `Household` and the renewal path never touches it, so without the skip **a household that once wrote a note produces a fresh unacknowledged snapshot every renewal, indefinitely** — a queue that refills annually with re-reads of identical sentences, which is how a list stops being worked.

The doc says plainly that the skip **displaces rather than eliminates** the edit-materiality question: exact equality across cycles instead of "was this edit material" against a live string. Genuinely easier. Not gone.

## The snapshot now has a write site

The previous revision added three columns, said the snapshot was taken "at submit", and **nothing wrote it**. That is a different write from the acknowledgement (the board's action on `noteAckById`/`noteAckAt`). Both sites are specified, and both were verified against `origin/main`:

- **`beginRenewal` — as the review described.** #1499 removed the note read; the membership lookup selects `householdId` alone, so renewals need a **new read as well as a new write**, and the write must land **inside** the conditional `updateMany` on `PENDING_RENEWAL` whose comment exists so a double-submit audits once. Outside it, a double-submit snapshots twice.
- **`submitIntake` — *not* as described, and the doc corrects it.** `submitIntake` does **not** write `intakeNotes`; that write is in `saveIntake`, earlier, so the snapshot cannot ride it. The site is still cheap, for better reasons: `loadUserWithHousehold` already pulls the household with a full `include` (**no new read**), and `submitIntake` already issues the `orgMembershipProcess.update` that advances `INTAKE` (**no new write** — one more field on `data`).

## Citations

The doc carries **no line numbers**, and rules are cited by section heading plus quoted text. The review of this PR supplied the live counterexample: it quoted the access rule at `membership.md:302`, #1579 moved it to 308, and by the time this revision was written it had moved again to **314**. Three numbers, one unchanged rule. A heading and a quotation survive that; a reader who cannot find the quoted sentence learns the rule changed instead of silently reading the wrong line.

## Open questions

One remains, and it needs a human, not a reviewer:

**What happens if the list is not worked.** With no gate nobody is stranded, but the note is still unread. The byte-identical skip means what remains on the list is genuinely unread, which makes the question sharper rather than softer. A staleness notice to the board is the cheap answer. This is a commitment question, not a technical one.

## Placement

`docs/in-design/` per `DOCUMENTATION_STANDARD.md` §4, no `Status:` header per §6.7. At merge, the standing rules extract into `docs/rules/membership.md` and this working doc is deleted (§6.9).

**No issue exists for this** — #900 and #907 are merged PRs, not issues. The `active-is-bg-cleared` defect the doc uses as evidence is filed as #1603, so it survives this doc's deletion.

---
_Posted by Claude._


thpr (commented):
Reviewed against `main` at d4b38c5b. Every code citation in the doc checks out — `external.ts:112`, `intake.ts:392`, `lifecycle.ts:152-157`, `household/settings/route.ts:28`, `schema.prisma:507`, `conventions.md:41`, and all five `ACTIVE` entry edges with their guard sites verbatim from `TRANSITIONS`.

**The diagnosis is right, and it is the valuable half of this doc.** The #1499 hole is real: with the shortcut restored, `bgClearedAt` is stamped at intake, `awaitingBgReview.where` requires `bgClearedAt: null`, and the row never enters the queue. #1499's body says the note "is still surfaced to the reviewers" — true only for rows that reach the queue, which the fresh-check case never does. That is exactly the case the field exists for.

What follows is what needs fixing before this is buildable.

## Design holes

**No holding state is named.** "Does not become `ACTIVE`" needs a destination. The precedent is already in the table this doc quotes: `activate()` forks `PENDING_PAYMENT` → `ACTIVE` or `PENDING_BG_CLEARANCE` (`lifecycle.ts:305-306`). The doc never says where a note-blocked process sits instead. That answer pulls in new `TRANSITIONS` edges, a regenerated `docs/generated/lifecycle/membership.md`, and the `isLegalTransition`/reachability oracles — none of which appear in Cost.

**Cost contradicts the decided design.** It says "one predicate change in `advanceExternalIfComplete`", which is left over from the before-payment variant the doc rejects two sections earlier. Under the decision, `advanceExternalIfComplete`'s `holdForNote` gets *deleted* (as #1499 does); the predicate lands at `activate()` / `clearBackgroundCheck` / `overrideBlocked`.

**The mechanism the doc asks for already exists.** It proposes "one predicate every edge calls, with a drift test". `lifecycle.ts:270-277` already has `INVARIANTS` + `validate`, including `active-is-bg-cleared` — the precise sibling of the proposed rule — and `classify()` is an exhaustive switch that fails to compile when a status is added. The ask is one `INVARIANTS` entry plus the existing `TRANSITIONS` oracle, not a new pattern. Worth naming them, and worth stating the caveat: `INVARIANTS` is a detection oracle, not an enforcement gate, so it covers the drift test but not the gate itself.

**`PERSON_BG` has no membership.** `orgMembershipId` is nullable, and `review.ts:256/304` is explicitly the PERSON_BG edge. A predicate that reads the household note through `orgMembership` has to handle null.

**In-flight rows are unaddressed.** Processes parked at `PENDING_BG_REVIEW` today by `holdForNote` — #1499 ships no migration and says so explicitly. This design inherits them without saying where they land.

## Schema and security

- **The sketch will not compile.** Prisma needs both sides: `intakeNoteAcknowledgements IntakeNoteAcknowledgement[]` on `OrgMembershipProcess`, and the `"IntakeNoteReads"` back-relation on `Person`.
- **No `/// @sensitivity:` annotations.** The model it mirrors annotates every field (`schema.prisma:508-521`). The prose says `noteRead` is pii; per `docs/security/SECURITY-POLICY.md` the schema comment *is* the mechanism, so the sketch should carry it.
- **"One board list" is the expensive line in Cost.** `schema.prisma:285-292` carries an explicit GUARD: no household-scoped pii token may be granted to a lead/keyholder-facing view returning Household rows — reviewers read `intakeNotes` through `everyones:pii` on `GET /api/membership/reviews`. A new board-facing surface needs its own grant, and per AGENTS.md the registry entry ships in its own PR *before* the route. Cost undercounts by a PR.
- **A second copy of PII with no retention story.** `noteRead` outlives the household clearing `intakeNotes` to null, and outlives a household merge. Storing text rather than a hash is the right call — a hash cannot answer "what did they read" — but the doc should say what deletes it.
- **No index or uniqueness**, and "a later edit re-opens the obligation" implies several rows per process. Latest-wins is never stated.

## Documentation standard

The PR body cites §4 and §6.7, so the standard was read — but §4.1 is unmet in three ways:

1. **"The problem statement must be readable by someone who has never opened the code. No symbol names, no file paths, no line numbers."** The Problem section is built from `advanceExternalIfComplete`, `external.ts:112`, `intake.ts:392`, `lifecycle.ts:152-157`, `bgClearedAt` and `awaitingBgReview`. The plain-English version is already in the doc's own closing line — "One charges a family a background check to have a sentence read; the other loses the sentence." Lead with that and move the citations into Design.
2. **Objective and Executive summary are missing.** §4.1 orders Problem → Objective → Executive summary → design.
3. **"Provenance goes last."** Line 2 is provenance (#900/#907/#1499), and "Relationship to #1499" sits mid-design. Both belong in an appendix.

## A simpler shape worth arguing against

Snapshot the note onto the process at submit, and put `noteAckById` / `noteAckAt` on `OrgMembershipProcess`. No new model, no join table, three columns.

That dissolves three things this doc spends its length on: the "editable afterwards" problem (a snapshot cannot be edited), open question 2 (material-vs-typo edits — there are no re-open semantics to define), and open question 4 (per-cycle falls out, because each process carries its own snapshot). The audit answer gets stronger, not weaker: the note as submitted, rather than as of whenever someone happened to read it.

The price is that a household editing the note after submit does not reach the reader. That is arguably correct — the note is part of that application — but it is an argument the doc should make rather than an option it skips.

## Smaller points

- `renewal.ts:192` is the `hasNote` binding; the shortcut disqualification is `clearNow = bgFresh && !hasNote` at `renewal.ts:203`.
- Open question 1 (nobody reads it) and open question 4 (per-cycle re-reading) compound: every household that ever wrote a note stalls at `ACTIVE` on every renewal, waiting for someone to re-read the same sentence. Separately they read milder than they are together.
- Clearing the note to null self-unblocks the gate. Probably fine — they wrote it, they can withdraw it — but it should be stated rather than left implicit in "a live note".

## Verdict

Accurate, well-cited, and the critique of both `main` and #1499 holds up line by line. Not mergeable as it stands: resolve the Cost/design contradiction, name the holding state, point "one predicate" at the existing `INVARIANTS`, and fix the §4.1 opening.

---
_Posted by Claude._

thpr (commented):
Re-reviewed against the rewrite. Everything from the first pass is addressed, and the reversal from "gate before `ACTIVE`" to "surface, do not gate" is the right call — the diagnosis was that something is invisible, not that something is stuck, and the gated version was answering a question the merge of #1499 had already withdrawn.

Verified the new claims against `main`: the six `ACTIVE` edges, `markContractSigned` reaching `ACTIVE` on the signature alone, `scanLifecycleViolations` running without a kind filter, the `docs/rules/membership.md` quote, the three `intakeNotes` surfaces, and both shortcut disqualifications gone from `submitIntake` and `beginRenewal`. All accurate.

Three things below. Citations here carry no line numbers, for the reason §"A note on citations" gives — I quoted `membership.md:302` at this PR earlier in the week, #1579 landed an edit above it, and the rule is now at 308. The section is right and I was the counterexample.

---

## 1. Open question 3 is a Policy escalation, not a documentation edit

The access rule reads:

> - An intake note is readable by the household's leads and the reviewers, not by
>   its other members.  [Decision — *Policy: Membership Policy, Art. VI §VI.2; Records Policy, Art. IV*]

The tier tag is the load-bearing part. `DOCUMENTATION_STANDARD.md` §6.1: a **Policy**-tier rule is not negotiable in a PR at all — stop and escalate.

Open question 1 (board members only, or background-check reviewers too?) decides this rule's fate, and the two answers are not symmetric. Reviewers are already inside the rule, so a reviewer-only list needs no change to it. Board members are not, so a board-facing list amends a rule anchored to two board policies by article and section.

The doc currently says "the rule changes and `docs/rules/membership.md` changes with it." It cannot change that way. On the board-reader branch this is a board decision first and a doc edit second, and the design is blocked on it rather than merely informed by it — which is a stronger claim than "answered before the registry change" and worth stating as such, since it is the one item here that cannot be resolved by anyone reading this PR.

Worth checking the tag rather than the file, incidentally: the neighbouring staleness rule a few lines up is a plain `[Decision]` with one policy anchor. Same file, different negotiability.

## 2. The snapshot has no write site, and "per-cycle for free" is free only on the read

The model section adds three columns and describes the snapshot as taken "at submit", but nothing writes it. That is a **different write from the acknowledgement** — Cost's "one mutation to record the acknowledgement" is the board's action on `noteAckById`/`noteAckAt`. The snapshot is a system write at process advance, and it needs a site per kind:

- **`submitIntake`** is cheap: it already writes `intakeNotes` on the household in the same call, so the snapshot rides along in that transaction.
- **`beginRenewal`** is not. #1499 removed the note read — the membership lookup now selects `householdId` alone, where it previously pulled `household: { select: { intakeNotes: true } }` to compute `hasNote`. So renewals need a new read as well as a new write, and the write has to land inside the existing conditional `updateMany` on `PENDING_RENEWAL`, whose comment exists precisely so a double-submit audits once. Attached outside it, a double-submit snapshots twice.

The larger point is the second half of the claim. Nothing re-prompts the family at renewal — `beginRenewal` is a signature flow, and `intakeNotes` lives on `Household`, editable from household settings at any time and never cleared by anything. So the sentence snapshotted onto a renewal process is whatever has been sitting on that row, possibly for years.

A household that once wrote *"treat us as a volunteer household"* therefore produces a fresh unacknowledged snapshot **every renewal, indefinitely**, with nothing marking it as the same text a board member has already acknowledged three times. Dropping the gate makes that non-blocking, not harmless: it is a queue that refills annually with re-reads of identical sentences, which is how a list stops being worked — and the list being worked is the entire enforcement mechanism this design has.

The cheap fix lives at the write site: skip the snapshot when the text is byte-identical to the most recent acknowledged snapshot on that membership. Worth saying plainly that this brings back a smaller version of the question the snapshot is credited with eliminating. It is genuinely easier — exact equality across cycles, rather than "was this edit material" against a live string — but the doc reads as though the question is gone, and it is displaced rather than gone.

## 3. A pii column on `OrgMembershipProcess` moves the boundary cost, and Cost prices only half of it

`OrgMembershipProcess` carries **no `pii` field today**. Its ceiling is `personal`, held by `shopifyInvoiceUrl` alone; everything else is `internal` or `public`. `intakeNoteSnapshot` raises that ceiling for the first time.

Four registered routes return the model in their view bag. Raising the ceiling means re-clearing all four against the new tier plus regenerating `classifications.ts` — a boundary change, which per AGENTS.md ships in its own PR. Cost budgets the new list's grant and not this.

More importantly, **the reasoning that keeps the note narrow today does not transfer**. `registry.ts` argues explicitly that `Household.intakeNotes` is out of reach for `programLeadMentor` and `programCoreVolunteer` because Household binds no scope beyond `their_households`, so a `their_program_households` token resolves to nothing on a Household row. That argument is about the `Household` model's bindings. On `OrgMembershipProcess` it has to be re-made from that model's bindings, which are none — `scopeBindings.ts` lists it among the unbound models, annotated *"board/admin today; a household-facing status route is plausible."*

So the conclusion still holds, for a different reason than the one written down. What changes is the future: the bindings file already anticipates the household-facing route, and whoever adds it will be binding a model that now carries pii, with the reasoning that kept the note narrow living in a comment attached to a different model. Worth a sentence in the doc so that person finds it.

**A correction I owe this PR.** I recommended snapshot-on-process over the separate `IntakeNoteAcknowledgement` model, and on this axis the separate model was cheaper: it would have bound no scopes, appeared in no existing view bag, and left all four registered routes untouched. I still think the snapshot is the right shape — it removes the re-open semantics, the edit-materiality comparison, and the join table — but it buys that simplicity with boundary surface, and neither the doc nor my first review priced it.

---

The `active-is-bg-cleared` defect the doc uses as evidence is now filed as #1603, so it survives this working doc's deletion at merge.

_Posted by Claude._

thpr (approved):
Approving. All three findings from the last pass are closed, and the citations behind the Policy branch check out verbatim — §3.6's "sometimes **stricter** than the policy requires… the risk is someone relaxing it while believing they are aligning to policy", §3.7's "a reader who finds their answer stops reading", §3.5 on hints, §6.13 on mechanism. Not editing `membership.md` in this PR is the right call, and fixing the replacement text here so the implementation PR transcribes rather than re-derives is better than either alternative.

The correction on `submitIntake` is accepted and mine was the error: the `intakeNotes` write is in `saveIntake`, not `submitIntake`, and the reasoning the doc replaced it with is stronger — `loadUserWithHousehold` already pulls the household with a full `include`, and `submitIntake` already issues the process update, so the snapshot is one more field on an existing `data`. The note that the advance is an unconditional `update` by id rather than a status-conditional `updateMany`, and that the double-submit race there is pre-existing, is worth having.

## One gap, non-blocking

The four-route table under "The boundary cost of a `pii` column here" gives each route's band and stops there. Its own header says all four "are re-cleared against the new tier", but the result of re-clearing is never stated — the reader gets the inputs and has to finish the analysis themselves.

The result is favourable and worth a sentence, because the next person otherwise redoes it: every audience across the four routes is inside the amended access rule. `POST /api/membership-audit/person-agreement` is sysadmin/board; `GET /api/membership-ops/applications` and `GET /api/finance-ops/membership-payment-plans` are both `anyRole: ['isSysadmin', 'isBoardMember']`; `GET /api/membership/reviews` is background-check reviewer plus board. So raising the ceiling exposes the snapshot to nobody new, and on the first route it is stripped outright for want of a pii band.

That turns the boundary PR from a question into transcription, which is the same favour the fixed rule text does for the implementation PR.

## A caveat for the owner rather than the author

The Policy branch rests on one claim nothing in this repository can check: that `Membership Policy, Art. VI §VI.2` and `Records Policy, Art. IV` permit board members to read intake notes. The doc handles that correctly — it records who read the policies, states that the conclusion was not inferred from the rule's wording, and says under §3.10 that re-verification means opening the documents on Drive rather than any local copy.

Flagging it not as a defect but because it is load-bearing and unverifiable in review: every reviewer of the implementation PR will be taking it on the owner's word, and the doc is the only place that word is written down. That is the right place for it to be.

---
_Posted by Claude._

PR #1526, opened 2026-08-06
security(isolation): single source of truth for the boundary-file set

dkaygithub:
## What

Follow-on to review **finding 3** on #1482. "What is a boundary file" was defined **twice** — `is_boundary` in the workflow and `isBoundaryPath` in `boundary-decommission.js` — kept in sync by hand (and mirrored again by a hand-written test).

This makes the workflow's `is_boundary` the **single source of truth**. It passes the matched set to the certifier as `--boundary <files> -- <violations>`; `isBoundaryPath` is deleted. `certifyDecommission` now takes `boundaryChanged` as a parameter, the CLI slices it from `argv`, and tests supply it explicitly.

## Why

The duplication fails **open**, asymmetrically: if the shell gains a boundary entry the JS lacks, the certifier's `boundaryChanged` misses that file and the "engine files never qualify" rejection silently does not fire. Fixing #1482's finding 1 (adding `middleware.ts`) was itself a live example — it required editing *both* lists by hand.

## Base / stacking

Branched off `boundary-decommission-exception` (#1482), because `isBoundaryPath` and the certifier lib **do not exist on `main`** — they're introduced by #1482. **Retarget this PR's base to `main` once #1482 merges.**

## Deliberately out of scope

Finding 3 also suggested deduping `exportsVerb`, which copies `VERB_EXPORT_RE` from `check-route-coverage.ts`. Skipped here on purpose: those are two independent scripts across a runtime boundary (`check-route-coverage.ts` runs under ts-node; this lib runs on bare node with zero deps), and `exportsVerb` feeds the **route-kill** check — whose failure mode is fail-*closed* (an un-registered live route 500s via `handler.ts`), so a drift there has no exposure consequence. Coupling the two scripts for that regex costs more than it buys. Happy to add it if you'd rather.

## Note on finding 2

Declining finding 2 (route-kill "fails open" on a bad path guess) — separate reasoning posted on #1482. Short version: removing a route's registry entry fails **closed** (`handler.ts` returns 500 when `getRoute` misses), so a mis-certified route "kill" is at worst an availability regression, never an exposure widening — outside what the isolation check protects.

## Verification

Targeted jest run for `boundary-decommission` is 16/16 green. CLI smoke-tested on bare node with the new `--boundary` flag; workflow YAML re-parses.


thpr (commented):
Two findings. The refactor itself is right — deduping the *policy* predicate while leaving the mechanical git-diff derivation in place is the correct split, and I traced the fail-closed property: with `--boundary` absent, `boundaryChanged` is `[]`, so the non-declarative loop no-ops, both `includes(bindings/registry)` are false, and `certifyDecommission` falls through to `'no whole-entry removals found in the boundary diff'` → `ok: false`. Omitting the flag rejects rather than certifies. Good.

## 1. The seam that moved has no test coverage

Deleting the `isBoundaryPath` test is correct — the function is gone — but that test was the only mechanical check that the two definitions agreed, and nothing replaced it. All six updated cases hand `boundaryChanged` straight to `certifyDecommission`, which exercises the library, not the shell → argv → lib wire that is now the sole carrier of the boundary set. The duplication that used to be tested has become a wire that isn't.

`check-boundary-decommission.js` exports nothing and runs `process.exit(main(process.argv.slice(2)))` at module scope, so the argv slicing can't be unit-tested in-process. Pulling it into an exported `parseArgs(argv)` makes it reachable:

```js
expect(parseArgs(['--base', 'X', '--boundary', 'a', 'b', '--', 'v']))
    .toEqual({ baseSha: 'X', boundaryChanged: ['a', 'b'], violations: ['v'] });
```

Worth adding a `boundaryChanged: []` → `ok: false` case alongside it, to pin the fail-closed behaviour above so it can't regress silently.

## 2. Empty-array expansion is unreachable only by accident

`"${BOUNDARY_FILES[@]}"` expands under `set -euo pipefail`. Safe on ubuntu-latest's bash 5.x regardless, and unreachable in practice — `BOUNDARY_TOUCHED` can be set by `RETIER` alone with zero `is_boundary` matches, but the certifier call is guarded by `[ -z "$RETIER" ]`, so an empty array never reaches it.

That means the invariant "`BOUNDARY_FILES` is non-empty at the call site" rests entirely on the re-tier guard several lines away. One condition makes it local:

```bash
if [ -z "$RETIER" ] && [ "${#BOUNDARY_FILES[@]}" -gt 0 ] &&
   node checkin-app/scripts/check-boundary-decommission.js \
```

thpr:
One more, cosmetic — not worth blocking on.

`--boundary` slurps positionally up to the `--`, so flag order matters: `--boundary a b --base sha -- v` swallows `--base` and `sha` into `boundaryChanged`. It fails closed (both get flagged as non-declarative boundary files and the certification is refused), and the workflow is the only caller and never emits that order, so there's no live exposure. Noting it mainly because the failure is silent-ish for anyone smoke-testing the CLI by hand and reaching for a different argument order.

If you take the `parseArgs` extraction from the review above, rejecting a `--`-prefixed token inside the boundary slice is a one-liner there. Equally fine to leave as-is.

thpr:
One item, and it's worth catching here rather than after the rebase — because it will not show up *as* a conflict.

## `HEAD` in the CLI is the merge ref

[#1501](https://github.com/innovationtreehouse/checkin/pull/1501) landed on `main` after this stack branched, and changed what `HEAD` means in this workflow: on a `pull_request` event `actions/checkout` resolves `refs/pull/N/merge` — the branch merged with main's tip — so `HEAD` carries main's advance. That fix replaced every `...HEAD` in the YAML with an explicit `$HEAD_SHA`.

`check-boundary-decommission.js` still does the thing #1501 removed:

```js
const mergeBase = git(['merge-base', baseSha, 'HEAD']).trim();
const changed   = git(['diff', '--name-status', `${mergeBase}..HEAD`])
…
readHead: p => gitShow('HEAD', p),
```

The rebase conflict from #1501 is confined to the workflow YAML. This file has none, so whoever resolves the YAML hunk never looks at it, and the semantics #1501 just removed get re-introduced in the one code path that fix didn't cover.

### What it does when main drifts from `base.sha`

- `changed` picks up main's files, so a boundary file **main** touched lands in `boundaryChanged` and the certification is refused as non-declarative — naming a file this PR never touched.
- `readHead` returns the *merged* `scopeBindings.ts` / `registry.ts`, so main's independent additions read as `entry added: …` and the diff is refused.

Both fail closed, so this is a flaky-refusal bug rather than a hole: the decommission exception intermittently stops firing for correct PRs, with a message that points at the wrong change.

One narrow direction reversal, for completeness: `violations` comes from the workflow (correct SHAs post-#1501) while `deleted` is built from the CLI's `changed` (wrong ones). If `main` deletes a file this PR *modified*, the CLI sees `D <path>`, `deleted.has(v)` is true, and modified app code gets admitted as "a deletion". Contrived — it normally implies a merge conflict — but it is a genuine flip.

### Fix

Pass `--head "$HEAD_SHA"` alongside `--base`, use it for both the diff and `gitShow`, and drop the internal `merge-base` call — `"$BASE_SHA"..."$HEAD_SHA"` is merge-base→head already, and matching the workflow's own two SHAs exactly is the point. This PR is already editing both the argv parsing and the workflow's invocation line, so it is the cheap place to land it.

---

Separately: **withdrawing finding 2.** Traced it properly and the decline in the PR body reaches the right answer. Removing a whole `defineRoute` never widens on either path — a `handler()` route 500s when `getRoute` misses, and a `withAuth` route has nothing reading the spec at runtime — so a falsely certified route-kill unlocks exactly what a correctly certified one unlocks, and does not expand the blast radius beyond a legitimate decommission. That is finding 1's territory, and finding 1 is fixed.

Minor correction to the reasoning, not the conclusion: the "fails closed — `handler.ts` returns 500" mechanism only covers 12 of the 16 registered endpoints. `GET /api/programs/[id]`, `GET /api/events/[id]`, `GET /api/programs/[id]/eligible-participants`, and `PATCH|DELETE /api/attendance/manual/[id]` are still on `withAuth` and never call `getRoute`, so removal there is a silent no-op rather than a 500. Doesn't change the outcome — worth knowing only so that sentence doesn't become load-bearing for some later relaxation.

thpr:
Re-reviewed at `0614981a`. The three earlier items are handled — `parseArgs` extracted and exported behind a `require.main` guard, the `${#BOUNDARY_FILES[@]}` guard in place, stray-flag rejection added. The refactor itself is right: `isBoundaryPath` is gone and `boundaryChanged` arrives from the workflow.

Two things.

## 1. Security test files are classified as boundary files, so a decommission PR blocks itself

`is_boundary` matches `checkin-app/src/security/*` with only `generated/` carved out. Shell `case` globs match `/`, so all ten files under `checkin-app/src/security/__tests__/` land in `BOUNDARY_FILES` and reach the certifier, which rejects anything outside `DECLARATIVE_FILES`. Verified against this branch:

```
boundaryChanged: [scopeBindings.ts]                          => ok: true

boundaryChanged: [scopeBindings.ts,
                  src/security/__tests__/visit-household-lead-scope.test.ts]
  => ok: false
     "…visit-household-lead-scope.test.ts: non-declarative boundary file —
       direction undecidable, never decommission-eligible"
```

That path isn't contrived. `visit-household-lead-scope.test.ts` is a per-model scope oracle for `Visit` — exactly what a `Visit` decommission has to update or delete — and rule 7 in `DEPLOY_MIGRATION_ORDER_OF_OPERATIONS.md` names "plus test-oracle updates" as part of the single decommission PR. Deleting the oracle instead of editing it doesn't help: `is_boundary` classifies by path regardless of git status, so a `D` row poisons the set identically.

Net effect: the exception refuses to fire for the shape it was built for, and the failure message blames a test file.

This is pre-existing from #1482 — `isBoundaryPath` carried the same `startsWith('checkin-app/src/security/')` — but this is the PR that makes the workflow the single owner of that set, so it is the natural place to fix.

Cheapest fix is one arm above the existing one, mirroring how `generated/` is already carved out:

```bash
checkin-app/src/security/__tests__/*) return 1 ;;
```

`is_companion` already admits those paths, so they remain legal companions.

One side effect worth naming so it is a decision rather than a surprise: a PR touching *only* `src/security/__tests__/` would stop tripping isolation altogether. Given `is_boundary`'s own comment scopes it to "policy/runtime files — the security boundary itself", a test file sitting in that set reads as misclassification rather than intent — but it is a change to the base rule, not just to the exception.

## 2. Updating my earlier `HEAD` comment — this refactor fixes half of it

Moving `boundaryChanged` to the workflow means it and `violations` now come from `"$BASE_SHA"..."$HEAD_SHA"`, the correct pair post-#1501. The failure mode I described where main touching `handler.ts` poisons `boundaryChanged` and refuses a clean PR is gone. Incidental to the refactor, but real.

What remains is that `certifyDecommission` now draws its inputs from **two different diffs**: `boundaryChanged` and `violations` from the workflow's correct pair, but `changed` — and therefore the `deleted` set — plus every `readHead` from the CLI's `mergeBase..HEAD`, where `HEAD` is still the merge ref:

- `readHead(scopeBindings.ts)` returns the *merged* file, so an entry `main` added independently reads as `entry added: …` and the diff is refused.
- `deleted` is built from the merge ref while `violations` is built from the PR's real head — so if `main` deletes a file this PR modified, `deleted.has(v)` is true and modified app code is admitted as "a deletion".

Same fix as before, and smaller now that half is already done: pass `--head "$HEAD_SHA"`, use it for both the diff and `gitShow`, and drop the internal `merge-base` call.

thpr:
Superseded by #1590, merged as `72d98303`. That PR carries all eleven commits from this stack with their original authorship, rebased onto `main`, plus the two remaining review items: the `src/security/__tests__/` carve-out in `is_boundary` (security tests were classified as boundary files, so a decommission blocked itself on its own scope oracle) and rule 7 / AGENTS.md rewritten to describe the certifier that actually shipped. Verified on `main`.

Closing rather than merging because the stack was rooted on `boundary-decommission-exception` — merging a stacked PR into a branch nothing will ever merge reports `MERGED` while the code never reaches `main` (#1447).

Issue #1529, opened 2026-08-06
Audit Results Class B


jee7s:
Domain-register audit — **Class B: code violates a written rule.** 28 findings, these are the work. ✓ = verified in code directly.

Cite these as **B1**–**B28**. Narrative for each is in the comment below.

```
ID   SEV   FINDING                                    SITE
───  ────  ─────────────────────────────────────────  ──────────────────────────
B1   CRIT  ✓ Merged person keeps role-bearing sess.   auth-options.ts:374
B2   CRIT  ✓ Double-tap force-closes the building     scan-service.ts:96
B3   HIGH  ✓ Intake notes leak to all household mbrs  intake.ts:126
B4   HIGH  Lead can create an OPEN proxy visit        attendance/manual:75
B5   HIGH  isYouth defaults to adult, not youth       time.ts:167
B6   HIGH  Failed fetch renders "no violation"        attendance/current:81
B7   HIGH  COI missing on mark-contract / bg-consent  applications/external:38
B8   HIGH  ✓ Enrolment does not fail closed (#1435)   participants/route.ts:148
B9   HIGH  Kiosk minimization keyed on signature      attendance/route.ts:36
B10  MED   Roster mark skips check if programId null  events/[id]/attendance:52
B11  MED   Last valid contact removable via upsert    emergencyContacts/svc:201
B12  MED   24h cap not enforced on machine closers    scan-service.ts:161
B13  MED   Existence check precedes authz (404/403)   attendance/route.ts:181
B14  MED   Cron withdraws live pickup authority       trusted-adult/svc:551
B15  MED   Nightly DOB purge writes no audit row      cron/nightly:84
B16  MED   SYSTEM_ACTOR = 0, actors indistinguishble  11 sites
B17  MED   Board cannot read its own audit log        system-status/audit-log:16
B18  MED   Merge captures loser pre-image only        merge/route.ts:455
B19  MED   Unarchive walks AuditLog, throws if absent archive.ts:56
B20  MED   ARCHIVED -> * missing from TRANSITIONS     lifecycle.ts:293
B21  MED   Unset year boundary becomes now()          renewal.ts:190
B22  MED   Dateless program -> status-only discount   orgMembership.ts:168
B23  MED   ✓ emailFrom falls back to resend.dev       config.ts:212
B24  MED   calculateAge UTC-gates lead promotion      time.ts:145
B25  LOW   Volunteer designation writes raw email     volunteer-designations:29
[...]

PR #1542, opened 2026-08-06
fix(age): roll birthdays over at org-zone midnight, not UTC (B24)

thpr:
Class-B domain-register audit finding **B24** — "calculateAge UTC-gates lead promotion" (#1529).

## The defect

`calculateAge` read UTC calendar fields on both sides of the comparison. That is correct for the DOB — a calendar date stored at UTC midnight, whose local fields read a day early west of UTC — but the default `asOf = new Date()` is a real instant, not a day. Reading its UTC fields rolled the birthday over at UTC midnight: 7 PM Chicago the evening before, 6 PM in winter.

Someone therefore turned 18 about five hours early, and with them:

- `my-household/page.tsx:408` / `:425` — adult status and the **promote-to-household-lead** control
- `lib/household/leads.ts:121` — the lead-promotion guard, via `isYouth`
- `membership-ops/participants/import/route.ts:72` — the `>= 18` adult classification on import
- `lib/person/adultDob.ts:34` — the `> 25` rule that strips a stored DOB

The function's own comment already named this ceiling and the upgrade path. This is that upgrade.

## The fix

New `orgCalendarDay(instant = new Date())` in `lib/time.ts`: the calendar day an instant falls on **in the organisation's zone**, returned at UTC midnight — the same shape every `@db.Date` column reads back as. `calculateAge`'s default becomes `asOf = orgCalendarDay()`.

The comparison body is untouched: both sides are still read via `getUTC*`, now because both sides are genuinely days rather than one day and one moment.

Done in the default rather than per-caller. The ten call sites that pass nothing are exactly the ones that were wrong; asking each to remember is the drift this codebase avoids elsewhere.

## Why APP_TIMEZONE and not the configured display zone

`#1522` added `setDisplayTimezone` / `getDisplayTimezone` to this same file, so the obvious move is to read the org's configured zone here. It is not usable, and using it would be a regression rather than an improvement:

- `setDisplayTimezone` has exactly one non-test caller — `components/TimezoneProvider.tsx`, a `"use client"` component. Nothing on the server ever calls it.
- The two server callers that need the zone show the established shape: `lib/notifications.ts:160` and `api/attendance/manual/[id]/route.ts:149` each `await resolveDisplayTimezone()` and pass `{ timeZone }` per call, both commented that the provider never runs there.
- The configured value only exists behind `resolveDisplayTimezone()`, which is **async** (a Prisma read behind a 60s cache). `calculateAge` is synchronous and called from render bodies.

So `getDisplayTimezone()` here would give the browser the configured zone and the server the `APP_TIMEZONE` fallback — the two disagreeing about who is 18. That is worse than today's consistent-but-early behaviour. `orgCalendarDay` pins `APP_TIMEZONE` and says so; the settings-aware variant has to be async, so it is marked as the upgrade for when the org actually moves zones.

## Callers that do not change

`ProgramRosterTab.tsx:221` (`program.startAt`), `lib/programAge.ts:66` (`program.asOf`) and `lib/membership/personBgCheck.ts:66` (`boundary`) pass a real calendar date. An explicit `asOf` is still read exactly as stored — deliberately: pushing a UTC-midnight day back through the org zone would shift it to the previous day and break the program age gates.

## Blast radius

Behaviour moves only inside the window between org-zone evening and UTC midnight on the eve of a birthday.

Direct: `import/route.ts:72`, `my-household/page.tsx:42` (displayed age), `:408`, `:425`, `adultDob.ts:34`.

`isYouth` (`time.ts:168`) calls `calculateAge` with no `asOf`, so it inherits the fix and its callers move with it: `household/leads.ts`, `programAge.ts` `isAdult`, `api/roles`, `api/profile`, `api/profile/onboarding-status`, `getFullAttendance`, `AdminEditHouseholdModal`, `attendance/current`, `membership-audit/broken`, `membership-ops/participants/new`.

`isYouth`'s `unknownIs` default is untouched — that is finding B5, with its own owner.

## Tests

New `calculateAge default asOf` block in `src/lib/__tests__/time.test.ts`, on fake timers:

- 00:30 UTC on the birthday — 7:30 PM Chicago the evening before — asserts age **17** and `isYouth` true
- 05:00 UTC — midnight Chicago — asserts the birthday does roll over to 18
- a January pair, so the CST/CDT offset difference is covered rather than assumed

Plus an `orgCalendarDay` block, and the old "rolls over at UTC midnight" ceiling test rewritten as "reads an explicit asOf as the calendar day it is".

Verified the tests fail without the fix: reverting the default to `new Date()` fails exactly the two eve-of-birthday cases (2 failed / 31 passed). The roll-over-on-the-day direction passes either way, as it should.

```
time.test.ts:        33 passed
npm run test:ci:     268 suites, 2548 passed
test:integration:    131 suites, 1340 passed, 6 skipped  (local PG, migrate deploy)
tsc --noEmit, lint:  clean
```

## Docs

One clause added to `docs/conventions.md` under "A day is never taken from a moment", naming `orgCalendarDay()` as the single sanctioned seam that turns an instant into a day — so the next reader does not read the helper as a violation of the rule directly above it.

No migration, no schema change.

### DIFF
--- a/checkin-app/src/lib/__tests__/programAge.test.ts
+++ b/checkin-app/src/lib/__tests__/programAge.test.ts
@@ -1,4 +1,5 @@
 import { checkProgramAge, isKnownAdult, validateProgramAgeBounds, MAX_PROGRAM_AGE } from "@/lib/programAge";
+import { orgCalendarDay } from "@/lib/time";
 
 // as-of date pins calculateAge so the DOB cases don't drift with wall-clock time.
 const asOf = "2026-01-01";
@@ -45,9 +46,12 @@ describe("checkProgramAge", () => {
 });
 
 describe("isKnownAdult", () => {
+  // n years ago today, as a calendar date at UTC midnight — how a DOB is stored.
+  // Built from a moment instead, the 18 case is a birthday landing at whatever
+  // time the suite happens to run, and the boundary it is testing moves.
   const yearsAgo = (n: number) => {
-    const d = new Date();
-    d.setFullYear(d.getFullYear() - n);
+    const d = orgCalendarDay();
+    d.setUTCFullYear(d.getUTCFullYear() - n);
     return d;
   };
 
--- a/checkin-app/src/lib/__tests__/time.test.ts
+++ b/checkin-app/src/lib/__tests__/time.test.ts
@@ -1,4 +1,4 @@
-import { formatDate, formatTime, formatDateTime, formatVisitRange, APP_TIMEZONE, setDisplayTimezone, getDisplayTimezone, toDatetimeLocal, fromDatetimeLocal, formatDateOnly, parseDateOnly, isYouth, calculateAge } from '../time';
+import { formatDate, formatTime, formatDateTime, formatVisitRange, APP_TIMEZONE, setDisplayTimezone, getDisplayTimezone, toDatetimeLocal, fromDatetimeLocal, formatDateOnly, parseDateOnly, isYouth, calculateAge, orgCalendarDay } from '../time';
 
 describe('calendar-date helpers', () => {
   it('stores a picked date at UTC midnight', () => {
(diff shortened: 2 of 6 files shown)


thpr:
CI failed on the first push, in a way that is worth recording because the failing test was right to fail.

`src/lib/__tests__/programAge.test.ts` built its fixtures as `new Date()` minus N years, so the "exactly 18" case was a birthday landing at whatever time of day the suite happened to run. The CI runner started at **02:38 UTC — 9:38 PM Chicago**, squarely inside the window this PR changes: the DOB's UTC day was already tomorrow, the org-zone `asOf` was still today, and `isKnownAdult` correctly said 17.

Fixed by making the fixture a day rather than a moment — `orgCalendarDay()` stepped back whole years in UTC, which is the shape a DOB is actually stored in. The exact-boundary case is now pinned instead of drifting with the clock. Reproduced the failure locally under fake timers at the CI instant, confirmed old fixture false / new fixture true, then dropped the probe.

The other now-relative DOB fixtures in the suite (10, 12, 15, 17, 30, 40 years) sit years away from any boundary and are unaffected.

Also rebased onto `main` (#1477 and #1527). Full unit suite green at 2570.

---

**One thing for the reviewer, deliberately not in this PR.** #1477 landed `lib/membership/personAgreementTriggers.ts` while this was open, and its `inAgreementAgeBand` / `hasKnownAdultAge` take `now: Date` and pass that instant straight to `calculateAge`. An explicit `asOf` is read as given, so those two keep the UTC rollover this PR removes from the default — the nightly job would start asking someone to sign their own agreement at 7 PM the evening before they turn 18, which is a smaller version of the failure that file's own doc comment says the feature exists to prevent.

The fix is `orgCalendarDay(now)` at those two call sites. Left out on purpose: it changes the semantics of a feature that landed hours ago, with its own tests, and does not belong in a fix scoped to B24's default. Happy to do it here if you would rather it not wait.

thpr:
Folded the #1477 caller in, as asked — `e47f7134`.

`inAgreementAgeBand` and `hasKnownAdultAge` in `lib/membership/personAgreementTriggers.ts` passed their `now` instant straight to `calculateAge`. An explicit `asOf` is read as given, so those two kept the rollover the default no longer has: the nightly sweep would open a `PERSON_AGREEMENT` from 7 PM the evening before someone turns 18 — a minor asked to sign a contract they cannot be bound by, which is the failure that file's own doc comment says the rule exists to prevent. Five-hour window, and the sweep runs inside it.

Both now resolve `now` through `orgCalendarDay()` before the age math. Nothing else in that file changes: `recentProgramWhere`, `agreementCycleFloor` and `autoPopulationWhere` compare instants to instants, which is correct.

The unit fixtures moved from `new Date("2026-08-02")` — a bare UTC midnight, which stands for 7 PM on the 1st once the zone is honored — to a real mid-morning instant on the 2nd. New case pinning the behaviour: at 9 PM local on the eve of the 18th birthday, both rules say no. It fails without the change (1 failed / 10 passed) and passes with it.

```
personAgreementTriggers.test.ts:  11 passed
npm run test:ci:                  270 suites, 2571 passed
npm run test:integration:         132 suites, 1355 passed, 3 skipped
tsc --noEmit, lint:               clean
```

`personAgreementTriggers.integration.test.ts` builds its people at 17/20/30, years clear of either edge, so it is unaffected either way.

jee7s (approved):

PR #1545, opened 2026-08-06
fix(attendance): program-less events are board/sysadmin only (B10)

thpr:
Addresses finding **B10** of the Class-B domain-register audit, #1529. That issue tracks many findings, so no closing keyword here — it stays open for the rest.

## The defect

`POST /api/events/[id]/attendance` applies its only authz on `participantIds` inside the `if (programId != null)` branch: the enrollment/volunteer filter that rejects targets not on the program's roster. Nothing else in the handler validates them.

The actor gate admitted keyholders alongside admin/board:

```ts
const isSysAdminOrBoardOrKeyholder = auth.user.isSysadmin || auth.user.isBoardMember || auth.user.isKeyholder;
```

On a program-less event the target filter is skipped, so for those events `participantIds` was entirely unvalidated. The transaction then adopts an overlapping unassociated visit into the event or creates a synthetic one — meaning any keyholder could fabricate attendance and presence records for **any person in the system**.

The skip's comment asserted such events were "reachable only by admin/board". That was true of the lead-mentor branch (`event.program` is null, so `isLeadMentor` is false) but not of keyholders, who are neither admin nor board.

## The fix

Narrow the actor rather than adding a target filter:

```ts
const isPrivileged = auth.user.isSysadmin || auth.user.isBoardMember
    || (event.programId != null && !!auth.user.isKeyholder);
```

Keyholders keep everything they can do today on program events, where the enrollment filter already bounds their targets to that program's roster. Board and sysadmin keep the unfenced path, which they already hold by written rule — `docs/rules/attendance-checkin.md:116`: "The board and sysadmins edit or delete any visit, and record one for someone else at any past time — the walk-in nobody badged in." No new rule is invented; the actor set is narrowed to the population the rule already grants it to.

The misleading comment at the filter is replaced with the present-tense constraint.

No schema change, no migration, nothing under `src/security/`.

## Why program-less events are an admin/board surface

- `GET /api/events`, whose only job is listing standalone events, is already `withAuth({ roles: ['isSysadmin', 'isBoardMember'] })`.
- `GET /api/events/mine` filters `programId: { in: [...] }`, so a null-program event can never appear there. A keyholder reaching one was API-only, not a UI workflow.
- `POST /api/events` cannot be used by a keyholder to create one: `isLeadMentor` is only set inside `if (programId)` after confirming the caller leads *that* program, so a falsy `programId` leaves the gate demanding `isSysadmin || isBoardMember`.

## For the reviewer

**The same assumption exists one file over and is correct there.** `PATCH /api/events/[id]` (`manualEditAttendance`) carries the same `programId != null` skip and nearly the same comment, but its gate is `isSysAdminOrBoard || isLeadMentor || isCoreVolunteer` — no keyholder — so the parenthetical holds. Deliberately not touched. `GET /api/events/[id]` likewise has no keyholder. The attendance POST was the only `isKeyholder` under `src/app/api/events/`.

`src/lib/attendanceConflicts.ts` and `src/app/api/nav/todo-counts/route.ts` both scope to led programs via `leadMentorId`; neither depends on keyholders being able to POST here.

## Tests

Added to `src/app/__tests__/eventAttendanceAPI.integration.test.ts`: a standalone (`programId: null`) event and a keyholder who is neither board nor sysadmin.

- **Fails without the fix** — verified by reverting the route: `Expected: 403, Received: 200`. The keyholder wrote the visit for an unrelated person.
- **Passes with it** — 403, with `{visits, audits}` counts re-queried and unchanged, so the guard is proven to run *before* the write rather than after.
- Companions: the same keyholder still succeeds on a program event for an enrolled target; a board member still succeeds on the program-less event.

```
Unit         (npm run test:ci):              268 suites, 2543 passed
Integration  (npm run test:integration):     131 suites, 1343 passed, 6 skipped
tsc --noEmit:                                no new errors
```

### DIFF
--- a/checkin-app/src/app/__tests__/eventAttendanceAPI.integration.test.ts
+++ b/checkin-app/src/app/__tests__/eventAttendanceAPI.integration.test.ts
@@ -25,6 +25,10 @@ describe('Event Attendance API Integration Tests', () => {
     // Cross-tenant attacker: lead of a DIFFERENT program (IDOR boundary).
     let foreignLeadId: number;
     let foreignProgramId: number;
+    // Keyholder who is neither board nor sysadmin, plus a standalone
+    // (program-less) event — the unfenced-target boundary.
+    let keyholderId: number;
+    let standaloneEventId: number;
 
     beforeAll(async () => {
         // Clean up any leaked state
@@ -32,7 +36,7 @@ describe('Event Attendance API Integration Tests', () => {
             where: { person: { email: { contains: 'event-attendance-test' } } }
         });
         await prisma.event.deleteMany({
-            where: { name: 'Attendance Test Event' }
+            where: { name: { in: ['Attendance Test Event', 'Attendance Test Standalone Event'] } }
         });
         await prisma.program.deleteMany({
             where: { name: 'Attendance Test Program' }
@@ -89,6 +93,11 @@ describe('Event Attendance API Integration Tests', () => {
         });
         foreignProgramId = foreignProgram.id;
 
+        const keyholder = await prisma.person.create({
+            data: { email: 'keyholder-event-attendance-test@example.com', name: 'Keyholder Att Test', isKeyholder: true, household: { create: { name: "Test HH" } } }
+        });
+        keyholderId = keyholder.id;
+
         const now = new Date();
         const pastStart = new Date(now.getTime() - 2 * 60 * 60 * 1000); // 2 hours ago
         const pastEnd = new Date(now.getTime() - 1 * 60 * 60 * 1000); // 1 hour ago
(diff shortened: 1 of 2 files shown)


jee7s (approved):

thpr:
Closing as obsolete. #1553 deleted the route this hardens — `checkin-app/src/app/api/events/[id]/attendance/route.ts` and its only test file, `eventAttendanceAPI.integration.test.ts` — so both files this PR touches are gone from `main` and the whole diff lived inside them. Rebasing would resurrect a deliberately deleted endpoint.

Deletion is the better fix, and #1553 reached the same finding independently: its message notes the gate "admitted keyholders (who cannot read the roster at all — GET 403s them) and refused core volunteers, the exact inverse of the screen that does the job." Same defect, resolved by removing the endpoint rather than narrowing its gate — which also takes out the bulk-mark path this PR left in place.

B10 is resolved on `main`. Verified there: no code references the deleted route, and the live correction path (`PATCH /api/events/[id]` → `manualEditAttendance`) gates on `isSysAdminOrBoard || isLeadMentor || isCoreVolunteer` with no keyholder, so the unfenced-target class does not exist on it.

Branch `claude/goofy-greider-09c368` left in place rather than deleted; nothing on it is needed.

PR #1558, opened 2026-08-07
fix(audit): persist significance on every visit edit and delete path (#1523)

### LINKED ISSUE #1523: Persist visit-correction significance on every edit and delete path
## Summary

`newData.significance` is written by exactly one of the seven files that audit a `Visit` edit or delete. Verified on `main` at `d4b38c5b`:

| Route | Visit audit writes | persists `significance` |
|---|---|---|
| `attendance/manual/[id]/route.ts` | 2 | **yes** |
| `attendance/manual/route.ts` | 1 | no (a create has no prior value) |
| `events/[id]/route.ts` (`manualEditAttendance`) | 3 | **no** |
| `events/[id]/attendance/route.ts` (roster mark) | 2 | no (moves no clock) |
| `facility/visits/route.ts` | 2 | **no** |
| `facility/visits/insert/route.ts` | 1 | no (a create) |
| `my-programs/conflicts/resolve/route.ts` | 1 | **no** |

## Why it matters

#1478 introduced `PROXY_MULTIPLIER = 2` and a `byProxy` option specifically to weight an adult changing another person's record — and in the same PR added three writers in `events/[id]/route.ts`, where a program lead corrects a member's attendance, none of which score. One file in the system reaches the multiplier.

The AT12 correction-review screen (#1258) reads `newData.significance.flagged` for its default view. A flagged lens that silently omits the staff and lead paths reads as *"no significant corrections this month"* when what actually happened is that the paths capable of the largest corrections never scored themselves.

**It cannot be fixed at read time.** Some of these rows carry no `oldData` and some carry no times at all, so there is nothing to recompute from. A row that never scored cannot be scored retroactively — the gap is only closable at the write.

## Scope

Add `editSignificance` / `deleteSignificance` (both exported from `checkin-app/src/lib/visit/significance.ts`, both taking the `byProxy` option) at five call sites:

- `facility/visits/route.ts` — PATCH and DELETE
- `my-programs/conflicts/resolve/route.ts` — DELETE
- `events/[id]/route.ts` — the `manualEditAttendance` EDIT and DELETE

All five already carry the `oldData` those calls need. Roughly four lines each.

**Out of scope:** creates and the roster mark. A create has no overwritten value and the significance model defines nothing for it; the roster-mark edit changes only which event a visit belongs to and moves no clock, so its delta is zero by construction. Whether a roster mark deserves a fabricated score is an open product question on #1258, not this issue.

## Origin

Raised as item 3 of a review on #1478 and flagged there as non-blocking. #1478 merged with items 1 and 2 fixed — every visit audit write now stores the subject person — and this one unaddressed, so it has no host. Filed so it has an owner rather than living in a merged PR thread.

Design context: `docs/in-design/ATTENDANCE_CORRECTION_REVIEW.md` §3 (open in #1497), which gates on this.

---
_Posted by Claude._

### LINKED ISSUE #1619: fix(attendance): the roster mark writes WEB, so nothing in the app ever writes LEAD_MARKED
`LEAD_MARKED` has four readers and **zero writers**. Verified against `origin/main`
@ `fbc59f7` on 2026-08-09 — `git grep LEAD_MARKED -- checkin-app/src` returns only
readers (`api/facility/trends/route.ts`, `facility-ops/visits/page.tsx`,
`lib/email-templates/base.ts`, `lib/visit/significance.ts`) plus the schema and the
two `20260803000*` migrations. The only rows that carry it are the ones the
`20260803000100` backfill rewrote.

The roster mark — `manualEditAttendance` in
`checkin-app/src/app/api/events/[id]/route.ts` — stamps `WEB` instead:

```ts
// checkin-app/src/app/api/events/[id]/route.ts (~line 346, the `times` object)
const times = {
    arrivedAt: arrival!,
    departedAt: dep,
    arrivedVia: "WEB",
    departedVia: departedAt ? "WEB" : null
} satisfies Prisma.VisitUncheckedUpdateInput;
```

This is a lost edit, not a design choice.
`docs/designs/1256_ATTENDANCE_CORRECTION_SURFACE.md` § "The `VisitSource` split"
lists this writer's migration as `SYSTEM → LEAD_MARKED` on **both** fields, and the
schema says the same (`LEAD_MARKED = staff asserted presence; a roster mark writes
BOTH fields at the event window`). The 3-way split was a per-writer edit; the other
two writers (`closeAllOpenVisits` → `FACILITY_CLOSE`, the nightly sweep →
`AUTO_CLOSE`) got theirs. This one was missed when the roster mark moved out of the
since-deleted `POST /api/events/[id]/attendance` into `manualEditAttendance`.

#### Impact

A program lead marks twenty participants Present for a three-hour session. Sixty
placeholder hours — nobody badged in — land in the Participation Trends chart as
measured facility time, because the exclusion meant to catch them
(`arrivedVia notIn ["LEAD_MARKED", "SYSTEM"]`) matches no live row.
`lib/visit/significance.ts` also weighs the mark at 1 (self-report) instead of 2
(staff observation), so a member correcting a lead's mark under-flags to the board.

#### ⚠️ This is NOT the one-line change it looks like

Do **not** simply swap `"WEB"` for `"LEAD_MARKED"` in that `times` object. The same
object is applied by two branches, and one of them is editing somebody else's real
arrival:

```ts
// ~line 334 — adopt an ordinary open walk-in into this event
const adoptable = openVisit !== null && (
    openVisit.associatedEventId === eventId ||
    (dep === null && openVisit.associatedEventId === null)
);
const existingVisit = adoptable ? openVisit : await tx.visit.findFirst({ ... });
```

An adopted visit is a **real, measured arrival** — someone badged in at the kiosk
(`SCANNER`) or self-checked-in (`WEB`), and the lead is folding that existing open
visit into the event rather than creating a second one. Stamping `LEAD_MARKED`
unconditionally relabels that genuine arrival as staff-asserted, and — with this
PR's NULL-admitting exclusion now live — **drops it out of the hours chart
entirely**. That trades the current over-count for an under-count on precisely the
people who did badge in. The same applies to a visit already tied to this event
that a lead is only correcting the times on.

#### The rule that is actually correct

`arrivedVia` describes **how the arrival was recorded**, not who last touched the
row. Corrections are tracked in `AuditLog` (`newData.type =
"lead_attendance_correction"`), not by rewriting the source.

- **Create branch** (no existing visit — staff asserted the whole visit):
  `arrivedVia: "LEAD_MARKED"`.
- **Update branch** (adopted walk-in, or an existing visit for this event):
  leave `arrivedVia` alone. A row that already carries `LEAD_MARKED` stays
  `LEAD_MARKED` for free; a `SCANNER`/`WEB` arrival keeps its measured source.
- **`departedVia`**: `departedAt ? "LEAD_MARKED" : null` on **both** branches — if a
  lead types a departure time, that departure is staff-asserted whatever the arrival
  was. This does not reach the trends filter, which keys on `arrivedVia` only.

So `times` can no longer be one shared object across both branches — that split is
the actual work here.

Keep the `AuditLog` `newData` in step with whatever is really written, so the
correction surface does not claim a field changed when it did not.

#### Acceptance criteria

- [ ] A roster mark that creates a visit writes `arrivedVia: "LEAD_MARKED"`, and
      `departedVia: "LEAD_MARKED"` when the lead supplied a departure.
- [ ] A roster mark that **adopts an existing open walk-in** leaves its
      `arrivedVia` as `SCANNER`/`WEB`. Guard test — this is the regression this
      issue exists to prevent.
- [ ] A roster mark that edits an existing visit for the event does not overwrite
      that visit's `arrivedVia`.
- [ ] `GET /api/facility/trends` excludes the created roster marks and still counts
      the adopted walk-in. Expect the chart's hours to drop where leads mark rosters
      — that is the fix landing, not a new bug.
- [ ] `lib/visit/significance.ts` scores a created roster mark at 2.

#### Coordination

- `checkin-app/src/app/api/events/[id]/route.ts` is also touched by open PR #1558
  (persist significance on every visit edit/delete path). Land after it, or
  coordinate — the two overlap on the same transaction, and #1558's significance
  persistence is what makes the corrected weighting durable.
- Reader side already shipped in this PR (trends exclusion admits NULL
  `arrivedVia`); no reader change is needed for this.

#### Out of scope / open question

Roster marks already written as `WEB` cannot be recovered from the source field —
they are indistinguishable from a member's own self-report by source, by
`associatedEventId`, and by times. The only discriminator is the
`newData.type = "lead_attendance_correction"` audit row. A one-time backfill keyed
on that is possible (a migration is a fair place to read `AuditLog` JSON, unlike the
hot trends read), but it is a separate decision from fixing the writer — history
stays over-counted until someone takes it.

---

Filed from the review of #1608, which fixed the reader side (the trends exclusion
now admits a NULL `arrivedVia`) and left this writer-side half unfixed by design.
The text above is @jee7s's, reproduced from that PR's description so it survives
the merge.

🤖 Filed by [Claude Code](https://claude.com/claude-code)

jee7s:
Closes #1523. Closes #1619 — #1622 merged into this branch and carries that fix;
it targeted a non-default base, so its own closing keyword never registered.
App code only — no boundary file, so isolation is a no-op.

## Why

`newData.significance` was written by exactly one of the seven files that audit a `Visit` edit or delete. #1478 introduced `PROXY_MULTIPLIER = 2` and a `byProxy` option precisely to weight an adult changing another person's record — and then added three writers where a program lead corrects a member's attendance, none of which scored. One file in the system reached the multiplier.

This matters beyond symmetry: the AT12 correction-review screen (#1258) reads `newData.significance.flagged` for its default view, and a flagged lens that silently omits the staff and lead paths reads as *"no significant corrections this month"* when what happened is that the paths capable of the largest corrections never scored themselves. It cannot be recovered at read time — rows that never scored carry no `oldData` in some cases and no times at all in others.

## What changed

Five call sites across four files:

| Route | Action | byProxy |
|---|---|---|
| `facility/visits` PATCH | EDIT | `auth.user.id !== previous.personId` |
| `facility/visits` DELETE | DELETE | `auth.user.id !== removed.personId` |
| `events/[id]` `manualEditAttendance` | DELETE (per row) | `userId !== targetId` |
| `events/[id]` `manualEditAttendance` | EDIT | `userId !== targetId` |
| `my-programs/conflicts/resolve` | DELETE | computed from the visit |

## Two things worth a reviewer's attention

**`facility/visits` PATCH now scores against the in-lock re-read, not the pre-lock snapshot.** `existing` is read outside the advisory lock; the transaction re-reads `current` and derives the new times from it. Scoring against `existing` would understate a racing SCANNER close to zero, because `minutesBetween` returns 0 on a null side. The transaction now returns `{ visit, previous }`, and `secondaryAffectedEntity`, `oldData` **and** the score all come from `previous`.

That also closes a stale-snapshot bug #1478 left behind — the audit row's `oldData` was the pre-lock read.

**The roster-adopt EDIT is deliberately left unscored.** `events/[id]/attendance` sets `associatedEventId` and moves no clock, so a score would be a structural zero. Putting a zero inside the lens would render as "reviewed and found insignificant" when nothing was reviewed. It gets a comment and a negative test assertion pinning that it stays absent.

## Adjacent defect — now fixed here

This section originally recorded a reopen scoring 0 as out of scope. `47e43550` closed it instead, so the note is kept corrected rather than deleted.

`events/[id]/route.ts` used to accept a Present mark with a blank departure against a *closed* visit, nulling its `departedAt` and putting someone back on the live in-the-building roster hours after they left — and it scored 0 doing it, because `minutesBetween` returns 0 on a null side. It now refuses that edit outright (`existingVisit?.departedAt && !dep`), matching what `facility/visits` PATCH already did.

What remains is the narrower case where the visit is *genuinely open*: `deleteSignificance` yields `{score: 0, flagged: true}` for a row with no `departedAt`, so deleting a live scanned visit reads as insignificant. That is a semantics change to weights the helper's own docstring calls "v1 defaults awaiting board calibration", so it is deliberately not in this PR — filed as #1630.

## Tests

Seven assertions across five existing integration suites, no new files. All five hand-computed scores were re-derived against the actual fixtures rather than trusted from comments.

No `node_modules` in this checkout when the section above was written, so that
round was not executed locally. The review round below **was** — see
"Verification" at the end.

---

# Review round 2 (thpr, 2026-08-10T02:17:31Z)

Four items, all addressed in `6249f2d1`. Nothing else touched.

## 1. `facility/visits` PATCH no longer stamps `WEB`

Confirmed before changing it: `WEIGHTS` defines `WEB` as "the member's own prior
self-report" (weight 1), and `facility/trends` drops only
`arrivedVia IN (LEAD_MARKED, SYSTEM)`, so a `WEB` arrival counts as measured
hours. The divergence was real.

**The value chosen is `LEAD_MARKED`, stamped per field.** `VisitSource` has no
board-specific member — `SCANNER`, `WEB`, `LEAD_MARKED`, `FACILITY_CLOSE`,
`AUTO_CLOSE`, `SYSTEM` — and `LEAD_MARKED` is the only "staff observation" value
in it. A board member editing in facility-ops is staff, so it is the right one.

**Why per field, and why that is the honest analog of `events/[id]`.** The rule
next door is *stamp only where the value is genuinely staff-asserted*: the create
branch stamps `arrivedVia: "LEAD_MARKED"` because that mark produced the arrival;
the update branch omits `arrivedVia` entirely so an adopted walk-in keeps its
`SCANNER`; and `departedVia` becomes `LEAD_MARKED` only when the lead actually
typed a departure.

`facility/visits` PATCH is update-only, so a literal copy of "never restamp
`arrivedVia` on an update" would be wrong — it would mean the route could never
record a staff-asserted arrival at all. But the route already encodes the real
predicate more precisely than `events/[id]` can: it writes a field **only when
the caller supplied it** (`parsedArrived ? … : {}`). That conditional *is* the
per-field "did staff type this" test, and a side not sent keeps its existing
`via` — which is exactly the property "an adopted walk-in keeps its `SCANNER`"
was protecting. So the change is the one-line value swap, with the existing
conditional doing the discrimination.

Corroborating evidence that this shape is the intended one: `attendance/manual/[id]`
uses the identical conditional-spread with `WEB`, and there `WEB` is *correct* —
that is the member/household-lead self-service path. `facility/visits` PATCH is
board/sysadmin-only (`roles: ['isSysadmin', 'isBoardMember']`); it had inherited
the member path's spelling.

I did not find an honest reason to deviate from the review's framing.

Note the score of *this* edit is unchanged — `editSignificance` weighs the OLD
values' sources. What changes is the next correction's weight (1 → 2) and
trends membership, which is precisely the divergence the review described.

## 2. The open-visit zero — recorded, not fixed

`significance.ts` is untouched. The assertion in
`adminVisitsAPI.integration.test.ts` no longer reads the 0 as intended; it is
re-recorded as unresolved and points at the issue below.

Filed by thpr as **#1630**, near-verbatim from the text that stood here. The
test comment now names it.

## 3. `TYPE_LABEL`

`staff_correction` → "Staff correction", `staff_removal` → "Staff removal",
`conflict_resolution` → "Conflict resolution". Three lines, matching the
sentence-case of the existing entries.

## 4. The `raise`/`raised` sweep

All nine sites, using the review's replacements verbatim. `grep -i rais` now
returns nothing in either document.

Line 156 of the register deliberately did **not** gain "nor shows on the
screen" — the source suppression zeroes only that field's term, so a
simultaneous arrival move can still carry the row over threshold. It reads
"adds nothing to the score however large the correction".

Two coordinated wordings worth noting: line 144 becomes "shows on **the review
screen**", which is what gives line 148's "Showing on **that screen**" its
antecedent.

### The non-blocking §3.9 note: cut

The closing clause "; sending those would train everyone to ignore the alert and
cost it the signal" is gone. It stated the reason rather than the rule — no
change can violate it — and the `[Decision — deliberate limit]` tag already
carries that weight. The bullet now ends "on the board's own corrections the
board would be telling itself."

## Found, deliberately not touched

`checkin-app/src/app/api/facility/visits/insert/route.ts:67-68` stamps
`arrivedVia: "WEB"` / `departedVia: "WEB"` on a visit a staff member inserts for
**another person**. That looks like the same defect as item 1 — a staff-typed
time recorded as the member's self-report — and it is pinned by an assertion in
`adminVisitsAPI.integration.test.ts`. It was not in the review, so it is
untouched and reported rather than fixed.

## Verification

Real `npm ci` (the worktree's `node_modules` was a symlink to another
checkout), `prisma generate`, Postgres 15 in Docker.

- **Red/green on item 1** — against the *pre-fix* source, `adminVisitsAPI`
  fails 2 / 24 with `Expected: "LEAD_MARKED" / Received: "WEB"` (the new test
  and the pre-existing `departedVia` assertion). With the fix: 24 / 24.
- `npm run test:integration` over `adminVisitsAPI`, `facilityTrends`,
  `visitWriteLock`, `auditLog`, `myProgramsConflictsResolve`,
  `eventCancelAndManualAttendance`, `attendanceManual` — 9 suites, 83 tests, green.
- `npm run test:ci` — 278 suites, 2710 tests, green.
- `npx tsc --noEmit` clean; `npm run lint` (`--max-warnings 0`) clean.

---

# Review round 3 (thpr, 2026-08-10T20:08:31Z)

`a2e6a586` answered B by adding a `STAFF_CORRECTED` enum value and a migration.
**That commit has been removed from this branch** — `git reset --hard 6249f2d1`, then
force-pushed with `--force-with-lease`. It duplicated #1624, which you had filed
nineteen minutes earlier having already considered and rejected patching this by
adding a value, and it pre-empted the board decision #1624 says gates the backfill.
Nobody had reviewed it, so it is gone from the history rather than reverted on top of
it: there is no added-then-deleted migration directory for you to decode.

What is here instead is `ebfdf542` and `8f8d2162`. No schema change, no migration, no
`VisitSource` edit — `git diff origin/main origin/claude/1523-significance --
checkin-app/prisma/` is empty.

## B. The regression `6249f2d1` introduced — fixed without touching the enum

Confirmed as described: `facility/trends` drops `arrivedVia IN (LEAD_MARKED, SYSTEM)`
(`route.ts:80-81`), so once the PATCH restamped, a board member fixing a bad `SCANNER`
arrival deleted that visit's hours from the trend.

**The PATCH no longer restamps `arrivedVia` at all.** The value already on the row is
how the arrival was measured, and correcting the time does not change that — so a
corrected `SCANNER` arrival stays `SCANNER` and stays counted. A typed departure still
stamps `LEAD_MARKED`; `trends` keys on arrival only, so that side is unaffected either
way.

This is the rule `events/[id]` already follows — its update branch omits `arrivedVia`
entirely so an adopted walk-in keeps its `SCANNER`. Round 2 argued the two routes
should legitimately differ, because `facility/visits` PATCH can tell "did staff type
this field" from the `parsedArrived ? … : {}` conditional. It can, but that is the
*who* axis, and `arrivedVia` is not where the *who* axis lives. The two staff paths
now agree instead of diverging.

### What this does not fix

Your first review's item 1 stands, and I want to be plain about it rather than let the
green test imply otherwise. Leaving `arrivedVia` alone fixes the **corrected-real-visit**
case — the live regression. It does not fix the case where staff type an arrival onto a
row whose `arrivedVia` is already `WEB`: that row still claims the member self-reported
a time they had no part in.

There is no honest value to write there. `WEB` claims self-report; `LEAD_MARKED` drops
the visit out of trends. With the current enum there is no value meaning "staff-corrected
but still measured" — which is precisely #1624's finding. That residue is #1624's, behind
the board decision it names, and I have not invented a workaround for it here.

**Item A (`facility/visits/insert` stamps `WEB`) is untouched** — you folded it into
#1624 yourself, and the same argument applies: no existing value is right there either.

### The test gap you named

Correct — the trends suite built every fixture by hand, so both behaviours passed it.
The new case creates a `SCANNER` visit, drives the real `PATCH` route, and re-reads
`trends`. Against the pre-fix stamp:

```
● Facility trends API › keeps a corrected scanner visit in the hours after a staff PATCH
    Expected: 1
    Received: 0
      > 265 |         expect(after.totals.uniqueVolunteers).toBe(1);
```

The visit vanishes from the trend, exactly as you described. With the fix it is counted
at its corrected 1h. The `adminVisitsAPI` case pins the stamp itself
(`SCANNER` in, `SCANNER` out) and fails the same way.

## Non-blocking

**`TODO(#TBD)` → `#1630`.** The `TODO(...)` form and the pointer into this PR's
description are both gone. The comment now opens `UNRESOLVED (#1630)` above the
sentences that already carry the meaning.

I kept the number rather than dropping the line entirely, which is the narrower reading
of AGENTS.md than yours. Its rule is headed "No history in comments" and bans narrating
how the code got here; a live pointer to an open decision is neither history nor a TODO,
and `#NNNN` in a comment is ordinary in this repo — around twenty sites on `main`,
including several in these same integration suites. Without it a reader gets "0 is what
ships, not what is right" and no way to find where that gets settled. Say the word and
it comes out.

## Not done, and it was done before — the register bullet

Restoring "…; sending those would train everyone to ignore the alert and cost it the
signal" to `docs/rules/attendance-checkin.md:153` was part of `a2e6a586`, so removing
that commit removed it too. I did not re-apply it: the instruction for this pass was the
enum removal and the narrow regression fix, nothing else. It is a one-line doc edit and
still worth making — flagging rather than quietly re-adding it.

## Found, deliberately not touched

- `docs/security/presence-identity-tiering.md:97` still describes `VisitSource` as
  "Enum `SCANNER`/`WEB`/`SYSTEM`". Stale since the 3-way split, not from this PR. The
  tiering conclusion ("stay public") is unaffected.
- The drain-window finding `a2e6a586`'s round-3 notes raised — that
  `20260803000100_visit_source_split_backfill`, **already on `main`**, rewrites `SYSTEM`
  rows to `LEAD_MARKED` and so breaks a v1.1.1 client's `Visit` reads during the deploy
  drain — was found while probing that commit's migration. This PR no longer has a
  migration, so it is not this branch's concern, but the finding was about `main` and
  stands on its own. Not filed; recorded here so it is not lost.

## Verification (round 3)

Real `npm ci` at the monorepo root (root `node_modules` was empty; the install is
hoisted there, and `prepare: husky` fails harmlessly under a worktree), Node 24 via
nvm, `prisma generate`, Postgres 15 in Docker.

- **Red → green, both directions measured.** With the fix reverted to `6249f2d1`'s
  stamp and the new tests in place: `Tests: 2 failed, 28 passed` — the trends case
  (`uniqueVolunteers` 1 → 0) and the `adminVisitsAPI` stamp case. Nothing else moved.
  With the fix: 30/30.
- Integration over `facilityTrends`, `adminVisits`, `auditLog`,
  `myProgramsConflictsResolve`, `eventCancelAndManualAttendance`, `attendanceManual`,
  `visitWriteLock` — 9 suites, 84 tests, green.
- `npm run test:ci` — 278 suites, 2710 tests, green.
- `npx tsc --noEmit` clean (exit 0, no output); `npm run lint` (`--max-warnings 0`) clean.
- `prisma generate` leaves `src/security/generated/classifications.ts` byte-identical.
  `git diff --name-only origin/main...HEAD -- checkin-app/src/security/` is empty.

---
_Posted by Claude._


thpr (commented):
Verified the five scores by hand against the fixtures, traced all seven `Visit`-audit writers, and read both the governing design (`1256_ATTENDANCE_CORRECTION_SURFACE.md`) and the AT12 design in the open #1497.

## Correct as written

I re-derived all five expected scores against the actual fixtures — 0 / 0 / 20 / 60 / 720 all check out, and `Math.round` absorbs the `Date.now()` drift between fixture statements in the two cases where it could bite. The `existing` → `previous` swap in `facility/visits` PATCH is the right call, and the transaction's union return narrows properly. Coverage of the seven writers is complete: the three `CREATE`s are correctly out of scope, and the roster-adopt EDIT genuinely moves no clock.

## Findings

### 1. The delete floor now floods AT12's default lens

`deleteSignificance` returns `flagged: true` unconditionally, and #1497 §3 makes `newData.significance.flagged` the database-filtered predicate behind the screen's default view. `manualEditAttendance` Absent and `conflicts/resolve` are routine weekly lead workflow — every no-show marked absent is now a `flagged: true` row in the board's landing view.

That is the exact failure `1256` §"Which paths raise the live flag" names as its reason for keeping the floor out of the live email: *"emails the board on every ordinary correction, which trains everyone to ignore the alert and costs the signal its whole value."* #1497 does not address it — there is no floor or volume discussion in that doc. So this PR moves the noise out of the inbox and into the screen, without either design noticing the move.

Not necessarily a blocker, but it is a product decision being made implicitly. The lever is either scoring these paths without asserting `flagged` (leaving the floor to `attendance/manual/[id]`, where the design argues for it), or making AT12's default lens `flagged AND actorClass = self`. Either way it wants a line in #1497.

### 2. `facility/visits` PATCH/DELETE and `conflicts/resolve` still write no `newData.type`

They are the only `Visit`-audit writers without one — `facility/visits/insert` writes `staff_entry`, `events/[id]` writes `lead_attendance_correction`, `attendance/manual/[id]` writes `self_correction`. #1497 §2 uses `newData.type` as the drill-down label, so these three rows land there unnamed. This PR already rewrites all three `newData` payloads, so it is a token per site.

### 3. `checkin-app/src/lib/visit/significance.ts:63` is now false

`minutesBetween`'s comment says *"One side null only when CLOSING an open visit (reopening is rejected at the route)."* This PR adds a caller that reopens, and its own inline comment at `checkin-app/src/app/api/events/[id]/route.ts:381` says so. AGENTS.md asks for stale comments to go when the code is touched, and this is the comment documenting the function being wired up.

### 4. The reopen defect is worse than the PR describes

`1256` §2 lists "no reopening a closed visit" among the route-level validity checks, and `facility/visits` PATCH enforces it. `events/[id]` on a Present mark with no departure against a closed visit does not — it silently nulls `departedAt`. The follow-up issue should read "this path violates a documented validity rule", not "this path scores 0". The zero is the symptom.

### 5. `1256` §"Which paths raise the live flag" is now wrong, and untouched

It states that the `newData.significance` object *"that `attendance/manual/[id]` persists"* is a convenience for filtering, and that AT12 can recompute because "every path above writes the source fields into `oldData`". Six writers now persist it, and #1497 §3 explicitly rejects the recompute. The correction currently lives only in an unmerged PR. Either amend the line here, or note on #1497 that it owns the edit.

### 6. Minor: the `auditLog.integration.test.ts` assertion pins nothing

It passes identically against the pre-swap code, since `existing` and `previous` are the same row in that fixture. The PR comment says as much, so this is honest rather than wrong — but the behaviour change has no test that would fail on a revert.

## One overstatement

"It cannot be recovered at read time" holds for the `facility/visits` PATCH stale snapshot, and for the two row shapes #1497 §1 identifies. The other four call sites all carry times and sources in `oldData` and were fully recomputable. The issue text carries the same overstatement, so the PR is faithful to it — worth correcting in both places so the next reader does not inherit it.

---
_Posted by Claude._

thpr (commented):
This PR establishes a standing rule that isn't written down anywhere.

Before it, `newData.significance` was written by one path, so "raised" meant one thing: a member corrects or removes their own visit, the board gets an email. After it, five paths score and flag — but only that one path still notifies. `deleteSignificance` flags unconditionally, so every routine roster Absent and every conflict resolution now carries `flagged: true` while interrupting nobody.

`docs/rules/attendance-checkin.md` currently has one word covering both mechanisms:

> Every removal is raised whatever it overwrote, because erasing a record is notable at any size.

Suggested addition to Procedure → "The visit record", immediately after that line, since a qualification goes on the rule it qualifies (`DOCUMENTATION_STANDARD.md` §3.7):

> - Every change to a visit is recorded and weighed, whoever made it. The board is told at the time only when a member changed their own record, or their household lead changed it for them. A program leader marking who did not turn up, or clearing a duplicate visit, is recorded and reviewable but interrupts nobody.  `[Decision — deliberate limit]`

`[Decision — deliberate limit]` is the tag §3.6 defines for a chosen absence — "the one most likely to be 'fixed' by someone helpful." That is the risk this line exists to hold off: the next reader sees five paths scoring and one notifying, reads it as an oversight, and wires the rest in.

One note on placement. `DOCUMENTATION_MIGRATION.md` §5 moves `1256_ATTENDANCE_CORRECTION_SURFACE.md` into `docs/in-design/` on the ordinary extract-and-delete lifecycle, so the amendment in `c15596da` is invested in a doc with a scheduled end. The register is where this survives.

---
_Posted by Claude._

thpr (changes requested):
Blocking on the register line from the previous comment.

This PR is what establishes the rule — five paths now score and flag, one still notifies — and `docs/rules/attendance-checkin.md` says neither half. Per `DOCUMENTATION_STANDARD.md` §5, a change that alters a rule says so explicitly rather than leaving the reviewer to notice.

The code is sound; this is the register catching up to it.

---
_Posted by Claude._

jee7s:
Status note — this branch has moved a long way since the last review, so here is what is on it now.

**The register line that was requested is in.** `docs/rules/attendance-checkin.md` gained the entry recording that five visit-write paths score and flag while one only notifies, per `DOCUMENTATION_STANDARD.md` §5. It landed at 04:00Z on 08-09; the `CHANGES_REQUESTED` review is from 03:00:39Z, so the review predates the commit answering it.

**#1622 is merged into this branch** and carries the fix for #1619 — the roster mark stamped `arrivedVia: "WEB"`, so lead-marked attendance counted as measured facility hours in the board's trends chart. The rule implemented is the conditional one, not the one-liner: `LEAD_MARKED` on the create path only, `arrivedVia` left alone on update so an adopted walk-in keeps its `SCANNER`/`WEB`, and `departedVia` set on both when a departure is typed. Its test rejects the naive unconditional version as well as the status quo.

Because #1622 targeted this branch rather than `main`, its `Fixes #1619` never registered — GitHub only honors closing keywords against the default branch. I have added the reference to this PR's description instead, so #1619 closes when this merges. `closingIssuesReferences` now reads #1523 and #1619.

**Merging `main` in needed a manual graft, and it is worth knowing why.** #1604 moved the `conflicts/resolve` read inside `pg_advisory_xact_lock` and rewrote the surrounding lines to a new `current` symbol. This branch independently *inserted* four lines just below. Non-overlapping ranges, so `git merge-tree` exited 0 with no conflict — and the result mixed the two: `oldData` on the in-lock `current`, `newData` still on the pre-lock `visit`, which was in outer scope and therefore compiled, typechecked, and passed every existing test including this PR's own `score: 720` assertion. `deleteSignificance` reads exactly the four fields a concurrent departure mutates between the two reads, which is the race the lock was added to order. Fixed in `33d66f38`, with a regression test that holds the advisory lock on raw connections outside Prisma's pool and waits on `pg_locks` for deterministic ordering — it fails on the unfixed merge with `score: 0` against an expected `1080`.

**CI is 9/9 green** on `b5d1b9e3`.

No action taken on the review state itself — dismissing or re-reviewing is yours. Flagging only that the two defects this PR closes are the last v1.1 gaps on `main`: the corrections feed omitting every privileged edit, and a Present mark with a blank departure reopening a closed visit onto the live in-the-building roster.

---

thpr (approved):
Resolved by `0b1c23f5`. The rule sits directly after the significance line it qualifies, tagged as a deliberate limit, and it picked up a second reason I had missed — on the board's own corrections the board would be telling itself. That one comes straight from the governing design and is worth having in the register.

One note, not blocking: the closing clause — "sending those would train everyone to ignore the alert and cost it the signal" — is the reason rather than the rule. No change can violate it, which is the `DOCUMENTATION_STANDARD.md` §3.9 test, and the `[Decision — deliberate limit]` tag already carries the same weight by definition. The three sentences before it are each violable on their own. Cutting the last one loses nothing; leaving it costs little.

## What changed underneath this PR

`#1560` landed the corrections route **and page** on `main` while this branch was in review, so the screen this work was gated on now exists:

- `checkin-app/src/app/api/facility/corrections/route.ts` — `flagged=true` filters on `newData.significance.flagged`, pinned to `tableName: 'Visit'`
- `checkin-app/src/app/facility-ops/corrections/page.tsx` — the flagged-only toggle initialises to `true`

So the page opens on the flagged view by default. Because `deleteSignificance` flags unconditionally, that view now contains every roster Absent, every conflict resolution and every board visit delete, alongside the member self-corrections that were previously all it held. With a 500-row cap and newest-first ordering, a busy month of routine lead removals can push real member corrections past the cut.

That is not a defect in this PR — the scoring is right, and the register now states the split correctly, which is what makes the consequence legible at all. Whether the screen's default should be flagged-only or flagged-plus-self belongs on #1258; the route already has the hook, with `kind`/`actorClass` filtering stubbed above the map.

Approving.

---
_Posted by Claude._

thpr (changes requested):
Still under review - claude is reckless

thpr (changes requested):
The core change is right, and I checked it rather than taking the body's word for it: enumerating every `tableName: "Visit"` audit writer gives five EDIT/DELETE sites, all five now score, and the three CREATE sites correctly don't. All five hand-computed scores re-derive against their fixtures (20, 60, 720, 1080, and the two zeros). The in-lock `previous` swap in `facility/visits` PATCH and the reopen guard in `events/[id]` are both real bug fixes on their own.

Four things to change. Two are small enough to land here; two are follow-ups.

---

### 1. `facility/visits` PATCH still stamps `WEB` — the same defect this PR fixes one file over

`checkin-app/src/app/api/facility/visits/route.ts:92-93`:

```ts
...(parsedArrived ? { arrivedAt: nextArrived, arrivedVia: "WEB" } : {}),
...(parsedDeparted ? { departedAt: nextDeparted, departedVia: "WEB" } : {}),
```

The reasoning this PR adds to `events/[id]` — that a time a staff member typed is staff-asserted, not the member's self-report — applies here unchanged. `WEIGHTS` defines `WEB` as "the member's own prior self-report" (weight 1), and `facility/trends` counts `WEB` as measured hours.

So after this merges the two staff paths disagree on identical input: a lead-typed arrival drops out of trends and weighs 2 on the next correction; the same arrival typed by the board stays in trends and weighs 1. That divergence is created by this PR landing the fix on only one of them.

Worth doing here, since it is the same one-line change already made next door.

### 2. `deleteSignificance` scores 0 for an open visit

`checkin-app/src/lib/visit/significance.ts:88`:

```ts
const duration = visit.departedAt
    ? minutesBetween(visit.arrivedAt, visit.departedAt)
    : 0;
```

An open visit yields `{ score: 0, flagged: true }` — rendered as `Flagged · 0`.

§2 of the design doc makes the score the strength: its table reads `delete a LEAD_MARKED or SCANNER visit | high | yes — strong`, and the floor supplies only *whether*. For an open visit the strength is zero, so a deleted live scanned check-in is indistinguishable from a trivial deletion on the one axis the review screen sorts and scans by. The delete table never contemplates an open visit; the code resolves that unspecified case as "nothing was lost".

The asymmetry is sharp. A person badges in and is on site six hours. Delete it now: 0. Wait for the kiosk checkout and delete the identical row: 360 × 3 × 2 = 2160. And an open visit is the live roster of who is in the building — this same PR's `events/[id]` path refuses an Absent mark on one for exactly that reason — so the path scoring lowest is the one destroying the most safety-critical record.

The new concurrency test states the invariant precisely:

> Scoring the pre-lock row instead sees `departedAt: null` -> duration 0 -> score 0, an audit row claiming nothing was lost while its own `oldData` shows a 3h visit.

That is the defect. The PR fixes the racing form of it and leaves the case where the visit is genuinely open, where `oldData` shows a six-hour-old arrival and `newData` says 0 with no race required. It is reachable on `facility/visits` DELETE and `my-programs/conflicts/resolve` (an open visit is a legitimate conflict half), and pre-existing on `attendance/manual/[id]`, whose board email reads "(significance 0)".

The fix is to score an open visit against now — time on site is the magnitude — with `now` injected so the fixed-offset assertions stay exact:

```ts
const duration = minutesBetween(visit.arrivedAt, visit.departedAt ?? opts.now ?? new Date());
```

I am **not** asking for that here: it changes `attendance/manual/[id]` scores too, and it is a semantics change to weights the helper's own docstring calls "v1 defaults awaiting board calibration". It wants its own issue.

What I am asking for here is the test. `adminVisitsAPI.integration.test.ts:383` pins the 0 with a comment reading it as intended:

```
// A delete always flags (the floor) even though the visit never
// closed, so there's no duration to weigh — score is 0.
```

A green assertion documenting the questionable value as correct is what makes it hard to revisit later. Record it as unresolved and link the issue.

Lower priority, same shape: a visit with both ends machine-closed gets `weight = max(0, 0) = 0`, so it also renders `Flagged · 0`.

### 3. Three new `newData.type` values, none in the UI's label map

`staff_correction`, `staff_removal` and `conflict_resolution` are absent from `TYPE_LABEL` in `checkin-app/src/app/facility-ops/corrections/page.tsx:40`. `kindLabel` falls back to "Edit"/"Delete", so nothing breaks — but on the screen this PR exists to make honest, routine duplicate-conflict cleanup now renders identically to a board removal, and the `type` the writers went to the trouble of stamping is invisible. Three lines.

### 4. Drop "raise"/"raised" from both docs — it now means two opposite things

The design doc's §"Which paths raise the live flag" uses *raise* for the board email. The new `docs/rules/attendance-checkin.md` bullet uses *raised* for scored-and-shown, and "the board being told at the time" for the email. Same word, opposite referents, and the register is what future readers cite against the design doc.

The verb should name the action: email, or show on the review screen. Nine sites.

`checkin-app/docs/designs/1256_ATTENDANCE_CORRECTION_SURFACE.md`:

| Line | Now | Suggested |
|---|---|---|
| 292 | `A flag is raised **after** the edit lands` | `A change is scored **after** the edit lands` |
| 389 | `### Which paths raise the live flag` | `### Which paths email the board` |
| 394 | `deliberately do **not** raise it, for two reasons:` | `deliberately do **not** email, for two reasons:` |
| 437 | `with the flag raised quietly behind it` | `with the score written quietly behind it` |
| 581 | `This *is* the "raise it to the board" surface` | `This *is* the board-review surface` |

581 needs nothing further — its next clause already does the work ("the write-time notification (§2) is the push, AT12 is the pull").

`docs/rules/attendance-checkin.md`:

- **144** — `Every removal is raised whatever it overwrote` → `Every removal shows on the review screen whatever it overwrote`
- **146-148** (this PR's new bullet) — `weighed and raised on the same terms` → `weighed on the same terms`; `Being raised and the board being told at the time are separate` → `Showing on that screen and the board being emailed at the time are separate`; `reaches the board as it happens` → `emails the board as it happens`
- **156** — `so correcting one is never raised however large the correction` → `so correcting one adds nothing to the score however large the correction`

156 deliberately does not gain "nor shows on the screen": the source suppression zeroes only that field's term, so a simultaneous arrival move can still carry the row over threshold. "Never raised" was already imprecise there, and dropping the verb without replacing the claim is the honest fix.

---

To be explicit about one thing I checked and am **not** raising: the change to what the review screen displays is in spec, not a scope overrun. §AT12 on `main` already says the flag is read from the persisted `newData.significance`, that a persisted value is the only form the database can filter on, and names #1523 as the requirement this places on the writers. Notifications are untouched — `flagBoard` still has its two call sites in `attendance/manual/[id]` and this PR adds none — so the "staff and lead paths do not email the board" limit holds exactly as before.

---
_Posted by Claude._

jee7s:
All four items addressed in `6249f2d1` — five files, +70/−23. CI is 9 pass + 1 skipped `label`, the same baseline as before.

## 1. `facility/visits` PATCH — now `LEAD_MARKED`

`route.ts:92-93`. I checked the premise first: `WEIGHTS` does define `WEB` as the member's self-report at weight 1, and `facility/trends` excludes only `arrivedVia IN (LEAD_MARKED, SYSTEM)`, so a `WEB` arrival was counting as measured hours. The divergence you described is real.

On the value — `VisitSource` has no board-specific member, and `LEAD_MARKED` is its only staff-observation value.

On the shape, I did not copy `events/[id]`'s update branch literally. "Never restamp `arrivedVia`" is right there because that path also adopts walk-ins, but on a PATCH-only route it would mean a staff-asserted arrival could never be recorded at all. This route already encodes the discrimination more precisely: it writes a field only when the caller supplied it, so a side not sent keeps its existing `via` — the same property "an adopted walk-in keeps its `SCANNER`" protects. So the conditional stayed and only the value changed.

What convinced me: `attendance/manual/[id]` uses the identical conditional-spread with `WEB`, and there `WEB` is correct, because that is the member path. The staff route had inherited the member path's spelling.

Red/green: against the pre-fix file, `adminVisitsAPI` fails 2/24 — the new assertion plus the pre-existing `departedVia` one — with `Expected: "LEAD_MARKED" / Received: "WEB"`. 24/24 with the fix.

## 2. Open-visit zero — recorded, not fixed

`significance.ts` is untouched, as you asked. The comment at `adminVisitsAPI.integration.test.ts` no longer reads the 0 as intended; it records the value as unresolved and carries `TODO(#TBD)`.

The issue is written but **not filed** — I have no issue-write scope. Full title and body are in the PR description under "NOT YET FILED", carrying the 0-vs-2160 asymmetry, the live-roster argument, your `opts.now` shape, all three reachable sites, and the both-ends-machine-closed note. The `#TBD` needs swapping once it has a number.

## 3. `TYPE_LABEL`

Three entries added in `facility-ops/corrections/page.tsx`. Worth knowing: no test anywhere renders that page, so this is unverified beyond typecheck. It is pure data.

## 4. The `raise` sweep

All nine sites, your replacements verbatim. `grep -i rais` now returns nothing in either document. Line 156 deliberately did not gain "nor shows on the screen", for the reason you gave — the source suppression zeroes only that field's term.

One coupling to be aware of if either line is edited later: 144's "shows on **the review screen**" is what gives 149's "Showing on **that screen**" its antecedent.

**§3.9 note — I cut it.** The bullet now ends at "on the board's own corrections the board would be telling itself."

## Verification

The worktree's `node_modules` turned out to be a symlink to another checkout; replaced with a real `npm ci` before trusting any result. Postgres 15 in Docker.

`npm run test:ci` 278 suites / 2710 tests green · seven related integration suites 83 tests green · `tsc --noEmit` clean · `lint --max-warnings 0` clean · `git diff --name-only origin/main...HEAD -- checkin-app/src/security/` empty.

## Found, deliberately not touched

- **`facility/visits/insert/route.ts:67-68`** stamps `arrivedVia`/`departedVia` as `WEB` on a staff insert for another person — the same defect class as item 1, pinned by a test named "creates a closed WEB visit for another person". Not in your review, so left alone.
- **The PR body's "Adjacent defect, not fixed here" paragraph is now stale.** It says the reopen case is unchanged, but `47e43550` later added the reopen guard you credit above. I left that narrative alone rather than editing text you had already read.

---

thpr (changes requested):
6249f2d1 addresses all four. Checked each rather than taking the message's word: the `via` is stamped per field with a test pinning that an unsent side keeps its `SCANNER`, the open-visit assertion no longer reads 0 as intended, `TYPE_LABEL` gains the three values, and neither doc has a `rais` left in it.

Two things that fix surfaced. Both are consequences of moving the PATCH to `LEAD_MARKED`, so they belong with it rather than after it.

### A. `facility/visits/insert` still stamps `WEB`

`checkin-app/src/app/api/facility/visits/insert/route.ts:67-68`:

```ts
arrivedVia: "WEB",
departedVia: "WEB",
```

This is the walk-in insert — the board typing *both* ends of a visit no machine measured, for another person. It is more staff-asserted than the PATCH this commit just fixed, not less, and it now disagrees with its sibling in the same route family: an arrival the board corrects is `LEAD_MARKED`, an arrival the board invents is `WEB`. The existing test names the behaviour outright — "creates a closed WEB visit for another person" (`adminVisitsAPI.integration.test.ts:446`).

The consequences are the two this commit message already argues are wrong: the visit counts in `facility/trends` as measured hours, and it weighs 1 — "the member's own prior self-report" — on the next correction, for a time the member had no part in.

### B. Correcting an arrival now drops the visit out of trends

`facility/trends` excludes `LEAD_MARKED` outright (`route.ts:80-81`), so with the PATCH restamping, a board member fixing a bad `SCANNER` arrival removes that visit's hours from the trend entirely. Before this commit the `WEB` stamp kept it counted. A departure-only correction is unaffected, since `arrivedVia` is left alone.

That is the hazard `events/[id]` already names in its own comment — "trading an over-count for an under-count on exactly the people who did scan in" — reached here by correcting a time rather than by marking a roster. The person did badge in; the correction made their real hours stop counting, which is the opposite of what a correction is for.

The parity I asked for in the first review is real, but it has been reached by making corrected visits vanish from the hours rather than by counting them correctly. I do not think that is the trade you want, and the enum cannot currently tell a roster-window `LEAD_MARKED` from a precisely-corrected one — which is the actual root of it, and may be what needs resolving.

Nothing covers this: the trends suite builds its fixtures directly and never goes through PATCH, so both the old behaviour and the new one pass it.

### Not blocking

`TODO(#TBD)` in `adminVisitsAPI.integration.test.ts` dangles on an issue that does not exist yet, and AGENTS.md's comment rules exclude TODOs and PR numbers from source. I will file the issue from your PR-body text and post the number here; the UNRESOLVED sentences above it carry the meaning on their own once the TODO line goes.

The register bullet also lost its reason in the rewrite — "sending those would train everyone to ignore the alert and cost it the signal" was the *why* behind the limit, and a `[Decision]` entry without its why is the thing the register exists to prevent. Worth restoring in the new wording.

---
_Posted by Claude._

thpr:
Filed the open-visit scoring issue as #1630, near-verbatim from your description — swap it into the `TODO(#TBD)` in `adminVisitsAPI.integration.test.ts`, and drop the "issue text is in PR #1558's description" line with it.

Also filed #1624 for what sits underneath this and finding A: `VisitSource` fuses *how* a time was captured with *who* asserted it, which is why no existing value is right for a staff-typed time. Your own "found, deliberately not touched" note on `facility/visits/insert` is the same fusion — I raised it as A before spotting you had already flagged it, so treat those as one item.

Neither issue blocks this PR.

---
_Posted by Claude._

thpr (approved):
Approving. Everything from both earlier reviews is addressed, and a2e6a586 went to the root cause rather than patching the symptom — the enum could not tell a roster window from a corrected time, so it now can.

Verified rather than taken on trust:

- All five EDIT/DELETE Visit audit writers score; the three CREATE writers correctly do not. All five hand-computed scores re-derive against their fixtures.
- The in-lock `previous` swap and the reopen guard are real fixes on their own.
- `WEIGHTS` is a `Record<VisitSource, number>`, so tsc enforces the new value; the other three consumers (`SOURCE_META`, `sourceLine`, the trends comment) are all updated. Nothing else in the app enumerates `VisitSource`.
- The migration is `ADD VALUE IF NOT EXISTS`, unwrapped — correct, since Postgres forbids using a value in the transaction that added it.
- **The drain-window argument holds.** I checked: `20260803000100_visit_source_split_backfill` is on `main` and rewrites `SYSTEM` to `LEAD_MARKED`/`AUTO_CLOSE`, values only added by the migration immediately before it. So the deployed release already cannot read those rows, and `STAFF_CORRECTED` adds no new exposure. Accurate as written.
- The new trends test drives the real PATCH route instead of hand-building `arrivedVia`, which is the only way it could have caught the bug it covers. Right call.

Two things are knowingly imperfect, filed rather than held against this PR:

- **#1631** — correcting a roster-marked visit's *arrival* still promotes it into facility hours, because trends reads `arrivedVia` alone and per-field writes overwrite it. Correcting the departure does not. Pre-existing on `main` (`WEB` passed the filter too); this PR did not introduce it and does not close it.
- **#1632** — historical staff walk-in visits carry `WEB` and keep counting, while new ones carry `LEAD_MARKED` and do not, so the trend steps at the deploy. The rows are identifiable from the `type: "staff_entry"` CREATE audit rows, so a backfill is available; whether to run it is a reporting judgement, not a blocker.

Both fold into **#1624**, now the umbrella: existence, time-capture and actor are three separate facts sharing one per-field column, and the stamp on this route has now been `WEB`, `LEAD_MARKED` and `STAFF_CORRECTED` in three successive commits — each trading one error for the other. Settling what facility hours *measure* is what ends that.

Also open: **#1630**, the open-visit delete scoring 0 — swap that number into the `TODO(#TBD)` before merge.

This is better than what it replaces on every axis I checked. Ship it.

---
_Posted by Claude._

PR #1559, opened 2026-08-07
security(registry): register GET /api/facility/corrections (AT12, #1258)

jee7s:
Registry-first for AT12 (#1258), per the design approved in #1497. **One file, one `defineRoute` entry, no app code** — an unused entry is inert, and the isolation rule requires it to ship alone ahead of the handler.

The route it registers arrives in a follow-up PR.

## The entry

```ts
defineRoute({
    endpoint: 'GET /api/facility/corrections',
    authorize: { anyRole: ['isSysadmin', 'isBoardMember'] },
    envelope: null,
    returns: ['AuditLog', 'Person', 'Visit'],
    orderedView: [
        ['isSysadmin',    ['everyones:personal', 'everyones:internal', 'public']],
        ['isBoardMember', ['everyones:personal', 'everyones:internal', 'public']],
    ],
});
```

## Why each grant, since this is the PR that cannot be wrong

**`everyones:personal` is required, and omitting it was the bug thpr caught on #1497.** `Visit.arrivedAt` and `departedAt` are `personal` (`classifications.ts:281-282`); everything else the screen needs from `Visit` is `public`. Granting only `internal` + `public` returns the ids and the `arrivedVia`/`departedVia` provenance and **strips both timestamps** — the before/after column, which is the entire point of the screen, renders empty. `POST /api/facility/visits/insert` already grants `everyones:personal` to the same two roles.

**`everyones:internal` is required** because every `AuditLog` field is `internal`.

**`pii` is deliberately not granted.** Only `Person.name` is needed and that is `public`, so email cannot ship.

**`member` is deliberately not granted.** `ToolStatus.level` is the only `member`-tier field in the repo; nothing in `returns` carries it. The wave-1 registry review flagged unearned tier grants specifically, so this omits rather than copies the sibling route.

**`Visit` is in `returns`** because the handler synthesizes `Visit`-shaped entries carrying the four time-and-source values extracted from the audit blobs. Those are real `Visit` field names, so they classify as `Visit` data under an honest `personal` grant — which is what keeps the raw `internal` blobs, and the `personal` times hidden inside them, server-side.

## Why this passes CI while the route does not exist

- `orphan-registry` reports **warn**, not error, when a registered endpoint has no route file — `main()` exits non-zero only on `severity === 'error'`.
- `new-route-old-authz` cannot fire: it fires on route *files* absent from the registry, and there is no file yet.
- Both handler-driving suites skip it — `registryAuthz.integration.test.ts` via its `routeFileExists()` guard, `policy.contract.integration.test.ts` via its import `catch`.
- `deriveCtxNeeds` yields all-false (`anyRole` and `everyones` both contribute nothing), so `buildCallerContext` issues no extra queries and `ctxNeeds.test.ts`'s masking is a no-op.
- `validateRouteGrants` passes vacuously — its loop body is `if (scope === 'everyones') continue;` and every token here is `everyones`. Admission via `anyRole` is the entire subject boundary for this route.

## Note for the reviewer

**No registered route returns `AuditLog` today.** `GET /api/system-status/audit-log` predates the registry and is grandfathered in `legacy-authz-routes.txt`. This is the first, so the entry is worth more than a glance — and it is why the design settled the response shape before this PR rather than after.

---
_Posted by Claude._


thpr (commented):
Read the full boundary path against this entry — `core.ts`, `handler.ts`, `stripper.ts`, `scopes.ts` / `scopeBindings.ts`, the generated classifications, the four validator suites, `check-route-coverage.ts`, and the #1497 design doc.

Every claim in the entry and the PR body holds:

- All nine `AuditLog` fields are `internal`, so `everyones:internal` is the minimum that returns anything at all.
- `Visit.arrivedAt` / `departedAt` are `personal`; the rest of what the screen reads (`id`, `personId`, `arrivedVia`, `departedVia`) is `public`, and `deletedAt` / `deletedById` are `internal` and covered. Without `everyones:personal` the before/after column renders empty.
- `Person.name` is `public`, so withholding `pii` costs the view nothing.
- `ToolStatus.level` is indeed the only `member`-tier field in the repo, and nothing in `returns` carries it.
- `everyones` survives on the synthesized rows: `ROW_SCOPE_KEY` holds only `EmergencyContact`, so `makeScopesHeld` seeds `everyones` for `Visit` and `AuditLog` regardless of which columns the object carries.
- `deriveCtxNeeds` yields all-false — `anyRole` and every `everyones` token contribute `{}` — so the ctxNeeds equivalence test is a no-op here rather than silently passing for the wrong reason.
- The delivery guard is satisfied by `public` alone (`Person.id` / `name`), so it will not mask an over-narrow view later.
- No test carries a hardcoded registry-route count that this would need to bump; the only count assertion is on `SCOPE_BINDINGS`, which is untouched.

The grant is narrower than every sibling admin entry — the only one that drops both `pii` and `member`. There is no unearned tier here.

Two things that only fall out of reading the enforcement code, neither blocking.

**This entry forecloses one of the two options #1497 left open.**

The design says `before` "needs either a separate bag key or a documented convention, and that is the one piece the route PR still has to design." A separate bag key is not expressible. `stripBag` (`stripper.ts:29-32`) drops any top-level key that is not a model name, so the bag's keys are a subset of the 76 model names and there can be exactly one `Visit` key. The convention is the only surviving option, and because a before row and its after row carry the same `Visit.id`, no field can distinguish them — array position is the only discriminator, and it is the one that crosses the boundary intact, since `stripValue` maps arrays element-wise and preserves order. Worth stating in this comment rather than leaving it for the route PR to rediscover: this entry is what closes the door.

**Both handler obligations the comment states are unenforceable, and nothing fails if the route PR forgets them.**

"`AuditLog.newData` is REBUILT by the handler to `{ type, significance }`" and "raw `oldData`/`newData` must never leave the route" are real invariants, but `everyones:internal` permits both fields wholesale and field tiers do not reach inside a JSON blob. `policy.contract.integration.test.ts` asserts tier stripping, not blob contents, so a handler that passes `oldData` straight through ships a whole-row snapshot and every suite stays green. That is not fixable at the registry — it is the gap the design's option (1) analysis names — but the route PR should carry an explicit assertion on the response shape rather than inherit the invariant from a comment.

---
_Posted by Claude._

thpr (approved):
Both comment corrections verified. The `defineRoute` call is byte-identical across `448adaf5` — the diff hunk sits entirely above it, so no field, role, tier or token moved.

Re-checked the entry against the merge of `main`. It brought two new fields into this route's declared surface, `AuditLog.actorSystem` and `Visit.forceCloseWarnedAt`, both `internal` and both already covered by `everyones:internal`. No widening, and the justification for the grant still reads true field-for-field. `registry.ts` itself was untouched by the merge — diffing `origin/main` against the branch leaves this entry as the only delta, so nothing was dropped in the resolution.

One thing for the route PR rather than this one, surfaced by that merge. `lib/auditActor.ts` makes `actorId: 0` the sentinel for a system write, and person 0 does not exist, so such a row would fail `actorId === secondaryAffectedEntity` and bucket as staff-or-proxy with no `Person` to name it. Not reachable today: both `systemActor()` writes on the attendance side use `tableName: 'SYSTEM_NOTIFY'`, and the nightly `AUTO_CLOSE` path audits no `Visit` row at all, so nothing under the screen's `tableName: 'Visit'` filter carries actorId 0. Worth holding onto because #1523 is scoped to *every* Visit audit write — if that sweep pulls an automated path into Visit auditing, the actor axis mislabels silently rather than failing.

Approving. The grant is the narrowest of any admin entry in the registry — the only one that drops both `pii` and `member` — and it is inert until its route lands.

---
_Posted by Claude._

PR #1560, opened 2026-08-07
feat(facility-ops): attendance corrections review route + page (AT12, #1258)

### LINKED ISSUE #1258: Admin hour-correction review screen

jee7s:
Closes #1258. Implements the AT12 correction-review screen per the design approved in #1497.

**Draft, and deliberately so — this cannot go green until #1559 merges.** See "Why this is red" below. Opened now so #1258 has its implementation link rather than reading as unclaimed.

## Sequence

`#1558` (#1523, the significance gate) → `#1497` (design merges) → **`#1559` (registry entry, alone)** → rebase this onto `main` → this.

## What it builds

- `src/lib/timePeriods.ts` — extracts `PeriodType` and `getPeriodStart` out of `facility/trends`, which is the one correct org-timezone bucketing implementation in the repo. `formatPeriodLabel` deliberately stays behind: its `locale` parameter is required, so it does not typecheck on a client caller.
- `src/app/api/facility/corrections/route.ts` — the handler.
- `src/app/facility-ops/corrections/page.tsx` — the screen, plus `facilityNav` and `pageRegistry` wiring under the existing `BOARD` predicate.

## The response shape, which is the part worth reviewing

The design settled on a **synthesized bag**. `stripValue` copies any field present on the object and never checks where the value came from, so the handler builds `Visit`-shaped entries carrying the four time-and-source values **extracted from** the audit blobs. Those are real `Visit` field names, so they classify as `Visit` data under the honest `everyones:personal` grant — while the raw blobs never leave the handler.

Concretely:

- **`oldData` is never copied.** Not to the client, not into any intermediate.
- **`newData` is rebuilt**, not passed through: `{ type, significance }` only. This corrects the design's own §4, which called `newData` "the small blob" — it is not. `facility/visits` PATCH writes a whole `Visit` row into both blobs, so shipping `newData` whole would route `personal` times through the `internal` grant, which is the exact defect §4 rejects option 1 for.
- **`before`/`after` are a nested `[before, after]` pair** under the `Visit` key. `stripValue` hits `Array.isArray` first and re-enters with the same model name, so the pair walks element-wise as `Visit` and each element is stripped normally. This was the one piece §4 left to the route PR.

## No pagination in v1, and why

`total`, `page` and `pageSize` have no legal home: `stripBag` drops non-model top-level keys, and the strip runs before the envelope wraps. So the route takes a date range with `take: MAX_ROWS + 1` — the extra row is the "there are more" signal — and the client buckets what it receives.

## One deliberate rule departure, commented at the site

The `Person` lookup is **not** `LIVE_PERSON`-filtered. An audit row's actor or subject may since have been merged away, and the drill-down must still name them — design §1 requires this. `mergedIntoId` is selected so the UI can mark the row.

Worth knowing for whoever touches it: that query currently satisfies `livePersonDriftGuard` only because the guard's regex matches `mergedIntoId` anywhere in the captured argument. It passes by coincidence, not by decision. Narrowing the `select` would silently trip the guard, so the comment says to keep the field.

## CI status

The last run failed at **step 1, `npm run lint`** — `react-hooks/set-state-in-effect` on `facility-ops/corrections/page.tsx`, fatal under `--max-warnings 0`. That is now fixed by deriving the loading flag from what was last fetched rather than setting state in the effect body.

The earlier diagnosis in this body was wrong. `check-route-coverage` did not fail, because it never ran: the unit job's step order is Lint → Generate Prisma Client → Type check → Verify security classifications → **Check route coverage** → **Unit tests**, and lint aborts the job before any of them. Typecheck, the classification check, route coverage and the entire unit-tier security suite were therefore **unverified on this branch**, not passing.

**What will fail next, and why it is expected.** With lint green, route coverage becomes the first gate this branch can reach, and it fails with exactly one error:

```
✗ [new-route-old-authz] GET /api/facility/corrections is not in the security registry
```

That is #1559's registry entry, which ships alone by boundary-isolation rule and is not merged. Applying #1559's entry locally takes `npm run check-route-coverage -- --strict` to **0 errors, exit 0**. This PR cannot go green until #1559 lands on `main`; that is the designed sequence, not a defect here.

## Verified locally

- `npm run lint` → exit 0, clean.
- `npm test` → **2571/2571 tests pass**, 269/271 suites. The 2 failing suites (`s-read/diagnose`, `s-read/sync`) fail to *load* on `Cannot find module '@aws-sdk/client-lambda'` — declared in `package.json`, absent from the shared `node_modules`. Environmental, unrelated files.
- `npx tsc --noEmit` → only those same 2 errors.
- Route coverage both ways: without #1559's entry, 1 error and exit 1; with it applied, `--strict` clean.
- Registry entry reverted before commit — `git diff --name-only origin/main...HEAD -- checkin-app/src/security/` is empty. Boundary isolation intact.

**Not verified: the 10 integration tests never ran.** No database is reachable in this environment. The `500 [security] No registry entry` failures are *explained*, not disproven — the handler's data correctness is still unproven until #1559 lands and CI runs the suite.

## Design obligations discharged

§7 of `ATTENDANCE_CORRECTION_REVIEW.md` assigned three things to this PR, all now done: the design doc is deleted, the `secondaryAffectedEntity` invariant is extracted into `docs/rules/attendance-checkin.md`, and all three stale §4 claims in `1256_ATTENDANCE_CORRECTION_SURFACE.md` are corrected.

A fourth correction was needed and is included: the same sentence carrying the bad gate also specified "a read-only aggregation route … grouping by kind × actor-class × period … rendering counts". §4 of the governing design rejects exactly that shape — `stripBag` drops any key that is not a model name, so `buckets`/`total`/`page` cannot cross the boundary, and §4's option 2 (a separate counts endpoint) is rejected for the same reason. Leaving it standing would keep pointing implementers at a shape the boundary forbids.

Eight code comments citing the deleted doc as "design §1/§2/§3/§4" were retargeted — those numbers mean different things in the governing design, so the deletion would have silently turned them into wrong pointers.

## On aggregation

A total count is now shown (`500+` past the truncation cap, beside the existing notice). Buckets and a drill-down are deliberately absent: the first is unbuildable per §4 above, and the second has nothing to display — `WEIGHTS` and `FLAG_THRESHOLD` in `src/lib/visit/significance.ts` are module constants with no `BoardSettings` read, so "today's thresholds" are identical to the ones every persisted row was scored under. §3's payoff ("when the two disagree, the detail view shows both") is unreachable until threshold calibration lands, which §5 puts out of scope.

---
_Posted by Claude._


thpr (commented):
Traced the diff against `main`: the handler/stripper path, `scopeBindings`, every `tableName: 'Visit'` audit writer, the LIVE_PERSON drift guard, and the nav/registry wiring.

## The claims that needed checking, and hold

- **The nested `[before, after]` pair walks element-wise as `Visit`.** `stripValue` hits `Array.isArray` before the object branch and re-enters with the same `modelName`, so each element strips as a `Visit` row. Correct.
- **`Visit` is absent from `ROW_SCOPE_KEY`,** so `scopesHeld` seeds `everyones` unconditionally — the `ponytail:` caveat about a null `personId` is accurate, and in practice today's picks mostly carry no `personId` at all (only `facility/visits` PATCH/DELETE write whole rows into the blobs).
- **The three-key bag dodges the single-key unwrap** in `handler.ts`. The comment earns its place.
- **`secondaryAffectedEntity` is the subject on all ten `tableName: 'Visit'` writers,** so `actorId !== secondaryAffectedEntity` is a sound proxy test.
- **The LIVE_PERSON guard.** The PR body calls the pass a coincidence. It isn't, quite: `FILTERED_RE` is `/LIVE_PERSON|mergedIntoId/`, and selecting `mergedIntoId` in order to render "(merged)" *is* deliberate tombstone handling, which is what the guard tests for. It's also the only workable arrangement — the guard's third assertion (no stale allowlist entry) would fail on an ALLOWLIST entry for a site that isn't flagged. Worth softening the comment rather than leaving a note that reads as "this is load-bearing by accident".

## Findings

**1. `MAX_ROWS` is duplicated, and drift is silent and lossy.** The route and the page each hold `500` independently, bound only by a comment. Lower the server's cap alone and `allRows.length > MAX_ROWS` never fires: the overflow banner disappears while rows are still being dropped, so a truncated list renders as a complete one. `@/lib/timePeriods` is already the shared home for this pair — export the constant there and import it on both sides.

**2. A whole class of correction renders as a blank row.** `events/[id]/attendance/route.ts` (lines ~120 and ~150) writes `newData: { participantId, associatedEventId, synthetic }` — no `type`, no `arrivedAt`/`departedAt`, and no `oldData` at all on the EDIT. Through this screen that becomes: Kind falls back to `action` ("Edit"/"Insert"), `pickVisit` yields `{ associatedEventId }`, and **both Before and After render "—"**. A lead marking a no-badge attendee Present writes a real placeholder visit window (`LEAD_MARKED` on both via fields, a synthetic arrival→departure span), and the review screen shows it as though nothing changed.

That's a writer-side gap rather than a defect in this diff, and the "absent key means never touched, render —" rule is doing exactly what it was designed to do. But it's the sibling of what #1558 is already fixing on those same two lines, and it leaves the screen structurally blind to lead-marked attendance. Worth naming now rather than discovering it from the empty column.

**3. The refetch has no generation guard.** The effect keys on `period` and `flaggedOnly` with no abort or stale-response check, so fast toggling can land an out-of-order response and the stale table sticks until the next change. Minor, and the fix is a captured flag in the `.then`.

## Worth knowing, not a blocker

`AuditLog` has no index — none in `schema.prisma`, none in any migration. `where tableName = 'Visit' AND timestamp BETWEEN … ORDER BY timestamp DESC LIMIT 501` therefore sequentially scans and sorts the app-wide audit table on every load of a board screen. This is pre-existing (`/api/system-status/audit-log` has the same shape), so this PR only adds a second consumer of the cost. `@@index([tableName, timestamp])` is the obvious follow-up, in its own migration PR.

docs/rules/attendance-checkin.md
```diff
@@ -110,6 +110,13 @@ Things the app takes as true because they are handled outside it.
   reopened — and whose record it is. Integrity is after the fact rather than a
   gate: every change is audited, and a significant one is flagged to the board.  [Decision — *Principle: self-scope and repair*]
 
+- Every write that audits a visit stores the subject person — the one whose
+  attendance changed — in the audit row's secondaryAffectedEntity field, never the
```
thpr: Field names should not be in the domain rules

thpr (changes requested):

thpr (commented):
Reviewed the three commits since my last pass: the merge from `main`, the docs retirement, and the page fix.

## New defects

**1. The derived loading flag can latch on permanently.** `.finally(() => setLoadedQuery(query))` captures the *effect's own* `query`. When a slower earlier request settles after a newer one, `loadedQuery` ends on the old query while `query` holds the new one, so `loading` stays true with no fetch pending. `rows.length !== 0` at that point, so the `loading && rows.length === 0` spinner branch never renders: the table shows the previous query's rows beneath the new count line, and nothing recovers until the user touches a filter again.

The same race existed before this commit, but it settled `loading` to false and left only stale rows. Deriving the flag turns it into a stuck state, and the new count line turns those stale rows into a stated wrong number. A cleanup flag guarding all three setters fixes both halves:

```ts
useEffect(() => {
  if (!ready) return;
  let live = true;
  fetch(`/api/facility/corrections?${params}`)
    .then((res) => { if (!res.ok) throw new Error("request failed"); return res.json(); })
    .then((body: Payload) => { if (live) { setData(body); setError(""); } })
    .catch(() => { if (live) setError("Failed to load corrections."); })
    .finally(() => { if (live) setLoadedQuery(query); });
  return () => { live = false; };
}, [ready, period, flaggedOnly, query]);
```

**2. The error path asserts a wrong count.** On a failed fetch `data` keeps the previous payload, and `finally` still sets `loadedQuery = query`, so `loading` goes false and the count line renders the **old** rows' length carrying the **new** period word — directly under "Failed to load corrections." A stale number presented as this period's total is worse than no number. Clear `data` on error, or suppress the count while `error` is set.

**3. The docs commit records a route that this branch's own merge deleted.** The merge brought in #1553, which deleted `events/[id]/attendance/route.ts`. The docs commit landed after it and still cites "the events-attendance roster mark stores no times at all" as one of the two row shapes with nothing to recompute from, still lists events-attendance among the paths that land in the audit trail, and `ScoreBadge`'s comment still describes "the events-attendance roster edit, which moves no clock." No live writer has that shape any more.

The conclusion — persist significance rather than recomputing it — survives on the remaining shape (`facility/visits` edits written before the `oldData` fix). The evidence for it doesn't, and this doc is now the permanent home for that reasoning, so it's worth correcting before it sets.

## Landed in the merge, and worth a look

**`AuditLog` gained `actorSystem`, with `SYSTEM_ACTOR = 0` as the actorId sentinel** (`lib/auditActor.ts`). The route adds `r.actorId` to the person batch unconditionally, so a system-written `tableName: 'Visit'` row would resolve to no Person and render as Actor "Person #0" with an orange **Proxy** badge — an automated sweep displayed as one person editing another's record. Nothing writes that row today (the nightly force-close files under `SYSTEM_NOTIFY`), so this is latent rather than live. But the route now sits next to a sentinel it doesn't know about, and skipping `actorId === SYSTEM_ACTOR` in the batch — letting `actorClass` fall to "unknown", or labelling it System — is a two-line guard against a wrong answer that would otherwise arrive silently.

**The same merge added `@@index([actorId])` and `@@index([actorSystem])` to `AuditLog`.** Neither serves `where tableName = 'Visit' AND timestamp BETWEEN … ORDER BY timestamp DESC`, so the gap I raised last time is now a conspicuous omission rather than a general absence of indexes on the model.

## Where my earlier findings stand

- **The blank-row class is resolved,** by that same #1553 deletion. Every remaining `tableName: 'Visit'` writer carries times, or is a DELETE where an empty "after" is the right answer.
- **The `MAX_ROWS` duplication is unaddressed,** and the stakes went up: the count line now reads `MAX_ROWS` to choose between "500+" and an exact number, so drift between the two copies produces a wrong figure rather than just a missing banner.
- **The refetch race is unaddressed,** and finding 1 above is what the loading change did to it.

thpr (commented):
Reviewed `816dbb78`.

## Fixed, and correctly

- **The latching loading flag.** The `live` cleanup is the right shape: React runs cleanup before the next effect, so a superseded response sets nothing at all — not `data`, not `error`, not `loadedQuery`. The latch is gone, and the earlier stale-rows race with it.
- **The count on the error path,** suppressed via `{!error && …}`.
- **The references to the deleted roster-mark route,** gone from all three sites — the `ScoreBadge` comment, the design's Source list, and the two-shapes sentence. The extracted rule reads better without the column name, too.

## The error path still asserts something it cannot know

The fix has two halves, and the second reopened the defect one element lower. `.catch()` now calls `setData(null)`, so on a failed fetch `rows` is `[]` and `loading` is false — which sends the render into the else branch, where the table body hits `rows.length === 0` and prints **"No corrections in this range."** directly beneath the red "Failed to load corrections."

That is the same unfounded assertion this commit removed from the count line. Its own words — "a total is an assertion, and there is nothing to assert" — apply unchanged to an empty state. And in substance it is a step back from where the previous commit stood: the error path used to show stale rows, which at least look like data; it now tells a board member there were no corrections in a period it never successfully read. Gate the table, or at minimum its empty-state row, on `!error` alongside the count.

## The replacement doc claim does not hold either

The two-shapes sentence now reads "`facility/visits` edits written before the `oldData` fix stored no before-state, and a `DELETE` has no after-state by construction."

`deleteSignificance` (`lib/visit/significance.ts:86`) reads *only* the before-state — `arrivedAt`, `departedAt`, `arrivedVia`, `departedVia` — plus `byProxy`, which is exactly the actor-versus-subject comparison this screen already performs. Every DELETE writer puts those four fields in `oldData`. A delete is therefore the shape that most clearly *can* be recomputed at read time, not one that cannot.

The argument does not need it. The pre-fix `facility/visits` edits genuinely carry nothing to recompute from, and the "only a persisted flag is filterable in SQL" point stands on its own. But that one sentence has now been patched twice with a second example that does not survive checking, in the document that is the permanent home for the reasoning — worth either dropping the second example or replacing it with one that holds.

## Still open from the earlier passes

The `MAX_ROWS` duplication between route and page (the route is untouched here), the `SYSTEM_ACTOR` guard on the person batch, and the missing `(tableName, timestamp)` index.

thpr (commented):
Covers the two commits since my last review — `700cbdb1` (the dedupe and the system actor) and `930dd491`.

## Landed correctly

- **`MAX_ROWS` unified** in `lib/corrections.ts`, with `PeriodType` deduped alongside it via `import type`, so the page picks up no `date-fns-tz` bundle weight from `timePeriods`.
- **The `SYSTEM_ACTOR` guard** is right on both halves: the route drops the sentinel from the person batch, and the page gives it a System badge that sits off the self/proxy axis rather than defaulting to a side. `actorSystem` is `internal` in the generated classifications, so it is already inside the granted band and needs no registry change — that reasoning checks out. So does the claim that nothing writes a system `tableName: 'Visit'` row today: every `systemActor()` call site files under another table.
- **The error empty-state and the `DELETE` recompute claim** are both properly fixed.

## New

**1. The design doc now says "Two row shapes" and lists one.** The second example came out, the count did not, so the sentence reads as a dropped clause rather than a deliberate revision — in the document that is now the permanent home for the reasoning. Worth either restating it as one shape or restoring a second that holds. (The same edit also left one line running well past the wrap width the rest of the section keeps.)

**2. The `!error` gate swallows the loading indicator along with everything else.** `error` is cleared only inside the success `.then`, never when a request starts. So after a failed fetch, switching period leaves `error` still set while `loading` is true — and the new outer gate hides the `PageLoader` too. The user gets the stale red "Failed to load corrections." and nothing else for the full duration of the retry, with no indication that anything is in flight.

Moving the gate into the else-branch keeps the intent and restores the spinner. `rows` is `[]` on error, so a retry hits the loader branch first:

```tsx
{loading && rows.length === 0 ? (
  <PageLoader />
) : !error ? (
  <>{/* count + table */}</>
) : null}
```

**3. The system-actor branch ships without a test** (from `700cbdb1`). This route already has an integration suite carrying a case per response-shape decision, and the new branch is exactly that kind of decision: the sentinel excluded from the `Person` batch, `actorSystem` riding in the `AuditLog` entry, the row on neither side of the self/proxy axis. Seeding one row at `actorId: 0, actorSystem: 'cron:nightly'` is about ten lines, and it is the only thing that would catch the sentinel silently re-entering the batch later.

## Still open

The missing `(tableName, timestamp)` index on `AuditLog` — `main` has since added `@@index([actorId])` and `@@index([actorSystem])`, neither of which serves this route's access path.

dkaygithub (approved):

thpr (approved):
Approving. Covers `b4ba8312` and the two merges from `main`.

## All four items fixed, and correctly

- **The loader gate.** Traced every state: initial → loader; refetch with rows already on screen → stale table, no flash; error idle → alert only; error then a period switch → loader. That last one was the broken case, and it is fixed.
- **"One row shape."**
- **The index.** `@@index([tableName, timestamp])` with a hand-written additive migration. The index name matches Prisma's own default (`AuditLog_tableName_timestamp_idx`), so `migrate diff` sees no drift; the migration timestamp sorts newest; and `DEPLOY_MIGRATION_ORDER_OF_OPERATIONS.md` puts new indexes squarely in the "may ship in the same release as the code that starts using them" class. Nothing to flag.
- **The sentinel test** pins both halves — excluded from the person batch, `actorSystem` present in the entry, subject still resolving.

## Two follow-ups, neither blocking

**The merged registry entry now says something false.** #1559 landed carrying this above the `defineRoute`:

> `// GET /api/facility/corrections does not exist yet. Blocked on #1523 (persist significance on every Visit audit write) — the review is only meaningful once every write scores.`

This PR makes the first clause false, and it cannot be fixed here: the isolation workflow requires that any PR touching `src/security/**` contain nothing but tests, docs, schema annotations, or the workflow itself, so even a comment-only edit to `registry.ts` would force this whole feature diff to be boundary-only. It needs its own comment-only PR — worth doing rather than leaving a false statement standing in a CODEOWNERS-gated file whose purpose is to be audited. I am spinning that up separately.

**The second clause is still true, and it is a merge-ordering call rather than a defect.** #1558 (#1523) is open, so significance is written by `attendance/manual/[id]` alone. Merging this first ships a screen whose default flagged-only view shows self-corrections and nothing else — the registry author's own stated precondition, unmet. Worth making that an explicit decision about what v1 means rather than a side effect of review order. (The PR body's Sequence section reads as stale alongside it.)

*(Minor: the new test seeds `type: 'facility_close'`, a string no writer produces and which is not in `TYPE_LABEL`. It exercises the fallback correctly, but it coins a vocabulary term — `self_correction`, or no `type` at all, would say the same thing without inventing one.)*

PR #1563, opened 2026-08-08
docs(rules): retag the cross-household lead limit, drop the HOUSEHOLD_LEAD_MODEL design doc

thpr:
`checkin-app/docs/designs/HOUSEHOLD_LEAD_MODEL.md` described work that has shipped. Per `docs/DOCUMENTATION_STANDARD.md` §4, a merged working doc is distilled into `docs/rules/<domain>.md` and then deleted. This does that.

## Verified it actually shipped

Both legs of the expand-contract are merged and are ancestors of `main`:

- Expand — #917 `refactor(household-lead): collapse HouseholdLead join into Person.isHouseholdLead`, merge commit `86f84840`
- Contract — #947 `refactor(household-lead): contract phase — DROP HouseholdLead table + remove model`, merge commit `54ff61e9`

The contract leg was checked independently rather than taken from the doc: `gh pr view 947` reports MERGED, `git merge-base --is-ancestor 54ff61e9 origin/main` passes, and its migration `20260706130000_drop_household_lead` (`DROP TABLE "HouseholdLead"`) has since been folded into `20260711173200_coalesced_baseline`.

In the current tree: `Person.isHouseholdLead Boolean @default(false)` exists, there is no `HouseholdLead` model, no migration creates the table (the baseline creates only the boolean column and `Person_householdId_isHouseholdLead_idx`), and the auth/gate call sites read `isHouseholdLead` — `authClaims.ts`, `household/leads.ts`, `household/filters.ts`, `notifications.ts`, `emailRecipients.ts`, `trusted-adult/service.ts`.

## What was extracted

One retag. No new rules.

The doc's load-bearing claim — a lead always leads their own household, cross-household leads deliberately unsupported — was **already stated** in `docs/rules/people-households.md`. It was tagged `[Decision]`; it is now `[Decision — deliberate limit]`, which is what standard §3.6 reserves for an absence that was chosen and that someone helpful is likely to "fix".

Everything else failed the §3.9 test:

| Doc claim | Why it did not survive |
|---|---|
| Boolean, not a join table | Mechanism (§3.1) |
| No DB trigger | Mechanism, and a build convention rather than a domain rule (§3.4) |
| Fail-safe beats fail-open for an authz primitive | Already in `docs/rules/principles.md` under "Fail closed", and already cited by the neighbouring rule — duplicate |
| Rejected options (a2)/(b)/(c) | Tombstones: provenance, not constraints |

The doc's over-grant rationale for the limit was deliberately **not** carried across: it explains the rule but cannot be violated, so it is background rather than a rule. No new file, no new section, and no promotion to the Policy tier — no board policy names any of this.

## References swept

Repo-wide grep for `HOUSEHOLD_LEAD_MODEL` now returns zero.

- `checkin-app/src/lib/household/leads.ts` — two comments dropped the doc citation and the "a1" label, and now state the invariant directly
- `checkin-app/src/app/api/membership-ops/participants/[id]/household/route.ts` — same
- `docs/in-design/BG_PER_ADULT_SUBJECT.md` — dropped the trailing "Also see" pointer
- `docs/in-design/DOCUMENTATION_MIGRATION.md` — dropped the step-1 "not a source here" paragraph and the step-4 extract-then-delete entry; both are stale now that the step is done

## For the reviewer

The two `.ts` files are comment-only edits — no behaviour change, nothing to run beyond the usual checks.


docs/in-design/BG_PER_ADULT_SUBJECT.md
```diff
@@ -726,4 +726,4 @@ needs correcting, and INDEX's SA1 row needs to match #1260's reduced scope.
   not households" shift. Not a dependency.
 
 Unblocks **SA3** (household-composition sweeps), which needs a per-adult predicate.
-Also see: `personBgCheck.ts` policy block (the rule this violates) · `checkin-app/docs/designs/HOUSEHOLD_LEAD_MODEL.md`.
```
jee7s: #1561 deletes this file, FYI.

jee7s (approved):

PR #1574, opened 2026-08-08
🛡️ Sentinel: [MEDIUM] Fix stack trace exposure in database logs

dkaygithub:
🚨 Severity: MEDIUM
💡 Vulnerability: Raw error stack traces were being persisted to the `ErrorLog` database table, which feeds administrative dashboards.
🎯 Impact: Potential exposure of internal server information (directory structures, dependency versions) to users with dashboard access, violating defense-in-depth principles.
🔧 Fix: Removed stack trace persistence from `logBackendError` and instead log it to standard error for secure centralized aggregation.
✅ Verification: Verified that `logBackendError` no longer writes the `stack` field to the database.

---
*PR created automatically by Jules for task [9900129378056148480](https://jules.google.com/task/9900129378056148480) started by @dkaygithub*

### DIFF
--- a/.jules/sentinel.md
+++ b/.jules/sentinel.md
@@ -32,3 +32,8 @@
 **Vulnerability:** Found early returns checking buffer lengths (e.g., `providedBuffer.length !== expectedBuffer.length`) on webhook secrets and cron auth tokens before using `crypto.timingSafeEqual`.
 **Learning:** Returning early on length mismatch leaks the exact length of the expected secret.
 **Prevention:** Hash both the expected and provided secrets to a fixed length (e.g., using SHA-256) before passing them to `crypto.timingSafeEqual` to avoid leaking secret lengths while still safely catching any mismatch.
+
+## 2024-05-24 - Stack Trace Exposure in Database Logs
+**Vulnerability:** Raw error stack traces were being persisted to the `ErrorLog` table in the database, which feeds into administrative dashboards.
+**Learning:** Persisting raw error stack traces in database tables that feed application dashboards violates defense-in-depth principles, potentially exposing internal server information (like directory structures and dependency versions).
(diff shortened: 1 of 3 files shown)


thpr:
#1100 was the same finding and the same fix; it has been closed as a duplicate of this PR. Carrying forward the review points from there that are still open here.

**First, the standing decision.** This is at least the third time stack-trace persistence in `ErrorLog` has been filed. The position stated on #1100 was to accept the risk and annotate it rather than change the code, so that it stops resurfacing in audits. That decision has not been applied here — this PR fixes it again. The `.jules/sentinel.md` entry is the closest thing to it, and if the accept-the-risk route is taken, that entry (reworded to record the acceptance rather than the fix) is the part worth keeping.

The premise is also weaker than the PR body suggests: `ErrorLog.stack` is already `@sensitivity:internal` and reads are behind an admin gate, so this is not an exposure to ordinary users.

If the code change does go ahead, three things are unresolved:

1. **Raw `console.error` bypasses the `logger` seam.** `logger.error` is defined about fifteen lines above the edit in the same module. The comment says "central console", but `logger` is the only centralization seam there — if it is ever pointed at a real sink, the single line this PR exists to preserve silently misses it. Route through `logger.error`.

2. **The admin Errors panel still advertises the removed capability.** `checkin-app/src/components/admin/ErrorLogPanel.tsx:54` reads "Click a row to expand stack/context", and lines 82-94 render `e.stack`. With `stack` always null, every expansion is empty and the copy is a lie. Stacks would then live only in container stdout, which admins cannot reach and which rotates with the task. The panel and this change need to move together.

3. **The `console.error` is now unconditional.** It fires for non-`Error` throws too, where the second argument is the raw thrown value rather than a stack. The test patch here covers `logger.integration.test.ts`, which is the only suite that reaches the real `logBackendError` today — the other callers mock `@/lib/logger` — but widening the log surface widens what `jest-fail-on-console` will catch later. Gating on `stack` would keep it narrow.

Not a blocker to note, but the new `.jules/sentinel.md` heading is dated `2024-05-24`, two years off.

PR #1575, opened 2026-08-08
fix(membership-ops): name the reviewer's act, not its consequence

jee7s:
The second reviewer's background-check approval modal used **"clear"** throughout, and that word reads two ways: *attest the check came back clean*, or *remove the check from the record*. This modal sits beside "Reject", on an action it says cannot be undone. The wrong reading was available at exactly the wrong moment.

All three user-visible strings now say the same thing in the same words:

| | Before | After |
|---|---|---|
| Title | Clear this background check? | Attest this background check is clean? |
| Body | This clears the background check, records it against… | Attesting the check is clean records that against… |
| Button | Clear the check | Attest the check is clean |

Plus the first-reviewer copy, which ended "…before the check clears" → "…before the check is attested clean."

The title and the body's opening clause matter more than the button did: the title names the action before any consequence is described, and the first clause of the body is the one that sticks. The consequence list — payment opening, the family being emailed, the irreversibility — is unchanged.

"Attest" is also the domain's own verb rather than a softer synonym for it. This is a two-reviewer attestation model: `BackgroundCheckAttestation`, two APPROVEs recorded against a named subject. The copy now matches what the system records.

Behaviour is unchanged; only strings move. The three test assertions that drive the modal by title and button name move with them.

**Verified:** `jest src/app/membership-ops/review` 9/9, lint clean, tsc clean.

### DIFF
--- a/checkin-app/src/app/membership-ops/review/__tests__/page.test.tsx
+++ b/checkin-app/src/app/membership-ops/review/__tests__/page.test.tsx
@@ -83,7 +83,7 @@ describe("membership-ops/review page", () => {
     expect(await screen.findByText("Clear this background check?")).toBeInTheDocument();
     expect(screen.getByText(/emails the family/)).toBeInTheDocument();
     expect(fetchMock).toHaveBeenCalledTimes(1); // the queue GET only — nothing posted yet
-    fireEvent.click(screen.getByRole("button", { name: "Clear the check" }));
+    fireEvent.click(screen.getByRole("button", { name: "Attest the check is clean" }));
 
     await waitFor(() =>
       expect(fetchMock).toHaveBeenCalledWith(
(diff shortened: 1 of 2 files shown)


dkaygithub (approved):

thpr (approved):
Copy-only, and the sweep comes back clean: the four old strings appear nowhere outside the two files in the diff — no other test root, no flow test (HTTP-only, no DOM), no doc quoting the modal. The test selectors match the new strings exactly, and `getByRole` names being full-string matches means the confirm button "Attest the check is clean" doesn't collide with the row button "Attest — check is clean".

Checked against docs/rules/membership.md: "The background-check result is never stored. The record is that reviewers saw it and what they judged" [Decision, Policy-cited]. The new copy names the reviewer's judgement rather than a stored result, and the row button already read "Attest — check is clean" on main, so this introduces no drift from that rule.

One thing for whoever picks it up next rather than for this PR: "clear" survives everywhere else that's user-visible — JoinTreehouseBanner ("we'll email you when it clears"), the applications page ("once the background check clears"), plus bgClearedAt / PENDING_BG_CLEARANCE and the register's own "a clearance is dated when the second reviewer clears it". Act-versus-outcome is a defensible split, but nothing currently records that it's deliberate. A separate audit is going after exactly that and will land the choice in VOCABULARY.md.

Issue #1322, opened 2026-07-24
Vocab rename ledger (UNFINISHED.md)


thpr:
**Backlog item RB20** — from the backlog baseline (`docs/backlog/INDEX.md`, PR #1192).

Vocab rename ledger (UNFINISHED.md): leadMentor→ProgramLeader, staff→Treehouse Volunteer, dues→membership-fee, ToolLevel explicit rank, dependent retire+BUG-2, payment "certified"→"manual", etc.

- Origin: CHORE · Readiness: READY-FOR-DEV · Size: M

_Auto-created from the backlog INDEX; see the doc for bucket context and cross-links._

thpr:
**Backlog reconciliation (2026-08-01)** — RB20 vocab ledger gained the **'Review/Reviewer' term convergence** item (BG-reviewer / attestation-reviewer / membership-review / trusted-adult 'decider' — decide one canonical word or keep distinct), from UNFINISHED.md. Source: INDEX RB20.

jee7s:
**HOLD** — Ideation. A ledger of unrelated renames of very different risk. `leadMentor` is still live on `origin/main` (`Program.leadMentorId`, `checkin-app/src/app/api/programs/[id]/route.ts:133`), so the premise holds — but a schema rename with a migration and a rolling-deploy window is not the same unit of work as a copy change.

1. Should each rename be its own issue and PR — `leadMentor→ProgramLeader` alone is a schema rename with a migration?
2. Which renames are user-visible copy only, and which are schema- or API-breaking?
3. Is the Review/Reviewer convergence a decision (pick one canonical word, or keep them distinct) that must be answered before any rename starts?
4. Does `dues→membership-fee` include the Shopify product and variant naming, or app copy only?

---
_Posted by Claude._

PR #1579, opened 2026-08-08
docs(vocabulary): settle the background-check verbs

thpr:
The background-check flow used five words for three things, and it was not settled
which word named which.

`attest` named both the reviewer's act and — on the second click only — its
consequence: the same act, writing the same row with the same `result: "APPROVE"`,
was called "Record your approval?" for the first reviewer and "Clear this
background check?" for the second. The word changed with the outcome, not with the
act.

`clear` meant both *grant a clearance* and *erase*, twice inside one confirm modal:
"the check itself has not cleared yet" sits three lines above a button labelled
"Clear approvals".

`attest` also covered two acts that are not the reviewer's at all — the applicant's
consent self-attestation, and a Tool Certifier attesting a level change.

## The decision

One word per sense. A reviewer **reviews** the Averity report, **attests** to a
judgement about one adult, and two attestations produce a **clearance**.

- **Approve / Reject** demote from naming the act to being the value carried on an
  attestation.
- **"Clean" is banned.** The register records that the background-check result is
  never stored — only that reviewers saw it and what they judged. "The check is
  clean" asserts the content of a document the app never holds; "I approve this
  person" states the judgement, which is what is stored. That rule is Procedure
  tier (the app's stricter-than-policy expression), so this is a wording call
  rather than a policy escalation — but it is the reason for the call.
- **`clear` never means erase.** Undoing attestations is *discard*.
- The applicant **consents**; a Tool Certifier **certifies**.

This also resolves the existing "Review / Reviewer is overloaded" entry rather than
adding a second beside it. Three of its four flows turned out to be one flow — the
BG-reviewer role, the membership review queue and `BackgroundCheckAttestation.reviewer`
are the same people doing the same job (117 refs). Trusted-adult review keeps
**review** + **decider** (48 refs): one board member deciding with reasoning
recorded, not a two-of-N blind attestation. "Scholarship Review Team" is a third,
unrelated, and already canonical.

## Scope

**Glossary only — nothing is renamed.** The renames this implies are listed in
`UNFINISHED.md` so they can be spun off and sized separately. All of them are
ops-facing copy or internal identifiers; none touch `prisma/schema.prisma`,
`src/security/`, or any wire key.

Two things are recorded as **wrong and deliberately staying**, so nobody closes them
as oversights: `AttestationResult.result` (reads as "the check's result", the one
thing never stored) and `PENDING_BG_REVIEW` / `PENDING_BG_CLEARANCE`, which differ
by whether dues are paid rather than by review-vs-clearance. Renaming either reaches
the schema, a migration, and the security boundary.

## Relationship to #1575

Open PR #1575 changes the same modal. It fixes half of this — the modal stops
calling the act "Clear…" — while introducing "clean", which this decision bans. This
PR is based on `main`, not on that branch, and touches no source file, so the two do
not conflict either way.

Decision for #1322; no closing keyword, since the ledger it tracks covers renames
this PR does not perform.

### DIFF
--- a/checkin-app/docs/VOCABULARY.md
+++ b/checkin-app/docs/VOCABULARY.md
@@ -138,10 +138,11 @@ Rules:
    "I approve this person" states the judgement, which is what is stored.
    *(rename pending: the reviewer queue's confirm modal and row button)*
 3. **`clear` never means erase.** Undoing attestations is **discard**. `clear` is
-   reserved for the outcome above. *(rename pending: the "Clear approvals" button)*
+   reserved for the outcome above. Erasing a person's clearance date is **remove**.
+   *(rename pending: the "Clear approvals" button; the compliance page's erase path)*
 4. **"Attest" is the reviewer's word alone.** An applicant **consents**; a Tool
    Certifier **certifies** (see Shop & Certification).
-   *(rename pending: `selfAttestBgConsent` → `recordBgConsent`)*
+   *(rename pending: `selfAttestBgConsent` → `selfRecordBgConsent`)*
 5. **`PENDING_BG_REVIEW` and `PENDING_BG_CLEARANCE` do not mark review vs
    clearance** — they differ by whether dues are paid. A known misnomer; the enum
    stays, because renaming a status reaches the schema, a migration, and the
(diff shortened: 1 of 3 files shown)


thpr:
Follow-up issue for the renames this decision implies: #1580 — the reviewer queue's modal/button copy, "Clear approvals" → "Discard approvals", and `selfAttestBgConsent` → `recordBgConsent`. Kept out of this PR so the dictionary lands on its own; nothing there touches the schema, the security boundary, or a wire key.


jee7s (approved):

PR #1582, opened 2026-08-08
fix(membership): apply the settled background-check vocabulary

### LINKED ISSUE #1580: Apply the settled background-check vocabulary to the copy and identifiers
Follow-up to #1579, which settled the vocabulary but deliberately renamed nothing.

The dictionary now says: a reviewer **reviews** the Averity report, **attests** to a
judgement about one adult, and two attestations produce a **clearance**. `clear`
names the outcome and never means *erase*; the applicant **consents** rather than
self-attesting; "clean" is banned because it asserts the content of a report the app
never holds.

Three places still contradict that.

### 1. The reviewer queue names the act two wrong ways

`checkin-app/src/app/membership-ops/review/page.tsx`

- The confirm modal calls the act "Clear this background check?" / "Clear the check"
  — that is the outcome, not the act.
- The row button asserts "Attest — check is clean" — that is the report's content.

Both become *attest*. The modal body ("This clears the background check", "before
the check clears") is already correct and stays: it describes the consequence, which
is exactly what `clear` is now for.

8 assertions in `review/__tests__/page.test.tsx` match on these strings.

**Overlaps open PR #1575**, which rewrites the same modal — it removes
`clear`-as-act but introduces "clean". Whichever lands second resolves; the intent
here supersedes it.

### 2. "Clear approvals" means erase

`checkin-app/src/app/membership-ops/applications/page.tsx` — button, confirm label,
and the comment above them. Both senses collide inside one modal today: "the check
itself has not cleared yet" sits three lines above a button labelled "Clear
approvals". Becomes **Discard approvals**. No test assertions.

### 3. The applicant does not attest

`selfAttestBgConsent` → `recordBgConsent`. ~17 refs across 5 files. Internal only —
the `/api/membership/bg-consent` route path is a wire contract and does not change.

### Out of scope

`AttestationResult.result` and the `PENDING_BG_REVIEW` / `PENDING_BG_CLEARANCE`
misnomer are both recorded in `UNFINISHED.md` as wrong and deliberately staying:
renaming either reaches `prisma/schema.prisma`, a migration, and `src/security/`,
which carry their own migration-safety and boundary-isolation rules. Do not fold
them in here.

Nothing in this issue touches the schema, the security boundary, or a wire key.

thpr:
Fixes #1580. Applies the vocabulary settled in #1579 — a reviewer **reviews** the
Averity report, **attests** to a judgement about one adult, and two attestations
produce a **clearance**. `clear` names that outcome and never means *erase*; the
applicant **consents**.

Copy and identifiers only. No schema, no `src/security/`, no wire keys, no route paths.

## The reviewer queue

Scope added after @jee7s settled the deferred question on this PR in favour of the
dictionary. #1592 removed "clean" but replaced it with "passed", which states an
outcome of the report — the same claim, and the app never holds the report.

The objection #1592 was fixing was never the word *clearance*; it was `clear` as a
**verb** on a button reading as *reset*, next to Reject and on a screen whose sibling
control said "Clear approvals" for discard. The noun doesn't have that problem.

| | Was (#1592) | Now |
|---|---|---|
| Row button | Attest — the check passed | **Attest — approve this person** |
| 2nd-reviewer title | Attest that this background check passed? | **Approve the background check clearance?** |
| 2nd-reviewer confirm | Confirm — the check passed | **Attest — approve the clearance** |
| 1st-reviewer title | Record your approval? | **Record your attestation?** |
| 1st-reviewer confirm | Approve | **Attest** |

Row and modal keep distinct accessible names — the row proposes, the modal commits.
That was #1592's own finding and it stands.

## `clear` meant erase in two places

Both acted on the clearance itself.

`membership-ops/applications/page.tsx` — "Clear approvals" sat three lines below "the
check itself has not cleared yet" inside one confirm modal. Now **Discard approvals**.

`membership-audit/compliance/page.tsx` — the entire erase path for a person's
background-check date (handler, button, modal title, confirm label, toast) was built
on "clear". Now **Remove**, the verb its own body text already used. 8 sites, 4 assertions.

`membership-ops/participants/page.tsx` — "Clear the field to remove it" → "Leave the
field blank to remove it".

## The applicant does not attest

`selfAttestBgConsent` → `selfRecordBgConsent`, plus surrounding prose and the
page-local `bgAttesting` / `attestBgConsent`. Kept the `self` prefix rather than the
plain `recordBgConsent` named in #1580: `markBgConsent` is exported and called from
seven places, and a bare `recordBgConsent` wrapping it would be a coin flip at every
call site.

## Dictionary

Drops the three `(rename pending)` markers this PR satisfies, and records **"passed"**
alongside **"clean"** as words the dictionary rules out, with the reason, so it isn't
re-derived. Rule 3 now says explicitly that `clear` is fine as the outcome noun and
wrong as a button verb — the distinction the #1592 discussion turned on.

## Note on the audit

The compliance page's erase path was **not** in the audit behind #1579, and I reported
"0 test assertions" for the applications page when there were two — one sweep ran from
the wrong directory and returned a false zero. Re-swept; `clear`-as-erase is at zero
across the membership surface, with generic UI uses (`Clear filter`, `onClear`) and
correct outcome uses (`bgClearedAt`, "reviewers clear it") left alone.

## Checks

`tsc --noEmit` clean, `eslint --max-warnings 0` clean, full unit suite 2710 passed /
278 suites.

### DIFF
--- a/checkin-app/docs/VOCABULARY.md
+++ b/checkin-app/docs/VOCABULARY.md
@@ -131,18 +131,20 @@ adult) → two attestations → **clearance** (dated; the validity window runs f
 Rules:
 1. **A reviewer attests; two attestations clear the check.** One word per sense —
    never "approve" for the act, never "attest" for the outcome.
-2. **Never "clean."** The app never holds the report or its result, only that a
-   reviewer read it and what they judged
+2. **Never "clean," never "passed."** The app never holds the report or its
+   result, only that a reviewer read it and what they judged
    ([../../docs/rules/membership.md](../../docs/rules/membership.md), Procedure ›
-   Application and review). "The check is clean" asserts the document's content;
-   "I approve this person" states the judgement, which is what is stored.
-   *(rename pending: the reviewer queue's confirm modal and row button)*
-3. **`clear` never means erase.** Undoing attestations is **discard**. `clear` is
-   reserved for the outcome above. Erasing a person's clearance date is **remove**.
-   *(rename pending: the "Clear approvals" button; the compliance page's erase path)*
+   Application and review). Both words state an outcome of the report; "I approve
+   this person" states the judgement, which is what is stored. Policy's own
+   outcome vocabulary is approval, restriction, or denial — a pass/fail framing
+   has nowhere to put the middle one.
+3. **`clear` never means erase.** Undoing attestations is **discard**. Erasing a
+   person's clearance date is **remove**. `clear` is reserved for the outcome
+   above — and as a **verb on a button** it reads as *reset* even so, which is why
+   the reviewer queue names the outcome with the noun (*complete the clearance*),
+   not the verb.
 4. **"Attest" is the reviewer's word alone.** An applicant **consents**; a Tool
    Certifier **certifies** (see Shop & Certification).
-   *(rename pending: `selfAttestBgConsent` → `selfRecordBgConsent`)*
 5. **`PENDING_BG_REVIEW` and `PENDING_BG_CLEARANCE` do not mark review vs
    clearance** — they differ by whether dues are paid. A known misnomer; the enum
    stays, because renaming a status reaches the schema, a migration, and the
--- a/checkin-app/src/app/api/membership-ops/applications/external/route.ts
+++ b/checkin-app/src/app/api/membership-ops/applications/external/route.ts
@@ -43,7 +43,7 @@ export const POST = withAuth({ roles: ["isSysadmin", "isBoardMember"] }, async (
     //
     // The guard belongs on the route, not in the service: markContractSigned is also
     // the Zoho webhook's entry point (system actor), and markBgConsent is what
-    // selfAttestBgConsent calls, where the actor IS the applicant by design.
+    // selfRecordBgConsent calls, where the actor IS the applicant by design.
     const process = await prisma.orgMembershipProcess.findUnique({
         where: { id: processId },
         select: { orgMembershipId: true, subjectPersonId: true },
--- a/checkin-app/src/app/api/membership/bg-consent/route.ts
+++ b/checkin-app/src/app/api/membership/bg-consent/route.ts
@@ -1,7 +1,7 @@
 import { NextResponse } from "next/server";
 import { withAuth } from "@/lib/auth";
 import { logger } from "@/lib/logger";
-import { selfAttestBgConsent, ExternalError, type ExternalErrorCode } from "@/lib/membership/external";
+import { selfRecordBgConsent, ExternalError, type ExternalErrorCode } from "@/lib/membership/external";
 import { apiError } from "@/lib/api-response";
 
 export const dynamic = "force-dynamic";
@@ -19,7 +19,7 @@ const STATUS_BY_CODE: Record<ExternalErrorCode, number> = {
 
 /**
  * POST /api/membership/bg-consent — applicant-facing "I submitted my consent on
- * Averity" self-attestation (#875). Records consent on the caller's own process
+ * Averity" self-report (#875). Records consent on the caller's own process
  * awaiting external action via markBgConsent (applicant as audit actor), which
  * may advance the application to PENDING_PAYMENT. Honor-system by design; the
  * board's mark-bg-consent remains the backstop.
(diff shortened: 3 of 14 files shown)


jee7s (approved):

thpr:
@jee7s — flagging a scope question on this PR rather than acting on it, because the answer changes what belongs here.

**The state.** #1592 merged, so `membership-ops/review/page.tsx` on `main` says *"Attest — the check passed"* / *"Attest that this background check passed?"* / *"Confirm — the check passed"*. #1579 also merged, so `checkin-app/docs/VOCABULARY.md` on `main` now says:

> **Clearance / cleared** — the outcome: two attestations, dated.
>
> 2. **Never "clean."** The app never holds the report or its result, only that a reviewer read it and what they judged. "The check is clean" asserts the document's content; "I approve this person" states the judgement, which is what is stored. *(rename pending: the reviewer queue's confirm modal and row button)*

Those two are in tension on `main` right now. The dictionary names the outcome **clearance** and rules out stating the report's result; the queue states the report's result as **passed**. And rule 2 still carries a *(rename pending)* marker pointing at the very modal and button #1592 just rewrote — so the dictionary is, as written, asking for that screen to change again.

**What that implies for this PR.** The original review thread on #1592 had #1582 owning the reviewer-queue wording. I removed those changes when #1592 merged — `review/page.tsx` and its test are byte-identical to `main` here, and the commit that carried my alternative wording was dropped, not reworked. If #1579's dictionary is taken as settled, the consistent end state is that this PR (or a follow-up) re-words that screen to match it, which would overwrite the text #1592 just landed.

**I don't want to do that without discussion**, for two reasons. It would revert a change you authored and that was approved on its own merits, and the underlying question — whether the button states the reviewer's *judgement* or the check's *outcome* — was explicitly deferred rather than decided. Re-opening it by quietly editing the strings is the wrong way round.

So: three ways forward, and I have no stake in which.

1. **Dictionary wins.** The queue is re-worded to the dictionary's terms; scope is added here or split into a follow-up.
2. **Shipped UI wins.** `VOCABULARY.md` is amended — the outcome is named *passed* in UI copy, `clearance` stays the noun for identifiers and prose — and rule 2's *(rename pending)* marker is dropped.
3. **Neither yet.** Both stand, and the ledger records the contradiction as open so nobody "fixes" one side into the other by accident.

Option 3 is free and reversible, and stops the next person reading `VOCABULARY.md` from treating the pending marker as a work item. Happy to do whichever you prefer — this PR's other three surfaces (applications board, compliance page, consent rename) are unaffected either way and don't depend on the answer.

jee7s:
I am OK with something like "approve the background check clearance". My concern was with "clear the check" being ambiguous to whether the check state was being cleared or if the person got their background check clearance. That means "Dictionary Wins" but needs more exact wording on the action the reviewer is taking

thpr:
@jee7s — done, and thanks for the steer. Pushed as 9cc424b.

You were pointing at the verb, not the noun, and that distinction is what unlocked it. `clear` as a **verb on a button** reads as *reset* — worse here because the sibling screen's discard control literally said "Clear approvals". The **noun** carries no such reading, so the outcome can stay `clearance` without the ambiguity you flagged.

Your phrasing went straight into the title:

| | Was | Now |
|---|---|---|
| Row button | Attest — the check passed | **Attest — approve this person** |
| 2nd-reviewer title | Attest that this background check passed? | **Approve the background check clearance?** |
| 2nd-reviewer confirm | Confirm — the check passed | **Attest — approve the clearance** |
| 1st-reviewer title | Record your approval? | **Record your attestation?** |
| 1st-reviewer confirm | Approve | **Attest** |

On the exact action the reviewer is taking: the act is **attest**, the value they pick is **approve** or **reject**, and two approvals **complete the clearance**. So the row button names the act plus the judgement ("approve this person" — a person, not a document), and the confirm names the act plus what it completes. Your row-vs-modal accessible-name split from #1592 is preserved: the row proposes, the modal commits.

Also recorded **"passed"** in `VOCABULARY.md` next to "clean" as a word that's out, with the reason — both state an outcome of a report the app never holds, and policy's outcome vocabulary is approval / restriction / denial, which a pass-fail framing can't express. That way the next person doesn't re-derive it. The three `(rename pending)` markers this PR satisfies are dropped.

Full suite green: 2710 passed / 278 suites.

jee7s (approved):

PR #1584, opened 2026-08-08
refactor(types): consolidate SessionUser onto the NextAuth session shape

thpr:
Clears the `## 🟢 Dedup \`SessionUser\` type` item from `checkin-app/docs/designs/UNFINISHED.md` (VOCAB #8).

## What was wrong

Four independent `SessionUser` declarations described the same NextAuth session user, with different fields and different optionality:

| Where | Shape |
|---|---|
| `types/participant.ts:5` | the only exported one; role booleans **required**; doc comment said "Fields map 1:1 to Prisma **Participant** model" (that model is now `Person`) |
| `attendance/current/page.tsx:50` | local structural subset |
| `programs/[id]/page.tsx:50` | local structural subset (+ `ageBand`) |
| `AppFrame.tsx:54` | local structural subset (roles + `toolStatuses`) |

The actual source of truth is the `Session["user"]` module augmentation in `types/next-auth.d.ts` — it carries `denied`, `ageBand`, `programsLed`, `hd`, `emailVerified`, `canAccessStaging`, `impersonatedBy`, and more. Every one of the four was a hand-copied subset free to drift from it.

## What changed

`types/auth.ts` now exports the alias:

```ts
export type SessionUser = NonNullable<Session['user']>;
```

The three local declarations are gone, replaced by an import of it. Their `as SessionUser` casts became identities once the local types went, so they're dropped too — those call sites now read `session?.user?.isSysadmin` directly rather than casting first. (`AgeBand` in the programs page went unused along with them.)

`types/participant.ts` is **deleted** rather than renamed: its only other export, `BoardMember`, had zero importers — `safety/board-contacts/page.tsx:10` declares its own local copy. `types/index.ts` re-exports from `./auth` instead.

## The auth path keeps its own type — read this part

Collapsing everything onto `Session["user"]` does not typecheck: it produced 17 `tsc` errors in authorization code. `security/access-resolvers.ts` (role-token resolution, `buildCallerContext`) and `api/roles/route.ts`'s `RoleActor` consume `boolean`, not `boolean | undefined`, and `householdId` would widen to include `null`.

That required-flag shape is load-bearing, so it stays — as a separate, honestly-named type rather than bending either side:

```ts
export type AuthenticatedUser = SessionUser &
    Required<Pick<SessionUser, BusinessRole>> & { householdId?: number };
```

`AuthResult`, `getOptionalSessionUser`, and the security-test personas use `AuthenticatedUser`; the client/UI surfaces use `SessionUser`. Two types, two real meanings.

It is **derived, not hand-copied**, so new fields on the augmentation flow into it automatically — it narrows exactly two things and says which. Both narrowings were verified against runtime behaviour rather than assumed:

- **Five role flags always present as booleans** — `lib/authClaims.ts:44`, `for (const flag of ROLE_FLAGS) token[flag] = denied ? false : f[flag]`, on both jwt-callback branches (fresh sign-in and per-request refresh).
- **`householdId` is a plain Int, never null** — `prisma/schema.prisma:130` declares `householdId Int` (non-null), and `assignParticipantClaims` sets `token.householdId = p.householdId`. Kept **optional**, matching the previous contract, because the fail-closed `return {} as JWT` path leaves it unset.

**No authorization check was loosened, rewritten, or reinterpreted.** Worth a reviewer's eye specifically: `access-resolvers.ts:293` gates on `auth.user.householdId !== undefined`. Under a `number | null | undefined` type that test would let a `null` through as "has a household" — which is exactly why the narrowing stays rather than widening the type underneath the check.

## Verification

- `npx tsc --noEmit` clean. This is the main net for a pure type change, but not the whole story — mocked tests are blind to shape drift, so `SessionUser` was grepped across all three test roots (`src/**/__tests__`, `tests/`, `__tests__/`). That surfaced three importers beyond the ones the issue listed: `tests/security/personas.ts`, `tests/security/ctxNeeds.test.ts`, `tests/security/registryAuthz.integration.test.ts`. All three build auth-boundary users, so they moved to `AuthenticatedUser`.
- `eslint` clean on every touched file.
- Suites run, all green: `withAuth` (14), `shopNav` (12), `AppFrame` (13), `routeAuthDrift` (449), `pageRegistry` (3), `ctxNeeds` (8).
- Not run locally (DB-gated): `scanAuth` and `registryAuthz` are integration tests. Their edits are type-only and covered by `tsc`.
- Touches nothing under `checkin-app/src/security/`, so the boundary-isolation workflow does not apply.


jee7s (approved):

thpr:
Rebased onto `main` (11 commits). One conflict, in `checkin-app/docs/designs/UNFINISHED.md`: #1587 restructured the ledger while this branch deleted the `## 🟢 Dedup \`SessionUser\` type` section. Resolved by taking #1587's version wholesale and re-applying only the section deletion.

One extra edit came out of that, worth a reviewer's glance since it isn't part of the original change: #1587's *Considered and dismissed* list ends its **Account / Session / VerificationToken** bullet with "…is the SessionUser dedup item above", which dangles once that section goes. The bullet now states the smell directly — `userId Int @map("participant_id")` on `Account`/`Session` (`prisma/schema.prisma:1189`, `:1222`), a Participant-era physical column NextAuth calls "user" — and records that it stays put, because renaming it is a migration rather than a vocab fix. Verified those two `@map`s are still on `main`.

No code conflicts. Re-verified after the rebase: `tsc --noEmit` clean, and `withAuth` (14), `shopNav` (12), `AppFrame` (13), `routeAuthDrift` (449), `pageRegistry` (3), `ctxNeeds` (8) all pass, plus #1591's new `scopeOracleSet` (12).

Also checked the incoming renames for silent auto-merge survivors — `normalDuesCents`/`volunteerDuesCents` (#1585), `certifyPaymentPlan`/`certificationNote`/payment `"certified"` (#1586) — zero hits in any file this branch touches. #1590's whole-entity decommission rule covers `checkin-app/src/security/**`, which this PR does not enter.

PR #1591, opened 2026-08-09
test(security): pin the scope-oracle set so a deleted oracle fails loudly

jee7s:
Follow-up to **#1590**, which I reviewed and which is approved. Branched off #1590's head, so it rebases to nothing once that lands — **merge #1590 first**.

## What #1590 changed, and what it cost

#1590 stops classifying `checkin-app/src/security/__tests__/` as boundary files. That is correct: `src/security/*` matched the ten scope oracles, so a decommission that had to update its own oracle was refused by the exception built to permit it — and blamed a test file while doing it.

The side effect is disclosed in that PR, and it is bigger than "a PR touching only those tests no longer trips isolation." Driving the shipped `is_boundary`/`is_companion` directly:

```
boundary=no  companion=YES  src/security/__tests__/visit-household-lead-scope.test.ts
boundary=YES companion=YES  src/security/scopeBindings.ts
```

So an oracle is now a **companion** — it can ride along inside any feature PR, where before it forced isolation.

And nothing pinned that these files exist:

```
grep -rln "src/security/__tests__" checkin-app/tests checkin-app/scripts   → (empty)
```

Deleting `visit-household-lead-scope.test.ts` inside a 40-file feature diff trips neither isolation nor the suite — **a deleted test is invisible to a green run.** It does not fail; it just stops checking. CODEOWNERS still forces a security owner onto the review, but that is a human reading a large diff, which is the failure isolation exists to stop.

The exposure is a two-step: weaken the oracle now, where it rides along quietly; widen the policy later, where the widening is isolated and reviewed but the oracle no longer catches it, so CI reads green. Both steps need an owner's approval, so this is defence in depth, not an open door — but the mechanical backstop went away and nothing replaced it.

## The fix

One test, following `artifactsDrift.test.ts`'s shape: pin the eleven filenames, and assert each still contains an assertion.

Adding an oracle costs one line and makes a reviewer see the new file. Removing one has to be written down on purpose. That is the whole point.

## Verified by breaking it

| Edit | Result |
|---|---|
| Delete `visit-household-lead-scope.test.ts` | 2 failures, naming the missing file |
| Replace it with an assertion-free stub | 1 failure |
| Restore it | 12/12 green |

`tests/security` 611/611, lint clean, tsc clean.

## Deliberate limit

Presence, not depth — a file gutted to a single assertion still passes. It catches deletion and emptying, which are the invisible edits; a *weakened* assertion is a diff a reviewer can actually see. The comment names the upgrade path (per-oracle assertion counts) if that stops being true.


thpr (commented):
Two gaps, both one-line fixes, neither blocking. The shape is right, and it holds today: the eleven names match `origin/main` exactly, all eleven files contain `expect(`, none is skipped, nothing else in the repo enumerates them, and `checkin-app/tests/security/*` is already `is_companion`, so this file can never trip isolation itself.

## 1. `.skip` satisfies both assertions while the oracle stops checking

The `ponytail:` comment names the ceiling as "a file gutted to a single assertion still passes," which tells a later reader the residual risk is an oracle that is weaker but still running. The residual is larger than that. Change `describe(` to `describe.skip(` in any oracle: the file is still present, the source still contains `expect(`, both assertions pass, and nothing executes. The effect is the same as deleting the file, which is the case this test exists to catch.

The stated reason for accepting the documented residual is that a weakened assertion is a diff a reviewer can see. That reason is weaker for a skip. A skip reads as routine test hygiene, and Jest stops complaining on its own; a deleted file at least shows up as a removed entry in the file tree. No oracle uses `.skip` or `xit` today, so this lands green:

```ts
expect(src).not.toMatch(/\b(?:describe|it|test)\.skip\(|\bx(?:describe|it|test)\(/);
```

If you would rather not widen the scope, the alternative is to name skips in the comment as a known hole. I prefer the line; it is cheaper than the sentence explaining its absence.

## 2. The listing is non-recursive and `.test.ts`-only, so future oracles can land outside the pin

`readdirSync(ORACLE_DIR).filter(f => f.endsWith('.test.ts'))` sees less than Jest runs. The default `testMatch` picks up everything under `__tests__/`, including `.test.tsx`, `.spec.ts`, and subdirectories. An oracle added under any of those is absent from `present`, so the set equality still passes, the file never has to be written into `ORACLES`, and it stays outside the deletion guard permanently. The property this PR is buying, that adding an oracle costs one line and makes a reviewer see the new file, silently does not hold for it.

Deletion and rename of an existing oracle are caught either way; this affects additions only. `engines.node` is `>=24`, so the recursive form is available:

```ts
const present = readdirSync(ORACLE_DIR, { recursive: true })
    .map(String)
    .filter(f => /\.(test|spec)\.tsx?$/.test(f))
    .sort();
```

Recursive listing returns paths relative to `ORACLE_DIR`, so an oracle placed in a subdirectory would have to be written into `ORACLES` with its prefix. That is the behaviour you want; the current eleven are all top level, so they round-trip unchanged.

dkaygithub (approved):

PR #1592, opened 2026-08-09
fix(membership-ops): the check passed — it was not cleaned, and nothing was cleared

jee7s:
## The word

"Clear the check" carries **two** wrong readings, not one. Beyond *erase the check*, it also reads as *clear the current state* — reset, discard, start over. On a screen whose other action is **Reject**, that is arguably the more natural reading.

#1575 replaced it with "clean", which fixed the erase sense but left an odd claim: a background check is not clean or dirty. It **passed**, or it did not.

| | Now |
|---|---|
| Title | Attest that this background check **passed**? |
| Body | Attesting that the check **passed** records that against… |
| Button | Attest — the check **passed** |
| First reviewer | …before the check **passes**. |

The act is still attestation and the row written is still an attestation; only the outcome is renamed.

## Correction to an earlier version of this description

This PR previously claimed #1582 "branched before #1575 merged" and "will conflict". **That is false.** #1575's merge commit `b3ae31f4` is an ancestor of #1582 — `git merge-base --is-ancestor b3ae31f4 <1582 head>` returns true — and `git merge-tree` against `origin/main` is clean. GitHub's `CONFLICTING` flag on #1582 is stale.

**The correction makes the hazard worse, not better.** #1582 rewrites the same four strings and merges **without a conflict**, so nothing warns a reviewer that #1575's wording is being replaced. A silent revert is harder to catch than a noisy one.

## Three live proposals for the same four strings

| | Title | Confirm button |
|---|---|---|
| **#1575** (merged) | Attest this background check is clean? | Attest the check is clean |
| **#1582** (open) | Attest — this completes the background clearance? | Attest — complete the clearance |
| **#1592** (this) | Attest that this background check passed? | Attest — the check passed |

**#1579** writes the dictionary that #1582 implements, and rules "clean is out" — agreed, and this removes it. Where it and this PR differ is whether `clear`/`clearance` is safe in the *button*. #1579's position is that `clear` is fine once defined as *grant a clearance*. The counter-argument is that a definition in `VOCABULARY.md` does not travel to the person reading a confirm dialog, and this is the one place where the wrong reading is destructive and irreversible.

`pass` has no second meaning here. `clear` and `clean` both do.

A resolution that satisfies both: keep #1579's model — a reviewer **reviews**, **attests**, and two attestations produce a **clearance** — so `clearance` stays the noun in prose and identifiers, while the *button* states what the reviewer is asserting. #1582's own body text ("Attesting clears the background check", line 148) would then be the only line left to settle.

**This needs one human decision before #1579, #1582 or #1592 merges.** Whichever lands last silently overwrites the others.

## Verified

`jest src/app/membership-ops/review` 9/9, lint clean. Behaviour unchanged; only strings and the two assertions that drive them by name.

### DIFF
--- a/checkin-app/src/app/membership-ops/review/__tests__/page.test.tsx
+++ b/checkin-app/src/app/membership-ops/review/__tests__/page.test.tsx
@@ -80,10 +80,10 @@ describe("membership-ops/review page", () => {
     fireEvent.click(screen.getByRole("button", { name: "Attest — check is clean" }));
     // Pat already holds one approval, so naming Pat is the SECOND — the confirm names
     // the consequence of a clearing approve rather than a generic "are you sure?".
-    expect(await screen.findByText("Attest this background check is clean?")).toBeInTheDocument();
+    expect(await screen.findByText("Attest that this background check passed?")).toBeInTheDocument();
     expect(screen.getByText(/emails the family/)).toBeInTheDocument();
     expect(fetchMock).toHaveBeenCalledTimes(1); // the queue GET only — nothing posted yet
-    fireEvent.click(screen.getByRole("button", { name: "Attest the check is clean" }));
+    fireEvent.click(screen.getByRole("button", { name: "Attest — the check passed" }));
 
     await waitFor(() =>
       expect(fetchMock).toHaveBeenCalledWith(
@@ -106,7 +106,7 @@ describe("membership-ops/review page", () => {
     fireEvent.click(screen.getByRole("button", { name: "Attest — check is clean" }));
     fireEvent.click(await screen.findByRole("button", { name: "Cancel" }));
 
-    await waitFor(() => expect(screen.queryByText("Attest this background check is clean?")).not.toBeInTheDocument());
+    await waitFor(() => expect(screen.queryByText("Attest that this background check passed?")).not.toBeInTheDocument());
     expect(fetchMock).toHaveBeenCalledTimes(1); // the queue GET only
   });
 
--- a/checkin-app/src/app/membership-ops/review/__tests__/page.test.tsx
+++ b/checkin-app/src/app/membership-ops/review/__tests__/page.test.tsx
@@ -66,9 +66,9 @@ describe("membership-ops/review page", () => {
     renderPage();
     await screen.findByText("The Smiths");
 
-    expect(screen.getByRole("button", { name: "Attest — check is clean" })).toBeDisabled();
+    expect(screen.getByRole("button", { name: "Attest — the check passed" })).toBeDisabled();
     nameSubject("Sam Smith");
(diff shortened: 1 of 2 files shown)


thpr:
Your diagnosis was right and I've taken it. #1582 now says **"Attest — complete the clearance"** rather than "Attest and clear the check" (af3d780).

**On "clear the check".** You're right, and I'd underweighted it. The reset reading isn't hypothetical: the board's own start-over control on the applications board was labelled **"Clear approvals"** until #1582 renamed it to "Discard approvals". So a reviewer has seen that exact verb mean *discard* on an adjacent screen, and then meets it next to a "Reject" button meaning the opposite. "A definition in VOCABULARY.md does not travel to the person reading a confirm dialog" is the right test and I failed it.

**On the rebase warning.** Exactly right, and it caught something. Git flagged the title and the confirm label. It did *not* flag the modal **body** — #1575 put "clean" into lines my diff never touched, so they auto-merged in silently:

```
Attesting the check is clean records that against…
…before the check is attested clean.
```

A resolution that only cleared the marked hunks would have merged a PR whose stated purpose is removing "clean" while leaving two instances of it. Both are fixed and I grepped the word to zero.

**Where I landed differently: not "passed".** Three reasons, and the third is the one I'd most want your view on.

1. It makes the same *kind* of claim "clean" made. #1579's rule against "clean" comes from `docs/rules/membership.md` — the background-check result is never stored; the record is that reviewers saw it and what they judged. "The check passed" states a result. "I approve this person" states a judgement. The reason clean was out applies to passed.
2. The policy's own outcome vocabulary is **approval / restriction / denial** — *"A check may result in approval, a restriction on how the person may volunteer, or denial"* (Membership Policy, Art. VI §VI.2). "Pass" isn't in it; "approve" is, and it's the value the row actually stores.
3. **"Passed" is binary and the policy isn't.** A check can come back with a restriction on how someone may volunteer. Pass/fail has nowhere to put that; approve/reject at least doesn't foreclose it.

Your suggested resolution already pointed here — keep `clearance` as the noun in prose and identifiers, and make the button say what the reviewer is asserting. The noun is what fixes your objection: nothing reads "complete the clearance" as resetting something, because the ambiguity lives in the verb.

| | #1582 now |
|---|---|
| Title | Attest — this completes the background clearance? |
| Confirm | Attest — complete the clearance |
| Row button | Attest — approve this person |

One practical note: #1592 and #1582 now touch the same lines, so they conflict. If you're happy with the above, #1582 supersedes this and you could close it — your call, and the ambiguity you spotted is fixed either way. If you still prefer "passed" on the outcome question, I'd rather settle that here than have it resolved by whichever branch merges second.


thpr:
Approving as this is an improvement, but still would like to see thoughts on #3 above

thpr (approved):

Issue #1552, opened 2026-08-07
A household that loses its only checked adult mid-year keeps its membership — undecided what should happen


thpr:
Surfaced by the domain-register audit (#1528), and recorded until now in
`docs/rules/membership.md`'s `## Policy requirements not yet enforced here`
appendix, which is being retired — this issue is the record.

## The gap

Removing or demoting the household's only background-checked adult leaves the
household **active, with youth still enrolled**, and nothing responds. No block,
no grace period, no warning, no surface that reports it.

Policy requires at least one checked adult per member family, so the household is
out of compliance the moment it happens — the app simply does not notice.
— *Membership Policy, Art. VI §VI.1*

## Why it is filed as undecided rather than as a bug

**What the answer should be has never been agreed.** Three plausible ones, with
real trade-offs:

- **Block** — refuse the removal/demotion while youth are enrolled. Cleanest
  invariant, worst failure mode: a family cannot fix their own household
  composition without staff, and the app is not where that conversation belongs.
- **Grace period** — let it happen, start a clock, act if nobody re-checks.
  Needs a decision on the length and on what "act" means at the end of it.
- **Warn only** — surface it for follow-up and gate nothing. Consistent with how
  the register already handles the program-attached background-check obligation
  ("it is surfaced for follow-up and gates nothing — chasing it is a
  conversation, not a refusal at the door") and with *Principle: people decide
  about people*.

The third is the cheapest and most consistent with existing precedent. It is
still a choice somebody has to make.

## Related, and not the same

- #1261 — scheduled household-composition sweeps. That is a **mechanism** that
  could carry whichever answer is chosen; it does not decide this.
- #1260 / PR #1470 — clearance is being made per-adult rather than a blanket
  stamp on every lead. That makes "who is actually checked" answerable for the
  first time, which is what any of the three options above depends on. **Worth
  settling this after that lands**, not before.

## Whichever way it goes, the register changes

The answer belongs in `docs/rules/membership.md` Procedure, at the activation and
renewal rules it qualifies, tagged per `docs/DOCUMENTATION_STANDARD.md` §3.6 —
`[Unsettled — …]` while open, a plain `[Decision]` or
`[Decision — deliberate limit]` once answered.

PR #1595, opened 2026-08-09
fix(membership-ops): re-derive background-check state after a participant merge

### LINKED ISSUE #1396: Participant merge transfers the background check date but leaves the application at background-check review

jee7s:
Fixes #1396. Stacked on **#1589** (merged) — the conflict-of-interest guard it adds is load-bearing here, see below.

Follows the design already on main, `docs/in-design/MERGE_BG_CARRYOVER.md`, and deviates from it in two places that are called out rather than quietly taken.

## The bug

The merge moves `Person.lastBackgroundCheck` (newer of the two wins) and re-points the process row, but never re-evaluates the derived state that date feeds. A household that becomes background-check-fresh *because of* the merge keeps `bgClearedAt` null, so the board runs a 2-of-N review for a check that already exists and is still valid.

## The trap this had to avoid

**Stamping `bgClearedAt` alone is worse than the bug**, and that is what the issue's suggested direction proposes. `awaitingBgReview.where` carries `bgClearedAt: null`, so a stamped `PENDING_BG_REVIEW` row leaves the reviewer queue — and `attest`, `activate`, `advanceExternalIfComplete` and `overrideBlocked reset` all then refuse it. `INVARIANTS` only has `active-is-bg-cleared`, the other direction, so the reconciler would not even report it. The row would have no exit but archival. Today it is at least *visibly* parked.

## What it does

New `lib/membership/mergeBgAdvance.ts`, called post-commit from the merge route. It re-runs one predicate — no live intake note **and** `householdBgIsFresh` — then:

| state | action |
|---|---|
| `PENDING_EXTERNAL_ACTION` | stamp + `applyVolunteerStatus` + `advanceExternalIfComplete` |
| `PENDING_PAYMENT` (parallel track) | stamp + `applyVolunteerStatus`; status already correct |
| `PENDING_BG_REVIEW`, note gone, 0 attestations | real clearance via `clearBackgroundCheck` → `PENDING_PAYMENT` + `notifyPaymentOpen` |
| `PENDING_BG_REVIEW`, note live | untouched — audit row records that the merge made the household fresh *without* clearing it |
| any row with ≥1 attestation | untouched — a dedupe action does not finish a human's review |
| everything else | untouched |

**Advance only.** Every query and the CAS both filter `bgClearedAt: null`; it never unstamps. Retraction is #1552, open and undecided.

**`PERSON_BG` and `PERSON_AGREEMENT` are excluded structurally.** Every query is scoped through `orgMembership: { householdId }`, and both kinds carry `orgMembershipId: null`. This matters: `householdBgIsFresh` filters `isHouseholdLead: true`, so an unscoped version would clear **an adult child's own background check because a parent's is fresh**. There is a regression test for exactly that.

**It cannot fail the merge.** The merge has already committed; a throw here would report failure for work that succeeded. Errors go to `logBackendError`.

**#1589 is load-bearing and the coupling is invisible.** Without its guard, a board member could merge to seat a background-check date on their own household *and* now auto-clear the application off it. The dependency is commented at the call site.

## Deviations from the design, both deliberate

**1. `PENDING_EXTERNAL_ACTION` also advances.** The doc says stamp only, because the existing advance path "picks the stamp up when that arrives" (`:523-528`). It does not, when the contract is already signed: `markContractSigned` will not re-fire (`syncContractStatus` only calls it when `!contractSignedAt`), and a household that no longer needs a check will never mark consent. The row would sit at `PENDING_EXTERNAL_ACTION` with its gate met — bug #1396 one state later. `intake.ts:438` handles the identical case the same way, commenting "without this the process would strand here with its gate already met". The doc's non-atomicity objection assumed an in-transaction carryover; this is post-commit regardless, and `advanceExternalIfComplete` is idempotent and CAS-guarded.

**2. Tighter than the doc in two places.** The ≥1-attestation guard (the doc argues at `:374-376` that a lone APPROVE is not an obstacle — defensible, and cheap to relax), and the note clause applied at all four states rather than only the two the doc calls settled.

## Not built — needs a decision

**`PENDING_BG_CLEARANCE → ACTIVE` is absent.** It is the only state that can send anything, so building it forces the design's open question 2, which says to confirm with @thpr first. Skipping the emails is affirmatively wrong (`openPersonBgForNewMember` is "a compliance gap, not a courtesy"), so there was no neutral option.

Everything here is independent of that: `PENDING_BG_REVIEW` rows are always unpaid, so they converge to `PENDING_PAYMENT` and fire only `notifyPaymentOpen` — never `sendCongrats` or Trigger C.

Provenance (open question 1) takes the doc's own cheap default, the audit blob. It is a strict subset of option 2, so choosing that later is additive.

## Does it close the repro?

**Partially, and the headline repro was never the real bug.** As literally written — household at `PENDING_BG_REVIEW` — it now works only when the note that put it there has since been deleted; with a live note the application still parks by design.

But the thread says *"I believe we were at PENDING_PAYMENT when this happened."* **That case is fully fixed**: the stamp lands, the row drops out of the reviewer queue, and the redundant 2-of-N goes away. `PENDING_EXTERNAL_ACTION` is fixed and now converges instead of stranding. `PENDING_BG_CLEARANCE` is the gap above.

## Verification

- `npm run test:ci` — **2586 passed, 0 failed**. Includes `guardFromStateParity` (three new CAS from-states registered) and `artifactsDrift` (no `TRANSITIONS` change needed — every edge driven here is already declared).
- `npm run lint` clean; `tsc` clean apart from two pre-existing `@aws-sdk/client-lambda` errors in unrelated `s-read` files.
- `git diff --name-only origin/main...HEAD -- checkin-app/src/security/` → **empty**. No boundary change, so no ship-alone requirement.

**Not verified: the 10 integration cases have never been executed.** No database was reachable in the authoring environment. They cover every row of the matrix plus the `PERSON_BG` security case and the never-unstamped case; they typecheck and lint only. CI is their first real run.


thpr:
Three things that only show up reading the diff. None of them blocks.

### 1. The "never unstamps or re-stamps" case never reaches the code it names

`route.integration.test.ts:1137` sets `pMerge.lastBackgroundCheck = STALE`, so `householdBgIsFresh` is false and `advanceHouseholdBgAfterMerge` returns at the predicate — before any query runs. The assertion passes even if `bgClearedAt: null` were deleted from every `where` in the module, which is the one thing advance-only rests on.

The property is real; it just isn't under test. To pin it: keep `FRESH`, pre-stamp the row's `bgClearedAt`, assert the date is unchanged.

### 2. `mergeBgAdvance.ts:10` — the merge never promotes the keeper to lead

> moves the newer of the two `lastBackgroundCheck` dates onto the survivor and **can promote them to household lead**

`resolveKeeperUpdate` doesn't touch `isHouseholdLead`, and the tombstone CAS only *clears* the merged-away record's (route.ts's own step 6 comment says so).

Post-commit placement is still right for the reason the call-site comment gives — clearing the tombstone's flag changes the live-lead set `householdBgIsFresh` reads. It's only this line that's wrong, and a comment that doesn't match the code is the kind that gets trusted later.

### 3. The note branch audits one state of the four it suppresses

`noteHeldDisclosure` (`mergeBgAdvance.ts:159`) queries `fromWhere("PENDING_BG_REVIEW")` only, but the early return above it suppresses the carryover for every state. A household with a live note and a `PENDING_EXTERNAL_ACTION` row that this merge just made fresh gets no trail at all — which is the branch's stated purpose ("record that the merge made the household fresh so the trail explains why a covered application is still parked").

Behaviour is right either way: that row still reaches a human, since `advanceExternalIfComplete`'s `holdForNote` sends it to `PENDING_BG_REVIEW` once the contract lands. So this is trail coverage, not state. The test at `:1113` — "leaves every state alone while a live intake note holds the household, and audits why" — exercises only `PENDING_BG_REVIEW`, so it can't tell the two apart. Either widen the query or narrow the name and the comment to the state that's actually covered.

thpr (approved):
Approving. The three points above are addressed, and `486c4b61` fixes something bigger than any of them.

Re-read against the new commits:

- The flip gate is tight. It can only suppress a carryover when the keeper's household holds *another* live lead with a valid check — and a tombstone that is a lead in the keeper's household is already refused upstream by `isLead && householdOthersCount > 0`, so the canonical repro can't fall through it.
- `PENDING_BG_REVIEW` really is always unpaid (`activate()` stamps `paidAt` only from `PENDING_PAYMENT`/`BLOCKED`), so `clearHeldReview` cannot reach `sendCongrats` or Trigger C. The claim in the description holds.
- `notifyClearanceOutcome` is behaviour-identical at both extraction sites, including the PERSON_BG path where `householdId` is null and neither branch fires.
- PERSON_BG/PERSON_AGREEMENT exclusion is structural, not conventional — `orgMembership: { householdId }` cannot match a null `orgMembershipId`.
- Dropping the unsigned `PENDING_EXTERNAL_ACTION` stamp is the right call: `markContractSigned` still re-runs the advance, so nothing strands, and the note hold stays armed.

Two residuals, neither blocking:

**`CARRYOVER_STATES` is the one from-state list here not sourced from `fromWhere`**, and it reaches Prisma as `{ status: { in: [...] } }` through `uncleared`'s `from: object` — spreading a value typed `object` contributes no known properties, so tsc never checks those strings against `OrgMembershipProcessStatus`. `guardFromStateParity` doesn't cover it either; it's a select, not a CAS. Typing `from`/`extra` as `Prisma.OrgMembershipProcessWhereInput` would make the list checked for free, and would also restore checking on `{ attestations: ... }` / `{ contractSignedAt: ... }` at the other call sites.

**`noteHeldDisclosure`'s comment now overstates.** It says it audits every row a stamp would have reached, but unsigned `PENDING_EXTERNAL_ACTION` rows are excluded from stamping while still in `CARRYOVER_STATES`. The test uses a signed one, so it doesn't distinguish. Comment or query, either way.

The description is stale in two places against `486c4b61` — the `PENDING_EXTERNAL_ACTION` matrix row still says unsigned rows are stamped, and "Does it close the repro?" predates the flip gate. Worth refreshing before merge, since that section is the record of what this actually fixes.

PR #1597, opened 2026-08-09
fix(membership): do not tell a family a review completed when none did

jee7s:
Builds on **#1595**. Three corrections from an adversarial pass over that PR against `docs/in-design/MERGE_BG_CARRYOVER.md` and `docs/rules/membership.md`.

**Merge #1595 first.** This targets `main` so CI actually dispatches — a PR based on another open branch runs no workflows at all, which reads as green when nothing ran. The cost is that the diff currently shows #1595's commits alongside the ~60 lines that are new here; once #1595 merges, GitHub recomputes the merge base and it shrinks to this PR's own change. The dependency is real either way: every line below edits code #1595 introduces.

Same shape #1593 used over #1591.

## 1. It told a family a review completed when none did — the one to block on

The carryover converges a held application into `PENDING_PAYMENT`, which fires `notifyPaymentOpen`. On `main` that notice reads:

> **Subject:** Background check complete — you can now pay your membership dues
> Good news — your household's **background-check review is complete**.

On this path no reviewer read anything. The household's own prior check was recognised by a merge. The design anticipated it — *"`notifyPaymentOpen` is unreachable. Its only trigger would have been `PENDING_BG_REVIEW → PENDING_PAYMENT`, which the carryover no longer performs"* (`:718`) — but #1595 performs exactly that transition, so the send became reachable again with copy that is now false.

It also sits awkwardly with #1595's stated position. That PR defers `PENDING_BG_CLEARANCE` precisely because it would force the design's open question 2 on family-facing email, calling `notifyPaymentOpen` the safe residue. It is the one send whose *content* is wrong here.

New `notifyPaymentOpenOnExistingCheck`: *"your household already holds a current background check, so no new one is needed"* + the same payment link. The family still learns payment opened; the sentence is true. The reviewer path keeps the original copy unchanged.

## 2. The clearance was indistinguishable from a real one

`clearBackgroundCheck` audits `{ status, bgCleared: true, clearedPersonIds }`. A merge-derived clearance wrote that byte for byte — same event, same shape, no marker — while `docs/rules/membership.md` requires *"Every background-check decision … records a written reason."*

It now takes optional provenance, and the carryover passes the `{ via, sourcePersonId }` it already records on its other paths plus the reason. The reviewer path passes nothing and is unchanged.

Note this clearance names nobody (`clearedPersonIds: []`), which is deliberate and safe — no `Person.lastBackgroundCheck` is written, so no unchecked adult is marked cleared. The cost was attribution, and that is what this fixes.

## 3. The note-held branch owed reviewers a ping

The design leaves a note-held row parked on purpose, and pays for that no-op by promising the queue is told: *"do not stamp — `notifyReviewers()` so the queue sees the new evidence"* (`:407`), *"discloses + pings reviewers"* (`:511`). #1595 wrote an audit row and stopped. An audit row is not a surface anyone watches, so it calls `notifyReviewers()` too — zero-arg, already exported on `main`.

## What this does not fix

The adversarial pass raised more, deliberately left out of a narrow correction:

- **The intake-note gate should be deleted entirely once #1499 lands.** #1499 writes *"An intake note does not hold the application"* into `docs/rules/membership.md:181` and already annotates the design as **STALE** on the note clause. After it, #1595's gate would be the only intake-note control-flow read left in the app, and would leave note-bearing households unstamped — including `applyVolunteerStatus`, so a pre-designated volunteer household is billed non-volunteer fees. That is a bigger change than this PR, and it belongs after #1499 merges.
- Declaring #1595's post-commit move and its SYSTEM audit actor on the `advanceExternalIfComplete` path as deviations.
- Amending `MERGE_BG_CARRYOVER.md` (`:718`, `:523-528`, `:710`, `:767-770`, `:804`), which now asserts several things that are false about shipped code.

## Verification

- `jest src/lib tests/security` — **1512 passed**, 97 suites.
- `jest src/lib/membership` — 158 passed.
- `npm run lint` clean; `tsc` clean apart from two pre-existing `@aws-sdk/client-lambda` errors in unrelated `s-read` files.
- No `checkin-app/src/security/` change.

### DIFF
--- a/checkin-app/src/lib/membership/mergeBgAdvance.ts
+++ b/checkin-app/src/lib/membership/mergeBgAdvance.ts
@@ -3,7 +3,7 @@ import { logBackendError } from "@/lib/logger";
 import { fromWhere } from "@/lib/membership/lifecycle";
 import { advanceExternalIfComplete } from "@/lib/membership/external";
 import { householdBgIsFresh, nextBoundary } from "@/lib/membership/renewal";
-import { applyVolunteerStatus, clearBackgroundCheck, notifyClearanceOutcome } from "@/lib/membership/review";
+import { applyVolunteerStatus, clearBackgroundCheck, notifyClearanceOutcome, notifyPaymentOpenOnExistingCheck } from "@/lib/membership/review";
 
 /**
  * A participant merge moves the newer of the two `lastBackgroundCheck` dates onto
@@ -58,7 +58,7 @@ export async function advanceHouseholdBgAfterMerge(
 
         await stampPendingExternal(householdId, actorId, source);
         await stampParallelPayment(householdId, actorId, source);
-        await clearHeldReview(householdId, actorId);
+        await clearHeldReview(householdId, actorId, mergedPersonId);
     } catch (error) {
         await logBackendError(error, "membership merge background-check advance", { householdId, mergedPersonId });
     }
@@ -144,7 +144,7 @@ async function stampParallelPayment(householdId: number, actorId: number, source
  * moves its status, leaving no exit but archival. So run the real clearance, which
  * stamps and converges in one write. Zero attestations, re-read under the lock.
  */
-async function clearHeldReview(householdId: number, actorId: number): Promise<void> {
+async function clearHeldReview(householdId: number, actorId: number, mergedPersonId: number): Promise<void> {
     const from = fromWhere("PENDING_BG_REVIEW");
     for (const process of await uncleared(householdId, from)) {
         const outcome = await prisma.$transaction(async (tx) => {
@@ -155,8 +155,12 @@ async function clearHeldReview(householdId: number, actorId: number): Promise<vo
                 select: { status: true, bgClearedAt: true, _count: { select: { attestations: true } } },
             });
             if (!fresh || fresh.status !== "PENDING_BG_REVIEW" || fresh.bgClearedAt || fresh._count.attestations > 0) return null;
-            return clearBackgroundCheck(tx, process.id, actorId);
+            return clearBackgroundCheck(tx, process.id, actorId, undefined, {
+                via: "merge",
+                sourcePersonId: mergedPersonId,
+                reason: "still-valid background check carried in by participant merge; no reviewer attested",
+            });
         });
-        if (outcome) await notifyClearanceOutcome(outcome);
+        if (outcome) await notifyClearanceOutcome(outcome, { paymentOpenNotice: notifyPaymentOpenOnExistingCheck });
     }
 }
--- a/checkin-app/src/lib/membership/review.ts
+++ b/checkin-app/src/lib/membership/review.ts
@@ -356,7 +356,7 @@ export async function attest(
  * than counting attestations — a BLOCKED process carries at most one APPROVE per
  * subject, so counting there always yields nobody.
  */
-export async function clearBackgroundCheck(tx: TxClient, processId: number, actorId: number, subjectOverride?: number[]): Promise<ClearanceOutcome> {
+export async function clearBackgroundCheck(tx: TxClient, processId: number, actorId: number, subjectOverride?: number[], provenance?: ClearanceProvenance): Promise<ClearanceOutcome> {
     const process = await tx.orgMembershipProcess.findUnique({
         where: { id: processId },
         include: { attestations: true },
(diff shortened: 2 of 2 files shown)


thpr (commented):
Reviewed the commit that is this PR's own (`4fc6b8ac`, two files); the rest of the file list is #1595's.

The three corrections land. The new copy is true on the path it fires from, both new parameters default to the old behaviour so the reviewer path is genuinely untouched, and `clearHeldReview` still runs under the FOR UPDATE + re-read CAS. Checked the sibling sends while I was in there: `sendCongrats` claims nothing about a review, and neither `submitIntake`'s fresh-check shortcut nor `advanceExternalIfComplete` mails the family at all — so `notifyPaymentOpen` was the only reachable place the false sentence could reach a household. The scoping is right.

One finding worth blocking on, two small ones.

## 1. The new reviewer ping fires when there is nothing to review

`mergeBgAdvance.ts:52-57` — `notifyReviewers()` sits outside `noteHeldDisclosure` and is unconditional. The branch it lives in is gated on household facts only: `householdBgIsFresh` plus a non-empty `intakeNotes`. Neither asks whether a process exists, and `noteHeldDisclosure` loops `uncleared(householdId, fromWhere("PENDING_BG_REVIEW"))`, which can legitimately return nothing.

So an already-ACTIVE household whose lead holds a current check and whose intake notes are still on file — the ordinary post-intake state, `intakeNotes` is never cleared — makes any participant merge into that household email **every** reviewer and board member "An application is ready for your background-check review", linking to a queue that shows them nothing. Once per merge.

That is the same failure this PR exists to fix, pointed at the other audience: a notice whose subject line asserts an event that did not happen. It also undercuts finding 3's own premise — a ping that regularly fires on an empty queue stops being a surface anyone watches, which is the thing an audit row was faulted for.

The fix is the count the loop already has:

```ts
if (household.intakeNotes?.trim()) {
    // The queue is why this row stays parked, so the queue is told the
    // evidence changed. An audit row is not a surface anyone watches.
    if (await noteHeldDisclosure(householdId, actorId, source)) await notifyReviewers();
    return;
}
```

with `noteHeldDisclosure` hoisting its `uncleared(...)` result into a const and returning `processes.length > 0`.

## 2. `provenance?: object` is wider than what it writes

`review.ts:362` and `:409`. `object` accepts any shape, and the spread lands *after* `status` / `bgCleared` / `clearedPersonIds`, so a later caller can overwrite the audit row's own verdict — write `bgCleared: false` onto a real clearance — with no type error and no test failure. Naming the shape (`{ via: string; sourcePersonId: number; reason: string }`) or spreading it first closes it for free, and the annotation doubles as the documentation of what an auditor should expect to find in the row.

## 3. None of the three changes has a test

The commit touches no test file. Each change is a one-assertion addition to a test that already exists in #1595's suite:

- The provenance keys: `"clears a PENDING_BG_REVIEW row with no attestations"` already reloads the row; the audit `toMatchObject({ via: "merge", sourcePersonId })` next to it is the same assertion `"attributes the carryover to the merging board member"` already makes for the stamp path. Right now that attribution test covers `stampBgCleared` only, so the clearance path — the one this PR just gave provenance to — is the one path with none.
- The ping and the new subject line: `sendEmail` is already mocked at the top of that suite, so both are assertable on the mock's calls, and the assertion for #1 above is "the note-held branch with no held process sends nothing".

Without them the copy can drift back to `notifyPaymentOpen` on a later refactor and nothing notices, which is precisely the regression class this PR is correcting.

thpr (changes requested):
Marking the review above as request-changes: finding 1 (the unconditional `notifyReviewers()` in the note-held branch) ships a false "an application is ready for your background-check review" to every reviewer and board member on any merge into a note-bearing household with no held process — the same class of untrue notice this PR fixes for families. Findings 2 and 3 are non-blocking.

thpr (approved):
All three addressed in 583aa041.

The ping now follows `noteHeldDisclosure`'s own count, so it fires only at a queue that has the row to point at, and the comment states the reason rather than the mechanism. `ClearanceProvenance` closes the audit-row spread. The unit suite pins the note-held branch in both directions without a database — held row → one audit plus one ping, no held row → nothing sent — and asserting `logBackendError` was not called means a silent throw cannot fake either pass. The integration test now covers the clearance's attribution and the payment-open subject, which is what would catch the copy drifting back.

Nothing further from me.

PR #1600, opened 2026-08-09
chore(fr7): finish the Fee/FeePayment sweep the drop leaves behind

jee7s:
Finishes the `Fee`/`FeePayment` sweep that **#1444** leaves behind. Found by an adversarial pass over that PR.

**Merge #1444 first.** Every change here is only correct once the tables are gone — deleting the CI fixture earlier would drop real coverage, and the runbook entry it removes is still true until then. Targets `main` so CI actually dispatches; the changes are independent of #1444's diff, so there is no conflict either way.

## The one that matters — a silent, permanent CI regression

`.github/workflows/migration-safety.yml` seeds rows into `Fee` and `FeePayment` so a destructive migration on them fails in CI rather than on dev. #1444 touches **zero** files under `.github/`.

That job builds its base database from `merge-base(origin/main, HEAD)`. Once the drop is on main, **every later PR that adds a migration** runs:

```
INSERT INTO "Fee" (...)          →  42P01: relation "Fee" does not exist
INSERT INTO "FeePayment" (...)   →  42P01
```

And it **stays green**. The heredoc runs bare `psql`, whose documented exit status is 0 unless `ON_ERROR_STOP` is set — so the job prints two errors and passes, permanently, in the one place whose entire purpose is being trustworthy about destructive migrations.

This is verbatim the class `DEPLOY_MIGRATION_ORDER_OF_OPERATIONS.md` rule 2 leads with: *"Shell scripts with raw SQL — PR #917's own cutover sweep missed `full_reset_and_dev_init.sh`'s raw `INSERT INTO "HouseholdLead"`."*

Both inserts go. `ON_ERROR_STOP=1` goes in, so the next fixture referencing something that no longer exists fails loudly instead of rotting quietly. **The `Event` and `RSVP` top-ups carry the 2026-07-02 incident coverage and are untouched** — RSVP depends only on `Event`, so nothing real is lost.

## The rest

**`checkin-app/tests/security/scopeValidators.test.ts`** asserted `expect(errs).not.toContain('Fee.participantId — no such field')`. With no `Fee` model that can never be flagged, so the assertion is true regardless of the code — coverage weight for a table that cannot be populated, which is the grievance #354 was filed about, surviving in the file that checks the bindings. The RSVP half is real and stays. #1444 rewrote this exact Fee/RSVP narrative in three sibling files and missed this one in the same directory.

**`docs/ops/legacy-cleanup.md`** is an **on-call runbook** stating *"the models still exist and are still referenced… the tables are live in the schema today."* False the moment #1444 merges. The file's own opening rule is *"When an entry's count reaches zero, delete the entry"*, and the exit condition the entry states is exactly what shipped.

**`docs/backlog/INDEX.md`** — the FR7 row still called the `DROP TABLE` "the remaining release" and pointed at `PROGRAM_PAYMENT_VISIBILITY.md`, which does not exist on main.

**`docs/security/auth-consistency-analysis.md`** — one clause, "Fee is unbound".

**`docs/in-design/TOMBSTONE_REMOVAL.md`** — `feePayment` in the merge-repoint list (already wrong before the drop; #1404 removed that write) and a disposition row now moot.

**`docs/designs/PROGRAM_INSTANCE_RESTRUCTURE.md` is annotated, not rewritten.** Its M2 step repoints `Fee`/`FeePayment` to `ProgramInstance` and now has nothing to repoint — but it is a live proposal owned elsewhere, and reworking its migration plan is its owner's call, not the drop's.

## Not addressed here, worth stating

The drop caps rollback at exactly one release, and nothing says so. `v1.1.1` is clean; `v1.1.0` still carries `include: { fees: true }` in `GET /api/programs/[id]`, and `rel/1.0`'s merge route has 8 `feePayment` references. There is no down-migration anywhere in the repo and no snapshot step in `deploy-prod.yml`. The deploy doc does not require one, so #1444 is compliant — but the constraint is new and undocumented.

## Verification

- `jest tests/security/scopeValidators` — 13/13.
- `npm run lint` clean.
- `migration-safety.yml` parses as YAML.
- No `checkin-app/src/` change at all; docs, one test, one workflow.

Relates to #354.


thpr:
Two things from reading the diff, neither blocking.

**`docs/backlog/CUJS.md` is in this sweep's blast radius and isn't in it.** Two stale model references survive, one directory over from the `INDEX.md` row this PR does fix:

- `CUJS.md:60` — "Set capacity/fee/dates (create form + PATCH; `Fee` model)". `Fee` is gone; the mechanism is `Program.orgMemberPriceCents` / `nonOrgMemberPriceCents`.
- `CUJS.md:127` — cites "only a free-text `FeePayment.quickBooksInvoice` column" as the evidence for zero QuickBooks integration. The conclusion survives; the column doesn't.

CUJS is the doc flow tests get chosen from, so a stale mechanism citation there costs more than a doc nit.

**`ON_ERROR_STOP=1` hardens a constraint the new comment doesn't state.** The fixture runs against the *base* schema, before the PR's own migrations apply. A fixture line naming a table that doesn't exist *yet* previously printed an error and passed; now it fails the job. So a future PR that adds a table and a fixture row for it in the same PR will fail migration-safety for a reason that looks unrelated to its diff. That's the right failure mode, but it deserves a clause in the block comment — the added text only explains the drop direction, not the "base schema only" rule.

thpr (changes requested):
Requesting changes on the two items in https://github.com/innovationtreehouse/checkin/pull/1600#issuecomment-5229453669 — the missed `docs/backlog/CUJS.md` references, and the unstated "base schema only" constraint that `ON_ERROR_STOP=1` now enforces.

The core fix is right and verified: bare `psql` exits 0 on error, the base DB is built at merge-base (post-drop since #1444 landed as fa21c636), so without this the fixture would print two 42P01s and pass forever. `Event`/`RSVP` coverage is intact and the `scopeValidators` assertion it drops was vacuous once the model left the schema.

thpr (approved):
Both items addressed in 5fde4d88, and addressed correctly.

- `CUJS.md:60` now names `Program.orgMemberPriceCents`/`nonOrgMemberPriceCents` — the real mechanism.
- `CUJS.md:127`'s dead-column evidence is replaced with "no QB model, field, client, or credential in the schema or `src/`". I checked: zero `quickbooks|qb` hits across `schema.prisma` and `checkin-app/src`, so the claim holds and no longer depends on a column that was dropped.
- The fixture comment now states the base-schema constraint outright, including that a PR adding a table cannot seed it there. That was the exact trap I was worried `ON_ERROR_STOP` would spring silently.

Remaining `Fee` references in `PROGRAM_INSTANCE_RESTRUCTURE.md` are the deliberate annotate-don't-rewrite call, consistent with the note at the top of that file.

Nit, non-blocking: `docs/backlog/INDEX.md:80` (the **P29** row) still points at `docs/designs/PROGRAM_PAYMENT_VISIBILITY.md`, which does not exist on this branch. This PR removed that same dead pointer from the FR7 row at :131 and left this one. Worth folding in while the file is open, or splitting off — either way it doesn't hold up the merge.

PR #1605, opened 2026-08-09
fix(membership): close the two background-check clearance gaps

jee7s:
Two release-blocking gaps in the background-check tracks. They share a root — clearance now covers one named adult instead of a household — so they ship together.

> **Revised after review (c19d3d2a).** thpr's blocking finding was correct and the fix for defect 1 has been replaced, not patched. The full reasoning is in "Which path was taken, and why" below. The short version: the earlier revision made a subject-less approval count toward the adult a second reviewer named. That minted a clearance backed by one reviewer's reading, with no provenance on the audit row, invisible to the `blanketStamped` worklist, and suppressing the next real check. **It is gone.** The genuine half — `attest()` answering `approvals: 2` while nothing cleared — is fixed by keeping the reported count subject-scoped. Defect 2's bound was also tightened past the review's own suggestion; see below.

## Defect 1 — `attest()` reported a clearance that had not happened

**Reproduction** (real Postgres, `bgPerAdultSubject.integration.test.ts`): a two-lead household INITIAL at PENDING_PAYMENT with `bgConsentAt` set, carrying one APPROVE whose `subjectPersonId` is null. A second eligible reviewer approves and names the lead.

On `main`: `attest()` returns `{ status: 'PENDING_PAYMENT', approvals: 2 }` — it tells the reviewer two approvals are recorded — while `bgClearedAt` stays null and the named lead's `lastBackgroundCheck` stays null. `approvals` counted every APPROVE on the process; `clears` required two approvals on the *same named* subject. The two could disagree, and did.

Null-subject approvals are not a corner case: migration `20260802120000` adds the column nullable with no backfill, so this is what a review in flight across that deploy looks like.

**The fix.** `approvals` is derived from the same subject-scoped count that decides `clears`, so the two cannot disagree. A reported 2 always means cleared; the reviewer in this scenario is told **1**, which is true — one person has read that adult's report.

`approvalsForSubject` counts **named** approvals only. An approval that named nobody counts toward nobody, including toward the adult a later reviewer names. That preserves what the column's own migration comment already promised ("its rows read as legacy — clearance stamps nobody for them") and what `docs/rules/membership.md` already stated ("Whoever approves first names the person whose report they read, and a later reviewer confirms that same person").

**How such a review finishes.** The board resets it and two reviewers redo it. `reset` reaches a review that is still in progress, not only a BLOCKED one (`review.ts:508`), and the board can see the process on `/api/membership-ops/applications`, which returns every non-ACTIVE process with its attestations. With three or more non-cohabiting reviewers a third simply approves the named adult and it clears normally; the deadlock binds only when the pool is effectively two. A new integration test walks that recovery end to end.

## Which path was taken, and why

Two paths were on the table. **This PR takes the first.**

| | Path A — refuse the unnamed approval (**taken**) | Path B — count it, with provenance |
|---|---|---|
| In-flight cohort | board reset + two fresh attestations | clears with no redo |
| Record left behind | two real attestations, one named adult | one reviewer's reading, tagged `via: "unnamed-approval"` |
| Rule change needed | none — the rule already says this | an exception written into `membership.md` |

Path B was rejected on the evidence, not on taste:

- **The redo is small and bounded.** The subject column landed 2026-08-02 with no backfill. The affected set is household reviews holding exactly one APPROVE at that moment that are *still uncleared* — a week's worth of half-attested applications. Anything that already had two approvals cleared under the old counting before the column existed.
- **The dead end it was meant to escape is not one.** The board sees the process and can reset it, and reset reaches an in-progress review. Both were verified on `origin/main`, and the reset path already has a test (`membershipBgNonBlocking.integration.test.ts:361`).
- **Path B reaches an outcome `overrideBlocked` explicitly refuses the board** — "`approve` stays BLOCKED-only — force-clearing a review still open to its second reviewer is what the two-reviewer rule forbids" (`review.ts:479`). The board has full authority, a named subject, a conflict-of-interest gate and an audit trail, and still may not do it. Counting would have done it with no board involvement and no record.
- **Its own defence relied on the artifact we already called untrustworthy.** "The legacy approval covered a household check that blanket-stamped both leads, so attributing it to one adult is strictly narrower" takes the blanket stamp as evidence — the exact thing `blanketStamped` exists to remediate.
- **The three compounding harms are all real**, verified against `origin/main`: `clearBackgroundCheck` was called without `provenance`, whose own comment says the row is then indistinguishable from a real two-reviewer clearance (`mergeBgAdvance.ts:158` passes it for exactly this situation); `blanketStamped` filters on `attestations.every(a => a.subjectPersonId === null)` and `stamped.length > 1` (`compliance/route.ts:195`), and a Path-B clearance carries one *named* attestation and stamps one lead, so it fails both; and `householdBgIsFresh` reads `lastBackgroundCheck`, so the stamp suppresses the next real check for `bgRecheckMonths`.

Because Path A is what the rule already says, no exception is written into `membership.md`. One bullet is **added** recording the decision positively, so the next reader does not have to re-derive it.

## Defect 2 — the second signing adult belonged to no track

**Reproduction** (`personBgTriggers.integration.test.ts`): a household with two leads, neither in any program. The review names and stamps one. On activation, `openPersonBgForNewMember` runs.

On `main` the unnamed lead gets `personBgCountFor(...) === 0` — no stamp and no PERSON_BG row. Both openers filter on `PROGRAM_ATTACHED_WHERE`, and the parent who signs, drops off, and may hold a keyholder role is in no program.

**The fix.** A new `BG_OBLIGATED_WHERE` in `lib/person/filters.ts` — program-attached **or** a signing adult of an **ACTIVE member household** — used by both openers and by the compliance worklist.

**The membership bound, and why it is tighter than the review's suggestion.** The review is plainly right that a bare `{ isHouseholdLead: true }` is unbounded: `runPersonBgAnnualSweep` runs it over all Persons on a daily cron, and `isHouseholdLead` marks the lead of every household the app creates. Its suggested bound was `household: { orgMembership: { isNot: null } }`. Verified against `origin/main`, that is too loose for the very population the finding names:

- `startIntake` **anchors an OrgMembership at status NONE** the moment a household begins an application (`intake.ts:157`), so every abandoned intake has a row.
- Denying a household writes `status: "DENIED"` (`households/route.ts:251`); revocation writes `REVOKED`.
- The bulk import creates them wholesale (`participants/import/route.ts:111`).

`isNot: null` would still sweep in the leads of abandoned, denied and revoked applications — the exact people the finding says should owe nothing. The bound taken is `orgMembership: { status: "ACTIVE" }`, which mirrors the sibling agreement track's `autoPopulationWhere` (`personAgreementTriggers.ts:188`) and matches the rule the obligation states: *activating a membership* opens it.

Trigger C is safe under this bound on both activation paths — `payment.ts:215` and `clearBackgroundCheck` each set `OrgMembership.status = "ACTIVE"` inside the transaction, before the opener runs after commit.

**Why not stamp both leads.** That is what the narrowed stamp exists to prevent. *"A check covers the adult who took it. One person's clearance never satisfies another's."* The lead the review named is stamped and therefore FRESH, so `openPersonBg`'s dedup guard skips them — only the unnamed lead gets an obligation.

**Why the compliance route is in this diff.** `peopleNeedingBgCheck` is the only surface with the "Record external check & submit" button, i.e. the only way a PERSON_BG ever reaches `bgConsentAt` and becomes queue-visible. It inlined the same program-attached OR. It now shares `BG_OBLIGATED_WHERE`, so what the triggers open is what the board can act on.

## Defect 3 — board-facing copy described the old population

`compliance/route.ts` (the header list and the section-4 comment) and the user-visible `compliance/page.tsx` both said "program-attached". A board member would have read that over a list that now includes household leads. Both now name the real population.

## Proof each test fails without the fix

Run against a Docker `postgres:15`, tests held fixed while the source is swapped.

**Against `origin/main`'s source** — 3 failed, 27 passed:

| test | failure |
|---|---|
| `an approval naming nobody counts toward nobody, and reports the count it counted` | `Expected: 1 / Received: 2` — the reported symptom exactly |
| `Trigger C covers the signing adult the household clearance did not name` | `Expected: 1 / Received: 0` |
| `Trigger A covers a member household lead with no program attachment` | `Expected: 1 / Received: 0` |

**Against this PR's own previous revision (755b2434)** — 3 failed, 27 passed, i.e. the review's findings are now regression-tested:

| test | failure |
|---|---|
| `an approval naming nobody counts toward nobody…` | `bgClearedAt` set — the one-reviewer clearance |
| `a review holding an approval that named nobody exits by board reset…` | reset rejected: the review had already been force-cleared |
| `Trigger A leaves alone the leads of households that never became members` | `Expected: 0 / Received: 1` — the unbounded lead arm |

Guards (green both ways, labelled as guards in-file): nobody but the named adult is ever stamped; the board-reset escape hatch reaches an in-progress review and two fresh attestations then clear it.

## Verification

- `npm ci` from clean, then `npx prisma generate`.
- `npx tsc --noEmit --pretty false` — **no errors** (the two `@aws-sdk/client-lambda` errors reported on the previous revision were stale `node_modules`, not real).
- `npm run lint` — clean.
- `npm run test:ci` — **276 suites, 2691 tests, all pass.** No suite fails to load.
- `npm run test:integration` against Docker `postgres:15` — **136 suites, 1444 tests, all pass.**

## Not verified

- No manual UI pass. The compliance page renders `{p.programName && ...}`, so a lead with no program degrades to "Household #N" — read, not exercised.
- Production volume of the widened worklist is still unmeasured, but it is now bounded by ACTIVE membership rather than by nothing: `peopleNeedingBgCheck` and `peopleMissingDob` grow by the signing adults of current member households who are stale or have no recorded age. Both are warn-only and gate nothing. The first `person-bg-annual` cron run after deploy opens a backlog for those leads; idempotent, so re-runs add none.
- The migration is untouched; nothing here needs a backfill or a schema change.

## Not undone

The four items the review checked and found correct are unchanged: `attest`'s `clears` and `clearBackgroundCheck`'s stamp still cannot disagree; two approvals still means two distinct, non-cohabiting reviewers; no new null-subject household approvals are creatable; and `submitPersonBgForReview` remains population-agnostic.

### DIFF
--- a/checkin-app/src/app/__tests__/bgPerAdultSubject.integration.test.ts
+++ b/checkin-app/src/app/__tests__/bgPerAdultSubject.integration.test.ts
@@ -12,7 +12,7 @@
 import { POST as ATTEST } from '@/app/api/membership/reviews/route';
 import { POST as OVERRIDE } from '@/app/api/membership-ops/applications/review-override/route';
 import { GET as COMPLIANCE } from '@/app/api/membership-audit/compliance/route';
-import { eligibleReviewProcessIds } from '@/lib/membership/review';
+import { eligibleReviewProcessIds, overrideBlocked } from '@/lib/membership/review';
 import { bgFreshThreshold, personBgVerdict } from '@/lib/membership/personBgCheck';
 import { householdBgIsFresh } from '@/lib/membership/renewal';
 import prisma from '@/lib/prisma';
@@ -218,42 +218,58 @@ describe('Per-adult background-check subjects', () => {
     });
 
     // ── An approval that names nobody ────────────────────────────────────────
-    it('an approval naming nobody counts toward the adult the second reviewer names', async () => {
+    it('an approval naming nobody counts toward nobody, and reports the count it counted', async () => {
         const hh = await twoLeadHousehold('Legacy');
-        // An approval whose subject column is null: what every review in flight across
-        // the subject migration carries, since the column was added without a backfill.
+        // An approval whose subject column is null: what a review in flight across the
+        // subject migration (20260802120000, nullable, no backfill) carries.
         await prisma.backgroundCheckAttestation.create({ data: { processId: hh.processId, reviewerId: rev1, result: 'APPROVE', subjectPersonId: null } });
-        (sendEmail as jest.Mock).mockClear();
 
         as(rev2, { isBackgroundCheckReviewer: true });
         const res = await ATTEST(req({ processId: hh.processId, result: 'APPROVE', subjectPersonIds: [hh.alex] }) as never);
         expect(res.status).toBe(200);
 
-        // Two distinct reviewers approved and exactly one adult was named, so the review
-        // is complete and stamps that adult only.
+        // GUARD. Alex holds one approval, not two — the subject-less row vouches for no
+        // one, so nothing clears and nobody is stamped. Nothing records whose report rev1
+        // read; counting it would date Alex's clearance on rev2's reading alone.
+        const proc = await prisma.orgMembershipProcess.findUnique({ where: { id: hh.processId } });
+        expect(proc?.bgClearedAt).toBeNull();
+        expect(await bgDate(hh.alex)).toBeNull();
+        expect(await bgDate(hh.sam)).toBeNull();
+
+        // The reported defect: attest() answered `approvals: 2` here while nothing
+        // cleared, telling the reviewer the review was done. The count it reports is now
+        // the count that decided `clears`, so the two can never disagree.
+        expect((await res.json()).outcome.approvals).toBe(1);
+    });
+
+    it('a review holding an approval that named nobody exits by board reset, then clears on two real readings', async () => {
+        const hh = await twoLeadHousehold('LegacyReset');
+        await prisma.backgroundCheckAttestation.create({ data: { processId: hh.processId, reviewerId: rev1, result: 'APPROVE', subjectPersonId: null } });
+        as(rev2, { isBackgroundCheckReviewer: true });
+        await ATTEST(req({ processId: hh.processId, result: 'APPROVE', subjectPersonIds: [hh.alex] }) as never);
+        (sendEmail as jest.Mock).mockClear();
+
+        // Both reviewers have spent their one attestation, so neither queue offers it —
+        // but the board sees it on /api/membership-ops/applications and `reset` reaches a
+        // review still in progress, not only a BLOCKED one. GUARD on that escape hatch:
+        // it is why the unnamed approval can be refused rather than counted.
+        expect(await eligibleReviewProcessIds(rev1)).not.toContain(hh.processId);
+        expect(await eligibleReviewProcessIds(rev2)).not.toContain(hh.processId);
+        await overrideBlocked(hh.processId, board, 'reset');
+        expect(await prisma.backgroundCheckAttestation.count({ where: { processId: hh.processId } })).toBe(0);
+
+        as(rev1, { isBackgroundCheckReviewer: true });
+        await ATTEST(req({ processId: hh.processId, result: 'APPROVE', subjectPersonIds: [hh.alex] }) as never);
+        as(rev2, { isBackgroundCheckReviewer: true });
+        await ATTEST(req({ processId: hh.processId, result: 'APPROVE', subjectPersonIds: [hh.alex] }) as never);
+
+        // Two people read Alex's report, and only Alex is stamped.
         const proc = await prisma.orgMembershipProcess.findUnique({ where: { id: hh.processId } });
         expect(proc?.bgClearedAt).not.toBeNull();
         expect(proc?.status).toBe('PENDING_PAYMENT');
         expect(await bgDate(hh.alex)).not.toBeNull();
         expect(await bgDate(hh.sam)).toBeNull();
-        // The clearance owes the family the payment-open notice.
         expect((sendEmail as jest.Mock).mock.calls.some((c: unknown[]) => String(c[1]).includes('you can now pay your membership dues'))).toBe(true);
-    });
-
-    it('leaves no review that reports two approvals with nobody left able to finish it', async () => {
-        const hh = await twoLeadHousehold('NoDeadEnd');
-        await prisma.backgroundCheckAttestation.create({ data: { processId: hh.processId, reviewerId: rev1, result: 'APPROVE', subjectPersonId: null } });
-
-        as(rev2, { isBackgroundCheckReviewer: true });
-        const outcome = (await (await ATTEST(req({ processId: hh.processId, result: 'APPROVE', subjectPersonIds: [hh.alex] }) as never)).json()).outcome;
-
-        // Both reviewers have now spent their one attestation, so the process is gone from
-        // both queues. A reported approval count of 2 has to mean cleared, or the review
-        // is stranded with no one able to act on it.
-        expect(await eligibleReviewProcessIds(rev1)).not.toContain(hh.processId);
-        expect(await eligibleReviewProcessIds(rev2)).not.toContain(hh.processId);
-        expect(outcome.approvals ?? 0).toBeLessThan(2);
-        expect((await prisma.orgMembershipProcess.findUnique({ where: { id: hh.processId } }))?.bgClearedAt).not.toBeNull();
     });
 
     // ── Board force-approve ──────────────────────────────────────────────────
--- a/checkin-app/src/app/__tests__/bgPerAdultSubject.integration.test.ts
+++ b/checkin-app/src/app/__tests__/bgPerAdultSubject.integration.test.ts
@@ -16,6 +16,7 @@ import { eligibleReviewProcessIds } from '@/lib/membership/review';
 import { bgFreshThreshold, personBgVerdict } from '@/lib/membership/personBgCheck';
 import { householdBgIsFresh } from '@/lib/membership/renewal';
 import prisma from '@/lib/prisma';
+import { sendEmail } from '@/lib/email';
 import { getServerSession } from 'next-auth/next';
 
 jest.mock('next-auth/next', () => ({ getServerSession: jest.fn() }));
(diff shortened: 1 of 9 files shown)


thpr (changes requested):
Defect 1's fix can record an adult as background-checked on one reviewer's reading, and the record it leaves is indistinguishable from a genuine two-reviewer clearance. That is the blocking finding; the rest is smaller.

## 1. A clearance backed by one reviewer, with nothing left behind to find it

`approvalsForSubject` counts a null-subject APPROVE toward the adult the second reviewer names. So: rev1 approved before subjects existed, naming nobody. rev2 approves naming Alex. Alex reaches two, `clearBackgroundCheck` stamps `Alex.lastBackgroundCheck`.

Exactly one person ever read Alex's report. Nothing in the data says whose report rev1 read — if it was Sam's, Alex is now recorded as checked on nobody's reading of Alex.

The PR's defense is that the legacy approval covered a household check that blanket-stamped every lead, so attributing it to one adult is strictly narrower. That takes as evidence the exact artifact `blanketStamped` exists because we decided it was not real coverage.

Three things then make the outcome unrecoverable:

**No provenance.** `review.ts:329` calls `clearBackgroundCheck(tx, processId, reviewerId)` with `provenance` undefined, and the actor is a reviewer. Per that function's own comment — "provenance is set when something other than a two-reviewer attestation produced this clearance; without it the row is indistinguishable from one" — the audit row cannot be told apart from a real one. `mergeBgAdvance.ts:158` passes provenance for precisely this situation.

**Outside the remediation worklist.** `blanketStamped` filters on `attestations.every(a => a.subjectPersonId === null)` and `stamped.length > 1` (`compliance/route.ts:195`). These clearances carry one named attestation and stamp one lead, so they fail both conditions. The list whose job is untrustworthy stamps structurally cannot see the ones this PR mints.

**The stamp suppresses the next real check.** `householdBgIsFresh` reads `lastBackgroundCheck`, so the next renewal skips a fresh check for `bgRecheckMonths`.

It also reaches an outcome the code explicitly forbids the board from reaching. `review.ts:479`, on the override: "`approve` stays BLOCKED-only — force-clearing a review still open to its second reviewer is what the two-reviewer rule forbids." The board — full authority, an explicitly named subject, a conflict-of-interest gate, an audit trail — may not clear a review holding one approval. Counting now does it with no board involvement and no record that it happened.

**The dead-end premise is also overstated.** The stranded process is invisible to the two reviewers, but not to the board: `/api/membership-ops/applications` returns every non-ACTIVE process with its attestations and the household leads a board action selects from (`applications/route.ts:22`), and `reset` reaches a review that is still in progress, not only a BLOCKED one (`review.ts:508`). With three or more non-cohabiting reviewers a third simply approves the named adult and it clears normally — the deadlock binds only when the pool is effectively two.

### What I would do instead

The genuine half of defect 1 is that `attest` reported `approvals: 2` while nothing cleared. Making `approvals` subject-scoped fixes that on its own — the reviewer is told 1, and the board resets the review so two reviewers redo it. One-time, bounded, auditable, and no false record.

If the in-flight cohort should clear without a redo, make it a recorded judgment rather than an invisible one. `clearBackgroundCheck` takes `provenance` independently of `subjectOverride` (that is exactly what `mergeBgAdvance` does), so it is one argument:

```ts
{ via: "unnamed-approval", sourcePersonId: reviewerId, reason: "counted an approval recorded before subjects existed; one reviewer read this adult's report" }
```

plus a sibling of `blanketStamped` keyed on that provenance so the board can confirm them. Either way, `docs/rules/membership.md:206` — "Whoever approves first names the person whose report they read, and a later reviewer confirms that same person" — needs the exception written down; defect 2 got its register entry and this one did not.

## 2. `BG_OBLIGATED_WHERE` has no member-household bound

`filters.ts:34` is a bare `{ isHouseholdLead: true }`, and `runPersonBgAnnualSweep` runs it over all Persons on a daily cron (`personBgTriggers.ts:80`).

`isHouseholdLead` is not "signing adult on a membership application". It is the lead marker on every household the app has created: imported legacy households, the unclaimed and broken ones `UNCLAIMED_OR_BROKEN_HOUSEHOLD_WHERE` chases, program-only families, and the leads of DENIED, REVOKED, and abandoned applications. Program attachment is what kept the sweep's unbounded population defensible; the lead arm has no equivalent property.

Each such lead with a date of birth gets a PERSON_BG plus its audit row, closable only by a board member recording an external check and two reviewers attesting — for people who never applied. The rest land on `peopleMissingDob`, which for imported adults is likely most of them.

The sibling track already bounds this: `autoPopulationWhere` carries `household: { orgMembership: { status: "ACTIVE" } }` (`personAgreementTriggers.ts:186`). The rule this PR writes scopes the obligation to "activating a membership ... for its signing adults", and a lead of a household with no membership is neither. Trigger C is bounded by construction; Trigger A is not, so the bound belongs in the filter:

```ts
OR: [PROGRAM_ATTACHED_WHERE, { isHouseholdLead: true, household: { orgMembership: { isNot: null } } }]
```

## 3. Board-facing copy still says program-attached

The population changed; the descriptions did not. `compliance/route.ts:29-31`, `compliance/route.ts:89` ("Subject = union of ProgramParticipant / ProgramVolunteer / Program.leadMentor"), and the user-visible `compliance/page.tsx:306` and `:417`. A board member reads "Program-attached people 18 or older" over a list of household leads.

## Checked, no finding

- `attest`'s `clears` and `clearBackgroundCheck`'s stamp cannot disagree: a second approval naming a different adult is rejected, so `establishedSubject` always resolves to the id `approvalsForSubject` counted.
- Two approvals still means two distinct, non-cohabiting reviewers.
- No new null-subject household approvals are creatable — `review.ts:312` is the only writer and a household APPROVE requires exactly one name — so the legacy allowance is self-limiting.
- `submitPersonBgForReview` is population-agnostic, so the "Record external check & submit" button does work for a lead with no program, as the description says.

---

thpr (approved):
All three findings addressed at c19d3d2a, the recommended way rather than the cheap one.

**The clearance path.** `approvalsForSubject` counts named approvals only, so nothing mints a clearance the record cannot account for — the provenance gap, the `blanketStamped` blind spot and the suppressed next check never come into play. What remains in `review.ts` against `main` is the reporting fix alone: `approvals` is subject-scoped, so a reported 2 always means cleared, which was the genuine defect. `subjectsWithTwoApprovals` is behaviourally identical to `main`'s map version — `attest` rejects a second approval naming a different adult, so `establishedSubject` is the only candidate it could resolve to. The exception is written into the register at `membership.md:211`.

**The lead arm.** Bounded to ACTIVE member households, mirroring `autoPopulationWhere`. Going past my suggested `isNot: null` is the right call: `startIntake` anchors an abandoned application's membership at `NONE`, so the weaker bound would still have swept in abandoned intakes. `OrgMembershipStatus` is NONE/ACTIVE/REVOKED/DENIED with no expiry state, so an ACTIVE membership does not blink off during a renewal and no lead drops out of the population mid-cycle.

**One thing I checked that the ACTIVE bound newly depends on.** `openPersonBgForNewMember` queries on the global client. Had it run inside the activating transaction, the lead arm would read a membership that is not yet ACTIVE and match nobody — defect 2 silently unfixed. It doesn't: both call sites (`payment.ts:236`, `review.ts:428`) are post-commit side effects, and the transaction sets `orgMembership.status = ACTIVE` before returning. The Trigger C test exercises that ordering for real, creating the membership at `NONE` and letting `activate` flip it. Worth keeping in mind if a future change ever moves those triggers inside the transaction.

Tests: the deleted assertions are back as guards, the board-reset escape hatch is covered end to end through `overrideBlocked`, and Trigger A has a negative case over NONE, DENIED, REVOKED and no membership at all. Copy on the route and both page sections now describes the population actually listed.

---

Issue #1583, opened 2026-08-08
Drop the BoardSettings membership-fee @map shims (contract stage)


thpr:
The "dues" → "membership fee" rename shipped the Prisma/API/UI half only. Both
`BoardSettings` fields still carry an `@map` to their original physical column, so
the database columns are still named `normalDuesCents` / `volunteerDuesCents`:

```prisma
standardMembershipFeeCents  Int @default(0) @map("normalDuesCents")
volunteerMembershipFeeCents Int @default(0) @map("volunteerDuesCents")
```

## Why the physical rename was deferred

A column rename is a contract-stage change under
[`DEPLOY_MIGRATION_ORDER_OF_OPERATIONS.md`](checkin-app/docs/DEPLOY_MIGRATION_ORDER_OF_OPERATIONS.md)
rule 3 ("a rename of anything old code touches").

Migrations run to completion *before* `aws ecs update-service` starts the rolling
deploy, and old tasks keep serving live traffic for the whole drain window. About
25 call sites read `BoardSettings` with a bare `findUnique`/`upsert` and no
`select`, so the previous release's Prisma client hand-selects `"normalDuesCents"`
by name. Renaming the columns in the same release would fail every one of those
reads with `42703 undefined column` for the entire drain — nav todo-counts,
intake, renewal, payment, and the background-check triggers all sit on that path.
Same failure shape as #917.

Shipping the `@map` instead makes the rename release schema-compatible with the
release it replaces, at zero migration cost.

## What to do, once the shimmed release is fully rolled out

1. Drop both `@map(...)` arguments and the explanatory comment above them in
   `checkin-app/prisma/schema.prisma` (`model BoardSettings`).
2. Hand-write the migration — never let Prisma generate it, since its diff engine
   has no `RENAME` primitive and emits `DROP COLUMN` + `ADD COLUMN`, which is
   silent data loss on a populated table (rule 4). `BEGIN`/`COMMIT` because a
   multi-statement migration can partially apply on Postgres (rule 5):

```sql
BEGIN;
ALTER TABLE "BoardSettings" RENAME COLUMN "normalDuesCents" TO "standardMembershipFeeCents";
ALTER TABLE "BoardSettings" RENAME COLUMN "volunteerDuesCents" TO "volunteerMembershipFeeCents";
COMMIT;
```

3. `npx prisma generate` and confirm `src/security/generated/classifications.ts`
   is unchanged — the classification keys are Prisma field names, which do not
[...]

PR #1606, opened 2026-08-09
fix(migrations): keep OrgMembershipProcess.certifiedById on disk through the drain window

jee7s:
`20260808120000_org_membership_process_manual_payment_by` (#1586) renames
`OrgMembershipProcess.certifiedById` to `manualPaymentById` with a bare
`ALTER TABLE ... RENAME COLUMN`, in the same release as the code that reads the new
name. That is release-blocking: `deploy-prod.yml` runs `prisma migrate deploy` to
completion *before* `aws ecs update-service`, so the previous release serves live
traffic against the renamed column for the entire drain window.

This PR holds the physical column on its old name with `@map` and adds a
compensating migration that renames it back. The Prisma field stays
`manualPaymentById`; nothing else in the app changes.

## The evidence

On a real Postgres 15: a fresh database migrated to **v1.1.1**'s chain, then main's
19 pending migrations, then **v1.1.1's own generated Prisma client** run against it.

| Probe (v1.1.1 client) | origin/main | this branch |
|---|---|---|
| `orgMembershipProcess.findMany()` | **THROW** — `The column OrgMembershipProcess.certifiedById does not exist in the current database.` | OK — row returned, `certifiedById: 4242` |
| `orgMembershipProcess.findUnique({id:1})` | **THROW** — same error | OK |
| `boardSettings.findUnique({id:1})` | OK | OK |
| `person.findMany({take:1})` | OK | OK |

`boardSettings` reading fine on both sides is the control: those two fields went
through the *same* vocabulary rename one PR earlier (#1585) and shipped `@map`
shims, so they survive the drain window. `OrgMembershipProcess` is the only model
that breaks.

The blast radius is wide because almost none of v1.1.1's ~83 `OrgMembershipProcess`
call sites narrow with a `select` — a bare `findMany`/`findUnique` names every
column in the client's model, `certifiedById` included. The payment and
Shopify-webhook paths are among them.

This branch's own client was also checked against the same database:
`findMany` with and without a `select`, a `where: { manualPaymentById }` filter, and
an `update` of the field all succeed and round-trip the value. New code is unaffected
by the `@map`.

## The rule it violates

`docs/DEPLOY_MIGRATION_ORDER_OF_OPERATIONS.md` **rule 3 — "Contract last, in a
separate, later release"**, which names "a rename of anything old code touches" as
destructive and forbids shipping it alongside the cutover that feeds it. #1586's own
commit message acknowledges being contract-class and ships anyway.

Rule 4 is satisfied by that migration (it is a hand-written `RENAME`, not Prisma's
generated `DROP`+`ADD`) — rule 4 is about not losing data, rule 3 is about
sequencing, and only rule 3 is broken here. `migration-safety.yml` is green on
#1586 for the same reason: it replays migrations on a populated DB, which is not a
sequencing test. The doc says so explicitly.

## Why `@map`

`@map` decouples the logical field name from the physical column, so both releases
read the same disk. Old code selects `certifiedById` and finds it; new code writes
`manualPaymentById` and Prisma translates. The physical rename becomes a
contract-stage change for a release after this one, when nothing reading the old
name is still running.

This is the shape #1585 already used for `BoardSettings.normalDuesCents` /
`volunteerDuesCents`, for the identical reason. Its contract-stage follow-up is
tracked as #1583; `docs/designs/UNFINISHED.md` now lists `manualPaymentById`
alongside those two in the same ledger entry.

## What happens to the offending migration — it is NOT deleted

Deleting the directory is unsafe here, and quietly so. #1586 merged at 01:21 UTC on
2026-08-09 and dev auto-deploys from `main`; commit `08c7b84e` (02:49 UTC) has
`831c7162` as an ancestor and its **Deploy dev** run succeeded, so
`20260808120000` is already applied to dev and its column is already renamed there.

I built a database in exactly that state and tried the deletion:

- `prisma migrate deploy` **passes** — "34 migrations found ... No pending
  migrations to apply." A ledger row whose directory is gone raises nothing.
- But dev's physical column stays `manualPaymentById` while `schema.prisma` now
  expects `certifiedById`, so the *new* client throws
  `The column OrgMembershipProcess.certifiedById does not exist` on every read and
  write of the model. Verified — all four probes throw.

So deletion trades a loud, one-release prod outage for a silent, permanent dev
outage with a green deploy. Editing the file in place is worse: `MIGRATION_COALESCE_FLOW.md`
and rule 6 both note the ledger checksum is a sha256 of `migration.sql`, and editing
a migration already recorded as applied breaks it.

The remedy is a compensating migration,
`20260808120100_org_membership_process_restore_certified_by_column`, a single
`ALTER TABLE ... RENAME COLUMN` back. `20260808120000` is left byte-identical
(`git diff origin/main` on that directory is empty).

Its timestamp sorts it **immediately after** the forward rename rather than at the
end of the chain, per @thpr's review. `migrate deploy` runs the chain sequentially
and unwrapped, and prod's v1.1.1 stops at `20260724120000`, so all four migrations
are pending together there: at the original `20260809120000` the column would have
sat on its new name across `20260808130000_drop_fee_feepayment` **and**
`20260809090000_audit_log_table_timestamp_index` — a non-concurrent `CREATE INDEX`
on `AuditLog` whose duration scales with the table — with v1.1.1 still serving.
That shortened the outage to the index build rather than closing it. Adjacent, the
window is two back-to-back metadata-only `ALTER`s. Out-of-order insertion is safe:
`migrate deploy` applies whatever is pending and no environment has the old name in
its ledger (the migration is unreleased and unmerged), and nothing in the tree
references it.

Behaviour per environment:

- **prod / any fresh DB** (neither applied): both run, net effect zero, column ends
  as `certifiedById`.
- **dev** (forward already applied): the compensating one is the only pending
  migration; verified it applies alone and cleanly.
- **CI drift check**: `prisma migrate diff --from-config-datasource --to-schema
  prisma/schema.prisma --exit-code` after deploying the full chain reports
  "No difference detected".

A rename-forward/rename-back pair in the history is the cost. It is transient — the
next coalesce (`MIGRATION_COALESCE_FLOW.md`) regenerates the baseline from
`schema.prisma`, where the pair collapses to a column named `certifiedById`.

Single-statement migration, so no `BEGIN`/`COMMIT` wrapper (rule 5 applies to
multi-statement files).

## Verification

- `npx prisma validate` — valid
- `npm run lint` — clean
- `npx tsc --noEmit --pretty false` — only the two pre-existing
  `@aws-sdk/client-lambda` errors in `s-read` (environmental, absent dependency)
- `npm test` — 2656 passed, 0 failed; the same two `s-read` suites fail to *load*
  for the same missing module
- `npx prisma generate` leaves `src/security/generated/classifications.ts`
  byte-identical (`@map` changes the column, not the logical field name or its
  `@sensitivity` tier), so no security-boundary file is touched:
  `git diff --name-only origin/main...HEAD -- checkin-app/src/security/` is empty
- No raw SQL depends on the column name — the three `$queryRaw` statements against
  `OrgMembershipProcess` select `id` only; `scripts/` and `prisma/seed.ts` never
  name it

## What I could not verify

- **The real dev database was not inspected.** It is VPC-only. That dev has the
  migration applied is inferred from a successful **Deploy dev** run on a commit
  descended from #1586, plus the workflow's documented "migrate to completion, then
  deploy" shape — not from reading dev's `_prisma_migrations`. If it somehow did not
  apply, the compensating migration would fail there with `column "manualPaymentById"
  does not exist`; worth a glance at the dev migrate-task logs before merge.
- **Prod was assumed not to have it.** #1586 has not been released; prod's last
  release is v1.1.1, whose chain stops at `20260724120000`.
- **The ~83 call-site count is #1586's, not re-counted here.** The failure does not
  depend on it — a bare `findMany` with zero call sites still throws.
- **`#1583`'s issue body still names only the two `BoardSettings` columns.** I added
  `manualPaymentById` to the `UNFINISHED.md` entry that points at it; the issue itself
  needs the third column added by hand (I did not have issue-write scope).
- **The integration and flow suites were not run** — they need a seeded DB and a live
  dev server respectively. The unit suite and the drift check both passed against a
  real Postgres.

### DIFF
--- a/checkin-app/docs/designs/UNFINISHED.md
+++ b/checkin-app/docs/designs/UNFINISHED.md
@@ -125,12 +125,15 @@ and a payment-plan approval. _(VOCAB #17)_
 
 ---
 
-## 🟢 Drop the `BoardSettings` membership-fee `@map` shims (contract stage)
+## 🟢 Drop the vocabulary-rename `@map` shims (contract stage)
 
-"dues" → "membership fee" shipped the Prisma/API/UI half; both fields still `@map`
-to the old physical columns, which a rolling deploy can't rename in the same
-release. Rename them once that release is out — [#1583](https://github.com/innovationtreehouse/checkin/issues/1583).
-_(VOCAB #6)_
+"dues" → "membership fee" (`BoardSettings.standardMembershipFeeCents`,
+`.volunteerMembershipFeeCents`) and "certified" → "manual"
+(`OrgMembershipProcess.manualPaymentById`) shipped the Prisma/API/UI half; all
+three fields still `@map` to the old physical columns, which a rolling deploy
+can't rename in the same release. Rename them once that release is out —
+[#1583](https://github.com/innovationtreehouse/checkin/issues/1583).
+_(VOCAB #6, #17)_
 
 ---
 
--- a/checkin-app/prisma/migrations/20260808120100_org_membership_process_restore_certified_by_column/migration.sql
+++ b/checkin-app/prisma/migrations/20260808120100_org_membership_process_restore_certified_by_column/migration.sql
@@ -2,4 +2,9 @@
 -- `manualPaymentById` and reaches it through @map, so the previous release —
 -- which serves traffic against this schema for the whole rolling-deploy drain
 -- window — keeps reading OrgMembershipProcess (rule 3).
+--
+-- Timestamp deliberately sorts immediately after 20260808120000 (the forward
+-- rename), not at the end of the chain: `migrate deploy` runs the chain
+-- sequentially and unwrapped, so anything in between — a DROP, an index build —
+-- leaves the column on its new name for that long, and old code 500s meanwhile.
 ALTER TABLE "OrgMembershipProcess" RENAME COLUMN "manualPaymentById" TO "certifiedById";
--- a/checkin-app/prisma/migrations/20260809120000_org_membership_process_restore_certified_by_column/migration.sql
+++ b/checkin-app/prisma/migrations/20260809120000_org_membership_process_restore_certified_by_column/migration.sql
@@ -1,0 +1,5 @@
+-- Puts the column back on its original name. The Prisma field stays
+-- `manualPaymentById` and reaches it through @map, so the previous release —
+-- which serves traffic against this schema for the whole rolling-deploy drain
+-- window — keeps reading OrgMembershipProcess (rule 3).
+ALTER TABLE "OrgMembershipProcess" RENAME COLUMN "manualPaymentById" TO "certifiedById";
--- a/checkin-app/prisma/schema.prisma
+++ b/checkin-app/prisma/schema.prisma
@@ -502,8 +502,11 @@ model OrgMembershipProcess {
   shopifyOrderId        String?
   /// @sensitivity:internal
   paidAt                DateTime?
+  // The @map holds the physical column on its old name: most reads of this model
+  // carry no select, so the previous release reads it through a rolling deploy's
+  // drain window. Rename it in a later release (contract stage).
   /// @sensitivity:internal
-  manualPaymentById     Int?
+  manualPaymentById     Int?                    @map("certifiedById")
   /// @sensitivity:internal
   certificationNote     String?
   /// Write-never as of the outreach engine (PR-2): the renewal sweep and the go-live


dkaygithub (approved):

thpr (commented):
The `@map` shape is right, and the surrounding checks hold up — no FK or index rides on the column (the model has no relation field for it, only `@@index([orgMembershipId])` / `[status]`), the three `$queryRaw` statements are `SELECT id ... FOR UPDATE`, and the generator keys off DMMF `field.name` plus `///` docs, so neither the `//` comment nor the `@map` can move `classifications.ts`. v1.1.1 really does stop at `20260724120000` and has no runtime `prisma.fee` caller, so `drop_fee_feepayment` is safe and `OrgMembershipProcess` is the only model that breaks.

One thing the timestamp gives away, though.

## The restore sorts last, so the window is still open for the length of a `CREATE INDEX`

`20260809120000` lands after `20260808130000_drop_fee_feepayment` and `20260809090000_audit_log_table_timestamp_index`. `migrate deploy` runs those sequentially with nothing wrapping the chain — `drop_fee_feepayment`'s own header says as much ("Prisma does not wrap migration files in a transaction on Postgres"), which is why it hand-writes `BEGIN`/`COMMIT` for itself.

So prod's sequence is: rename forward (commits) → drop two tables → build a non-concurrent index on `AuditLog` → rename back. `deploy-prod.yml` runs the whole chain to completion at line 241 and only reaches `aws ecs update-service` at line 307, so v1.1.1 is serving traffic throughout. For the entire `AuditLog` index build — a duration that scales with the table, not a constant — every bare `findMany` / `findUnique` on `OrgMembershipProcess` throws, payment and Shopify-webhook paths included. That is the same outage the PR removes, shortened from the drain window to the index build rather than closed.

The fix is free: rename the directory to `20260808120100_org_membership_process_restore_certified_by_column` so the pair is adjacent, and the window collapses to two back-to-back `ALTER`s. Out-of-order insertion is fine for dev — `migrate deploy` sorts *pending* migrations by name and applies them; it only fails on applied-but-missing directories or a checksum mismatch, neither of which a new file triggers.

---

thpr (approved):
The renumber closes it. I re-checked the out-of-order claim I made in my earlier comment against a real Postgres 15 rather than leaving it as an assertion, since dev is the environment it would have bitten:

| Check | Result |
|---|---|
| `main`'s chain on an empty DB (reproduces dev) | column is `manualPaymentById` |
| then this branch's chain — the new migration sorts *before* three already-applied ones | applies, no warning and no error; `20260808120100` is the only pending one; column back to `certifiedById` |
| `_prisma_migrations` afterwards | `20260808120100` recorded last, `started_at` after `20260809090000` — position in the directory doesn't matter, only presence |
| this branch's full chain on an empty DB (prod / any fresh DB) | column is `certifiedById` |
| `migrate diff --from-config-datasource --to-schema prisma/schema.prisma --exit-code`, both databases | "No difference detected", exit 0 |

`migrate deploy` sorts *pending* migrations by name and ignores where they fall relative to applied rows, so dev takes the restore alone and cleanly and prod nets to zero. With the pair adjacent, prod's exposure is two back-to-back `ALTER`s instead of an `AuditLog` index build.

## Non-blocking: dev takes the mirror-image break during its own drain window

Not a reason to hold this, and the renumber doesn't change it either way — recording it so it isn't mistaken for a regression when it shows up.

Dev already has `20260808120000` applied and is serving a build from `main`, which has no `@map` and so reads column `manualPaymentById`. `deploy-dev.yml` has the same shape as prod — `migrate deploy` at line 155, `update-service` at line 219 — so this release renames the column to `certifiedById` while the *current* dev task is still up, and dev 500s on `OrgMembershipProcess` until the new tasks take over. Self-healing, dev only, and unavoidable for the release that introduces the shim; the "applies alone and cleanly" line in the description is a statement about the SQL, which is true.

Approving.

---

PR #1608, opened 2026-08-09
fix: two oversight surfaces that stay green while broken (cron health, trends hours)

jee7s:
Two monitoring surfaces that stayed reassuring while the thing they watch was broken.

## Defect 1 — cron health stayed green through a total failure

**Reproduction.** `cron/nightly` wraps each abandoned-visit checkout in
`Promise.allSettled`, logs every rejection, and returns
`{ success: true, facilityClose: { checkedOutCount: 0 } }` with status 200 — even
when every single checkout rejected. `withCron` judged the run by
`res.status < 400` alone, so `recordCronRun(job, startedAt, true)` wrote a fresh
success into `CronRunLog`, `getCronJobStatuses` reported the job healthy with a
new `lastSuccessAt` and no `lastError`, and the board's red System Status pill —
the app's only signal that the sweeps still run — stayed green. Covered by the
new integration test: two open visits, both checkouts forced to reject, sweep
closes nobody, ledger row `success: true` with no error on `main`.

**What I chose and why.** The response body already carried counts, so the honest
signal was half-written. A route now reports the rows it could not process as a
top-level `failed`, and `withCron` reads that count off the response body.

The ledger records **two** things, not one, because "did it run" and "did it work"
are different questions:

| column | means | drives |
|---|---|---|
| `success` | the run COMPLETED | `lastSuccessAt`, `stale`, "this job stopped running" |
| `error` | the run was not clean | `lastError`, the red line in the panel |

A partial sweep therefore records **`success: true` WITH an error**. No migration:
`CronRunLog` already had both columns, so the distinction was expressible in the
schema as it stands — only the two `///` doc comments needed to say so.

The response the caller receives is untouched — a partial sweep still answers 200
and still did its partial work, and infra's retry semantics do not change; only
the health verdict does.

Alternatives rejected: returning 5xx on inner failure (turns a partial sweep into
a retry/alarm storm and hides the counts), and a per-route ledger call (every new
cron route would have to remember it — the same reason `withCron` owns the
recording today).

**Every route under `api/cron/` was checked**, not just the one reported:

| route | swallows rows? | change |
|---|---|---|
| `nightly` | `Promise.allSettled` | reports `failed` |
| `scholarship-grace-expiry` | per-row `try/catch` | reports `failed` |
| `reconcile-shopify` | per-order `try/catch` in `runReconcile` | `ReconcileResult.failed`, spread into the body |
| `lifecycle-reconcile` | per-row `try/catch` | none — see below |
| `pending-participants`, `post-event`, `membership-renewals`, `person-bg-annual`, `trusted-adult-expiry` | no | none needed: a throw propagates to `withCron`, which already records a failure |

`lifecycle-reconcile` isolates rows too, but a failed I1 heal leaves the row
off-diagram, so the scan that follows re-finds it and reports it to the
sysadmin/board channel as a violation. It is not a blind spot, so it is left
alone rather than grown a second signal.

### Review round 2 — why the two signals are split rather than collapsed

The first version of this PR recorded a partial sweep as `success: false`. That
was wrong in the other direction, and @thpr caught it: `lastSuccessByJob()`
aggregates only `success: true`, so a single **permanently**-failing row — a visit
whose `processVisitCheckout` always throws, or reconcile's cursor-frozen order,
which is *by design* re-scanned every run until it heals — would freeze
`lastSuccessAt` forever. After 48h the job reads as stale, the panel row goes red,
and `countStaleCronJobs`' badge says **"N job(s) not running"** about a job that
ran last night and swept everything else. One silent failure traded for another.

A job that ran, completed, and processed 9 of 10 rows is not a job that did not
run, so it does not get to say so. Both readers were changed to keep the two
apart:

- **`getCronJobStatuses`** takes `lastError` from the **latest run**, instead of
  the latest row with `success: false`. An error on a completed run now surfaces.
  This also *deleted* the "failed since last success" timestamp comparison — it is
  implied once the row being read is the latest one.
- **`countStaleCronJobs` → `countUnhealthyCronJobs`**, counting stale **OR**
  erroring. This half is not optional: without it a nightly sweep that fails every
  row runs forever behind a green pill, which is precisely defect 1 above. The
  payload field (`configHealth.staleCronJobs` → `unhealthyCronJobs`) and the badge
  label follow — "unhealthy" is true of both states, where "not running" was false
  for one of them. Which one it is stays a panel-level detail, one click away.

The Scheduled Jobs panel now distinguishes them visually: a **red timestamp** means
the job stopped running (check the scheduler); a **red line under a green
timestamp** means it ran but could not finish its work.

## Defect 2 — the trends exclusion matches no live writer

**Reproduction.** A program lead marks twenty participants Present for a
three-hour session. `GET /api/facility/trends` excludes
`arrivedVia notIn [LEAD_MARKED, SYSTEM]` to keep synthetic staff marks out of
measured hours, but the roster mark writes `arrivedVia: "WEB"`
(`events/[id]/route.ts:349`), so all sixty placeholder hours land in the
Participation Trends chart as measured facility time. Verified on
`origin/main` @ `fbc59f7` (2026-08-09): `git grep LEAD_MARKED` finds no writer
anywhere in `src/` — only the schema, the two `20260803000*` migrations, the
trends filter, `visit/significance.ts`, the source-icon map, and tests. The
exclusion reaches only rows the `20260803000100` backfill rewrote. The comment
above it asserted the opposite ("real visits use SCANNER/WEB on this field"),
which is exactly what makes it invisible: `WEB` is what the roster mark writes.

**Cross-file dependency I could not fix.** The writer lives in
`events/[id]/route.ts`, which this change does not own and which open PR #1558 is
also editing. It is a follow-up, written out in full below.

No reader-side substitute is honest. Roster-marked visits are indistinguishable
from a member's own self-report by source (`WEB`/`WEB`), by `associatedEventId`
(`attendance/manual` and `/api/scan` both set it), and by times (a lead can type
any). The only durable discriminator is the `newData.type =
"lead_attendance_correction"` audit row, and joining the hours report to
`AuditLog` JSON is precisely the "sum visits plus overlays" the design rejected —
plus it would still miss the adopt-an-existing-visit branch, and would be dead
weight the day the writer is fixed. So the comment now states the ceiling in
present tense rather than asserting a filter that does not fire.

**What I did fix on the reader side.** The same clause silently dropped every
visit whose arrival source was never recorded: `arrivedVia` is nullable, and in
SQL a NULL never satisfies a `NOT IN`. Verified empirically against Postgres 15
through the generated client — Prisma emits
`"arrivedVia" NOT IN (…)` for `notIn` (and `NOT "arrivedVia" IN (…)` for
`NOT: { in }`), and both drop NULL rows. So untagged legacy visits vanish from
the board's hours chart entirely, under-reporting facility time with no
indication. The filter now admits NULL explicitly, which matches how
`visit/significance.ts` already reads an untagged source ("untagged legacy value
≈ self-reported" — real, trusted least). The new test also asserts the
`LEAD_MARKED` exclusion still fires, so the fix does not widen the hole it sits
next to.

---

# Required follow-up — filed as #1619

The writer-side fix is [#1619](https://github.com/innovationtreehouse/checkin/issues/1619),
filed by @thpr from the text that used to be reproduced here. It carries the part that
matters: the `⚠️ This is NOT the one-line change it looks like` section, the create/adopt
rule (leave `arrivedVia` alone on the update branch, so an adopted walk-in keeps its real
measured source), the five acceptance criteria, the #1558 coordination note, and the open
question about already-written `WEB` rows.

Only the NULL under-count is fixed in this PR. The `WEB` over-count is #1619.

---

## Tests fail without the fixes

Each was run against `origin/main`'s source with the new tests in place
(`git checkout origin/main -- <source files>`), then against this branch:

| test | on `origin/main` | on this branch |
|---|---|---|
| `cronAuth.test.ts` › records a partial sweep as a run that completed AND names what failed | **fail** — ledger row has no error | pass |
| `cronNightlyAPI` › records the run as unclean, not as stopped, when every checkout fails | **fail** — `data.failed` `undefined`, ledger error null | pass |
| `cronNightlyAPI` › keeps checking out good visits when ONE visit fails | **fail** — `data.failed` `undefined` | pass |
| `cronScholarshipGraceExpiry` › isolates a per-row failure | **fail** — `data.failed` `undefined`, ledger error null | pass |
| `reconcile.integration` › counts an order whose forward pass throws | **fail** — `r.failed` `undefined` | pass |
| `facilityTrendsAPI` › counts an untagged arrival and still drops the staff-marked one | **fail** — `uniqueVolunteers` 0, hours 0 | pass |

The three guards for the round-2 split were verified red the same way, by mutating
the logic back and confirming each one bites:

| mutation | test that fails |
|---|---|
| `withCron` records `success: healthy` again (one flag) | `cronAuth` › records a partial sweep as a run that completed AND names what failed |
| `getCronJobStatuses` reads `lastError` only from `!success` rows | `cronRuns` › still surfaces the error, so the panel does not read as healthy |
| `countUnhealthyCronJobs` filters on `stale` only | `cronRuns` › counts toward the nav badge even though nothing is stale |
| the panel's own badge count filters on `stale` only | `SystemHealthPanels` › counts a job that ran-but-failed toward the badge |

`CronRunsBox` had no test at all before this. It derives its badge count client-side,
so it can drift from `countUnhealthyCronJobs` and leave the nav pill saying 2 while
the panel it links to shows nothing — now guarded, along with the two-state render.

Companion assertions that pass on both are boundary guards for the new branches,
not regression tests: `records a clean success when the body reports zero failed
rows` (the `> 0` boundary), `names a failure even when the ledger row recorded no
message` (`error` is nullable), and `is counted once when the job is BOTH stale and
failing` (no double-count in the badge).

## Verification

- **CI: 9/9 checks green** on `aba5650`, including the user-journey flow tests.
- `npm run lint` — clean.
- `npx tsc --noEmit --pretty false` — clean.
- Unit: `npm run test:ci` — **276 suites, 2701 tests, all passing**.
- Integration: full `npm run test:integration` against a throwaway Docker
  Postgres 15 — **136 suites, 1444 tests, all passing**.
- Schema drift: CI's exact check reproduced locally —
  `prisma migrate deploy` onto a clean DB, then
  `prisma migrate diff --from-config-datasource --to-schema prisma/schema.prisma --exit-code`
  → *"No difference detected"*, exit 0. The `///` comment edit emits no SQL.
  The `schema.prisma` edit is `///` doc comments only: it emits no SQL (no migration
  in this repo's history contains a `COMMENT ON`), and `prisma generate` reproduced
  `src/security/generated/classifications.ts` byte-identical, since neither
  `@sensitivity:` annotation changed.

**Not verified:** whether any NULL `arrivedVia`
rows exist in the production DB; the squashed baseline
makes the field nullable with no backfill, and no current writer omits it, so the
affected rows are legacy ones. The follow-up writer fix above is unverified by
construction — it lives in a file this PR does not touch.

### DIFF
--- a/checkin-app/prisma/schema.prisma
+++ b/checkin-app/prisma/schema.prisma
@@ -1262,9 +1262,14 @@ model CronRunLog {
   startedAt  DateTime
   /// @sensitivity:internal
   finishedAt DateTime @default(now())
+  /// Whether the run COMPLETED — not whether it was clean. A sweep that isolates
+  /// per-row failures still completed, so it lands here as true with `error` set.
+  /// Drives "is this job still running" (lib/cronRuns.ts lastSuccessByJob).
   /// @sensitivity:internal
   success    Boolean
-  /// Message from the throw that failed the run; null on success.
+  /// Why the run was not clean — the throw's message, the HTTP status, or the count
+  /// of rows the sweep could not process. Null means the run did all of its work.
+  /// Drives "is this job still working", independently of `success`.
   /// @sensitivity:internal
   error      String?
 
--- a/checkin-app/src/app/__tests__/cronNightlyAPI.integration.test.ts
+++ b/checkin-app/src/app/__tests__/cronNightlyAPI.integration.test.ts
@@ -213,8 +213,10 @@ describe('Cron Nightly API Integration Tests', () => {
                 // Both visits are still open — nothing was swept.
                 expect(await prisma.visit.count({ where: { id: { in: [one.id, two.id] }, departedAt: null } })).toBe(2);
 
+                // The run COMPLETED (so the job never reads as "stopped") but was not
+                // clean — both facts recorded, on the two separate ledger columns.
                 const ledger = await prisma.cronRunLog.findFirst({ where: { job: 'nightly' }, orderBy: { id: 'desc' } });
-                expect(ledger?.success).toBe(false);
+                expect(ledger?.success).toBe(true);
                 expect(ledger?.error).toBe('2 item(s) failed');
             } finally {
                 mockBadVisitIds.clear();
--- a/checkin-app/src/app/__tests__/cronNightlyAPI.integration.test.ts
+++ b/checkin-app/src/app/__tests__/cronNightlyAPI.integration.test.ts
@@ -187,6 +187,40 @@ describe('Cron Nightly API Integration Tests', () => {
         const systemNotifyCount = () =>
             prisma.auditLog.count({ where: { tableName: 'SYSTEM_NOTIFY' } });
 
+        // The System Status pill is the only in-app signal that the sweeps still
+        // work, and it reads the CronRunLog. A night that checked NOBODY out must
+        // not land there as a fresh success.
+        it('records the run as a failure when every checkout fails', async () => {
+            await closeAllOpenVisits();
+            await prisma.auditLog.deleteMany();
+
+            const arrivedAt = new Date(Date.now() - 2 * 60 * 60 * 1000);
+            const one = await prisma.visit.create({ data: { personId: normalUserId, arrivedAt } });
+            const two = await prisma.visit.create({ data: { personId: keyholderId, arrivedAt } });
+
+            mockBadVisitIds.clear();
+            mockBadVisitIds.add(one.id);
+            mockBadVisitIds.add(two.id);
+
+            try {
+                const res = await GET(cronReq());
+                expect(res.status).toBe(200);
+
+                const data = await res.json();
+                expect(data.facilityClose.checkedOutCount).toBe(0);
+                expect(data.failed).toBe(2);
+
+                // Both visits are still open — nothing was swept.
+                expect(await prisma.visit.count({ where: { id: { in: [one.id, two.id] }, departedAt: null } })).toBe(2);
+
+                const ledger = await prisma.cronRunLog.findFirst({ where: { job: 'nightly' }, orderBy: { id: 'desc' } });
+                expect(ledger?.success).toBe(false);
+                expect(ledger?.error).toBe('2 item(s) failed');
+            } finally {
+                mockBadVisitIds.clear();
+            }
+        });
+
         it('keeps checking out good visits when ONE visit fails, still notifies the board, and returns 200', async () => {
             await closeAllOpenVisits();
             await prisma.auditLog.deleteMany();
@@ -220,8 +254,10 @@ describe('Cron Nightly API Integration Tests', () => {
                 const data = await res.json();
                 expect(data.success).toBe(true);
 
-                // 2 good visits checked out; the bad one is NOT counted as success.
+                // 2 good visits checked out; the bad one is NOT counted as success,
+                // and the run reports it so the ledger can mark the sweep unhealthy.
                 expect(data.facilityClose.checkedOutCount).toBe(2);
+                expect(data.failed).toBe(1);
 
                 // Board still notified even though the isKeyholder's checkout failed
                 // (notification derives from the abandoned-visit list, not results).
--- a/checkin-app/src/app/__tests__/cronNightlyAPI.integration.test.ts
+++ b/checkin-app/src/app/__tests__/cronNightlyAPI.integration.test.ts
@@ -189,8 +189,9 @@ describe('Cron Nightly API Integration Tests', () => {
 
         // The System Status pill is the only in-app signal that the sweeps still
         // work, and it reads the CronRunLog. A night that checked NOBODY out must
-        // not land there as a fresh success.
-        it('records the run as a failure when every checkout fails', async () => {
+        // not land there as a clean success — nor as "job stopped running", which is
+        // a different failure the board would chase into the scheduler.
+        it('records the run as unclean, not as stopped, when every checkout fails', async () => {
             await closeAllOpenVisits();
             await prisma.auditLog.deleteMany();
 
--- a/checkin-app/src/app/__tests__/cronScholarshipGraceExpiry.integration.test.ts
+++ b/checkin-app/src/app/__tests__/cronScholarshipGraceExpiry.integration.test.ts
@@ -159,10 +159,12 @@ describe('Cron Scholarship-Grace-Expiry API Integration Tests', () => {
                 expect(data.processed).toBe(2);
                 expect(data.released).toBe(1);
                 // Isolating the row keeps the sweep going; it does not make the run
-                // healthy, so the count is reported and the ledger records a failure.
+                // healthy. The run completed, so `success` stays true — the ledger
+                // records the failure on `error` instead, which is what the panel reads.
                 expect(data.failed).toBe(1);
                 const ledger = await prisma.cronRunLog.findFirst({ where: { job: 'scholarship-grace-expiry' }, orderBy: { id: 'desc' } });
-                expect(ledger?.success).toBe(false);
+                expect(ledger?.success).toBe(true);
+                expect(ledger?.error).toBe('1 item(s) failed');
 
                 // Inside the window: untouched.
                 const inside = await prisma.programParticipant.findUnique({ where: { programId_personId: { programId, personId: ids.inside } } });
--- a/checkin-app/src/app/__tests__/cronScholarshipGraceExpiry.integration.test.ts
+++ b/checkin-app/src/app/__tests__/cronScholarshipGraceExpiry.integration.test.ts
@@ -158,6 +158,11 @@ describe('Cron Scholarship-Grace-Expiry API Integration Tests', () => {
                 // "gone" throws on delete() and is isolated, not counted as released.
                 expect(data.processed).toBe(2);
                 expect(data.released).toBe(1);
+                // Isolating the row keeps the sweep going; it does not make the run
+                // healthy, so the count is reported and the ledger records a failure.
+                expect(data.failed).toBe(1);
+                const ledger = await prisma.cronRunLog.findFirst({ where: { job: 'scholarship-grace-expiry' }, orderBy: { id: 'desc' } });
+                expect(ledger?.success).toBe(false);
 
                 // Inside the window: untouched.
                 const inside = await prisma.programParticipant.findUnique({ where: { programId_personId: { programId, personId: ids.inside } } });
--- a/checkin-app/src/app/__tests__/facilityTrendsAPI.integration.test.ts
+++ b/checkin-app/src/app/__tests__/facilityTrendsAPI.integration.test.ts
@@ -9,8 +9,8 @@
  * path: bucketing visits by period, the volunteer/participant split (by program
  * enrollment — ProgramParticipant, NOT age), structured (event-associated) vs
  * unstructured hours, the `arrivedVia=LEAD_MARKED` exclusion (synthetic "marked
- * present" visits, plus its pre-split `SYSTEM` spelling), the `programId`
- * filter, and the invalid-period 400.
+ * present" visits, plus its pre-split `SYSTEM` spelling) and the untagged-arrival
+ * case it must not swallow, the `programId` filter, and the invalid-period 400.
  */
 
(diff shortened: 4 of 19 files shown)


thpr (commented):
## Verification

The per-route audit table in the description holds up independently: `post-event`, `membership-renewals`, `person-bg-annual`, `trusted-adult-expiry` and `pending-participants` genuinely have no swallowing `try/catch`, and none of their result shapes already carries a `failed` key that would collide with the new health signal. `reconcileReversals` has no per-row catch either, so it propagates to `withCron` as the table says. The `notIn` sweep is complete — the only other two in `src/` (`membership-ops/applications`, `finance/matchAudit`) sit on non-nullable enums, so the NULL hole is unique to trends.

## Findings

### 1. A permanently-failing row now pins a job as "not running", forever

`src/lib/cronAuth.ts:89` together with `src/lib/cronRuns.ts:71`.

`lastSuccessByJob()` aggregates only `success: true`, so `lastSuccessAt` freezes the moment any run reports `failed > 0`. One poison row — a visit whose `processVisitCheckout` always throws, or reconcile's cursor-frozen order, which is *by design* re-scanned every single run until it heals — makes the job stale after 48h, red in `SystemHealthPanels.tsx:193`, and counted by `countStaleCronJobs`, whose badge title reads `"N job(s) not running"` for a job that ran last night and swept everything else.

The ledger has one boolean per run and no third state, so "ran, mostly worked" is unrepresentable. `lastError` does render alongside it ("Failing since: 1 item(s) failed"), which softens the panel row, but `stale` and the nav badge are now wrong in exactly the way the 48h constant's own doc comment says they must not be — "generous enough that a daily job can miss a run, a deploy window, or a clock skew without alarming, and strict enough that a job which genuinely stopped surfaces within two days".

Worth resolving one way or the other before merge. Either keep `success: true` (the job *ran*) and set a non-null `error`, adjusting `getCronJobStatuses` to surface the latest row's error rather than "failed since last success"; or accept the conflation deliberately and say so in the `withCron` doc comment, since the next reader will otherwise arrive at it as a bug report.

### 2. The reported trends defect still ships unfixed

The title says "trends hours", but the live over-count — the placeholder hours per roster-marked session landing in the chart, because `events/[id]/route.ts:349` stamps `WEB` — is unchanged by this PR. That diagnosis checks out against `1256_ATTENDANCE_CORRECTION_SURFACE.md:490`, which lists that writer's target as `LEAD_MARKED` on both fields. Only the NULL under-count is fixed here.

Splitting the two halves is defensible, but the follow-up currently lives only in the PR description, which stops being reachable the moment this merges. File it as an issue and link it from here.

### 3. The proposed follow-up is not the one-liner the description describes

The `adoptable` branch (`events/[id]/route.ts:333-345`) adopts an existing open **walk-in** visit — `SCANNER`/`WEB`, a real measured arrival — into the event, and applies the same `times` object the create branch uses. Changing `times.arrivedVia` to `"LEAD_MARKED"` unconditionally would relabel that genuine arrival as staff-asserted, and with the exclusion in this PR now live, drop it out of the hours chart entirely. That trades an over-count for an under-count on precisely the people who did badge in.

The writer fix needs to stamp `LEAD_MARKED` on the create branch, and on updates of a row that already carries `LEAD_MARKED`, while preserving an adopted visit's own source. Worth writing into the follow-up note so whoever picks it up does not ship the literal one-line version.

## Smaller notes

- `reportedFailures` reads `failed` by name off an arbitrary body. No collision today — I checked all seven `...result` spreads — but `reconcile-shopify` acquires the field purely by spread, so a future sweep returning `failed` with a different meaning ("failed payments found") would silently flip its health. The JSDoc on `ReconcileResult.failed` naming `lib/cronAuth.ts` is the right mitigation; nothing more needed.
- `res.clone().json()` before returning `res` is the correct ordering, and the `catch → 0` fallback is right for a non-JSON body.
- The tests are well-targeted. The `failed: 0` companion case is the boundary guard for the new `> 0` branch, and the trends test scopes by its own `programId` rather than leaning on a shared month bucket, so it asserts exact counts instead of deltas. Cleanup covers both new rows.

Nothing in the diff itself is incorrect. Finding 1 is the one I would want answered before merge.

---

thpr (changes requested):
Requesting changes on the review above.

Blocking:

- **Finding 1** — `withCron` recording `success: false` on a partial sweep freezes `lastSuccessAt`, so one permanently-failing row leaves the job reading as "not running" in the nav badge and the System Status panel indefinitely. Needs either a separate ran-vs-healthy signal, or an explicit decision recorded in the `withCron` doc comment.
- **Finding 2** — file the `events/[id]/route.ts:349` writer fix as an issue and link it here, so the follow-up survives the merge of this description.
- **Finding 3** — carry the `adoptable`-branch caveat into that issue: the literal one-line change would relabel real walk-in arrivals as staff-asserted and drop them from the chart.

Findings 2 and 3 are cheap (an issue with the caveat written into it). Finding 1 is the one that needs a call.

---

thpr (approved):
Re-reviewed at `aba56502`. All three findings are addressed. Approving.

**Finding 1 — resolved, and resolved at the right layer.** Splitting `success` ("the run completed") from `error` ("the run was clean") is the correct shape: `lastSuccessAt` keeps advancing on a partial sweep, so the poison-row freeze cannot happen, while `countUnhealthyCronJobs` still lights the pill. I checked the derivation rather than the description:

- `withCron` — `ran = res.status < 400`, error set from `failed` when non-zero and from the status otherwise. The 5xx path still records `success: false`, so a job that genuinely stops still goes stale. Confirmed by the new `records a handler error envelope as a run that did NOT complete` case.
- `getCronJobStatuses` — `unclean = latest.error !== null || !latest.success`, read off the latest run rather than "failed since last success". Equivalent for the old cases (a failed run's row satisfies both clauses), and it is what makes the completed-but-unclean row visible at all. The nullable-`error` fallback to `"unknown error"` is guarded.
- `countUnhealthyCronJobs` = `stale || lastError`, and `CronRunsBox` derives its badge with the same predicate. The duplication is client-side and unavoidable, and it has a test pinning the two together, which is the right call — that is exactly the pair that would drift into "pill says 2, panel shows nothing".

The `staleCronJobs` → `unhealthyCronJobs` rename on the `/api/nav/todo-counts` payload is complete: `git grep` for `staleCronJobs`, `countStaleCronJobs`, and the old "not running" label strings across the whole branch returns nothing, and neither `todo-counts` nor `config-health` appears in `src/security/`, so there is no registry surface to follow. The only remaining `CronRunLog` reader is `cronRuns.ts` itself.

**Findings 2 and 3 — as resolved as this PR can make them.** The issue text in the description is complete and correct, including the part that matters: the create/adopt split, with "leave `arrivedVia` alone on the update branch" stated as the rule and the adopted-walk-in guard listed as an acceptance criterion. That is the version that does not regress. Filing it is still owed — I will do that separately and link it back here, since it is the one thing that does not survive the merge of this description.

Two notes, neither blocking:

- `lastRunByJob` dropped the `where: { success: false }`, so it now scans the whole table where the old query rode the `(job, success, finishedAt)` index prefix. At ~9 rows/day under a 90-day TTL that is ~800 rows, and `useTodoCounts` fetches on mount and on mutation events rather than on a timer, so this is genuinely nothing — worth knowing only if the ledger ever grows a per-row-per-job shape.
- `distinct` over `orderBy: [job, finishedAt desc]` picks arbitrarily between two rows of the same job sharing a `finishedAt`. For daily sweeps that cannot happen; noting it so it is a known property rather than a surprise.

The schema edit is `///` doc comments only, which emit no SQL and leave the generated classifications byte-identical — no migration implication.

---

thpr:
Filed the follow-up as #1619 — https://github.com/innovationtreehouse/checkin/issues/1619

Your text, reproduced as written, including the `⚠️ This is NOT the one-line change it looks like` section, the create/adopt rule, the five acceptance criteria, the #1558 coordination note, and the open question about already-written `WEB` rows. Nothing edited beyond a footer pointing back here.

That clears the last item from my review — approval above stands. The `NOT YET FILED` block in the description can be replaced with a link to #1619 whenever you next touch it.

---

PR #1610, opened 2026-08-09
fix(trusted-adult): don't tell a family a renewed adult is de-authorized

jee7s:
## The bug

`runExpirySweep` sent the "no longer authorized at the front desk" email **per lapsed REVIEW**, but `/api/trusted-adults/operational` lists a trusted adult while **ANY** of its reviews is `APPROVED`. Those two rules disagreed, so the ordinary renew-early flow mailed a false de-authorization notice.

The path, all on `origin/main`:

1. A trusted adult's INITIAL review is APPROVED with `reviewBy = T`.
2. At `T-30d` the sweep warns the family "expiring soon".
3. The family renews. `renewTrustedAdult` only refuses when the *latest* review is `PENDING_BOARD_REVIEW`/`PENDING_SUBJECT_ACTION`, so this is allowed.
4. The board approves the RENEWAL. `decideReview`'s APPROVE branch stamps a fresh `reviewBy` on the new row and never touches prior approvals — the adult now carries **two** APPROVED reviews.
5. At `T` the sweep's lapsed query matches the **old** review and emails the family that the adult is no longer authorized — while `/operational` still lists them, correctly, because the renewal is live.

The family is told they lost someone the front desk will still release their kids to.

## The fix

The root cause is two definitions of "authorized". `/operational` had it inline (`reviews: { some: { status: "APPROVED" } }`), `decideReview`'s deny branch had a second hand-rolled copy (with a `// Mirror /operational's rule` comment — already a warning sign), and the sweep had no notion of it at all.

Extracted `AUTHORIZED_REVIEW` into `src/lib/trusted-adult/filters.ts`, following the existing `LIVE_PERSON` precedent in `src/lib/person/filters.ts`. All three sites now read it:

- `/api/trusted-adults/operational` — the authority, unchanged in behaviour.
- `decideReview` — both the deny notice and Deny & Revoke.
- `runExpirySweep` — **the behaviour change**: after the flip to `EXPIRED` commits, re-ask whether the adult still has any APPROVED review. Only the review that leaves them with *no* approval at all earns the de-authorization notice.

The check runs after the flip, so the row being expired is already excluded, and the `continue` sits after `expired++` — the returned `expired` count and the per-review `EXPIRED` bookkeeping are unchanged. If several reviews for one adult lapse in the same sweep, the family gets exactly one notice, not one per row.

## Red / green

Two checks in `trustedAdultService.integration.test.ts`, both proved red by checking out `origin/main`'s `service.ts` underneath them.

**1. `a renewed adult whose old review lapses gets no de-authorization email; a genuinely lapsed one still does`**

Seeds two adults in one sweep — one renewed (stale INITIAL + live RENEWAL), one genuinely lapsed — and asserts both directions, so the fix cannot be "stop emailing".

RED on `origin/main`:
```
  expect(jest.fn()).not.toHaveBeenCalledWith(...expected)
  Expected: not Anything, StringContaining "Renewed Grandma", Anything
  Received
    1: "sweeplead-...@ex.com", "Trusted adult approval expired: Renewed Grandma ...",
       "<p>The board's approval of Renewed Grandma ... has expired. They are no longer
        authorized at the front desk. ...</p>"
  Number of calls: 2
```
Two emails on main — the genuinely lapsed family *and* the renewed one. On this branch, one: only the genuinely lapsed family.

**2. `two lapsed reviews on ONE adult expire both rows but send exactly one de-authorization email`** (added for review feedback)

"One notice per ADULT, not one per row" is a property of statement **order**: each row's flip commits before the next row's authorization check reads, so only the last row to lapse finds no approval and mails. Batch the flips into a single upfront transaction — a plausible perf refactor of a nightly sweep — and every row sees zero approvals, so the family gets one email per lapsed row again. Nothing pinned that, and the first test seeds two *different* adults so it doesn't cover it.

This one seeds **one** adult with **two** lapsed APPROVED reviews. The `expired === 2` assertion pins the other half of the contract — `continue` stays after `expired++`, so the returned count and the per-review `EXPIRED` flips stay identical to main.

RED on `origin/main`:
```
  expect(jest.fn()).toHaveBeenCalledTimes(expected)
  Expected number of calls: 1
  Received number of calls: 2
```
(`run.expired === 2` and both rows `EXPIRED` pass on *both* sides — that is the point: only the email count moves.)

GREEN on this branch: `Tests: 30 passed, 30 total`.

## Other verification

- `npm run test:integration -- --testPathPatterns trustedAdultService` — 30 passed.
- `npm test -- --testPathPatterns "trusted|pickup|todo-counts"` — 6 suites, 32 tests passed.
- `npx tsc --noEmit` — exit 0, zero output.
- `npx eslint` on the changed files — clean.
- Drift-guard suites pass.

## Known gap: "expiring" is still per-REVIEW on two other surfaces

This PR fixes the *de-authorization* notice, which asks **"is this adult authorized?"** — the question `AUTHORIZED_REVIEW` answers. Two neighbours ask a *different* question, **"is this adult's authorization running out?"**, and both still answer it one row at a time:

- **The sweep's warning email** (`service.ts`, the `expiring` query). Renew more than 30 days out, get board approval, and the stale row still mails "A trusted adult approval is expiring soon".
- **`/api/nav/todo-counts`** (`route.ts:202-208`, `:237-238`) — thpr's review finding, confirmed. The same stale row pushes one "Renew an expiring trusted adult" todo *per row*, so a family that renewed early and was approved is told to go renew someone they already renewed, while `/operational` correctly still lists that adult.

**`AUTHORIZED_REVIEW` does not fix either, and I checked rather than assumed.** todo-counts' filter is `status: { in: APPROVED_STATUSES }` where `APPROVED_STATUSES = ["APPROVED"]` (`route.ts:43`) — already byte-for-byte `AUTHORIZED_REVIEW`. Substituting it changes no behaviour at all; it would only *look* like a fix while the family keeps getting nagged.

What both surfaces actually need is the other predicate: exclude self, compare `reviewBy` across the adult's APPROVED rows — "is any approval outliving this one?". Folding that in here would have added back the second definition of "authorized" this PR exists to delete, and would have fixed one of the two surfaces while leaving its identical twin broken.

Secondary, same call site: todo-counts' `reviewBy: { not: null, lte: warnThreshold }` has no `gt: now` (the sweep's `expiring` query does), so a row that lapsed before the nightly sweep ran is still labelled "expiring". Adding `gt: now` does **not** fix the renewal nag — the stale row's `reviewBy` is in the future for the whole 30-day warn window.

**Follow-up:** one shared "expiring adult" predicate, applied to the warning email and todo-counts together.

### DIFF
--- a/checkin-app/src/lib/trusted-adult/service.ts
+++ b/checkin-app/src/lib/trusted-adult/service.ts
@@ -3,6 +3,7 @@ import { personActor, systemActor, type AuditActor } from "@/lib/auditActor";
 import { escapeHtml } from "@/lib/email-templates/base";
 import { emailBoardMembers, emailHouseholdLeads } from "@/lib/emailRecipients";
 import { isTrustedAdultConflict } from "@/lib/trusted-adult/conflict";
+import { AUTHORIZED_REVIEW } from "@/lib/trusted-adult/filters";
 import { config } from "@/lib/config";
 import { validateContact } from "@/lib/trusted-adult/contact";
 import type { TxClient } from "@/lib/db-client";
@@ -401,7 +402,7 @@ export async function decideReview(reviewId: number, boardMemberId: number, inpu
         // one live would keep the adult on the pickup list after a full rejection.
         if (input.decision === "DENY" && input.revokePriorApprovals) {
             const live = await tx.trustedAdultReview.findMany({
-                where: { trustedAdultId: review.trustedAdultId, status: "APPROVED" },
+                where: { trustedAdultId: review.trustedAdultId, ...AUTHORIZED_REVIEW },
                 orderBy: { id: "desc" },
             });
             if (live.length) {
@@ -434,7 +435,7 @@ export async function decideReview(reviewId: number, boardMemberId: number, inpu
         } else {
             // Mirror /operational's rule: a surviving APPROVED review = still authorized.
             const live = await prisma.trustedAdultReview.findFirst({
-                where: { trustedAdultId: head.trustedAdultId, status: "APPROVED" },
+                where: { trustedAdultId: head.trustedAdultId, ...AUTHORIZED_REVIEW },
                 orderBy: { reviewBy: "desc" },
             });
             if (live) {
@@ -562,7 +563,16 @@ export async function runExpirySweep(now: Date) {
         expired++;
         // Notify only after the flip commits: the status change is the authoritative act,
         // the email a courtesy on top, so a slow or failing provider can't hold the tx.
-        // The flip is its own dedup — an EXPIRED row is no longer selected by this query.
+        // The flip is its own dedup — an EXPIRED row is no longer selected by this query —
+        // and it also excludes this row from the authorization check below.
+        const stillAuthorized = await prisma.trustedAdultReview.findFirst({
+            where: { trustedAdultId: r.trustedAdultId, ...AUTHORIZED_REVIEW },
+            select: { id: true },
+        });
+        // One lapsed review is not a lapsed adult: a renewal approved before this row's
+        // date still authorizes them, and /operational still lists them. Only the review
+        // that leaves the adult with no approval at all earns the de-authorization notice.
+        if (stillAuthorized) continue;
         const name = r.trustedAdult.trustedAdultName?.trim() || "your trusted adult";
         await notifyHouseholdFamily(
             r.householdId,
--- a/checkin-app/src/app/__tests__/trustedAdultService.integration.test.ts
+++ b/checkin-app/src/app/__tests__/trustedAdultService.integration.test.ts
@@ -721,4 +721,39 @@ describe('runExpirySweep edge cases', () => {
         expect(sendEmail).not.toHaveBeenCalledWith(expect.anything(), expect.stringContaining('Renewed Grandma'), expect.anything());
         expect(sendEmail).toHaveBeenCalledTimes(1);
     });
+
+    // Pins "one notice per ADULT, not one per row" — a property of statement ORDER: each
+    // row's flip commits before the next row's authorization check reads, so only the last
+    // row to lapse finds no approval and mails. Batch the flips into one upfront
+    // transaction and every row sees zero approvals: one email per row again. The
+    // `expired` assertion pins the other half — `continue` stays AFTER `expired++`, so
+    // per-review bookkeeping is byte-identical to the pre-fix behaviour.
+    it('two lapsed reviews on ONE adult expire both rows but send exactly one de-authorization email', async () => {
+        const now = new Date();
+        const ta = await prisma.trustedAdult.create({
+            data: { householdId, trustedAdultName: `Twice Lapsed ${SWEEP_TAG}`, trustedAdultEmail: 'twice@example.com', familyContext: 'ctx', disclosedById: leadId },
+        });
+        // A year-old original and the renewal that replaced it, both now past reviewBy.
+        const older = await prisma.trustedAdultReview.create({
+            data: { householdId, trustedAdultId: ta.id, kind: 'INITIAL', status: 'APPROVED', reviewBy: new Date(now.getTime() - 400 * DAY) },
+        });
+        const newer = await prisma.trustedAdultReview.create({
+            data: { householdId, trustedAdultId: ta.id, kind: 'RENEWAL', status: 'APPROVED', reviewBy: new Date(now.getTime() - 1 * DAY) },
+        });
+
+        const run = await runExpirySweep(now);
+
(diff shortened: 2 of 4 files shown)


checkin-app/src/lib/trusted-adult/service.ts
```diff
@@ -562,7 +563,16 @@ export async function runExpirySweep(now: Date) {
...
+        // The flip is its own dedup — an EXPIRED row is no longer selected by this query —
+        // and it also excludes this row from the authorization check below.
+        const stillAuthorized = await prisma.trustedAdultReview.findFirst({
+            where: { trustedAdultId: r.trustedAdultId, ...AUTHORIZED_REVIEW },
+            select: { id: true },
+        });
+        // One lapsed review is not a lapsed adult: a renewal approved before this row's
+        // date still authorizes them, and /operational still lists them. Only the review
+        // that leaves the adult with no approval at all earns the de-authorization notice.
+        if (stillAuthorized) continue;
```
thpr, line 575: This is where the "one notice per adult, not one per row" claim lives, and it depends on the flip of each row committing before the next row's check. If several reviews for one adult lapse in the same sweep, rows not yet processed are still `APPROVED`, so only the last one to flip emails. Move the flips into one upfront transaction and that inverts silently — every row would then see no surviving approval and email. Worth naming the constraint here, since the test seeds two different adults and doesn't hold this down.

thpr (commented):
Two findings from reading the diff — neither blocks the fix, which holds.

**The same per-review bug survives in `todo-counts`, and it isn't in the known-gap list.**

`src/app/api/nav/todo-counts/route.ts:202-208` counts `APPROVED` reviews with `reviewBy: { not: null, lte: warnThreshold }` — no `gt: now`, so already-lapsed rows count too — and `:237-238` pushes one "Renew an expiring trusted adult" todo per row. A family that renewed early and had the board approve it still sees that todo off the stale row: the renewal is live, `/operational` lists the adult, and the household is told to go renew someone they already renewed. Same per-review-vs-per-adult confusion as the sweep, same user-visible lie, a different surface. The known-gap section names only the sweep's warning email, so this one would go unrecorded. Worth folding into the same follow-up rather than this PR.

**The "one notice per adult, not one per row" property is load-bearing on statement order and nothing pins it.**

It holds only because each row's flip commits before the next row's `stillAuthorized` check runs — the rows not yet processed are still `APPROVED`, so only the last one to flip emails. Batch the flips into a single upfront transaction (a plausible perf refactor of a nightly sweep) and every row sees zero `APPROVED` reviews, so the family gets one de-authorization email per lapsed row instead of one. The new test seeds two *different* adults, so it doesn't cover this. Either add a same-adult two-lapsed-rows case, or state the ordering constraint in the comment so the next person doesn't refactor it away.

---

thpr (changes requested):
Requesting changes on the two findings above.

1. `todo-counts` (`src/app/api/nav/todo-counts/route.ts:202-208`, `:237-238`) still nags a family to renew a trusted adult they already renewed — the same per-review-vs-per-adult split this PR exists to close. Fix it here or record it alongside the warning-email gap so it doesn't get lost.

2. Pin the "one notice per adult, not one per row" property: a same-adult two-lapsed-rows case in the test, or the ordering constraint stated at `src/lib/trusted-adult/service.ts:575`. Right now nothing stops a later refactor from turning it into one email per row.

---

thpr (approved):
Both findings satisfied at `c8b20c20`.

**The one-notice-per-adult property is pinned.** The new test seeds one adult with two lapsed `APPROVED` rows (−400d and −1d) and asserts `expired === 2`, both rows `EXPIRED`, no surviving approval, and `sendEmail` called exactly once. Neither row sits in the warn window, so the count is unambiguous. It covers both halves — the notice dedup and the `continue`-after-`expired++` bookkeeping — and it goes red on `main` at 2 emails. `Run Tests (integration)` is a real gate on this repo, so the property is enforced rather than asserted in a PR body.

**`todo-counts` is recorded rather than fixed, and the write-up is better than my comment was.** Checking that `APPROVED_STATUSES` is already byte-identical to `AUTHORIZED_REVIEW` is the part that matters: substituting the shared constant there would have looked like a fix while the family kept getting nagged. Same for `gt: now` — the stale row's `reviewBy` is in the future for the whole 30-day warn window, so it doesn't touch the renewal case. Both correct. The real fix is the other predicate — "is any approval outliving this one?" — and folding it in here would have reintroduced the second definition of "authorized" this PR exists to delete.

One thing outstanding, not blocking: the follow-up now lives only in this PR's body, and a known gap recorded there evaporates once it merges. Filed as innovationtreehouse/checkin#1618 so the two surfaces get the shared expiring predicate together.

---

PR #1611, opened 2026-08-09
fix(security): serve manual-attendance corrections through handler()

jee7s:
## The defect

`PATCH` and `DELETE /api/attendance/manual/[id]` are both registered in `src/security/registry.ts` with `their_own:personal` + `led_households:personal`, but the route was implemented with legacy `withAuth`. `handler()` — and therefore `stripBag()` — never ran, so the declared policy was **inert**: the registry described a boundary nothing enforced.

Re-verified against `origin/main` (`fbc59f743`) with `git show origin/main:<path>`, not a working-tree grep.

### The leak, observed directly

A household lead PATCHes a member's visit. Same mocked request, run against each version of `route.ts`:

```jsonc
// origin/main (legacy withAuth) — 200
{ "visit": { "id": 42, "personId": 8,
             "deletedAt": null, "deletedById": null, "forceCloseWarnedAt": null,
             "associatedEventId": null, "arrivedAt": "2026-07-20T14:05:00.000Z",
             "departedAt": "2026-07-20T16:00:00.000Z",
             "arrivedVia": "WEB", "departedVia": "WEB" },
  "flagged": false }

// this branch (handler) — 200
{ "visit": { "id": 42, "personId": 8,
             "arrivedAt": "2026-07-20T14:05:00.000Z",
             "departedAt": "2026-07-20T16:00:00.000Z",
             "arrivedVia": "WEB", "departedVia": "WEB",
             "associatedEventId": null } }
```

All three `internal` columns — `deletedAt`, `deletedById`, `forceCloseWarnedAt` — were on the wire, against a registry comment that says *"'internal' tombstone fields stay stripped"*.

## The fix

Both verbs converted to `handler()` returning a `ModelBag`, matching the already-converted routes (`api/profile`, `api/facility/corrections`, `api/membership-audit/person-agreement`, `api/trusted-adults/mine`): `apiError(...)` returns become `throw badRequest/notFound/unauthorized(...)`, and the response is the bag rather than a hand-built body.

**No boundary file is touched.** `src/security/**` is untouched — the registry entries already exist and already say the right thing; the route simply never asked. `security-boundary-isolation.yml` sees `BOUNDARY_TOUCHED=0` and short-circuits. The changed files are the route, its test, and `scripts/migrated-routes.txt`.

### Are the registry tokens right for this endpoint?

**PATCH — yes.** Checked against the bindings, not the prose. The response is one `Visit` row; the caller is either the row's owner (`their_own`: `Visit.personId === selfId`) or the lead of that member's household (`led_households`: `Visit.personId ∈ ledHouseholdMemberIds`) — both bound on `Visit` in `scopeBindings.ts`, so neither grant silently resolves to nothing. `arrivedAt`/`departedAt` are `personal`, so `their_own:personal` + `led_households:personal` is exactly what keeps the times visible to those two subjects and nobody else; `member`/`public` cover `id`/`personId`/`arrivedVia`/`departedVia`/`associatedEventId`; the three tombstone columns are `internal` and granted to no one here. `scopeValidators.test.ts` already asserts the seam for `PATCH`.

**DELETE — not wrong, but vacuous, and its comment is now stale.** The tombstone response carries no model data, so the tokens are inert by construction. Two follow-ups, both needing a `registry.ts` edit that must ship alone, so **deliberately not done here**:

1. The entry's comment says *"the response is `{ success, flagged }`, no model data"*. That shape is not expressible through `handler()` — `stripBag` drops any top-level bag key that is not a model name (the same constraint the `GET /api/facility/corrections` entry already documents for pagination). The body is now `{}`.
2. `led_households` in the `orderedView` makes `deriveCtxNeeds` set `ledHouseholdMembers: true`, so every delete pays a `prisma.person.findMany` roster query for a response that ships nothing.

## Response-shape changes

| verb | before | after |
|---|---|---|
| `PATCH` | `{ visit: <raw row incl. deletedAt, deletedById, forceCloseWarnedAt>, flagged }` | `{ visit: <stripped to the granted view> }` |
| `DELETE` | `{ success: true, flagged: true }` | `{}` |

`flagged` and `success` are dropped because a non-model scalar has no legal home in a bag. **Nothing consumes them**: both callers (`src/app/attendance/my-visits/page.tsx:56`, `src/app/attendance/household/page.tsx:56`) read only `res.ok` and `data.error`, then reload. The board flag is unaffected — it is an email plus `auditLog.newData.significance`, both still written.

Error bodies are unchanged (`handler()` maps `ApiResponseError` through the same `apiError` helper), and the `logBackendError` call the legacy wrapper gave these routes is preserved by re-throwing non-`ApiResponseError`s after logging.

## The regression check

Two tests in `src/app/api/attendance/manual/[id]/__tests__/route.test.ts`, both edges of the same grant:

1. **"strips the internal tombstone columns while keeping the times the lead may see"** — a lead PATCHes a member's visit; the times must survive (`led_households:personal` resolves) *and* all three `internal` columns must be gone.
2. **"strips the times when the roster does not contain the visit's person"** — same lead session, `ledHouseholdMemberIds` without the member; the times must strip. `visitSubject` (DB leadship) and `ledHouseholdMemberIds` (session `householdLead` + `householdId`) are different sources and can disagree on a stale session; when they do, the grant must not resolve.

Test 2 caught test 1 passing for the wrong reason: the household-lead block inherited the global `tx.visit.update` mock, whose row carries `personId = OWN_ID`, so the times were surviving via `their_own`, not `led_households`. Fixed by pointing that block's `update`/`processVisitCheckout` mocks at the MEMBER's row, so test 1 now genuinely depends on the grant it names. `baseVisit` also gained `forceCloseWarnedAt: null` so the fixture carries every column a real `visit.update` returns.

### Before — `origin/main`'s `route.ts` restored, both tests in place

```
✕ strips the internal tombstone columns while keeping the times the lead may see (4 ms)
✕ strips the times when the roster does not contain the visit's person (2 ms)

  Expected path: not "deletedAt"
  Received value: null

  Expected path: not "arrivedAt"
  Received value: "2026-07-20T14:05:00.000Z"

Tests: 2 failed, 19 skipped, 21 total
```

Test 1 gets *past* its `arrivedAt`/`departedAt` assertions before failing — the failure is precisely the leak, not an incidental shape difference.

### After — this branch

```
PASS src/app/api/attendance/manual/[id]/__tests__/route.test.ts
Tests: 21 passed, 21 total
```

## Verification

- `npx tsc --noEmit --pretty false` — only the 2 known-environmental `@aws-sdk/client-lambda` errors.
- `npm run test:ci` — **2658 passed**, 0 test failures; the only 2 suite failures are the known `s-read` load errors (`Cannot find module '@aws-sdk/client-lambda'`).
- `npm run test:integration` against a throwaway Docker Postgres — **136 suites / 1440 tests green**, including `visitTombstoneCheckin.integration.test.ts` (imports `DELETE`), `tests/security/policy.contract.integration.test.ts` and `registryAuthz.integration.test.ts`.
- `npm run check-route-coverage` — **0 errors**. Both verbs dropped off the `[unmigrated]` list; the only remaining `attendance/manual` mention is the sibling `POST /api/attendance/manual`, out of scope and correctly still in `legacy-authz-routes.txt`.
- `npx eslint` on both changed source files — clean.

## Out of scope, but found while working

`POST /api/facility/visits/insert` has the same defect class: registered in the registry, implemented with `withAuth`, returns `NextResponse.json({ visit })`. Its declared view grants `everyones:internal`, so the tombstone columns it ships are *declared* rather than leaked — the registry comment says as much — but `stripBag` still never runs, so the entry is inert and any future tightening of the `Visit` tiers would not reach it. Separate PR.

## Not verified

Not exercised against a running app — no browser check of the two correction pages. The claim that neither page reads `flagged`/`success` comes from reading both call sites, not from running them.

### DIFF
--- a/checkin-app/scripts/migrated-routes.txt
+++ b/checkin-app/scripts/migrated-routes.txt
@@ -7,3 +7,5 @@
 GET /api/profile
 GET /api/programs/[id]
 GET /api/safety/board-contacts
+PATCH /api/attendance/manual/[id]
+DELETE /api/attendance/manual/[id]
--- a/checkin-app/src/app/api/attendance/manual/[id]/__tests__/route.test.ts
+++ b/checkin-app/src/app/api/attendance/manual/[id]/__tests__/route.test.ts
@@ -265,6 +265,11 @@ describe("household-lead correction of a member's visit", () => {
         (visitSubject as jest.Mock).mockImplementation(leadScope);
         visitFindUnique.mockResolvedValue(memberVisit);
         tx.visit.findFirst.mockResolvedValue({ id: 42 });
+        // The row written back is the MEMBER's. Leaving the default (personId =
+        // OWN_ID) let their_own resolve on it, which masks what led_households
+        // does or doesn't grant — the whole point of this block.
+        tx.visit.update.mockImplementation(async (args: { data: Record<string, unknown> }) => ({ ...memberVisit, ...args.data }));
+        (processVisitCheckout as jest.Mock).mockResolvedValue([{ ...memberVisit }]);
     });
 
     it("edits a member's visit, locking and matching on the MEMBER, auditing the actor", async () => {
@@ -320,6 +325,23 @@ describe("household-lead correction of a member's visit", () => {
         for (const internal of ["deletedAt", "deletedById", "forceCloseWarnedAt"]) {
             expect(visit).not.toHaveProperty(internal);
         }
+    });
+
+    // The other edge: led_households is gated on the roster, not on "is a lead".
+    // visitSubject (DB leadship) and ledHouseholdMemberIds (session householdLead
+    // + householdId) are different sources and can disagree on a stale session —
+    // when they do, the grant must not resolve and the times must not ship.
+    it("strips the times when the roster does not contain the visit's person", async () => {
+        mockSession.mockResolvedValue({ user: { id: OWN_ID, householdLead: true, householdId: HOUSEHOLD_ID } });
+        personFindMany.mockResolvedValue([{ id: OWN_ID }]); // MEMBER_ID absent
+
+        const res = await PATCH(req("PATCH", { arrivedAt: "2026-07-20T14:05:00Z" }), ctx as never);
+        expect(res.status).toBe(200);
+        const { visit } = await res.json();
+
+        expect(visit).not.toHaveProperty("arrivedAt");
+        expect(visit).not.toHaveProperty("departedAt");
+        expect(visit.personId).toBe(MEMBER_ID); // public tier still rides through
     });
 
     it("404s a visit belonging to someone outside the lead's household", async () => {
--- a/checkin-app/src/app/api/attendance/manual/[id]/__tests__/route.test.ts
+++ b/checkin-app/src/app/api/attendance/manual/[id]/__tests__/route.test.ts
@@ -9,6 +9,11 @@
  * emails the board, a delete never removes the row, closing an open visit
  * routes through processVisitCheckout, and every lock/where keys on the visit's
  * PERSON (not the actor) so a lead's write serializes against the member's kiosk.
+ *
+ * Both verbs go through @/security/handler, so the registered policy is what
+ * shapes the response: PATCH returns the `visit` envelope stripped to the
+ * registry's their_own/led_households:personal grant, DELETE ships no model
+ * data at all.
  */
 
 import type { NextRequest } from "next/server";
@@ -40,6 +45,9 @@ jest.mock("@/lib/prisma", () => ({
     default: {
         visit: { findUnique: jest.fn(), update: jest.fn() },
         auditLog: { create: jest.fn() },
+        // buildCallerContext reads the led-household roster (the route's only
+        // ctxNeeds prefetch) to resolve led_households.
+        person: { findMany: jest.fn() },
         $transaction: jest.fn(async (cb: (t: unknown) => unknown) => cb(tx)),
     },
 }));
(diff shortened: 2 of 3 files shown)


thpr (approved):
Reviewed at `b3ae7625`. Correct fix at the right layer.

What I verified independently of the PR body:

- **Envelope math holds.** `envelope: 'visit'` + bag `{ Visit }` → `{ visit: <stripped> }`; `envelope: null` + `{}` falls to the `keys.length !== 1` branch → body `{}`.
- **Status codes unchanged for every auth class.** Old `withAuth({})`: unauthenticated 401, kiosk 403. New `resolveAccess('authenticated')` is `auth.type === 'session'`, and `handler` maps a non-`unauthenticated` denial to 403 — same two outcomes. Error bodies also unchanged, since both paths end at `apiError`.
- **`flagged`/`success` are genuinely dead.** Both call sites (`my-visits/page.tsx:56`, `household/page.tsx:56`) read `res.ok` and `data.error`, then reload.
- **The chunked-close path still strips correctly.** `processVisitCheckout` returns full `visit.update`/`create` rows, so `personId` is present and `their_own`/`led_households` resolve on the surviving chunk.
- **Boundary isolation.** `is_boundary()` covers `src/security/**`, `src/middleware.ts` and three scripts; `scripts/migrated-routes.txt` is not in that set.

The follow-up commit is the one that matters. In `38ffa253` the lead block's `tx.visit.update` mock still returned `personId: OWN_ID`, so **`their_own`** resolved on the member's row and "keeps the times the lead may see" passed without `led_households` ever firing — the wrong-reason pass the original body claimed was impossible. `b3ae7625` returns the member's row and adds the roster-miss negative case, so the pin now tests what it says.

Two follow-ups filed rather than held against this PR:

- **#1614** — `visitSubject` authorizes on DB `canManage = isHouseholdLead || isSysadmin`, while the view resolves `led_households` off the session claim `auth.user.householdLead`. A sysadmin who is not a household lead (or a lead on a stale JWT) passes authz, the write lands, and the response comes back with no `arrivedAt`/`departedAt`. Latent only — both callers discard the body — but it is newly reachable now that the policy is live.
- **#1615** — the `DELETE` entry's comment still describes `{ success, flagged }`, and its `led_households` grant is inert yet still makes `deriveCtxNeeds` bill a `person.findMany` roster query per delete. Registry-only, so correctly out of this diff.

Smaller, non-blocking: `if (auth.type !== 'session') throw unauthorized()` is now unreachable but earns its place narrowing `auth.user` for tsc; a real fault logs twice (`logBackendError` then `handler`'s `console.error`), which the comment states is intended; and `my-visits/__tests__/page.test.tsx:25` still mocks `{ flagged: false }` for a response nothing reads.

Approving.

---
🤖 Review by [Claude Code](https://claude.com/claude-code) on behalf of @thpr

PR #1612, opened 2026-08-09
fix(membership): close three lifecycle gaps — merge carryover, individual agreement, legacy restore

jee7s:
Three independent defects in `lib/membership/**`, each re-verified against `origin/main` (`fbc59f74`) with `git show origin/main:<path>` before writing a line. All three were real. Built on #1595 and #1597 — neither is reverted.

---

## 1. Merge carryover skipped `PENDING_BG_CLEARANCE` (`mergeBgAdvance.ts`)

`advanceHouseholdBgAfterMerge` acted on `PENDING_EXTERNAL_ACTION`, `PENDING_PAYMENT` and `PENDING_BG_REVIEW`. A household that pays while its check is outstanding lands at `PENDING_BG_CLEARANCE` (`paidAt` set, `bgClearedAt` null, `orgMembership` INACTIVE). After a merge moved a still-valid `lastBackgroundCheck` onto the survivor, `householdBgFresh` returned true and all three stamp helpers ran — and the row matched none of their `fromWhere` clauses. The paid household stayed inactive with a valid check on file.

`MERGE_BG_CARRYOVER.md` (it lives at `docs/in-design/`, not `checkin-app/docs/designs/`) calls this out directly: its state matrix gives `PENDING_BG_CLEARANCE` **"stamp + `ACTIVE`" + flip `OrgMembership.status`, converging via the existing `clearBackgroundCheck` edge**, and its Reachability section says the parallel-track states are where essentially all the fix's value lives.

The fix follows the doc rather than inventing a path: `clearHeldReview` already did exactly this for `PENDING_BG_REVIEW`, so it is generalised over a two-entry `HELD_STATES` table. `clearBackgroundCheck` sees `paidAt` set and converges to `ACTIVE`, flips the membership, applies volunteer status and audits with #1597's provenance, in one write. `notifyClearanceOutcome` then fires `sendCongrats` + Trigger C post-transaction — the doc's recommendation. `sendCongrats` says "your membership is now active" and claims nothing about a review, so it does not reintroduce the #1597 problem; `notifyPaymentOpen` is unreachable from this state.

No new machine edge: `PENDING_BG_CLEARANCE → ACTIVE via clearBackgroundCheck` is already in `TRANSITIONS`, so `docs/generated/lifecycle/membership.md` needed no regeneration (`artifactsDrift` passes untouched).

### How the scoping hole was avoided

The prior spec's hole came from `householdBgIsFresh` filtering `isHouseholdLead: true` — a household-wide answer applied to a person-scoped row would clear an adult child's own check off a parent's date. Three things prevent it here, and none of them is new code:

- **The candidate query is unchanged.** Every acting state goes through `uncleared()`, whose `where` is `orgMembership: { householdId }`. `PERSON_BG` and `PERSON_AGREEMENT` both carry `orgMembershipId: null` (`personBgTriggers.ts:52`, `personAgreementTriggers.ts:128`), so they are excluded *structurally*, not by an added filter that could be dropped later. I widened the status set only, never the scope.
- **No person's date is written.** `clearBackgroundCheck` stamps `lastBackgroundCheck` for `subjectsWithTwoApprovals(attestations)`. The carryover runs only on rows with **zero** attestations (re-checked under `FOR UPDATE`), so that list is empty and no `person.updateMany` fires. The carried date is never overwritten with `now()`, and no household member is dated off a lead's check.
- **Both exclusions are now pinned by tests**, including a new `PERSON_AGREEMENT` case — the kind that shares `PENDING_EXTERNAL_ACTION` with an acting state and so is the one a scope widening would actually catch.

---

## 2. Settled `PERSON_AGREEMENT` violated `active-is-bg-cleared` (`lifecycle.ts`)

`markContractSigned` writes `{ contractSignedAt, status: 'ACTIVE', stageEnteredAt }` for `PERSON_AGREEMENT`, leaving `bgClearedAt` null. `scanLifecycleViolations` selects no kind and filters none, so the `lifecycle-reconcile` cron reported every adult child who signed to the sysadmin/board channel, permanently — noise on a channel whose value is that it is quiet.

**I scoped the invariant by kind rather than scoping the scanner, and rather than giving `PERSON_AGREEMENT` its own invariant set.** The reasoning, not the diff size:

- `docs/rules/membership.md` states two separate rules. Activation: *"payment and review run in parallel, and membership activates on whichever finishes last."* Individual agreement: *"An adult child of a household lead signs their own membership agreement"*, and *"It gates nothing."* `active-is-bg-cleared` is the **first** rule. It was written as if it were about rows; it is about memberships. Applying it to a row with no payment track and no check is a category error, not a missing exemption.
- **The machine already knew this.** `TRANSITIONS` carries a `kind`-tagged edge `PENDING_EXTERNAL_ACTION → ACTIVE, event: markContractSigned, kind: PERSON_AGREEMENT` (added in #1477). The transition table and the invariant list disagreed about the same machine. Putting `kind` on `ClassifyRow` makes them agree at the definition layer, which is where `lifecycle.ts`'s header says the one definition lives.
- **Scoping the scanner was the wrong layer.** `lifecycleDrift.ts` opens with *"This module NEVER re-derives an invariant."* A `kind: { in: [...] }` filter there is precisely a re-derivation — it encodes "which kinds owe a clearance" in the consumer, where the panel, the System Status route and the cron would each have to repeat it, and it would silently drop `intake-is-unpaid` coverage for those rows too.
- **A separate invariant set was an abstraction for one row shape.** The two kinds share `intake-is-unpaid`; only one clause differs. A second machine for one differing clause is more surface, not less.

`PERSON_BG` is *not* relaxed — `clearBackgroundCheck`'s `PERSON_BG` branch stamps `bgClearedAt` before flipping ACTIVE, so it genuinely owes the stamp. `SETTLES_ON_SIGNATURE` names the one kind that does not, and the test asserts `PERSON_BG` and `RENEWAL` still fail without it. **It is an exemption list, not an allowlist** (review, round 2): a fifth kind owes the stamp by default, and has to be exempted deliberately. Over the four declared kinds the two spellings are equivalent, which the exhaustive matrix below proves rather than asserts.

`classify` got the same clause: it and `validate` are the drift oracle for the same rows, and leaving `classify` returning `null` for a row `validate` calls clean would have been a new disagreement. The exhaustive `stateSpaceSafety` matrix now runs over status × **kind** × 4 flags and proves they agree on all 160 tuples.

---

## 3. `RENEWAL_PENDING_BG` archivable but not restorable (`archive.ts`)

`RESTORABLE_STATUSES` was `new Set(ARCHIVABLE_STATUSES)` on the rationale that "restore targets are exactly the statuses archive can come from — one list, so the two can never disagree." They disagreed: `archiveApplication` gates on `status === 'ACTIVE'` alone, so a renewal parked at the legacy `RENEWAL_PENDING_BG` is archivable today and writes `archivedFromStatus='RENEWAL_PENDING_BG'`; migration `20260806160000` backfills exactly that value for rows archived earlier. Those rows could never be unarchived, against `membership.md`'s *"An abandoned application is archived, never destroyed, and restores to the status it held before."*

The restore set is now derived from the **actual gate** — every status except `ACTIVE` and `ARCHIVED` — which is the eight rel/1.1 listed and the eight the migration backfills. That keeps the "one source" property the original comment wanted, but sources it from the rule that decides the question instead of from a list that answers a different one.

**I deliberately did not add `RENEWAL_PENDING_BG` to `ARCHIVABLE_STATUSES`.** That would have been the tempting one-word fix, and it is wrong twice: it injects `ARCHIVED → RENEWAL_PENDING_BG` into `TRANSITIONS`, which makes the status reachable and deletes the unreachable-legacy signal `machineSpecs.ts` says the report exists to surface (breaking two existing assertions and forcing an artifact regeneration for a pair of arrows nothing draws). The false premise in that list's comment — "nothing is ever archived from it" — is corrected in place, pointing at the gate.

**The divergence is now declared, not merely intended** (review, round 2). The gap that reasoning creates — a runtime restore edge the table does not carry — was left invisible to every test, and `lifecycle.test.ts:220` asserted the *opposite* of what ships while passing green. `RESTORABLE_STATUSES` moves into `lifecycle.ts` beside the other status lists and is exported, so the pure unit suite can see it; test #9 pins it to `[...ARCHIVABLE_STATUSES, ...LEGACY_STATUSES]` and records the undeclared edge instead of denying it. The two lists are linked again — that is what `ARCHIVABLE_STATUSES` was for — while `TRANSITIONS` stays silent, so the unreachable report is exactly as it was.

---

## Per-test table

Red = run against `origin/main`'s source with these tests in place; green = this branch. Full logs kept.

| # | Test | File | on `origin/main` | on branch | Kind |
|---|---|---|---|---|---|
| 1 | `activates a paid PENDING_BG_CLEARANCE household, without dating anyone's own check` | `merge/__tests__/route.integration.test.ts` | **FAIL** | PASS | regression (defect 1) |
| 2 | `never touches a PERSON_AGREEMENT re-pointed onto the survivor` | `merge/__tests__/route.integration.test.ts` | PASS | PASS | **guard** — pins the scope that must not widen |
| 3 | `migrated guards consume fromWhere(<from>)` (new `PENDING_BG_CLEARANCE` row) | `lifecycle/__tests__/guardFromStateParity.test.ts` | **FAIL** | PASS | regression (defect 1, guard shape) |
| 4 | `a settled PERSON_AGREEMENT is not a violation — the signature alone completes it` | `app/__tests__/lifecycleReconcile.integration.test.ts` | **FAIL** | PASS | regression (defect 2, end-to-end through the scanner) |
| 5 | `active-is-bg-cleared is the MEMBERSHIP convergence, and holds for the other kinds` | `membership/__tests__/lifecycle.test.ts` | **FAIL** | PASS | regression (defect 2, unit) |
| 6 | `an ACTIVE PERSON_AGREEMENT is on-diagram without a clearance` | `membership/__tests__/lifecycle.test.ts` | **FAIL** | PASS | regression (defect 2, `classify` half) |
| 7 | `every (status × kind × 4 flags) tuple is on-diagram-clean or off-diagram-flagged` | `lifecycle/__tests__/stateSpaceSafety.test.ts` | n/a | PASS | **guard** — widens an existing matrix onto the new axis; cannot run on `origin/main` (imports `ALL_KINDS`, added here) |
| 8 | `round-trips a renewal archived from the legacy RENEWAL_PENDING_BG` | `app/__tests__/membershipArchiveAPI.integration.test.ts` | **FAIL** | PASS | regression (defect 3) |
| 9 | `GUARD: the restore set is the archivable set plus the legacy status, and nothing else` | `membership/__tests__/lifecycle.test.ts` | n/a | PASS | **guard** — added in review round 2; cannot run on `origin/main` (imports `RESTORABLE_STATUSES`, exported here) |

Six regression tests, three guards, labelled honestly — #2, #7 and #9 do not fail on `origin/main` and are not claimed as regressions.

The three guards were mutation-tested rather than assumed, since a guard that cannot fail is decoration:

| Mutation | Expected to break | Result |
|---|---|---|
| Add `RENEWAL_PENDING_BG` to `ARCHIVABLE_STATUSES` (the "tempting one-word fix") | #9 + the existing unreachable-legacy assertion | both **FAIL** ✔ |
| Narrow `RESTORABLE_STATUSES` back to `new Set(ARCHIVABLE_STATUSES)` (defect 3, restored) | #9 | **FAIL** ✔ |
| Add `PERSON_BG` to `SETTLES_ON_SIGNATURE` (relaxing the kind that must not be relaxed) | #5 | **FAIL** ✔ |

## Verification

Re-run after review round 2:

- `npm test` (full unit suite) — **276 suites / 2694 tests, all pass** (+1: guard #9)
- `npm run test:integration` — the three suites this PR touches (`membershipArchiveAPI`, `lifecycleReconcile`, `merge/route`) — **85 tests, all pass**
- `npx tsc --noEmit` — **0 errors**
- `npm run lint` — clean, `--max-warnings 0`
- `git diff --name-only origin/main...HEAD -- checkin-app/src/security/` — **empty**; the boundary-isolation gate is untouched
- No migration, no schema change, no generated-artifact change — round 2 touches three files: `lifecycle.ts`, `archive.ts`, `lifecycle.test.ts`

### DIFF
--- a/checkin-app/src/app/__tests__/lifecycleReconcile.integration.test.ts
+++ b/checkin-app/src/app/__tests__/lifecycleReconcile.integration.test.ts
@@ -134,6 +134,20 @@ describe('lifecycle reconciler — invariant-driven sweep over a real DB', () =>
         expect(after.paidAt).not.toBeNull();
     });
 
+    it('a settled PERSON_AGREEMENT is not a violation — the signature alone completes it', async () => {
+        // markContractSigned flips an individual agreement straight to ACTIVE with no
+        // bgClearedAt: it has no membership, no payment and no background-check gate.
+        // Judging it by the membership convergence reports every adult child who signs.
+        const proc = await prisma.orgMembershipProcess.create({
+            data: { kind: 'PERSON_AGREEMENT', status: 'ACTIVE', subjectPersonId: selfId, contractSignedAt: new Date() },
+        });
+        processIds.push(proc.id);
+
+        const { violations } = await scanLifecycleViolations();
+
+        expect(violations.filter((v) => v.key === `process ${proc.id}`)).toEqual([]);
+    });
+
     it('clean: a row that validates clean produces no violation', async () => {
         await prisma.programParticipant.create({
             data: { programId, personId: selfId, status: 'ACTIVE', inventoryHeldAt: null },
--- a/checkin-app/src/app/__tests__/membershipArchiveAPI.integration.test.ts
+++ b/checkin-app/src/app/__tests__/membershipArchiveAPI.integration.test.ts
@@ -225,6 +225,25 @@ describe('Membership application archive API', () => {
         expect(audit?.newData).toMatchObject({ status: 'PENDING_PAYMENT', unarchived: true });
     });
 
+    it('round-trips a renewal archived from the legacy RENEWAL_PENDING_BG', async () => {
+        // archiveApplication gates on ACTIVE alone, so this status is archivable today —
+        // and migration 20260806160000 backfills it for rows archived before the column
+        // existed. A restore list narrower than the archive gate strands them in ARCHIVED.
+        const hh = await prisma.household.create({ data: { name: `LegacyRenewal ${TAG}` } });
+        const m = await prisma.orgMembership.create({ data: { householdId: hh.id, status: 'ACTIVE' } });
+        const p = await prisma.orgMembershipProcess.create({ data: { orgMembershipId: m.id, kind: 'RENEWAL', status: 'RENEWAL_PENDING_BG' } });
+        asBoard(boardId);
+
+        expect((await ARCHIVE(req({ processId: p.id }) as never)).status).toBe(200);
+        expect((await prisma.orgMembershipProcess.findUnique({ where: { id: p.id } }))?.archivedFromStatus).toBe('RENEWAL_PENDING_BG');
+
+        const res = await UNARCHIVE(unarchiveReq({ processId: p.id }) as never);
+        expect(res.status).toBe(200);
+
+        const after = await prisma.orgMembershipProcess.findUnique({ where: { id: p.id } });
+        expect({ status: after?.status, from: after?.archivedFromStatus }).toEqual({ status: 'RENEWAL_PENDING_BG', from: null });
+    });
+
     it('refuses to unarchive a non-archived process (409 wrong_phase)', async () => {
         const { processId } = await makeProcess('NotArchived', 'PENDING_PAYMENT');
         asBoard(boardId);
--- a/checkin-app/src/app/api/membership-ops/participants/merge/__tests__/route.integration.test.ts
+++ b/checkin-app/src/app/api/membership-ops/participants/merge/__tests__/route.integration.test.ts
@@ -1173,6 +1173,23 @@ describe("Merge Participants API", () => {
             expect(await prisma.auditLog.count({ where: { tableName: "OrgMembershipProcess", affectedEntityId: process.id } })).toBe(0);
         });
 
+        it("activates a paid PENDING_BG_CLEARANCE household, without dating anyone's own check", async () => {
+            // The one state where the carried check is the SOLE block: dues are settled and
+            // the membership sits INACTIVE. A stamp that doesn't converge leaves it there.
+            const process = await makeApplication("PENDING_BG_CLEARANCE", { contractSignedAt: new Date(), bgConsentAt: new Date(), paidAt: new Date() });
+            const child = await prisma.person.create({ data: { name: "Adult Child", householdId, isHouseholdLead: false } });
+            extraPersonIds.push(child.id);
+
+            expect((await POST(mergeReq(pKeepId, pMergeId))).status).toBe(200);
+
+            const after = await reload(process.id);
+            expect({ status: after.status, cleared: after.bgClearedAt !== null }).toEqual({ status: "ACTIVE", cleared: true });
+            expect((await prisma.orgMembership.findUniqueOrThrow({ where: { id: membershipId } })).status).toBe("ACTIVE");
+            // The clearance carries zero attestations, so it names no subject and stamps
+            // no person. A household-wide stamp would date this child off a parent's check.
+            expect((await prisma.person.findUniqueOrThrow({ where: { id: child.id } })).lastBackgroundCheck).toBeNull();
+        });
+
         it("never touches a PERSON_BG re-pointed onto the survivor", async () => {
             // householdBgIsFresh asks about household LEADS. Answering it for an adult
             // child's own check would clear them because a parent's check is fresh.
@@ -1186,6 +1203,24 @@ describe("Merge Participants API", () => {
             const after = await reload(personBg.id);
             expect({ status: after.status, cleared: after.bgClearedAt, subject: after.subjectPersonId })
                 .toEqual({ status: "PENDING_BG_REVIEW", cleared: null, subject: pKeepId });
+        });
+
+        it("never touches a PERSON_AGREEMENT re-pointed onto the survivor", async () => {
+            // The other orgMembership-less kind, and the one that shares a status with an
+            // acting state. Every candidate query goes through orgMembership, which both
+            // PERSON kinds sit outside; widening that scope would sign an adult child's
+            // own agreement off a parent's background check.
+            const agreement = await prisma.orgMembershipProcess.create({
+                data: { kind: "PERSON_AGREEMENT", subjectPersonId: pMergeId, status: "PENDING_EXTERNAL_ACTION" },
+            });
+            createdProcessIds.push(agreement.id);
+            await makeApplication("PENDING_BG_CLEARANCE", { contractSignedAt: new Date(), bgConsentAt: new Date(), paidAt: new Date() });
+
+            expect((await POST(mergeReq(pKeepId, pMergeId))).status).toBe(200);
+
+            const after = await reload(agreement.id);
+            expect({ status: after.status, cleared: after.bgClearedAt, signed: after.contractSignedAt })
+                .toEqual({ status: "PENDING_EXTERNAL_ACTION", cleared: null, signed: null });
         });
 
         it("attributes the carryover to the merging board member, naming the record it came from", async () => {
--- a/checkin-app/src/lib/lifecycle/__tests__/guardFromStateParity.test.ts
+++ b/checkin-app/src/lib/lifecycle/__tests__/guardFromStateParity.test.ts
@@ -69,10 +69,11 @@ const GUARDS: GuardSite[] = [
     { file: 'app/api/finance-ops/membership-payment-plans/route.ts', machine: 'membership', from: 'PENDING_PAYMENT', edge: ['PENDING_PAYMENT', 'ACTIVE'], mode: 'spread' },
     { file: 'app/api/finance-ops/membership-payment-plans/refuse/route.ts', machine: 'membership', from: 'PENDING_PAYMENT', mode: 'spread' }, // deny: flag-only, no status edge
     // merge BG carryover — two flag-only stamps (advanceExternalIfComplete / activate own the
-    // edges those rows later take) and one real clearance on an already-declared edge.
+    // edges those rows later take) and two real clearances on already-declared edges.
     { file: 'lib/membership/mergeBgAdvance.ts', machine: 'membership', from: 'PENDING_EXTERNAL_ACTION', mode: 'spread' },
     { file: 'lib/membership/mergeBgAdvance.ts', machine: 'membership', from: 'PENDING_PAYMENT', mode: 'spread' },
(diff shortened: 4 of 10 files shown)


thpr (commented):
All three defects are real and I re-verified each against the branch source. The fixes match `docs/in-design/MERGE_BG_CARRYOVER.md`'s state matrix and the rules register. Three findings, all in the definition layer rather than the behaviour.

## 1. Defect 3 makes the machine definition lie, and a live test asserts the opposite of shipped behaviour

`archive.ts` can now restore `ARCHIVED → RENEWAL_PENDING_BG`. `TRANSITIONS` declares no such edge — deliberately, and the reasoning for keeping it out is sound. But `src/lib/membership/__tests__/lifecycle.test.ts:220` still reads:

```ts
// Nothing is ever archived from these two, so neither is a restore target.
expect(isLegalTransition('ARCHIVED', 'RENEWAL_PENDING_BG')).toBe(false);
```

That assertion passes on this branch and now documents the reverse of what the code does. The generated diagram carries seven `unarchiveApplication` arrows; the code has eight restore targets.

The stale comment is the smaller half. Before this PR, `RESTORABLE_STATUSES` was `new Set(ARCHIVABLE_STATUSES)` — the same array `TRANSITIONS` maps into its archive/unarchive edges, so code and table were locked together structurally. They now come from two independent lists (`ALL_STATUSES` on one side, `ARCHIVABLE_STATUSES` on the other) with nothing asserting any relationship between them. The divergence you intend is invisible, and so would be an unintended one.

Cheap pin that keeps the unreachable report exactly as it is: export `RESTORABLE_STATUSES` and assert it equals `[...ARCHIVABLE_STATUSES, ...LEGACY_STATUSES]`, then rewrite that test to record the deliberate gap instead of denying it.

## 2. `CONVERGING_KINDS` is an allowlist, so it fails open for any kind added later

`lifecycle.ts:242`. `active-is-bg-cleared` now holds whenever `!CONVERGING_KINDS.includes(r.kind)`. A fifth kind added to `OrgMembershipProcessKind` is exempt from the invariant by default — no compile error, no failing test (`classify` carries the same clause, so the exhaustive matrix stays green), and the reconciler silently stops reporting `ACTIVE` without a clearance for it.

Same diff size, fails closed:

```ts
/** The one kind that reaches ACTIVE on a signature, owing no clearance stamp. */
const SETTLES_ON_SIGNATURE: readonly ProcessKind[] = ["PERSON_AGREEMENT"];
```

A new kind then owes the stamp until someone deliberately exempts it, which is the direction an invariant should default in.

## 3. `ALL_KINDS` does not feed the kind enum-parity test

`lifecycle.test.ts:259` keeps its own inline `['INITIAL', 'RENEWAL', 'PERSON_BG', 'PERSON_AGREEMENT']` while the new export sits beside it unused by that assertion. Passing `ALL_KINDS` there removes the duplicate list and makes the new status × kind matrix provably exhaustive against the schema enum rather than exhaustive by hand.

## Checked, and not findings

- The congrats send on the new `PENDING_BG_CLEARANCE → ACTIVE` path is not a repeat of #1597: `activate()` returns `held` for that branch and sends nothing, so the carryover's `sendCongrats` is the first and only one. Trigger C is likewise a first fire.
- `HELD_STATES` omitting `RENEWAL_PENDING_BG` matches the design doc's state matrix, which gives that status **nothing**. Not an oversight.
- The merge route already carries the conflict-of-interest check (`route.ts:258`), so widening the carryover into the state where the doc says the value lives does not open the "clear your own family" hole rule 3 names.

---

thpr (changes requested):
Marking my earlier review as changes requested. The three defects and their fixes are right; what blocks is the definition layer the fixes leave behind.

**Blocking:**

1. `ARCHIVED → RENEWAL_PENDING_BG` is now a runtime transition with no declared edge, and `lifecycle.test.ts:220` asserts it is illegal — passing, while the code does it. Either declare the divergence somewhere a test can see (export `RESTORABLE_STATUSES`, assert it equals `[...ARCHIVABLE_STATUSES, ...LEGACY_STATUSES]`) or re-point that assertion at what ships. As it stands the restore set and the transition table are two unlinked lists, which is the drift `ARCHIVABLE_STATUSES` existed to prevent.

2. `CONVERGING_KINDS` as an allowlist silently exempts any future kind from `active-is-bg-cleared`, with no compile error and no failing test. Invert it to the one-kind exemption list so a new kind owes the stamp by default.

**Non-blocking:** finding 3 (`ALL_KINDS` feeding the enum-parity assertion) is a one-line cleanup — take it or leave it.

Nothing here needs the fixes themselves reworked; both are edits to the lists, not to the behaviour the tests pin.

---

thpr (approved):
Both blockers cleared in `0c6bccd8`, and the non-blocking third taken. Verified by reading the diff, not by running the suite.

**Restore set.** `RESTORABLE_STATUSES` now lives in `lifecycle.ts` beside the other status lists and is exported; the new GUARD asserts it equals `[...ARCHIVABLE_STATUSES, ...LEGACY_STATUSES]` and re-points the `:220` assertion to record the divergence instead of denying it. Keeping the list derived from the archive gate rather than hand-writing it is the better of the two shapes I suggested — it still tracks the thing that actually decides the question, and a status added to the enum grows the set and fails the GUARD until someone classifies it as archivable or legacy. `TRANSITIONS` and the unreachable-legacy report are untouched, which was the point.

**Kind exemption.** `SETTLES_ON_SIGNATURE` is equivalent to the old `!CONVERGING_KINDS.includes(...)` across all four declared kinds in both `classify` and the invariant, and a fifth kind now owes the clearance stamp by default. `PERSON_BG`'s assertion still fails the moment anyone adds it to the exemption list.

**Parity.** `ALL_KINDS` feeding the kind parity assertion closes the chain: a new kind fails parity, which forces `ALL_KINDS`, which puts it in the exhaustive status × kind × flags matrix.

No stale `CONVERGING_KINDS` references. `archive.ts` dropping the `OrgMembershipProcessStatus` import types through the union-parity equivalence, and `lifecycle.ts` stays client-safe.

Approving.

---

PR #1613, opened 2026-08-09
fix(time): resolve the display timezone through context, not module state

jee7s:
The organisation's display timezone lived in a mutable module-level variable in
`lib/time`, which `TimezoneProvider` wrote during render. On the server that module is a
per-process singleton shared by every concurrent request, so the zone in effect belonged
to whichever request rendered its provider most recently — not to the request whose page
was being formatted. Check-in and departure times rendered an hour off for the request
that lost the race.

The window is not theoretical. `resolveDisplayTimezone()` falls back to `APP_TIMEZONE`
(`America/Chicago`) whenever the settings read throws, and Aurora scales to zero here, so
a cold process serves some requests the fallback and others the configured
`America/New_York` while both write the same variable.

## The fix

The zone now flows through a React context, which is per-render-tree and therefore cannot
be shared between concurrent requests. This is the pattern `EnvProvider` already uses in
this repo for the same job — carrying a server-resolved value into client components.

The raw formatters in `lib/time` take the zone in their options, and the type makes it
**required**:

```ts
export type ZonedOptions = Intl.DateTimeFormatOptions & { timeZone: string };
export function formatDateTime(date: DateInput, options: ZonedOptions): string
```

That matters more than it looks. Removing the global without requiring the zone would
have traded a rare race for a common footgun: every one of the 39 client call sites that
today gets the right zone by accident would silently fall back to the compiled-in default
instead, which is exactly what `docs/conventions.md` ("A moment is displayed in the zone
the organisation configured, never one compiled in") forbids. Requiring it makes that a
compile error rather than a wrong time on a screen.

Client components get the formatters pre-bound to the tree's zone from `useOrgTime()`, so
call sites are unchanged:

```tsx
const { formatDateTime } = useOrgTime();
```

`setDisplayTimezone` and `getDisplayTimezone` are deleted. The two server callers
(`notifications.ts`, `attendance/manual/[id]/route.ts`) already passed `{ timeZone }`
explicitly and are untouched.

`formatDateOnly` remains UTC-pinned and is deliberately not part of this change — a day
still has no zone.

## Red, then green

The regression test is `src/components/__tests__/timezoneIsolation.test.tsx`. It renders
two trees, one per "request": request A is in `America/New_York` and suspends before its
leaf formats; request B then renders in the `APP_TIMEZONE` fallback; A's leaf finally
renders. That is the production interleave — provider A, provider B, then A's leaf —
made deterministic with a promise gate instead of timing.

Against `origin/main`'s source (`git checkout origin/main -- checkin-app/src`, scenario
written against main's API, since the fix necessarily changes how a leaf obtains the
zone):

```
FAIL src/components/__tests__/timezoneIsolation.test.tsx
  ✕ keeps request A in its own zone while request B renders in another
    Expected substring: "10:30"
    Received string:    "8/31/2026, 9:30:00 PM"
```

Request A rendered 9:30 PM — request B's Chicago — instead of its own 10:30 PM New York.
An hour off, which is the reported symptom.

On this branch:

```
Test Suites: 277 passed, 277 total
Tests:       2691 passed, 2691 total
```

**One honest caveat about that pair of runs.** The two files are not byte-identical, and
cannot be: the fix removes the global the old leaf read, so the leaf must obtain the zone
differently. What is identical between the red and green runs is the scenario — the same
two providers, the same interleave, the same instant, the same two assertions. Only the
three lines by which a leaf gets its zone differ. A single unchanged file cannot span
both APIs, and I would rather say that than imply a stronger proof than exists.

Control run: the full suite is green on pristine `origin/main` too (276 suites, 2691
tests), so the branch result is not being read against a broken baseline.

## Verified

- `npx tsc --noEmit` — clean, 0 errors. The two known-environmental
  `@aws-sdk/client-lambda` errors did not appear after a fresh `npm ci`.
- `npm run lint` — clean.
- `git diff --name-only origin/main...HEAD -- checkin-app/src/security/` — empty.

## Out of scope, but found while working

Three files hardcode `America/Chicago` inline rather than reading the configured zone, so
they will render the wrong time for an organisation in any other zone. They are a
separate defect from this race — no shared mutable involved — and fixing them here would
have buried this diff:

- `src/app/attendance/certifications/page.tsx:195`
- `src/app/my-programs/attendance/page.tsx:60`
- `src/app/program-ops/events/page.tsx:16`

## Not verified

- No browser or deployed-environment check; the evidence is the unit suite, tsc and lint.
- The SSR bleed is reproduced through client-render suspension in jsdom, not through an
  actual concurrent Next.js server render. The mechanism is the same shared module
  instance, but I did not stand up two real overlapping HTTP requests.
- `src/app/safety/trusted-adults/__tests__/page.test.tsx` failed once during a full run on
  this branch, on a `waitFor` against a fetch mock. It passes in isolation on both this
  branch and `origin/main`, and passed on a full re-run here. It has no reference to
  `lib/time` or `TimezoneProvider`. I believe it is a flake rather than fallout from this
  change, but I saw it fail once and am not going to claim more certainty than that.

### DIFF
--- a/checkin-app/src/components/TimezoneProvider.tsx
+++ b/checkin-app/src/components/TimezoneProvider.tsx
@@ -29,6 +29,8 @@ export function useDisplayTimezone(): string {
     return useContext(TimezoneContext);
 }
 
+type ZoneFreeOptions = Omit<Intl.DateTimeFormatOptions, 'timeZone'>;
+
 /**
  * The instant formatters bound to this tree's configured zone — the client-side seam
  * for `lib/time`, whose raw formatters take the zone explicitly. Safe to call from
@@ -37,8 +39,6 @@ export function useDisplayTimezone(): string {
  * `timeZone` is omitted from the options: the tree's zone always wins, so accepting
  * one and discarding it would be the same silent-wrong-zone footgun this seam closes.
  */
-type ZoneFreeOptions = Omit<Intl.DateTimeFormatOptions, 'timeZone'>;
-
 export function useOrgTime() {
     const timeZone = useDisplayTimezone();
     return useMemo(
--- a/checkin-app/src/components/TimezoneProvider.tsx
+++ b/checkin-app/src/components/TimezoneProvider.tsx
@@ -1,19 +1,52 @@
 "use client";
 
-import { setDisplayTimezone } from '@/lib/time';
+import { createContext, useContext, useMemo } from 'react';
+import {
+    APP_TIMEZONE,
+    formatDate,
+    formatDateTime,
+    formatTime,
+    formatVisitRange,
+    type DateInput,
+} from '@/lib/time';
 
 /**
- * Installs the organisation's configured display timezone (AppSettings.timezone) for the
- * instant formatters in lib/time, which client components call as plain functions from
- * render bodies, memos and callbacks alike.
+ * Carries the organisation's configured display timezone (AppSettings.timezone) into
+ * client components. The root layout (a server component) resolves it per request and
+ * feeds it in here; leaf components read it via `useOrgTime()`.
  *
- * A provider rather than a prop: the formatters are called from leaf components all over
- * the tree, most of them client pages with no server parent to thread a prop from. The
- * value is installed during render, not in an effect, so children — including their
- * server-rendered first pass — format in the configured zone from the start rather than
- * flashing the fallback.
+ * Context rather than module state: on the server one module instance is shared by every
+ * concurrent request, so an installed value belongs to whichever request wrote it last.
  */
+const TimezoneContext = createContext<string>(APP_TIMEZONE);
+
 export function TimezoneProvider({ value, children }: { value: string; children: React.ReactNode }) {
-    setDisplayTimezone(value);
-    return <>{children}</>;
+    return <TimezoneContext.Provider value={value || APP_TIMEZONE}>{children}</TimezoneContext.Provider>;
+}
+
+/** The organisation's display timezone for this render tree. */
+export function useDisplayTimezone(): string {
+    return useContext(TimezoneContext);
+}
+
+/**
+ * The instant formatters bound to this tree's configured zone — the client-side seam
+ * for `lib/time`, whose raw formatters take the zone explicitly. Safe to call from
+ * render bodies, memos and callbacks, since it returns plain functions.
+ */
+export function useOrgTime() {
+    const timeZone = useDisplayTimezone();
+    return useMemo(
+        () => ({
+            formatDate: (date: DateInput, options?: Intl.DateTimeFormatOptions) =>
+                formatDate(date, { ...options, timeZone }),
+            formatTime: (date: DateInput, options?: Intl.DateTimeFormatOptions) =>
+                formatTime(date, { ...options, timeZone }),
+            formatDateTime: (date: DateInput, options?: Intl.DateTimeFormatOptions) =>
+                formatDateTime(date, { ...options, timeZone }),
+            formatVisitRange: (arrived: DateInput, departed?: DateInput) =>
+                formatVisitRange(arrived, departed, timeZone),
+        }),
+        [timeZone],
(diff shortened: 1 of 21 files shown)


checkin-app/src/components/TimezoneProvider.tsx
```diff
@@ -1,19 +1,52 @@
...
+export function useOrgTime() {
+    const timeZone = useDisplayTimezone();
+    return useMemo(
+        () => ({
+            formatDate: (date: DateInput, options?: Intl.DateTimeFormatOptions) =>
+                formatDate(date, { ...options, timeZone }),
+            formatTime: (date: DateInput, options?: Intl.DateTimeFormatOptions) =>
+                formatTime(date, { ...options, timeZone }),
+            formatDateTime: (date: DateInput, options?: Intl.DateTimeFormatOptions) =>
+                formatDateTime(date, { ...options, timeZone }),
```
thpr: The spread order silently overrides a caller-supplied `timeZone`, and it inverts the precedence `lib/time` had on `main`.

On `main` the formatters did `{ timeZone: displayTimezone, ...options }` — the caller's option won. Here `timeZone` spreads last, so the org zone wins. No current call site passes `timeZone`, so nothing changes today; the issue is that the parameter type is `Intl.DateTimeFormatOptions`, which still advertises `timeZone` as accepted, and then discards it.

That is the one footgun class this PR set out to close — a zone that is wrong on screen with nothing failing at compile time. Making it a compile error is a one-word change:

```ts
formatDate: (date: DateInput, options?: Omit<Intl.DateTimeFormatOptions, 'timeZone'>) =>
    formatDate(date, { ...options, timeZone }),
```

(and the same for `formatTime` / `formatDateTime`).

checkin-app/src/components/__tests__/timezoneIsolation.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+// Request A's leaf suspends on a slow read and formats only once it resolves, so its
+// render lands after another request has already rendered.
+let release!: () => void;
+const gate = new Promise<void>((r) => {
+    release = r;
+});
+let open = false;
+void gate.then(() => {
+    open = true;
+});
```
thpr: The gate is module-scoped and single-use: `open` latches `true` after the first test and never resets.

A second `it` added to this file would find `LateStamp` never suspending, so both trees would render in order, the interleave the suite exists to reproduce would not happen, and the test would pass while asserting nothing. That is the failure mode worth avoiding in a regression guard specifically — it stays green either way.

Moving `gate`/`release`/`open` into the `it` body (or a `beforeEach`) keeps the scenario intact and makes the next test inherit a fresh gate.

thpr (commented):
Two small things, neither blocking. The fix itself is right and the sweep is complete — I checked that `etDisplayTimezone` has zero hits left anywhere including docs, that the only remaining raw-formatter callers are the two server ones already passing `resolveDisplayTimezone()`, and that `src/app/layout.tsx` is the only root layout, so provider coverage is total.

*Posted by Claude Code on behalf of @thpr.*

thpr (approved):
Both follow-ups land correctly.

The `Omit<Intl.DateTimeFormatOptions, 'timeZone'>` guard is real rather than decorative — `tsconfig.json` includes `**/*.tsx` with only `node_modules` excluded, so `tsc --noEmit` type-checks the test file and an unused `@ts-expect-error` would fail the build. It verifies in both directions: the directive fires if the option is accepted again, and errors as unused if the type ever stops rejecting it.

The rebuilt gate is right too — `open`, `gate` and `release` all reset per test, and `LateStamp` reads the current test's `gate` at throw time.

Non-blocking nit: the `type ZoneFreeOptions` line went in between the JSDoc block and `useOrgTime`, so that doc comment now documents the type alias and the hook the 19 call sites hover has none. Moving the `type` above the comment restores it.

*Posted by Claude Code on behalf of @thpr.*

Issue #1569, opened 2026-08-08
Index directory forks the section permission table instead of reading it


thpr:
The `/index` page directory decides what to show from its own copy of the app's
permission rules, not from the rules the sections actually enforce. The two have
already diverged, and the divergence hides pages from people who can reach them.

`checkin-app/src/components/pageRegistry.ts` declares its own predicates
(`BOARD`, `BOARD_ONLY`, `BOARD_OR_OPS`, `SAFETY`, `SHOP`, `SYSADMIN`,
`HOUSEHOLD_LEAD`, `SIGNED_IN`, `PUBLIC`) and maps every route onto one of them.
Only `leadsAnyProgram` is imported from shared code; everything else is a
retyped copy of what `NAV_ITEMS[].visible` in `AppFrame.tsx` and each section
`layout.tsx` already say. `shopRoles()` is exported from `src/lib/shopNav.ts`
and the registry reimplements its body inline rather than calling it.

This is what `docs/conventions.md` ("The server decides, the client mirrors")
rules out: the interface hides a control on the same rule the real gate
enforces, and the two are written against one source so they cannot drift.

## The drift that exists today

**A background-check reviewer sees no Membership Ops rows.** The
`membership-ops` layout admits `isSysadmin`, `isBoardMember`,
`isBackgroundCheckReviewer` and `isOperations`; the nav admits reviewers and its
`hrefFor` sends a reviewer-only user straight to `/membership-ops/review`. The
registry maps every membership-ops route to `BOARD` (sysadmin or board), except
`/membership-ops/participants`. So the Index hides the one tab a reviewer is
routed to. `RegistryUser` does not carry `isBackgroundCheckReviewer` at all, so
no per-route override can express it either.

**`/shop-ops/create` is listed to people the tab hides.** `SHOP_NAV_LINKS`
gates Create on `isAdmin` (sysadmin or board); the registry lists it under
`SHOP` (`isCertifier`). A certifier who is neither sees the link and cannot use
the page.

Neither is a security hole. Every target re-enforces its own gate on arrival, so
an over-listed row at worst produces a link that refuses. The reviewer case is
the one with user-visible cost: a page they may use is missing from the
directory that exists to find pages.

## What finishing it means

Have the registry read one source per section instead of holding a second copy.
`shopRoles()` is the shape to follow: a named predicate exported from the module
that owns the section, called by both the nav and the registry. Sections whose
[...]

PR #1623, opened 2026-08-10
feat(facility-ops): grant Operations the aggregate tools only (Trends + Print ID Badges)

dkaygithub:
## What

Grants the **Operations** role the two Facility Ops tools that read attendance in aggregate, and nothing else:

- 🖨️ Print ID Badges (`/facility-ops/print-badges`)
- 📈 Participation Trends (`/facility-ops/trends`) + `GET /api/facility/trends`

Plus `isOperations` on the `/facility-ops` layout gate, so those two are reachable, and the matching entries in the page directory.

## Why this is smaller than it was

#1633 records the decision, at `[Decision — Principle: least privilege]` tier:

> Operations reach attendance in aggregate only — the trends, and printing the ID badges. One person's record sits outside that reach: operations do not record, correct or remove a visit, do not read the raw badge events behind one, and do not review other people's corrections.

So the answer to **#1476** is **no**. This PR previously granted all four Facility Ops tools; that is now withdrawn down to the two the rule allows. **#1633 has already merged** (`ee33d50c`) — this PR must not land ahead of the decision it implements, and it hasn't.

### Withdrawn

| | |
|---|---|
| `/facility-ops/visits` page gate | reverted — operations does not reach one person's visit record |
| `GET/PATCH/DELETE /api/facility/visits` | reverted to `['isSysadmin', 'isBoardMember']` |
| `/facility-ops/badges` page gate | operations removed |
| `GET /api/facility/badges` | reverted to `['isSysadmin', 'isBoardMember']` |
| `pageRegistry` rows for Visits, Badges, and the `/facility-ops` index | back to `BOARD` |

`/facility-ops/badges` is the **raw badge-event log**, not badge printing. #1633's reviewer note calls that distinction out explicitly: reading it as "printing the ID badges" would grant precisely what the rule withholds.

The **`Add Visit` gating** added in review goes with it. It was there because the page gate had been widened past the insert gate; with the page back at `['isSysadmin', 'isBoardMember']` the two sets are identical again, so the conditional was dead code and the button renders unconditionally, as on `main`.

Review findings **1, 2 and 6** dissolve with the visits grant. One piece of finding 6 survives on its own merits: the `DELETE /api/facility/visits` row in `authzRoleRejection`'s `roleGated` sweep. That was never about the grant — the one destructive verb on the route simply had no 401/403 coverage in the harness at all. It stays, and now also carries the operations-is-denied assertion.

## What is kept

- Trends page gate + `GET /api/facility/trends`
- Print ID Badges — no page gate of its own; it rides the layout gate, and its APIs already admit operations (`/api/people/search`) or are a public catalog (`/api/programs`)
- `isOperations` on the `/facility-ops` layout gate
- `pageRegistry`: `BOARD_OR_OPS` on **Trends and Print Badges only**
- **Unrelated pre-existing fix:** `facility-ops/badges/page.tsx` gated on `['isSysadmin']` alone, while both the layout gate and `GET /api/facility/badges` admit board members — and `useRequireRole` has no implicit board/sysadmin fallthrough, so a board member was bounced off a page its API serves them. `isBoardMember` is restored. This bug has nothing to do with the operations grant; it is kept here because it is the line this PR edits.

## The dead tabs #1633 does not solve

Operations can now reach `/facility-ops` for two tools, but `SectionTabs` is not role-aware and the layout renders all five `FACILITY_NAV_LINKS`. Three of those (Visits, Badges, Corrections) eject operations to `/` — out of the section entirely. Before this PR operations could not reach the section at all, so the dead tabs did not exist for them; **this PR creates them**, and has to close them.

Fixed at the source rather than with a second list:

- `FACILITY_NAV_LINKS` now carries the `roles` each tool's page gates on, named for the rule: `FACILITY_RECORD_ROLES` (one person's record — Visits, Badges, Corrections) and `FACILITY_AGGREGATE_ROLES` (Print ID Badges, Trends).
- Each of those four pages passes that same array to `useRequireRole`. The tab and the gate are one array, so they cannot disagree.
- `FACILITY_SECTION_ROLES` — the layout gate — is **derived** as the union of the tab roles, so the section can never admit someone to an empty tab strip.
- The layout filters the strip through `visibleFacilityLinks(user)`.

No permission table is duplicated: this removes a copy rather than adding one. The `pageRegistry` test now reads its expectations off `FACILITY_NAV_LINKS` too, which is the narrow, section-scoped version of what **#1569** asks for globally.

## Boundary and register checks

- **`src/security/**` is untouched** — diff is empty.
- `registry.ts` (lines 77–82) records that the insert gate, the sibling `GET/PATCH/DELETE /api/facility/visits` gates, and the `facility-ops/visits` page gate must stay equal. Verified after this change: all three are `['isSysadmin', 'isBoardMember']`. **The comment stays true**, so nothing needs correcting there. (Its trailing "Widening to isOperations is an open decision, #1476" is now answered by #1633 — a comment-only edit in a boundary file, so it needs its own PR. Noted, not done here.)
- `docs/rules/attendance-checkin.md` needs no amendment: #1633 added the operations line beside the existing board/sysadmin rule rather than qualifying it, and this PR no longer contradicts either.
- No registered route is touched. `POST /api/facility/visits/insert` keeps its gate and is now asserted to deny operations.

## Tests

- **Layout:** operations is admitted; operations sees only Print ID Badges and Participation Trends, and *not* Visits / Badges / Corrections; a board member still sees all five.
- **Visits page / Badges page:** an operations user is redirected to `/`.
- **Badges page:** a board member is admitted (the pre-existing fix).
- **Trends page:** an operations user is admitted.
- **`pageRegistry`:** every Facility Ops row is listed to operations *iff* its page admits them — asserted against `FACILITY_NAV_LINKS`, not a retyped list. The index stays hidden. Board still sees everything.
- **`authzRoleRejection.integration.test.ts`:** the block that asserted operations clears four routes now asserts it clears **`GET /api/facility/trends`** and is **403'd on `GET /api/facility/badges`, `GET/PATCH/DELETE /api/facility/visits` and `POST /api/facility/visits/insert`**. The name-matching guard covers both lists.

Red-then-green proved on both halves: reverting the tab filter fails the layout tab assertion; re-widening the two route files fails four of the five withheld-route assertions.

Node 24, root-level `npm ci`: `tsc --noEmit` 0 errors, `lint` clean, **278/278 unit suites (2724 tests)**, **136/136 integration suites (1464 tests)**, `check-route-coverage` 0 errors.

## Follow-ups, not in this PR

1. **`registry.ts` comment** — "Widening to isOperations is an open decision, #1476" is answered. Boundary file, own PR.
2. **`AppFrame.tsx` nav** — the sidebar `Facility Ops` item is still board/sysadmin-only, so operations reaches the two tools from `/index` (or a URL), not the main nav. Not a regression and not a bounce; giving it a `hrefFor` like `/membership-ops` has is a product call.
3. **`/facility-ops` index** — a server `redirect()` to `/facility-ops/visits`, so an operations user who *types* `/facility-ops` still lands somewhere that ejects them. Nothing links them there (no nav item, no directory row, no tab), so it is a URL-typed dead end only. `shop-ops`' `defaultShopTab` is the shape of the fix if (2) ever lands.
4. **`docs/designs/1256_ATTENDANCE_CORRECTION_SURFACE.md`** — its §1 matrix still marks ops 🟡 *"open decision, #1476"* on facility-wide insert/edit/delete, and §6.1 / §"Audience" say the same. #1633 answers it ⛔. Doc-only, and #1633 deliberately scoped itself to the register; worth a sweep once #1476 closes.

Closes nothing — #1476 is answered by #1633; this is the residual grant it left implementable.

### DIFF
--- a/checkin-app/docs/VOCABULARY.md
+++ b/checkin-app/docs/VOCABULARY.md
@@ -131,18 +131,20 @@ adult) → two attestations → **clearance** (dated; the validity window runs f
 Rules:
 1. **A reviewer attests; two attestations clear the check.** One word per sense —
    never "approve" for the act, never "attest" for the outcome.
-2. **Never "clean."** The app never holds the report or its result, only that a
-   reviewer read it and what they judged
+2. **Never "clean," never "passed."** The app never holds the report or its
+   result, only that a reviewer read it and what they judged
    ([../../docs/rules/membership.md](../../docs/rules/membership.md), Procedure ›
-   Application and review). "The check is clean" asserts the document's content;
-   "I approve this person" states the judgement, which is what is stored.
-   *(rename pending: the reviewer queue's confirm modal and row button)*
-3. **`clear` never means erase.** Undoing attestations is **discard**. `clear` is
-   reserved for the outcome above. Erasing a person's clearance date is **remove**.
-   *(rename pending: the "Clear approvals" button; the compliance page's erase path)*
+   Application and review). Both words state an outcome of the report; "I approve
+   this person" states the judgement, which is what is stored. Policy's own
+   outcome vocabulary is approval, restriction, or denial — a pass/fail framing
+   has nowhere to put the middle one.
+3. **`clear` never means erase.** Undoing attestations is **discard**. Erasing a
+   person's clearance date is **remove**. `clear` is reserved for the outcome
+   above — and as a **verb on a button** it reads as *reset* even so, which is why
+   the reviewer queue names the outcome with the noun (*complete the clearance*),
+   not the verb.
 4. **"Attest" is the reviewer's word alone.** An applicant **consents**; a Tool
    Certifier **certifies** (see Shop & Certification).
-   *(rename pending: `selfAttestBgConsent` → `selfRecordBgConsent`)*
 5. **`PENDING_BG_REVIEW` and `PENDING_BG_CLEARANCE` do not mark review vs
    clearance** — they differ by whether dues are paid. A known misnomer; the enum
    stays, because renaming a status reaches the schema, a migration, and the
--- a/checkin-app/docs/designs/1256_ATTENDANCE_CORRECTION_SURFACE.md
+++ b/checkin-app/docs/designs/1256_ATTENDANCE_CORRECTION_SURFACE.md
@@ -289,8 +289,8 @@ the change is (below). No reason field.
 
 ### Significance: what gets flagged
 
-A flag is raised **after** the edit lands, when the change is large *relative to
-how authoritative the value it overwrote was*. Two inputs:
+A change is scored **after** the edit lands, when the change is large *relative
+to how authoritative the value it overwrote was*. Two inputs:
 
 - **Source trust-weight** — how much we trust the old value. Highest for
   `SCANNER` (a physical measurement), then `LEAD_MARKED` (another person's
@@ -386,12 +386,12 @@ machine-close suppression is a source rule, not a threshold, and is not a knob.
   plus the significance object so AT12 can filter to the flagged ones without
   re-deriving (§4).
 
-### Which paths raise the live flag
+### Which paths email the board
 
 The real-time board email fires on the **self and household-lead** corrections
 only (`attendance/manual/[id]`). The staff and lead paths — `facility/visits`
 `PATCH`/`DELETE`, `manualEditAttendance`, `my-programs/conflicts/resolve` —
-deliberately do **not** raise it, for two reasons:
+deliberately do **not** email, for two reasons:
 
 - **The delete floor would make it noise.** `deleteSignificance` always flags, by
   design: a member erasing their own visit should always be seen. But marking a
@@ -402,12 +402,17 @@ deliberately do **not** raise it, for two reasons:
 - **On `facility/visits` the actor is the recipient.** That route is gated to
   sysadmin/board, so a flag would be the board notifying itself.
 
-This costs AT12 nothing. §"No new model" below is explicit that significance is
-**recomputed** from the audit row's old/new values rather than read off it — and
-every path above writes the source fields (`arrivedVia`/`departedVia`) into
-`oldData`, so the correction-review screen can score all of them. The
-`newData.significance` object that `attendance/manual/[id]` persists is a
-convenience for filtering, not the mechanism.
+This costs AT12 nothing, but not for the reason first given here. The original
+argument was that the screen **recomputes** significance from the audit row's
+old/new values, making the persisted object a filtering convenience. That is no
+longer how it works: every edit and delete path persists `newData.significance`,
+and the review screen **reads** it. A stored value is also the only form the
+database can filter on, which is what lets the default view page at all.
+
+Two row shapes cannot be recomputed — `facility/visits` edits written before the
+`oldData` fix stored no before-state — but recompute is not the fallback for
+them either. They are simply unscored, and the screen says so rather than
+showing a zero.
 
 If the board later wants leads' corrections in the live feed, the lever is the
 recipient set (§6.3) plus a per-path threshold — not the delete floor.
@@ -429,7 +434,7 @@ then — YAGNI until they ask.)
 
 **"Correct my hours" UI** resolves to: show the member their own visits (already
 possible via the manual-entry surface); every field is inline-editable; a
-significant save just goes through, with the flag raised quietly behind it.
+significant save just goes through, with the score written quietly behind it.
 
 ### Abuse: it's audited, and the big moves are seen
 The manual-insert route already accepts that a member can inflate their own
@@ -573,8 +578,8 @@ corrections never score themselves". Rows that never scored are *outside* the
 lens, not "reviewed and found insignificant". AT12's default lens is "flagged
 changes" — big or high-trust-overwriting edits, and deletes — with the full
 correction feed behind a filter. A member whose edits flag often is a standing
-signal. This *is* the "raise it to the board" surface; the write-time
-notification (§2) is the push, AT12 is the pull.
+signal. This *is* the board-review surface; the write-time notification (§2)
+is the push, AT12 is the pull.
 
 **Audience:** board + sysadmin, and ops once #1476 widens the section gate. The
 existing `/system-status/audit-log` viewer is sysadmin-gated and buries visit
--- a/checkin-app/src/__tests__/pageRegistry.test.ts
+++ b/checkin-app/src/__tests__/pageRegistry.test.ts
@@ -41,3 +41,33 @@ describe('pageRegistry drift guard', () => {
     expect(hrefs.length).toBe(new Set(hrefs).size);
   });
 });
+
+// Operations reaches the four Facility Ops tools (and the index, which redirects
+// into Visits), so they must appear in the directory rather than being URL-only.
+// Corrections is the exception: its page gate stays board/sysadmin, so listing it
+// for operations would advertise a tab that ejects them — see #1476.
+describe('Facility Ops visibility for operations', () => {
+  const ops = { isOperations: true };
+  const entryFor = (href: string) => PAGES.find((p) => p.href === href)!;
+
+  it.each([
+    '/facility-ops',
+    '/facility-ops/badges',
+    '/facility-ops/print-badges',
+    '/facility-ops/trends',
+    '/facility-ops/visits',
+  ])('shows %s to an operations user', (href) => {
+    expect(entryFor(href).visible(ops, true, null)).toBe(true);
+  });
+
+  it('hides /facility-ops/corrections from an operations user (page gate is still board-only)', () => {
+    expect(entryFor('/facility-ops/corrections').visible(ops, true, null)).toBe(false);
+  });
+
+  it('still shows every Facility Ops entry to a board member', () => {
+    const board = { isBoardMember: true };
+    for (const p of PAGES.filter((p) => p.section === 'Facility Ops')) {
+      expect(p.visible(board, true, null)).toBe(true);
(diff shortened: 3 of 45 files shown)


thpr (commented):
Gate-only change, correct role key, tests mirror the existing patterns. Two substantive findings and four smaller ones, all from reading the surrounding code rather than the diff.

## 1. The "Add Visit" button 403s for operations

`facility-ops/visits/page.tsx` renders `Add Visit` unconditionally (line 289) and POSTs `/api/facility/visits/insert` (line 172), which this PR deliberately leaves at `['isSysadmin', 'isBoardMember']`. That is the same failure the PR body gives as the reason for granting PATCH and DELETE — "read-only would render buttons that 403" — and it ships. An operations user opens the modal, picks a person, submits, and gets `Failed to add visit.` with no explanation of why.

Granting insert is a boundary change: `POST /api/facility/visits/insert` is a registered route (`authorize` + `orderedView` in `src/security/registry.ts`), so it has to ship in its own PR under the boundary-isolation rule. The fix that fits inside this PR is to render the button only for sysadmin/board.

## 2. Contradicts a domain rule and an open board decision, without saying so

- `docs/rules/attendance-checkin.md` (line 154): "The board and sysadmins edit or delete any visit, and record one for someone else at any past time" — `[Decision]` tier. AGENTS.md requires raising a change that contradicts the register rather than proceeding through it. The register is not updated and the PR does not mention it.
- `src/security/registry.ts` (lines 77-82) records the invariant that the insert gate, the sibling `GET/PATCH/DELETE /api/facility/visits` gates, and the `facility-ops/visits` page gate "must stay equal (their drifting apart was AT13/#1259)", and that "Widening to isOperations is an open decision, #1476."

#1476 is still open. It says explicitly that "widening a permission gate is a board call, not a default worth taking silently inside a feature PR", and it specifies operations moving to granted on **insert-for-others, edit, and delete** together. This PR is that issue's implementation, filed before the decision is recorded and covering two of the three verbs — which produces exactly the insert-vs-edit drift the registry comment warns about, and leaves that comment stating something no longer true. (A comment-only registry edit is still a boundary-file change, so it needs its own PR.)

At minimum: reference #1476, get the board call recorded there, and update `attendance-checkin.md` in the same landing.

## 3. Operations gets access but no navigation

`src/components/pageRegistry.ts` lines 97-102 mark every facility-ops entry `visible: BOARD`. `BOARD_OR_OPS` already exists in that file (line 35). As it stands an operations user can only reach the four tools by typing the URL.

## 4. The Corrections tab is shown to operations, then ejects them

The layout renders all five `FACILITY_NAV_LINKS` tabs, but `facility-ops/corrections/page.tsx` (line 125) still gates on `["isSysadmin", "isBoardMember"]`. An operations user clicking Corrections is redirected to `/` — out of facility-ops entirely. #1476 flags that screen as the oversight half of the same surface, so it is worth deciding alongside rather than leaving as a dead tab.

## 5. The badges page now disagrees with its own API in the other direction

`facility-ops/badges/page.tsx` becomes `useRequireRole(['isSysadmin', 'isOperations'])` — board is absent, while both the layout gate and the newly widened `GET /api/facility/badges` admit board members. `useRequireRole` has no implicit board/sysadmin fallthrough, so a board member is bounced from a page the API serves them. Pre-existing, but it is the line this PR edits, and it is the same UI/API disagreement the PR sets out to fix.

## 6. DELETE has no gate coverage

`roleGated` in `authzRoleRejection.integration.test.ts` has no `DELETE /api/facility/visits` case, so the one destructive verb this PR grants is the one with no assertion behind it. The four names the new block filters on do all match existing cases, so the block covers what it claims — DELETE was simply never in the harness.

## Not a concern

- No boundary-isolation trip: visits `GET/PATCH/DELETE`, badges, and trends are unregistered legacy routes; only `visits/insert` is registered, and it is untouched.
- Print Badges genuinely needs no API change — `/api/people/search` already allows operations, and the page has no gate of its own.
- No handler branches on role, so nothing behaves differently for operations beyond the gate.

---

thpr (changes requested):
Escalating my review above to changes-requested, on findings 1 and 2:

1. The `Add Visit` button ships 403-ing for operations — the exact failure the PR body gives as its reason for granting PATCH/DELETE.
2. The widening contradicts `docs/rules/attendance-checkin.md` line 154 (`[Decision]` tier) and lands #1476 before that decision is recorded, covering two of the three verbs the issue specifies and creating the insert-vs-edit drift `registry.ts` warns against.

The rest are non-blocking.

---

jee7s:
Reworked against #1633, which merged and recorded the answer to #1476: operations reach attendance **in aggregate only**. Every finding below is resolved, but two of them are resolved by the grant shrinking rather than by a fix.

## What was withdrawn

- **`/facility-ops/visits`** — page gate and `GET`/`PATCH`/`DELETE /api/facility/visits`. Both files are byte-identical to the merge base again.
- **`/facility-ops/badges`** — page gate and `GET /api/facility/badges`. This is the raw badge-event log, not badge printing; the rule's wording distinguishes them and reading it the other way would grant exactly what it withholds.
- `pageRegistry` rows for Visits, Badges and the `/facility-ops` index, back to `BOARD`.

## What remains

Trends page + `GET /api/facility/trends`, print-badges, `isOperations` on the layout gate, and `BOARD_OR_OPS` on those two registry rows only.

## Finding by finding

**1 — Add Visit 403s for operations.** Dissolved. With the page gate back at sysadmin/board it is identical to the insert gate, so the `canAddVisit` guard added earlier was tautological dead code. Verified before deleting; it does not stand on its own.

**2 — contradicts a `[Decision]`-tier rule and lands #1476 early.** Resolved by #1633. This PR now implements that decision rather than pre-empting it. `docs/rules/attendance-checkin.md` needed no amendment — the existing board/sysadmin rule stays true, and the new line sits beside it.

**3 — access but no navigation.** Fixed, scoped to the two permitted tools.

**4 — the Corrections tab shows for operations, then ejects them.** Fixed at the source rather than by a second list. `FACILITY_NAV_LINKS` gained a `roles` field named for the rule (`FACILITY_RECORD_ROLES` / `FACILITY_AGGREGATE_ROLES`); each of the four gated pages passes that same array to `useRequireRole`; the layout filters the strip through it; and the section gate is **derived as the union**, so the section can never admit someone to an empty strip. The `pageRegistry` test reads its expectations from the same constant. Net one fewer copy of the permission table — the section-scoped version of what #1569 wants globally.

**5 — badges page drops board while its API serves them.** Fixed. Unrelated to the grant; called out in the PR body as a pre-existing bug.

**6 — `DELETE` has no gate coverage.** Fixed and **kept**, even though the grant it was added alongside is gone. That verb never had 401/403 coverage in the harness at all, independent of any widening — as the review itself put it, "DELETE was simply never in the harness." It now also carries an operations-is-denied assertion.

## Verified rather than assumed

`registry.ts` lines 77-82 assert that the insert gate, the sibling `GET/PATCH/DELETE` gates and the visits page gate must stay equal. With the grant withdrawn all four are `['isSysadmin','isBoardMember']`, so the invariant holds and the boundary file needed no change — `git diff origin/main...HEAD -- checkin-app/src/security/` is empty and the `isolation` job confirms it.

Red-then-green on both halves: reverting the tab filter fails the layout tab assertion; re-widening the two route files fails 4 of the 5 withheld-route assertions (insert was already correct).

## Follow-ups, none blocking this PR

Three of these are consequences of this PR landing rather than pre-existing gaps.

1. **`registry.ts:79` is now false** — "Widening to isOperations is an open decision, #1476" was answered by #1633. It is the comment a reader consults before touching those gates, and it is a boundary file, so it needs its own PR.
2. **`AppFrame.tsx:105` keeps the sidebar's Facility Ops entry board-only**, so operations reach the two tools through the `/index` directory and not the main nav. A `hrefFor` like `/membership-ops` has is the shape, but which landing page is a product call.
3. **`/facility-ops` is a server `redirect()` to visits**, so an operations user who types that URL lands somewhere that ejects them. Nothing links them there, so it is a typed-URL dead end only; `shop-ops`' `defaultShopTab` is the existing pattern if (2) lands.
4. **`docs/designs/1256_ATTENDANCE_CORRECTION_SURFACE.md`** still marks operations 🟡 "open decision, #1476" in its §1 matrix and two other places, where #1633 now makes it ⛔.

Worth saying plainly: with (2) and (3) outstanding, an operations user who looks for these tools in the obvious place does not find them. The grant is correct; the path to it is thin.

---

thpr (approved):
Approving. Re-read the rework against the code rather than the summary.

#1633 merged and recorded the rule (`docs/rules/attendance-checkin.md` line 166, `[Decision]`), and this PR now implements it rather than pre-empting it — that resolves my blocking finding 2. The grant shrank to match: the changed-file list carries no `api/facility/visits/route.ts`, no `api/facility/badges/route.ts` and nothing under `src/security/`, so the registry invariant holds untouched and the Add Visit 403 (finding 1) is dissolved rather than patched.

Checked independently of the summary:

- `GET /api/facility/trends` emits `buckets`/`totals` only — person ids are used inside the handler and never leave it. The aggregate grant is genuinely aggregate.
- print-badges calls only `/api/people/search`, which already admits operations.
- The sorted-name guard closes the silent-shrink hole in the filtered sweeps, which was the real risk in name-matched filters.

Deriving the section gate as the union of the tab roles is the right shape — it makes "admitted to a section with no openable tab" unrepresentable rather than merely tested.

Residual items are non-blocking and filed as #1639: the sidebar still never offers the section to operations, `/facility-ops` redirects them into a page that ejects them, print-badges declares roles it does not enforce, and the `registry.ts` line 79 / AT3 design-doc text still call #1476 open. #1476 itself is worth closing.

---

Issue #1627, opened 2026-08-10
Remove dotted lines from /facility-ops/print-badges prints


thpr:
Badge should not print the dotted lines, these are being printed on pre-cut badges so if the dotted lines don't align perfectly (they never will) they are ugly

Issue #1625, opened 2026-08-10
Show printed name on the /facility-ops/print-badges screen


thpr:
add a column - note the printed name should only be dependent on ACTIVE members, not which are being printed RIGHT NOW, so the current code has a bug, so this is (1) fix the bug (2) expose the final name as a column as proof it's fixed

PR #1634, opened 2026-08-10
feat(print-badges): hide inactive people behind a default-on filter

### LINKED ISSUE #1626: /facility-ops/print-badges needs a filter
Add a filter to hide inactive people (check box, defaults on)

jee7s:
Fixes #1626

Adds a **"Hide inactive"** checkbox (Mantine `<Checkbox>`, `useState(true)` — the issue says defaults on) to the existing `<Group>` beside the search box on `/facility-ops/print-badges`, following the pattern at `program-ops/events/page.tsx:61-65`.

The predicate is `p.isMember`, already on every row from `/api/people/search` (`route.ts:108`, `personRecordIsActiveOrgMember`). No new helper, no new server call.

## The bug the obvious version ships

`participants` is read in **five** places on this page. A filter that reaches only the table's `rows=` corrupts the other four:

| line | read | left unfiltered |
|---|---|---|
| 71 | `selectedIds.size === participants.length` (toggleAll) | select-all takes the wrong branch |
| 74 | `new Set(participants.map(...))` | select-all silently selects hidden people |
| 83 | `participants.filter(p => selectedIds.has(p.id))` | **the PDF still contains the hidden people** |
| 133 | header checkbox `checked=` | never reaches checked |
| 193 | `rows={participants}` | the one obvious site |

Line 83 is the real defect: select everyone, *then* tick "Hide inactive", and the badges still print for people the operator can no longer see. The button counts would lie about it too.

## The fix

One derived list, every read repointed:

```ts
const visible = hideInactive ? participants.filter(p => p.isMember) : participants;
const selectedVisible = visible.filter(p => selectedIds.has(p.id));
```

- `rows=`, select-all's new set → `visible`
- the generate guard, the badge payload, both button labels and both `disabled` props → `selectedVisible`
- `toggleAll` and the header checkbox → `selectedVisible.length === visible.length`, **not** `selectedIds.size === visible.length` — `selectedIds` can still hold hidden people, so comparing its size leaves the header checkbox unreachable and `toggleAll` on the wrong branch.

`selectedIds` is deliberately **not** pruned in a `useEffect` when the filter flips: more machinery, a state-sync bug class, and it destroys the selection when the operator toggles the filter back off. Hidden ids are inert while hidden.

`DataTable` is untouched — it has 7 consumers and already renders `emptyMessage` for an empty `rows`. Filtering is the caller's job.

## Tests — red then green

`checkin-app/src/app/facility-ops/print-badges/__tests__/page.test.tsx`, two new cases:

1. the checkbox renders **checked by default**, a row with `isMember: false` is absent, unchecking reveals it;
2. **the one that matters** — with the filter off, click select-all, tick "Hide inactive", assert the button reads `Generate Badge (1)` not `(2)`, plus the sticker button and the select-all checkbox state.

Proven red against a naive `rows`-only implementation (same checkbox, same `rows` filter, the other four reads left alone):

```
✕ drops hidden people from the selection count, so the PDF matches the button
  Unable to find an accessible element with the role "button" and name "Generate Badge (1)"
  ...
  Name "Generate Badge (2)":
```

— one visible row, button says 2. Green with this PR: 5/5 in that suite.

The two pre-existing fixtures are both `isMember: true`, so the default-on filter does not disturb the three existing tests.

Full local run on this branch: **tsc 0 errors, lint clean, 278 suites / 2712 unit tests passing** (baseline `origin/main` is 278 / 2710 — the two new cases are the delta). Integration suite unaffected and unchanged: no server, route, or schema code is touched.

## Scope

- `checkin-app/src/app/facility-ops/print-badges/page.tsx` and its test — nothing else.
- `checkin-app/src/security/**` diff is empty.
- Does **not** touch `BadgeDocument.tsx` or the `badgesWithQr` payload object literal (`page.tsx:96-98`), which #1627/#1629 owns. The nearest hunk here ends 5 lines above it.
- Does **not** touch name derivation — that is #1625.
- Known limitation, unchanged by this PR: `/api/people/search` still caps at `take: 200` ordered by descending id, so this filters the 200 rows that came back rather than fetching 200 members.

### DIFF
--- a/checkin-app/src/app/facility-ops/print-badges/__tests__/page.test.tsx
+++ b/checkin-app/src/app/facility-ops/print-badges/__tests__/page.test.tsx
@@ -35,6 +35,11 @@ const participants = [
   { id: 2, name: "Bo Board", email: "bo@example.com", isMember: true, isBoardMember: true },
 ];
 
+const withInactive = [
+  participants[0],
+  { id: 3, name: "Lapsed Larry", email: "larry@example.com", isMember: false },
+];
+
 describe("facility-ops/print-badges page", () => {
   it("loads and renders the participant roster", async () => {
     setSession({ id: 1, isSysadmin: true });
@@ -72,5 +77,37 @@ describe("facility-ops/print-badges page", () => {
     fireEvent.click(screen.getByRole("button", { name: "Generate Badge (2)" }));
 
     await waitFor(() => expect(global.URL.createObjectURL).toHaveBeenCalled());
+  });
+
+  it("hides inactive people by default and reveals them when unchecked", async () => {
+    setSession({ id: 1, isSysadmin: true });
+    mockFetchJson({ "/api/people/search": { people: withInactive } });
+    renderWithProviders(<PrintBadgesPage />);
+    await screen.findByText("Kim Keyholder");
+
+    const filter = screen.getByRole("checkbox", { name: /hide inactive/i });
+    expect(filter).toBeChecked();
+    expect(screen.queryByText("Lapsed Larry")).not.toBeInTheDocument();
+
+    fireEvent.click(filter);
+    expect(await screen.findByText("Lapsed Larry")).toBeInTheDocument();
+  });
+
+  it("drops hidden people from the selection count, so the PDF matches the button", async () => {
+    setSession({ id: 1, isSysadmin: true });
+    mockFetchJson({ "/api/people/search": { people: withInactive } });
+    renderWithProviders(<PrintBadgesPage />);
+    await screen.findByText("Kim Keyholder");
+
+    const filter = screen.getByRole("checkbox", { name: /hide inactive/i });
+    fireEvent.click(filter); // reveal the inactive person
+    fireEvent.click(screen.getByRole("checkbox", { name: "Select all" }));
+    expect(screen.getByRole("button", { name: "Generate Badge (2)" })).toBeInTheDocument();
+
+    fireEvent.click(filter); // hide again — the inactive person must leave the print run
+    expect(screen.getByRole("button", { name: "Generate Badge (1)" })).toBeInTheDocument();
+    expect(screen.getByRole("button", { name: "Generate Sticker (1)" })).toBeInTheDocument();
+    // and "select all" still reads as fully selected over what is actually visible
+    expect(screen.getByRole("checkbox", { name: "Select all" })).toBeChecked();
   });
 });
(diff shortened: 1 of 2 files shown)


thpr (commented):
The `visible` / `selectedVisible` split is right, and the five-read table in the description is accurate — I checked, there is no sixth read of `participants` left in the file, and `selectedVisible` keeps `visible`'s order so badge ordering is unchanged. Worth noting the old code was a bit worse than described: with a search that narrowed to a non-selected person, `selectedIds.size` kept both buttons enabled while `selectedParticipants` was `[]`, so clicking generated an *empty* PDF. This fixes that too.

One thing the diff introduces that I think needs an answer before merge.

### A default-on filter over a server-capped window makes the empty state lie

`emptyMessage="No participants found."` is unchanged, but `rows` is now `visible`. So a search whose only matches are inactive people renders that message — the row was fetched, then dropped client-side, and nothing on screen says so. On a badge-printing station the operator's read of "No participants found." is *this person is not in the system*, and the recovery (uncheck a box they have no reason to suspect) is invisible. That is a support call, and it is the failure mode a default-on filter always has: the state it hides is the state nobody thinks to look for.

The no-search default view has the same shape and is slightly worse, because the filter runs *after* the cap rather than before it. `/api/people/search` is `take: 200, orderBy: { id: 'desc' }` (`route.ts:52-53`), so the operator sees "the active members among the 200 newest people", not "200 active members", with no paging and no count. The description names the cap as a known limitation, which is fair — but the cap was benign while the page showed all 200 rows, and stops being benign once a default-on filter silently subtracts from it. An org whose recent intake skews non-member could land on a near-empty default view and no indication why.

Both go away with the hidden count on screen, which the page already has everything for:

```ts
const hidden = participants.length - visible.length;
```

Then either `label={hidden ? `Hide inactive (${hidden})` : "Hide inactive"}` on the checkbox, or swap the empty message when `hidden > 0` ("No active people match — N hidden by the filter"). Either is a couple of lines and keeps the filter client-side; pushing the predicate into the route so the cap applies to members would be the fuller fix, but I would not ask for it here.

### Minor — one stated invariant is not quite true

> `selectedIds` is deliberately not pruned … Hidden ids are inert while hidden.

Inert until select-all: `toggleAll` does `setSelectedIds(new Set(visible.map(p => p.id)))`, which replaces the whole set and so drops any hidden id that was selected before the filter went on. Same for the clear branch. The resulting behaviour is fine — arguably what you want — but it is a replace, not a merge, so the selection is not actually preserved across a filter toggle once select-all is touched. Only flagging it because the description leans on the opposite claim to justify skipping the `useEffect`, and skipping the `useEffect` is still the right call.

---

thpr (approved):
Approving. The filter itself is correct and the five reads are genuinely all repointed.

Finding #1 (the empty state saying "No participants found." when the filter, not the search, is what emptied the table) is filed separately as 1636 so it doesn't hold this up.

---

PR #1637, opened 2026-08-11
fix(print-badges): derive the printed name from the ACTIVE roster, not the print batch

### LINKED ISSUE #1625: Show printed name on the /facility-ops/print-badges screen

jee7s:
Fixes #1625

## The bug

`BadgeDocument` called `computeDisplayNames(badges)` where `badges` was the current
checkbox selection. John Smith printed **John** alone and **John S.** when John Doe was
also ticked — same person, two badges, decided by a checkbox.

The population it drew from was worse than the selection. `/api/people/search` has no
membership filter, caps at `take: 200` ordered by id desc, and re-queries on every
keystroke — so typing `smi` moved everyone's printed name.

## The fix

**`?roster=active` mode on `/api/people/search`**, returning every ACTIVE, live member as
`{id, name}` — no `q`, no cap.

A mode rather than a new route, deliberately:

- `GET /api/people/search` is already on the frozen legacy-authz baseline
  (`scripts/legacy-authz-routes.txt:118`), so extending it adds **no new method key** — no
  registry entry, no CODEOWNERS round-trip. A new route file is in neither the registry nor
  the baseline and trips `new-route-old-authz`, which blocks **even in advisory mode**
  (`check-route-coverage.ts:357-369`); CI runs it `--strict`.
- Its role set (`isSysadmin`/`isBoardMember`/`isOperations`) already covers this screen and
  stays correct if Operations is later granted `/facility-ops`.
- The response is a strict subset of what the route already returns, so it cannot widen
  exposure.

Interacts with open PR #1623, which grants **Operations** access to Print ID Badges and
states "no API change needed for Print Badges: `/api/people/search` already allowed
operations." That stays true here — the roster mode inherits the same role set. It would
have been falsified by the other candidate below.

`/api/shop/org-members` runs the identical query and was the other candidate, but it is
`authorize: 'certifier'` — it would 403 for Operations — and it is policed by a strip
assertion under `src/security/`, so re-aiming it at badge printing would mean editing the
security boundary. `git diff --name-only origin/main...HEAD -- checkin-app/src/security/`
is empty.

**`computeDisplayNames` is lifted to the page**, unchanged. The page computes the map once,
renders it in a new **Printed Name** column, and passes each value down as `displayName`.
One map — so the column is proof of what will print, not a second opinion. `BadgeDocument`
stops computing names entirely.

`name` stays on the badge payload solely for `StickerDocument`, which prints it raw and is
unchanged. Target shape is `{ id, name, displayName, qrDataUri }`; `year` is #1628's.

**Generate is held until the roster resolves.** Printing before it lands would emit
un-disambiguated names onto physical badges with nothing on screen to say so, and a bad
badge cannot be rolled back out of a laminator.

**An inactive person shows a bare first name.** They are not in the ACTIVE population, so
they are not grouped into it. The tempting alternative — disambiguate them *against* the
roster without letting them perturb it — is false as specified: grouping over the union
means the `group.length === 1` short-circuit fires off the merged group, so one inactive
"John Smithson" flips ACTIVE "John" to "John S.", reintroducing the exact bug. #1626
already hides inactives by default, so the cell is rarely seen.

## Red, then green

The obvious unit test in `badgeDisplayNames.test.ts` *is* red today, but it can never go
green: the fix is at the call site, not in the function. `computeDisplayNames` will always
disambiguate within whatever population it is handed — that is its contract. Shipping a
test that can only fail would be worse than none, so it is not here.

The honest regression test is at the page level, and it is **only writable because the
column exists** — that is the argument for the column.

Proved by rebuilding the pre-fix behaviour (printed name computed over the current print
batch, as `BadgeDocument` did) rather than by reverting:

```
● keeps the printed name fixed to the ACTIVE roster, not the current selection
    Expected: "John S."
    Received: "John"
● keeps the printed name fixed when the search box narrows the visible rows
    Expected: "John S."
    Received: "John"
Tests: 2 failed, 5 passed, 7 total
```

With the fix: `7 passed`.

Server-side halves of the same claim, in a new integration suite (verified red by disabling
the roster branch — 2 of 3 fail, the plain-search case still passes):

- ACTIVE members only; non-members and merge tombstones excluded.
- `roster=active&q=Smith` returns the same set as `roster=active` — a search cannot shrink
  the population a printed name derives from.

## What this does NOT fix

**The 200-row cap is still there.** It no longer affects the printed *name* — that now comes
from the unbounded roster. It still governs **which people the operator can select**: `rows`
still comes from `/api/people/search` with `take: 200, orderBy: { id: 'desc' }`. A member
outside the 200 highest ids and outside the current search string cannot be selected or
printed. Read this as "the printed name is no longer search-dependent", not "the operator
can now reach everyone."

**Residual drift, by design.** Making the name depend on the whole ACTIVE roster makes it
stable per print session, not stable forever. The day a "John Smithson" joins, John Smith's
correct name becomes `John Sm.` while his laminated badge still says `John S.`, and nobody
reprints the org when a family joins. The new column will show that disagreement. Stable
alternatives would be persisting the printed name on the Person row or dropping
population-dependence entirely — both more than this issue asked for.

## For #1628

- The roster branch is the place to add `renewed` + `membershipYear`. Legal here precisely
  because a legacy route has no stripper; under `handler()` the stripper drops non-model bag
  keys and non-`Person` fields, so both would silently vanish.
- `membershipYearLabel` still lives in `badgeNames.ts` with its single caller in
  `BadgeDocument`, ready to be retired.
- `ParticipantBadge` is `{ id, displayName, qrDataUri }`; `year: string | null` slots in.
- Merge blocker for that issue: read `BoardSettings.id=1 orgMembershipYearBoundary`. It is
  nullable with no seed, and every badge silently loses its year if it is null.

## Verification

tsc 0 · lint clean · 278 unit suites (2714 tests) · route-coverage `--strict` 0 errors ·
`src/security/` diff empty · related integration suites green (7 suites, 133 tests) plus
the new roster suite.

Manual acceptance still worth doing: generate a PDF and confirm the Printed Name column
matches the name on the badge face for the same person.

### DIFF
--- a/checkin-app/src/app/facility-ops/print-badges/page.tsx
+++ b/checkin-app/src/app/facility-ops/print-badges/page.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useState, useEffect, useCallback } from "react";
+import { useState, useEffect, useCallback, useMemo } from "react";
 import { useSession } from "next-auth/react";
 import { useRouter } from "next/navigation";
 import QRCode from "qrcode";
@@ -10,6 +10,7 @@ import { notifications } from "@mantine/notifications";
 import { DataTable, type DataTableColumn } from "@/components/admin/DataTable";
 import BadgeDocument from "@/components/admin/BadgeDocument";
 import StickerDocument from "@/components/admin/StickerDocument";
+import { computeDisplayNames } from "@/components/admin/badgeNames";
 
 type ParticipantRow = {
   id: number;
@@ -25,6 +26,9 @@ export default function PrintBadgesPage() {
   const router = useRouter();
 
   const [participants, setParticipants] = useState<ParticipantRow[]>([]);
+  // Every ACTIVE member org-wide — the population printed names disambiguate against.
+  // Separate from `participants`, which is whatever the search box last matched.
+  const [roster, setRoster] = useState<{ id: number; name: string }[] | null>(null);
   const [searchTerm, setSearchTerm] = useState("");
   const [selectedIds, setSelectedIds] = useState<Set<number>>(new Set());
   const [hideInactive, setHideInactive] = useState(true);
@@ -61,6 +65,30 @@ export default function PrintBadgesPage() {
     }
   }, [status, fetchParticipants]);
 
+  // Once on mount — deliberately NOT keyed on searchTerm. The printed name must not
+  // move when the operator types.
+  useEffect(() => {
+    if (status !== "authenticated") return;
+    const url = new URL('/api/people/search', window.location.origin);
+    url.searchParams.set('roster', 'active');
+    fetch(url.toString())
+      .then(res => res.json())
+      .then(data => setRoster(data.people ?? []))
+      .catch(e => {
+        console.error("Failed to load the active-member roster for badge names:", e);
+        setRoster([]);
+      });
+  }, [status]);
+
+  const printedNames = useMemo(() => computeDisplayNames(roster ?? []), [roster]);
+
+  // The badge name and this column read the same map, so the column is proof of what
+  // will print. Someone off the ACTIVE roster is not in that population at all, so
+  // they get a bare first name rather than being folded into it — folding them in is
+  // what lets a non-member move a member's name, which is the bug.
+  const printedName = (p: ParticipantRow) =>
+    printedNames.get(p.id) ?? ((p.name ?? '').trim().split(/\s+/)[0] || `User #${p.id}`);
+
   const toggleSelection = (id: number) => {
     const newSet = new Set(selectedIds);
     if (newSet.has(id)) newSet.delete(id);
@@ -94,10 +122,12 @@ export default function PrintBadgesPage() {
             margin: 1,
             color: { dark: '#000000', light: '#FFFFFF' }
           });
-          // `name` feeds both documents: BadgeDocument disambiguates it, StickerDocument prints it raw.
+          // `displayName` is the badge's; `name` is retained solely for StickerDocument,
+          // which prints the full name raw and is not changing.
           return {
             id: p.id,
             name: p.name ?? '',
+            displayName: printedName(p),
             qrDataUri,
           };
         })
@@ -152,6 +182,10 @@ export default function PrintBadgesPage() {
       render: (p) => <Text fw={600}>{p.name || 'N/A'}</Text>,
     },
     {
+      header: 'Printed Name',
+      render: (p) => <Text>{printedName(p)}</Text>,
+    },
+    {
       header: 'Membership',
       render: (p) => (p.isMember ? <Text c="green">Active</Text> : <Text c="red">Inactive</Text>),
     },
@@ -187,10 +221,12 @@ export default function PrintBadgesPage() {
           checked={hideInactive}
           onChange={(e) => setHideInactive(e.currentTarget.checked)}
         />
-        <Button onClick={() => generate('badge')} disabled={selectedVisible.length === 0 || isGenerating} loading={isGenerating}>
+        {/* Held until the roster lands: printing first would silently emit un-disambiguated
+            names onto physical badges, with nothing on screen to say so. */}
+        <Button onClick={() => generate('badge')} disabled={selectedVisible.length === 0 || isGenerating || roster === null} loading={isGenerating}>
           Generate Badge ({selectedVisible.length})
         </Button>
-        <Button color="grape" onClick={() => generate('sticker')} disabled={selectedVisible.length === 0 || isGenerating} loading={isGenerating}>
+        <Button color="grape" onClick={() => generate('sticker')} disabled={selectedVisible.length === 0 || isGenerating || roster === null} loading={isGenerating}>
           Generate Sticker ({selectedVisible.length})
         </Button>
       </Group>
--- a/checkin-app/src/app/facility-ops/print-badges/page.tsx
+++ b/checkin-app/src/app/facility-ops/print-badges/page.tsx
@@ -72,11 +72,17 @@ export default function PrintBadgesPage() {
     const url = new URL('/api/people/search', window.location.origin);
     url.searchParams.set('roster', 'active');
     fetch(url.toString())
-      .then(res => res.json())
+      // apiError returns valid JSON, so a 500 would resolve to `[]` and release the
+      // Generate guard — the roster must stay null on any failure.
+      .then(res => {
+        if (!res.ok) throw new Error(`roster request failed: ${res.status}`);
+        return res.json();
+      })
       .then(data => setRoster(data.people ?? []))
       .catch(e => {
         console.error("Failed to load the active-member roster for badge names:", e);
-        setRoster([]);
+        setRoster(null);
+        notifications.show({ color: 'red', message: 'Could not load the active member roster, so badge names cannot be resolved. Reload to retry.', autoClose: false });
       });
   }, [status]);
 
--- a/checkin-app/src/app/__tests__/peopleSearchRoster.integration.test.ts
+++ b/checkin-app/src/app/__tests__/peopleSearchRoster.integration.test.ts
@@ -1,0 +1,95 @@
+/**
+ * @jest-environment node
+ */
+/**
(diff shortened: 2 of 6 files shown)


jee7s:
Closing as a duplicate of #1638, which landed the same fix 11 minutes later and is the better-factored version — `computeDisplayNames(subjects, cohort = subjects)` resolving against the cohort minus self is cleaner than what this PR does, and it needs no server change.

One finding from this branch is worth carrying over and is left as a note on #1638: `/api/shop/org-members` is `authorize: 'certifier'` with no `isOperations` in its `orderedView`, and #1623 makes print-badges operations-reachable.

Branch `claude/1625-printed-name` left in place in case any of it is wanted.

jee7s:
Reopened after #1638 closed. Since that PR reached the same fix by a different route, here is the one place they diverged and why this branch went the way it did — it is the obvious question to ask of this diff, and the answer is about a state the repo is not in yet.

## Why not `/api/shop/org-members`

`api/shop/org-members/route.ts` already runs the identical query — `ACTIVE_ORG_MEMBER_PERSON_WHERE + LIVE_PERSON`, `select {id, name}`, unbounded — and it is already registered. Reusing it means no new mode on a legacy route and a smaller diff. That was the recommendation this branch started from, and #1638 took it.

It fails on authorization, and only in the future.

`registry.ts` gates that endpoint `authorize: 'certifier'`, with:

```
['isSysadmin',    [...]]
['isBoardMember', [...]]
['certifier',     ['member', 'public']]
```

No `isOperations` — and `certifier` is a `MAY_CERTIFY_OTHERS` toolStatus, not a role an operations user holds.

Today that is harmless: `/facility-ops` admits only sysadmin and board, so every caller of this screen is already in that view. **#1623 changes it.** That PR makes `/facility-ops/print-badges` reachable by `isOperations`, implementing the decision recorded in #1633, and its body states the premise it relies on:

> No API change needed for Print Badges: `/api/people/search` already allowed operations, and `/api/programs` (the trends dropdown source) is a public catalog.

A cohort fetch against `org-members` falsifies that. An operations user would load the page, the roster fetch would fail authorization, and — because this design deliberately holds badge generation until the cohort arrives — the Generate button would never enable. Not a wrong name on a laminated badge, but a tool that silently does nothing for one of the two roles #1633 grants it to.

`/api/people/search` is on the frozen legacy baseline (`scripts/legacy-authz-routes.txt:118`) and already admits operations, so a `?roster=active` mode inherits the role set this screen actually needs, now and after #1623.

**The alternative is still open and is arguably cleaner:** add `['isOperations', ['member', 'public']]` to the `org-members` `orderedView`, mirroring the certifier row, and take the cohort from there. It grants nothing this page does not already display. It is a boundary-file change, so it ships alone under the isolation rule — which is why it is not in this PR rather than why it is wrong.

## Worth taking from the closed #1638

Its `computeDisplayNames(subjects, cohort = subjects)` — resolving each subject against the cohort **minus itself, matched by id** — is a better factoring than this branch's, and it survives a defect our own plan had: an asymmetric variant that grouped over `population ∪ extras` still fired the `group.length === 1` short-circuit off the merged group, flipping an ACTIVE "John" to "John S." That is the bug, reintroduced. The minus-self shape does not have it, and the default argument keeps every existing call site unchanged. Happy to adopt it here if that reads better.

---

checkin-app/src/app/facility-ops/print-badges/page.tsx
```diff
@@ -61,6 +65,30 @@ export default function PrintBadgesPage() {
...
+    if (status !== "authenticated") return;
+    const url = new URL('/api/people/search', window.location.origin);
+    url.searchParams.set('roster', 'active');
+    fetch(url.toString())
+      .then(res => res.json())
+      .then(data => setRoster(data.people ?? []))
+      .catch(e => {
+        console.error("Failed to load the active-member roster for badge names:", e);
+        setRoster([]);
+      });
```
thpr: `setRoster([])` on failure re-enables Generate (the guard is `roster === null`), so a failed roster load prints bare first names onto badges. And `res.ok` is unchecked, so a 500 — which returns valid JSON from `apiError` — resolves through `data.people ?? []` to `[]` without reaching this `.catch` at all. See the review body for the fix.

thpr (changes requested):
One blocker. The rest of the diff is fine.

## The roster-failure path defeats the guard this PR is built on

Generate is gated on `roster === null`, but the failure path sets `[]`. So when the roster fetch fails the button re-enables and every badge prints an un-disambiguated first name — onto laminate, with a console line as the only trace. That is the outcome the PR body says the hold exists to prevent:

> Printing before it lands would emit un-disambiguated names onto physical badges with nothing on screen to say so, and a bad badge cannot be rolled back out of a laminator.

Two ways in, and only one of them reaches `.catch`:

- **No `res.ok` check.** A 500 from this route returns `apiError("Failed to fetch people", 500)` — valid JSON. `res.json()` resolves, `data.people ?? []` gives `[]`, and the `.catch` never runs.
- A genuine throw (network, non-JSON body) → `.catch` → `[]`.

Fix is to keep the sentinel distinct from a legitimately empty roster:

```ts
fetch(url.toString())
  .then(res => {
    if (!res.ok) throw new Error(`people/search?roster=active responded ${res.status}`);
    return res.json();
  })
  .then(data => setRoster(data.people ?? []))
  .catch(e => {
    console.error("Failed to load the active-member roster for badge names:", e);
    setRoster(null);
    notifications.show({ color: 'red', message: 'Could not load the active member roster, so badge names cannot be resolved. Reload to retry.', autoClose: false });
  });
```

A console line is not enough here — the operator needs to know why Generate is dead, or they will assume the page is broken rather than that the names are unsafe.

No test covers this path. The closed #1638 had one worth porting (`holds badge generation until the member roster has loaded`): mock the roster route away, spy on `console.error` so `jest-fail-on-console` stays quiet, assert the Generate button is disabled.

---

checkin-app/src/app/facility-ops/print-badges/page.tsx
jee7s: Fixed in b9d0f716.

- `!res.ok` now throws with the status in the message, so an `apiError` 500 reaches the `.catch` instead of resolving through `data.people ?? []`.
- The `.catch` sets `roster` back to `null`, so the `roster === null` guard on both Generate buttons holds.
- Added the red toast: `Could not load the active member roster, so badge names cannot be resolved. Reload to retry.` — `autoClose: false`, same pattern as the `generate()` catch. `console.error` kept.
- Ported the guard test from #1638 and pointed it at the `!res.ok` path specifically (ok:false, status 500, valid JSON body), since that is the case `.catch`-only handling never sees. The plain search still succeeds in that test, so rows render and can be selected; both Generate buttons assert disabled and the toast asserts red + persistent. Verified red against the old handler — it hangs waiting for a `console.error` that never fires.

`npx jest --testPathPatterns print-badges`: 8 passed. `tsc --noEmit`: clean.

thpr (approved):
Blocker addressed in b9d0f716. Verified at the PR head.

- `if (!res.ok) throw` — an `apiError` 500 now reaches `.catch` instead of resolving through `data.people ?? []` to `[]`.
- `.catch` sets `roster` back to `null`, so the `roster === null` guard on both Generate buttons holds. The failure mode is now "the tool refuses to print" rather than "the tool prints bare first names onto laminate."
- Persistent red toast, matching the `generate()` catch pattern — the operator learns why Generate is dead instead of assuming the page is broken.
- The ported guard test is pointed at the `!res.ok` path specifically, which is the right target: that is the case `.catch`-only handling never sees. Asserting both buttons disabled and the toast red + `autoClose: false` pins the behaviour rather than the implementation.

No other changes on the branch, and the `jest.mock("@mantine/notifications")` addition follows the pattern already used in ~10 other page tests.

Approving. The two things I raised and then set aside stay set aside: the bare-first-name path for off-roster people is moot given inactive people don't get badges, and the Printed Name column rendering bare names while the roster is null is cosmetic now that nothing can print in that state.

---

PR #1638, opened 2026-08-11
fix(print-badges): resolve badge names against the active membership

### LINKED ISSUE #1625: Show printed name on the /facility-ops/print-badges screen

thpr:
## What changed

A badge's printed name was computed over whichever people happened to be in the batch being printed (`computeDisplayNames(badges)` inside `BadgeDocument`). The same person therefore got a different name depending on who was printed alongside them: **"Kim"** when printed alone, **"Kim Ke."** when printed next to Kim Kowalski. A reprint could contradict the badge already clipped to somebody's shirt.

Disambiguation now runs against the **active membership** — the population a badge name has to be unique within — and never against the batch.

- `computeDisplayNames(subjects, cohort = subjects)` resolves each subject against the cohort **minus itself** (matched by id), so a person printed alone and the whole roster printed at once yield the same name. Default argument keeps the single-argument call sites behaving exactly as before.
- The print-badges page loads that cohort once from `GET /api/shop/org-members` — already the active-org-member roster, and both `isSysadmin` and `isBoardMember` hold the full view, which is exactly the gate on `/facility-ops`. No new route, so no security-registry change.
- Badge generation is held until the cohort arrives, rather than silently stamping names computed against an empty cohort. A failed roster load logs and notifies instead of printing wrong names.
- The resolved name is exposed as a **"Badge name" column** on the screen — the proof asked for in the issue, and the same value the PDF prints.
- `BadgeDocument` takes `displayName` as given instead of recomputing it, so the column and the PDF are one computation rather than two that can drift.

Stickers are deliberately unchanged: they carry the full name, not the disambiguated one.

## Why

Fixes #1625. The issue is two parts — (1) the name must depend only on the active membership, not on who is being printed right now, and (2) surface the final name as a column as proof. Both are here.

## Reviewer notes

- The cohort universe is `ACTIVE_ORG_MEMBER_PERSON_WHERE` (household membership `ACTIVE`), matching how the page's own Membership column derives `isMember`. Someone printable but not an active member (a guest badge) is still disambiguated *against* that cohort, so their name stays distinguishable from any member's — and stays a pure function of themselves plus the roster.
- `/api/people/search` caps at 200 rows, which is part of why the old batch-scoped computation could never be stable; the cohort fetch does not go through that route.
- #1634 also edits this page (hides inactive rows, #1626). The changes are disjoint in intent but both touch the column list — expect a small conflict there.

## Testing

- `npx tsc --noEmit` clean; eslint clean.
- `npm test -- --testPathPatterns "print-badges|BadgeDocument|badgeDisplayNames|StickerDocument"` → 14 passed, 4 suites.
- New cases: disambiguation against the cohort rather than the printed batch; a non-cohort subject resolved against the cohort; the page showing "Kim Ke." for a collision with a member who is *not* on screen; badge generation disabled until the roster loads.

### DIFF
--- a/checkin-app/src/__tests__/badgeDisplayNames.test.ts
+++ b/checkin-app/src/__tests__/badgeDisplayNames.test.ts
@@ -25,4 +25,23 @@ describe('computeDisplayNames', () => {
     it('shows bare first name for the entry lacking a last name; others disambiguate around it', () => {
         expect(run(['John', 'John Smith'])).toEqual(['John', 'John S.']);
     });
+
+    it('disambiguates against the cohort, not the printed batch', () => {
+        const cohort = [
+            { id: 1, name: 'John Smith' },
+            { id: 2, name: 'John Doe' },
+            { id: 3, name: 'Jane Roe' },
+        ];
+        // One badge printed alone still reads as it does when the whole cohort is printed.
+        expect(computeDisplayNames([cohort[0]], cohort).get(1)).toBe('John S.');
+        expect(computeDisplayNames(cohort, cohort).get(1)).toBe('John S.');
+        // A cohort member with a unique first name keeps the bare first name.
+        expect(computeDisplayNames([cohort[2]], cohort).get(3)).toBe('Jane');
+    });
+
+    it('disambiguates a non-cohort subject against the cohort', () => {
+        const cohort = [{ id: 1, name: 'John Smith' }];
+        // A guest badge (not an active member) still has to be distinguishable from the members.
+        expect(computeDisplayNames([{ id: 9, name: 'John Doe' }], cohort).get(9)).toBe('John D.');
+    });
 });
--- a/checkin-app/src/app/facility-ops/print-badges/__tests__/page.test.tsx
+++ b/checkin-app/src/app/facility-ops/print-badges/__tests__/page.test.tsx
@@ -40,10 +40,18 @@ const withInactive = [
   { id: 3, name: "Lapsed Larry", email: "larry@example.com", isMember: false },
 ];
 
+// The active membership — wider than what the search returns, and the only thing badge names
+// may depend on. Kim Kowalski is a member who is NOT in the search results above.
+const orgMembers = [
+  { id: 1, name: "Kim Keyholder" },
+  { id: 2, name: "Bo Board" },
+  { id: 4, name: "Kim Kowalski" },
+];
+
 describe("facility-ops/print-badges page", () => {
   it("loads and renders the participant roster", async () => {
     setSession({ id: 1, isSysadmin: true });
-    mockFetchJson({ "/api/people/search": { people: participants } });
+    mockFetchJson({ "/api/people/search": { people: participants }, "/api/shop/org-members": { orgMembers } });
     renderWithProviders(<PrintBadgesPage />);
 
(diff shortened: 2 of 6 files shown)


jee7s:
We built #1625 in parallel without seeing this — #1637, opened 11 minutes earlier. I've closed it as the duplicate; this is the better-factored version. `computeDisplayNames(subjects, cohort = subjects)` resolving each subject against the cohort **minus itself by id** is cleaner than what we did, and it avoids a defect our own plan had: an asymmetric variant that grouped over `population ∪ extras` still fired the `group.length === 1` short-circuit off the merged group, so one inactive "John Smithson" flipped an ACTIVE "John" to "John S." — the bug, reintroduced. Yours does not have that shape.

One finding from the closed branch is worth carrying over, and it is the only reason we did not use `/api/shop/org-members` ourselves.

## `org-members` has no `isOperations` in its view

`registry.ts` — `GET /api/shop/org-members` is `authorize: 'certifier'`, and its `orderedView` is:

```
['isSysadmin',    [...]]
['isBoardMember', [...]]
['certifier',     ['member', 'public']]
```

No `isOperations`, and `certifier` is a `MAY_CERTIFY_OTHERS` toolStatus rather than a role an operations user holds.

The reviewer note here says the full view is held by "`isSysadmin` and `isBoardMember`, which is exactly the gate on `/facility-ops`". That is true on `main` today. It stops being true when **#1623** lands: that PR makes `/facility-ops/print-badges` reachable by `isOperations`, implementing the decision recorded in #1633. Its body states the premise explicitly — *"No API change needed for Print Badges: `/api/people/search` already allowed operations."* A cohort fetch against `org-members` falsifies that, and an operations user hits the roster load failing rather than a wrong name — so it surfaces as "badge generation never enables", not as a silent mis-print.

Two ways out, neither of which belongs in this PR:

1. **Add `isOperations` to the `org-members` `orderedView`.** Boundary file, ships alone per the isolation rule. `['isOperations', ['member', 'public']]` mirrors the certifier row and grants nothing this page does not already show.
2. **Take the roster from `/api/people/search` instead**, which is on the frozen legacy baseline and already admits operations. That is what #1637 did, via a `?roster=active` mode. It needs no boundary change but does put a second mode on a legacy route.

(1) is cleaner if #1623 is landing anyway. Flagging rather than choosing, since it is your call and the sequencing is yours too.

## Two observations, useful whichever way that goes

**The 200-row cap still governs what is selectable.** Deriving the *name* from the cohort fixes the name, but `rows` still comes from `/api/people/search` at `take: 200, orderBy: { id: 'desc' }`. A member outside the 200 highest ids and outside the current search string cannot be selected or printed at all. Not this issue, and worth someone knowing.

**Cohort-derived names are stable per session, not forever.** When a "John Smithson" joins, John Smith's correct badge name becomes `John Sm.` while the laminated one still reads `John S.` — and the new column will display that disagreement, which is arguably the column doing its job. Worth a sentence somewhere so the first person to notice does not read it as a regression.

Nothing here blocks. The approach is right and the column is the proof the issue asked for.

---

PR #1641, opened 2026-08-11
fix(badges): print the membership year only for households that renewed

### LINKED ISSUE #1628: Only renewed members should get the current membership year on the badge

jee7s:
Fixes #1628

> **Read the diff with this in mind.** This branches off `claude/1625-printed-name` (#1637, open) and targets `main`, so the diff below **includes #1637's commits** until that PR merges — at which point it self-cleans to just this change. #1637 establishes the `?roster=active` read and the `ParticipantBadge` payload shape this work extends; branching off `main` would mean re-deriving both and then conflicting. Targeting `main` is deliberate: stacked PRs in this repo run no CI, and zero checks read as green.
>
> **My own changes:** `lib/membership/renewal.ts`, `api/people/search/route.ts`, `facility-ops/print-badges/page.tsx`, `components/admin/BadgeDocument.tsx`, `components/admin/badgeNames.ts`, a `MAX_DATE` de-duplication across `lib/orgMembership.ts` / `lib/outreach/recipients.ts` / `api/membership-ops/households/route.ts`, and 5 test files.
>
> **The migration that was here is gone — it is now PR #1644**, split out at review request. Nothing in this branch depends on it.

---

## ⚠️ Every badge prints no year until #1644 lands

`orgMembershipYearBoundary` is null in every environment today, and a null boundary fails closed: `settledSince` becomes the max-date sentinel, nothing matches, and **no badge gets a year at all**. That is this PR's behaviour on merge day, and it is deliberate — a blank year is the renewal prompt the issue asked for, and the new **Year** column shows "Not renewed" on screen before anything reaches a laminator.

The boundary itself is now **PR #1644**, because setting it switches on annual renewal org-wide (cron sweep, renewal-status due dates, todo-counts, turning-18, outreach `nextDeadline`, `bgValidUntilBoundary`) rather than just badge years. When that PR merges, the badge behaviour becomes the more interesting one: a household earns the year by **settling the current renewal cycle**, renewal season opened **2026-07-01** (`RENEWAL_LEAD_MONTHS = 2` against a 1 Sep boundary), so only households that completed renewal in the last ~six weeks print a year. Everyone else — including a household fully paid through 2025-2026 whose membership does not lapse until 1 September — still prints none.

Either way the operator sees it on their monitor first, not on plastic.

## The predicate, and why it is not `ACTIVE`

**ACTIVE cannot be the discriminator.** `OrgMembershipStatus` is `NONE|ACTIVE|REVOKED|DENIED` — there is no grace or lapsed state, and `renewal.ts` documents *"No auto-revoke — manual admin action."* A household that sails past 1 September without renewing simply **stays ACTIVE**. ACTIVE is a strict superset of "renewed", which is exactly why a blank year works as a renewal trigger at all.

The predicate is a **terminal-ACTIVE `OrgMembershipProcess` on that household's membership with `stageEnteredAt >= settledSince`** — the same shape `lib/outreach/recipients.ts` uses for the "who still needs a renewal email" audience.

Two choices worth defending:

- **No `kind` filter.** `PERSON_BG`/`PERSON_AGREEMENT` carry `orgMembershipId: null` so they never appear under `orgMembership.processes`; what remains is `INITIAL` and `RENEWAL`. A family that **joined** this cycle has paid for this year exactly as much as one that renewed, so a `kind: "RENEWAL"` filter would blank every new member's badge. There is an integration test for precisely this, and I verified it is not decorative — adding `kind: 'RENEWAL'` to the query turns it red with `Expected: "2026-2027" / Received: null`.
- **`ARCHIVED` excluded.** `settledThisCycleWhere` uses `{ in: ["ACTIVE","ARCHIVED"] }`, but that predicate answers *"don't re-open this cycle"*, not *"did they pay"*. This follows `recipients.ts`, which uses ACTIVE only. Fail-closed: a wrong blank year prompts a renewal conversation, a wrong printed year suppresses one. One word to change if the board disagrees.

## `badgeYearCycle`, and the death of `membershipYearLabel`

`membershipYearLabel` hardcoded a **June** rollover (`now.getMonth() >= 5`) and never read `orgMembershipYearBoundary` — against the documented 1 September boundary it rolled over **three months early**, as its own comment admitted. It had one caller. **Deleted.**

Its replacement, `badgeYearCycle(configuredBoundary, now)`, sits beside `renewalWindow` in `lib/membership/renewal.ts` and is pure. The label flips at `windowStart`, so the badge starts advertising the coming year exactly when that year becomes renewable and keeps advertising it until the next window opens (boundary = 1 Sep):

| `now` | in season | label | `settledSince` |
|---|---|---|---|
| 2026-08-10 | yes | **2026-2027** | 2026-07-01 |
| 2026-09-15 | no | 2026-2027 | 2026-07-01 |
| 2027-06-15 | no | 2026-2027 | 2026-07-01 |
| 2027-07-15 | yes | 2027-2028 | 2027-07-01 |

Continuous across the boundary, one flip per year, and on today's date it reproduces the `2026-2027` in the issue text — so this PR is purely *who gets it*, not *what it says*. These four rows are the unit test, as `it.each` with an injected `now` and a fixed boundary.

The existing off-season sentinel (`new Date(8.64e15)` — now exported once as `MAX_DATE` from `renewal.ts`, replacing four inline copies, per review) is right for outreach — you only nag in season — and **wrong for badges**, which print year-round: it would blank every badge from ~2 Sep to ~30 Jun. `badgeYearCycle` walks back to the previous cycle's `windowStart` instead. The sentinel still appears, but only for the genuinely-unconfigured case (which is every environment until #1644).

## The 1 September boundary is PR #1644, not this one

`BoardSettings.orgMembershipYearBoundary` is `DateTime? @db.Date`, nullable, with no seed — only the settings UI has ever written it. Setting it is what makes any badge print a year, but it is also the null-gate on the whole renewal machine, so it lands on its own PR with its own ack: **#1644**.

**Null still fails closed.** No boundary ⇒ `settledSince` is the max-date sentinel (matches nothing), `label` is null, nobody is renewed, no badge gets a year. There is an integration test for it, and the route logs a warning at request time so the failure is visible from the app rather than only from the printer. That is why this PR is complete without #1644.

Two alternatives were rejected outright and stay rejected, since they would have kept the boundary in *this* PR:

- **A schema `@default`** fires on INSERT only. The `BoardSettings` singleton is created by `GET /api/settings/membership`'s upsert and already exists in every environment, so a default would never reach the existing row.
- **A code fallback** (`boundary ?? Sep 1`) deletes the fail-closed path. A genuinely unconfigured org would get a fabricated boundary and print a year nobody earned.

A **seed row** was also rejected — `prisma/seed.ts` never touches `BoardSettings` and seeds don't run against prod. The boundary is likewise **not** in the settings route's `DEFAULTS` constant; this codebase is explicit about not guessing board policy on bootstrap (`scholarshipDenialGraceDays`: *"null = feature off (never guess a default)"*).

## `year: string | null` renders *nothing*

`ParticipantBadge` gains `year: string | null`, per badge (one print run legitimately mixes renewed and not-renewed households). Null emits **no element at all** — `{badge.year ? <Text …> : null}`, not an empty `<Text>`. This matters: `headerRow` is `flexDirection: row` with `justifyContent: 'space-between'`, so an empty second child would still claim its slot and shove the logo off the left edge. Verified in the component rather than assumed — the test counts the spans inside each badge's header row and asserts `[1, 0]`.

`BadgeDocument.test.tsx`'s old blanket `getAllByText(/^\d{4}-\d{4}$/).length > 0` could not survive a conditional year, so it is now **fixture-driven**: one badge with `year: "2026-2027"`, one with `year: null`.

## Red, then green

Proved by **building the pre-fix behaviour**, not by reverting: `membershipYearLabel` restored with its June rollover, rendered unconditionally on every badge, `badgeYearCycle` swapped for the old label logic, and the roster branch handing the year to everyone.

All 7 new assertions went red, each for the right reason:

```
● badgeYearCycle › deep off-season, next spring
    - "label": "2026-2027",   ← boundary-driven
    + "label": "2027-2028",   ← June rollover, three months early
    - "settledSince": 2026-07-01   + "settledSince": 1970-01-01

● BadgeDocument › prints the year only on the badge that carries one
    Expected length: 1
    Received array:  [<span>2026-2027</span>, <span>2026-2027</span>]

● print-badges page › shows the membership year only for a household that settled this cycle
    Expected: "Not renewed"   Received: "2026-2027"

● roster › the membership year goes only to households that settled this renewal cycle
    Received: "2026-2027"     (expected null)
```

That last pair is the issue text verbatim — *"Current badges all print 2026-2027."*

Since the review: the roster fetch's fail-open handling is fixed on #1637 and merged forward here, off-roster people now disambiguate among themselves (new page test), and `peopleSearchRoster`'s teardown deletes the `BoardSettings` row it created rather than restoring a null onto it.

Green after restoring: **278/278 unit suites** (2735 tests) and **137/137 integration suites** (1477 tests), `tsc --noEmit` clean, `eslint --max-warnings 0` clean, `check-route-coverage --strict` 0 errors. No new route method key — `?roster=active` rides the existing legacy-baseline `GET /api/people/search`, so no registry entry and no boundary PR. `git diff` against `checkin-app/src/security/` is empty.

## Notes and limits

- **I could not read the prod `BoardSettings` row.** No DB access from here. It does not gate this PR — null fails closed either way — but if prod already holds a boundary that is *not* 1 September, #1644's `WHERE … IS NULL` leaves it alone and the label table above shifts accordingly. `badgeYearCycle` is boundary-agnostic; only the table's illustrative dates depend on 1 Sep.
- The `take: 200` cap on the *selectable* rows is unchanged and out of scope — the roster read that feeds names and years is uncapped, but which people the operator can tick still comes from the capped search.
- The screen's **Year** column is a small addition beyond the literal issue text. It is the same "column as proof" pattern #1637 introduced for the printed name, and it is what turns the blank-badge surprise into something visible before printing rather than after.


thpr (changes requested):
Changes requested. The badge logic is right — I walked `badgeYearCycle` against `nextBoundary`/`renewalWindow` for in-season, on-boundary, mid-boundary-day and deep off-season and it is continuous with one flip per year; the terminal-`ACTIVE` predicate matches `recipients.ts`; and the "no `kind` filter" argument holds, because `personAgreementTriggers.ts:128` really does write `orgMembershipId: null`.

The problem is that this PR is two changes, and the second one is much larger than its one line of SQL suggests.

## Split the migration out

`20260811130000_board_settings_year_boundary_default` is presented as "otherwise badges lose their year". That is true, and it is not what the statement does. `orgMembershipYearBoundary` is the null-gate on the entire renewal machine:

- `runRenewalSweep` (`renewal.ts:143`) returns `"no membership-year boundary configured"` today. With the boundary set and today already inside the Jul 1 window, the next `GET /api/cron/membership-renewals` opens a `PENDING_RENEWAL` process for **every ACTIVE membership** — the skip clause only spares memberships with an in-flight or settled-this-cycle RENEWAL, and with the boundary null none has ever been opened.
- `membership/renewal-status` starts showing members a due date. `nav/todo-counts` drops its unset-boundary todo. `membership-audit/turning-18` goes from an empty list to a populated one. `settings/outreach` gains a `nextDeadline`. The `lib/outreach/recipients` renewal audience becomes non-empty.
- `bgValidUntilBoundary` starts returning dates, so `membership-ops/households` rows gain `bgValidUntil` and `householdBgIsFresh` changes what `beginRenewal` pre-clears.

`docs/backlog/CUJS.md:227` says the sweep emails nothing today, so this is not a mass mailing. It is still a mass write of renewal processes plus member-visible "your renewal is due" state, on the same morning as the badge change.

The value itself is correct (`docs/rules/membership.md:19`) and the argument against a code fallback is sound — I am not asking for a different fix, I am asking for a different PR. Setting a `BoardSettings` policy value that switches on annual renewal org-wide should land on its own, where the body is about that and the board can ack it, not as a supporting detail in a badge fix. The badge change is fine without it: null already fails closed to a blank year, which is the same prompt the PR wants.

## Fix before merge: the roster fetch is fail-open

`print-badges/page.tsx:70-79` does not check `res.ok`:

```ts
.then(res => res.json())
.then(data => setRoster(data.people ?? []))
.catch(e => { console.error(...); setRoster([]); })
```

A 403 or 500 body parses to `{ error: "..." }`, so `data.people ?? []` yields `roster = []` rather than `null`. The `roster === null` guard on the two buttons then releases, and the run prints un-disambiguated bare first names and no year on every badge, with only a `console.error` to say so. That is exactly the outcome the button's own comment says the guard exists to prevent. The `.catch` arm sets `[]` deliberately and has the same effect.

Keep `roster` at `null` on `!res.ok` and on throw so the buttons stay disabled, and surface a red `notifications.show` — the page already uses that pattern for PDF failures.

## Smaller things

- `year` is derived from `OrgMembershipProcess.status` and `stageEnteredAt`, both `@sensitivity:internal`, and is returned to `isOperations`. Ops prints badges (#1623) so it probably has to see this, but this route's standard is to argue each field ops may see in place — the `opsOnly` comment block does it at length for everything else. Roster mode says nothing.
- `printedName` falls back to a bare first name for anyone off the ACTIVE roster, and `BadgeDocument` now prints `displayName` verbatim. With "Hide inactive" off, two non-member Johns in one run both print `John`; the old in-batch computation disambiguated them. The comment justifies excluding them from the roster population, not the collision among themselves.
- `new Date(8.64e15)` goes in inline as a magic number. `recipients.ts:26` already names this sentinel and notes it mirrors `orgMembership.ts:95` — that is a fourth copy.
- The `headerRow` comment overclaims: with `justifyContent: 'space-between'` and a single child the logo sits at flex-start either way, so an empty `<Text>` would sit at the right, not shove the logo off the left. The conditional render is still the right call; the stated reason is not.
- `setBoundary` upserts `{ id: 1 }`, and `prevBoundary` cannot distinguish "row absent" from "row present with a null value". If the row did not exist, `afterAll` restores the value but leaves the row, changing what any suite asserting on an unconfigured install sees.

Happy to re-review the badge change on its own as soon as the migration is its own PR.

---

jee7s:
Review addressed. Pushed `28150949`, `ad259144` (merge), `c755ea5f`, `0837707d`.

**Migration split out → #1644.** `20260811130000_board_settings_year_boundary_default` is `git rm`'d here and is now its own PR off `main`, timestamp unchanged (still distinct from #1640's `…120000`). Its body leads with what you said it is — the switch that turns annual renewal on org-wide (first cron sweep opening a `PENDING_RENEWAL` per ACTIVE membership, renewal-status due dates, todo-counts, turning-18, outreach `nextDeadline`, `bgValidUntilBoundary`), notes `CUJS.md:227` (no mail today), cites `docs/rules/membership.md:19` for the Sep-1 value, and asks for an explicit board/reviewer ack before merge. This branch is unaffected: null still fails closed to a blank year. PR body updated to say so.

**Roster fail-open** — fixed on #1637 (`b9d0f716`) and merged forward here. `!res.ok` throws with the status, `.catch` keeps `roster` at `null` so the button guard holds, plus a persistent red `notifications.show`. Guard test ported from #1638 and extended to the `!res.ok` path specifically.

**Smaller things:**

- **`year` argued in place** — the roster branch now carries the 4-line argument the `opsOnly` block gives every other ops-visible field: ops prints badges (#1623) so it has to see who renewed, derived from `status`/`stageEnteredAt` but exposing only the year string.
- **Off-roster collisions** — done, and it needed no signature change. Off-roster people now disambiguate among the fetched batch via a second `computeDisplayNames` map, consulted only after the roster map misses, so an off-roster name still cannot move a member's. New page test asserts `John N.` / `John G.` for two non-members while the one roster John stays bare `John`. Skipped: the map is empty while `roster === null`, so nothing prints batch-derived names during a roster failure.
- **`new Date(8.64e15)`** — now `export const MAX_DATE` in `lib/membership/renewal.ts`, beside the window functions whose null it stands in for. All four call sites converted (`orgMembership.ts`, `outreach/recipients.ts`'s local copy, `membership-ops/households` ×2, `people/search`); each already imported from `renewal.ts`, so it is one word per import line. No `8.64e15` left outside the definition. Server-only throughout — nothing crosses into the client page.
- **`headerRow` comment** — corrected. The conditional render stays; the reason is now that it keeps the row to a single child rather than a false claim about the logo being shoved off the left edge.
- **`setBoundary` teardown** — `peopleSearchRoster` records whether `BoardSettings` existed before `beforeAll`, and `afterAll` deletes the row if it did not, instead of restoring a null value onto a row the upsert created.

`tsc --noEmit` clean, `eslint --max-warnings 0` clean on the touched files, `print-badges | renewal | BadgeDocument | badgeDisplayNames`: 4 suites / 40 tests green. Integration suites are Docker-dependent and left to CI.

Threads left unresolved for you.

---

thpr (dismissed):
Approving. Re-reviewed at `0837707d`; all six items resolved, and I checked each rather than taking the summary's word for it.

- **Migration split.** `20260811130000_board_settings_year_boundary_default` is gone from this tree, and #1644 is one file off `main` — not stacked on this branch — with the renewal blast radius in the title. This branch still fails closed to a blank year with a null boundary.
- **Fail-open closed.** `!res.ok` throws, `.catch` returns `roster` to `null` so the Generate guard holds, red `autoClose: false` notification. The test drives a 500 with a *valid JSON body*, which is precisely the path `.catch` alone would have missed, and asserts both buttons disabled.
- **`MAX_DATE`.** One definition, four call sites converted, `git grep 8.64e15` returns only the definition. I checked the bundle and import-cycle hazard this could have introduced: `orgMembership.ts` already imported `renewal.ts`, so there is no new module edge, and all thirteen importers are server-side. Clean.
- Ops `year` argued in place, off-roster collisions fixed without letting a non-member move a member's name, test teardown now deletes the row it created, `headerRow` comment corrected.

## One nit, not blocking

`offRosterNames` is computed over `participants` — the search results — so an off-roster person's printed name still moves when the operator types. With both non-member Johns visible, one prints `John N.`; search away the other and he prints `John`. Two runs, two different badges for the same person.

That is the #1625 class, scoped to non-members, and it is strictly better than the previous behaviour where both printed `John`. Members are unaffected, since the roster map is consulted first. A stable fix would need a second uncapped fetch and is not worth it here. But the comment says "over the fetched batch" without naming the consequence — worth a clause, so the next reader does not have to rediscover that this cohort's names are search-dependent by design.

Also cosmetic: `printedName`'s deep fallback is now `User #${p.id}`, so while `roster === null` the Printed Name column reads `User #1`, `User #8` for every row rather than the old bare first name. Nothing can print in that state, so it is only a display question — but it reads as a resolved name rather than as "unresolved".

Neither needs a re-review. Land it whenever #1637 does.

---

jee7s:
Merged `main` (`ae33a703`) to clear the conflict #1637's squash-merge created: main carried #1637's content under a new sha, so the six files this branch shares with it conflicted against themselves. Resolved to the branch side for all six — main's version of each is byte-identical to #1637's branch head, which this branch already contains, so the branch is the superset (#1637 final + the year feature).

Verified the strong way: `git diff origin/main...HEAD --stat` is now only the year-feature delta — 12 files, no migration directory, no `schema.prisma`/`security/generated` from #1640, nothing #1637-only. `npx jest --testPathPatterns 'print-badges|renewal.test|BadgeDocument'` 35/35 pass and `tsc --noEmit` is clean; integration suites left to CI.

jee7s:
@thpr this needs reapproval as the setting to dismiss stale requests on new pushes was just enabled.

thpr (approved):
Approving. The logic is right and does what #1628 asked.

Verified by reading rather than by trusting the body:

- `badgeYearCycle` is continuous through the boundary in **both** `nextBoundary` branches. On the boundary day itself `nextBoundary` rolls to next year (strict `<` against a full timestamp), so `inSeason` goes false, `endYear` loses its `+1`, and the label and `settledSince` come out identical to the day before. One flip per year, at `windowStart`, as documented.
- The predicate is the right one. `OrgMembershipStatus` has no lapsed state and nothing revokes at the boundary, so ACTIVE genuinely cannot discriminate; terminal-ACTIVE process with `stageEnteredAt >= settledSince` and no `kind` filter matches `lib/outreach/recipients.ts` and correctly counts an INITIAL settled this cycle.
- `membershipYearLabel` is gone with no stale references, and `BadgeDocument` has one caller.
- Null `year` emits no element, so `headerRow`'s `space-between` keeps the logo at flex-start.
- Not a boundary PR: `GET /api/people/search` is on the legacy-authz baseline with no `src/security/` registry entry, and the `src/security/` diff is empty.
- The integration teardown correctly distinguishes "row absent" from "row present, value null" on the `BoardSettings` singleton.

## One thing in the body is no longer true

The body's central risk argument is that `orgMembershipYearBoundary` is null everywhere, so nothing prints a year on merge day and the interesting behaviour waits for #1644. **#1644 merged** — `01249093` on `main` sets the boundary to `2026-09-01`.

So on deploy the live branch is the interesting one, not the inert one. Today is in-season, `settledSince` is 2026-07-01, and nearly every badge prints blank until its household renews — including households paid through 2026-08-31, whose membership year does not end until then (`docs/rules/membership.md:19`).

That is the outcome #1628 asked for, so it does not change my approval. But the whole reason #1644 was split out was so the switch got its own board ack, and the badge half now arrives with no gate. Worth rewriting that section before merge so the deploy-day expectation matches what will actually happen — and worth noting the null-boundary integration test is now the only exercise of a path production will not take.

## Follow-on, not a blocker

#1651: off-roster printed names are computed over `participants`, which is the search-filtered, `take: 200` result, so a non-member's badge label moves with the search box and can collapse onto a member's. Filed separately because it is orthogonal to the year work and the neighbouring stability test only covers roster members.

---

PR #1643, opened 2026-08-11
fix(facility-ops): make the Operations grant reachable — sidebar entry and a live index

jee7s:
## What

#1623 gave operations two Facility Ops tools (Participation Trends, Print ID Badges — decision #1633) but no way to reach them: the sidebar's "Facility Ops" entry was board/sysadmin-only, and `/facility-ops` was a bare server redirect into `/facility-ops/visits`, a tab that ejects operations back to `/`. A typed URL was a dead end and the nav item never appeared.

## Reuse, not a third copy

Both fixes read from `lib/facilityNav.ts`, the table #1623 introduced — no new permission table, helper, or abstraction:

- The sidebar entry (`AppFrame.tsx`) now gates on `FACILITY_SECTION_ROLES`, the same derived union the section layout (`facility-ops/layout.tsx`) already gates on via `useRequireRole`.
- The index (`facility-ops/page.tsx`) redirects to `visibleFacilityLinks(session?.user)[0]?.href`, the same helper the layout uses to render tabs.

This is relevant to open issue **#1569** (the index directory forking the section permission table) — this PR does not make that fork worse; it reads the same source of truth `#1569` is about, it doesn't retype it.

## The index shape

Follows the existing `shop-ops` `defaultShopTab` precedent exactly: a client component using `useSession` + `router.replace` + `<PageLoader />`, because the destination depends on the session. The layout still enforces access.

## Behaviour preserved for board/sysadmin

`FACILITY_NAV_LINKS` is ordered Visit History, Raw Badge Events, Print ID Badges, Participation Trends, Corrections. Board/sysadmin's first visible link is still "Visit History", so `/facility-ops` lands them exactly where it always did (`/facility-ops/visits`). Operations' first visible link is "Print ID Badges" (`/facility-ops/print-badges`) — the first tab in that list they're actually allowed to open.

## pageRegistry

The index row's `visible` moved `BOARD` → `BOARD_OR_OPS` (already defined in the file) because the hub now lands operations on a tab they can open, not one that ejects them. The block comment's "the index, which redirects to /facility-ops/visits" claim was falsified by this PR — the index now redirects to the caller's first visible tab, so it's fixed to say that and to explain why the index row is listed to anyone the section admits.

## Tests

Three files, each written to fail on unmodified `origin/main` and pass with the fix:

- `checkin-app/src/components/__tests__/AppFrame.test.tsx` — new test: an operations-only user sees "Facility Ops" in the nav and does not see a board-only section ("Membership Audit"), proving the entry appeared because of the union, not a sloppy gate. Existing tests (plain signed-in user, certifier) are untouched and still assert Facility Ops is absent for them.
- `checkin-app/src/app/facility-ops/__tests__/page.test.tsx` — rewritten for the client component (mirrors `shop-ops/__tests__/page.test.tsx`): board/sysadmin → `/facility-ops/visits`, operations → `/facility-ops/print-badges`.
- `checkin-app/src/__tests__/pageRegistry.test.ts` — flipped the existing `it('hides the /facility-ops index from operations...')` (premise falsified by this PR) to assert it's now listed, and renamed it to say why.

### Red/green evidence

**AppFrame.test.tsx** — reverted `visible` to `(u) => !!u?.isSysadmin || !!u?.isBoardMember`:
```
TestingLibraryElementError: Unable to find an element with the text: Facility Ops.
  at Object.getByText (src/components/__tests__/AppFrame.test.tsx:105:19)
Test Suites: 1 failed, 1 total
Tests:       1 failed, 13 passed, 14 total
```
Restored → 14/14 pass.

**facility-ops/page.tsx** — reverted to the original `redirect("/facility-ops/visits")`:
```
● FacilityOpsIndex › redirects a board member / sysadmin to /facility-ops/visits
  expect(jest.fn()).toHaveBeenCalledWith(...expected)
  Expected: "/facility-ops/visits"
  Number of calls: 0

● FacilityOpsIndex › redirects an operations user to /facility-ops/print-badges — a page they can stay on
  expect(jest.fn()).toHaveBeenCalledWith(...expected)
  Expected: "/facility-ops/print-badges"
  Number of calls: 0

Test Suites: 1 failed, 1 total
Tests:       2 failed, 2 total
```
Restored → 2/2 pass.

**pageRegistry.test.ts** — reverted the index row to `visible: BOARD`:
```
● Facility Ops directory agrees with the section gates › lists the /facility-ops index to operations (it redirects into their first visible tab)
  expect(received).toBe(expected) // Object.is equality
  Expected: true
  Received: false
    at Object.toBe (src/__tests__/pageRegistry.test.ts:59:64)

Test Suites: 1 failed, 1 total
Tests:       1 failed, 9 passed, 10 total
```
Restored → 10/10 pass.

### Full verification (clean, on the restored diff)

- `npx tsc --noEmit` → exit 0
- `npm run lint` → exit 0
- Full unit run: `Test Suites: 278 passed, 278 total` / `Tests: 2728 passed, 2728 total`

Context: #1623, #1633.

### DIFF
--- a/checkin-app/src/__tests__/pageRegistry.test.ts
+++ b/checkin-app/src/__tests__/pageRegistry.test.ts
@@ -55,8 +55,8 @@ describe('Facility Ops directory agrees with the section gates', () => {
     expect(entryFor(href).visible(ops, true, null)).toBe(roles.includes('isOperations'));
   });
 
-  it('hides the /facility-ops index from operations (it redirects into Visits)', () => {
-    expect(entryFor('/facility-ops').visible(ops, true, null)).toBe(false);
+  it('lists the /facility-ops index to operations (it redirects into their first visible tab)', () => {
+    expect(entryFor('/facility-ops').visible(ops, true, null)).toBe(true);
   });
 
   it('still shows every Facility Ops entry to a board member', () => {
--- a/checkin-app/src/app/facility-ops/__tests__/page.test.tsx
+++ b/checkin-app/src/app/facility-ops/__tests__/page.test.tsx
@@ -1,15 +1,26 @@
-// eslint-disable-next-line @typescript-eslint/no-require-imports -- jest.mock factories run hoisted, before imports exist
+// eslint-disable-next-line @typescript-eslint/no-require-imports
 jest.mock("next/navigation", () => require("@/test-helpers/rtl").navMock());
-
-import { redirect } from "next/navigation";
-import { resetRtl } from "@/test-helpers/rtl";
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+jest.mock("next-auth/react", () => require("@/test-helpers/rtl").authMock());
+import { renderWithProviders, setSession, resetRtl, router } from "@/test-helpers/rtl";
 import FacilityOpsIndex from "../page";
 
 beforeEach(() => resetRtl());
 
-describe("facility-ops index page", () => {
-  it("redirects to the visits tab", () => {
-    expect(() => FacilityOpsIndex()).not.toThrow();
-    expect(redirect).toHaveBeenCalledWith("/facility-ops/visits");
-  });
+describe("FacilityOpsIndex", () => {
+    it("redirects a board member / sysadmin to /facility-ops/visits", async () => {
+        setSession({ id: 1, isSysadmin: true });
+        renderWithProviders(<FacilityOpsIndex />);
+
+        await Promise.resolve();
+        expect(router.replace).toHaveBeenCalledWith("/facility-ops/visits");
+    });
+
+    it("redirects an operations user to /facility-ops/print-badges — a page they can stay on", async () => {
+        setSession({ id: 2, isOperations: true });
+        renderWithProviders(<FacilityOpsIndex />);
+
+        await Promise.resolve();
+        expect(router.replace).toHaveBeenCalledWith("/facility-ops/print-badges");
+    });
 });
--- a/checkin-app/src/app/facility-ops/page.tsx
+++ b/checkin-app/src/app/facility-ops/page.tsx
@@ -1,5 +1,28 @@
-import { redirect } from "next/navigation";
+"use client";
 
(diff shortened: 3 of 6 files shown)


thpr (approved):
Approving. The fix is reachable end to end, not just wired.

What I checked beyond the diff:

- **The layout really is the gate.** `useRequireRole` returns `ready: false` until authenticated *and* authorized, and `facility-ops/layout.tsx` returns `null` in that state — so the index page never mounts for a caller with no facility role. The `?? "/"` fallback is unreachable defensive code rather than a second gate, which is the right shape.
- **Operations land somewhere live.** Both destinations behind the aggregate grant admit them at the API too: `GET /api/people/search` and `GET /api/facility/trends` each carry `roles: ['isSysadmin', 'isBoardMember', 'isOperations']`. Print ID Badges isn't a page that renders and then 403s on its first fetch.
- **The ordering claim holds.** `FACILITY_NAV_LINKS[0]` is Visit History on `FACILITY_RECORD_ROLES`, so board and sysadmin still land on `/facility-ops/visits` exactly as the old server redirect did.
- **No third nav surface.** Every `facility-ops` reference outside the section is `AppFrame.tsx`, `pageRegistry.ts`, and the two tests. Nothing in `src/security/` or `middleware.ts` reads the section gate, so this isn't a boundary change.

The server-to-client conversion matches `shop-ops/page.tsx` line for line, and there was no `metadata` export or other server-only surface on the old page to lose.

One nit, not a blocker, filed on #1569 where it belongs: the index row's gate is now derived on the `AppFrame` side (`FACILITY_SECTION_ROLES.some(...)`) and still hand-typed on the registry side (`BOARD_OR_OPS`), and the new `pageRegistry.test.ts` assertion hardcodes `.toBe(true)` rather than reading the union the way its sibling `it.each(FACILITY_NAV_LINKS)` test does. The two agree today; a future role added to any facility link would widen the nav and the layout while the directory silently stayed put, with the test still green. Details on the issue.

---

Issue #1520, opened 2026-08-06
Enforce variant⟺price invariant on Program PATCH (symmetric guard)


thpr:
Follow-up from PR #1351. Review finding #2 (refuted for the notice, but surfaced this latent gap).

## Invariant
The whole program-capacity/Shopify model assumes: **a program has a Shopify variant ⟺ it has a price.** This holds on every normal path — variants are minted only when `basePriceCents > 0` (`programs/route.ts`), and the sync-repair route refuses free programs (`sync-shopify/route.ts` → `isProgramCheckoutBroken` treats "free" as checkout-OK).

## Gap
`PATCH /api/programs/[id]` accepts raw `shopifyVariantId` / `shopifyOrgMemberVariantId` / `shopifyNonOrgMemberVariantId` (sysadmin/board only) with **no cross-check against price**. Two illegal final states slip through:
1. wiring a variant onto a **price-null** program;
2. nulling **both** prices while a variant stays stranded.

Either desyncs the capacity/oversell math (e.g. the freed-seat `notice` would fire for a program that never consumed a Shopify unit). Not reachable from the UI today (no client sends those fields), so this is **latent, not a live bug** — a privileged raw-API misconfiguration.

## Fix
Compute the **post-PATCH merged** state (current row + body deltas) and reject `400` when `variantSet && bothPricesNull` — **both directions**, not just free→variant. Guarding one direction is a false sense of safety.

## Notes
- Security-sensitive route → its own PR + isolated review per AGENTS.md boundary rules.
- Add route tests for both illegal directions + the two legal ones (priced+variant, free+no-variant).
- Low urgency (UI-unreachable); do it when the PATCH route is next opened.

Issue #1441, opened 2026-08-01
Program dates are optional; policy requires a start and a bounded end (policy divergence)


thpr:
**Backlog item P33** — from the backlog register (`docs/backlog/INDEX.md`), source: policy divergence (DIVERGENCE_HANDOFF).

Program dates are optional; policy requires a start and a bounded end (policy divergence) — POLICY SHORTFALL: both start and end can be absent and nothing rejects it. A start must be required; an absent end should fall to fiscal-year-end, not stay empty. While missing, three rules silently change meaning: (a) age eligibility falls back to the request moment — the exact thing the age rule exists to avoid (safety-adjacent); (b) the catalogue reads a missing end as running indefinitely; (c) member pricing falls back to status alone.

- Origin: FIX · Readiness: NEEDS-DESIGN · Size: M

_Auto-created from the backlog INDEX; see the doc + `backlog/sources/` reconciliation for context._

jee7s:
**HOLD** — the diagnosis is right and consequence (c) is worse than the issue says, but the proposed fallback presumes an answer that #1484 has not given yet.

Verified live on `origin/main`: `programCoverageDate` (`checkin-app/src/lib/orgMembership.ts:203`) is `p.endAt ?? p.startAt ?? null`, and `coversThrough` (`:168`) returns `true` when `through` is null. `GET /api/programs/[id]/discount-code` calls `isDuesSettledThrough(auth.user.id, programCoverageDate(program))`. So a program with no dates at all does not merely "fall back to status alone" — it grants the member discount **unconditionally**, in shipped money code, today.

The blocker for promotion is the fallback itself: "an absent end falls to fiscal-year-end" presumes a program's fiscal year is knowable, which is exactly the ambiguity #1484 exists to remove. Settle #1484 first.

To promote, answer:

1. Where does "fiscal-year-end" come from — the declared fiscal year #1484 proposes, or `BoardSettings.orgMembershipYearBoundary` directly?
2. What happens to programs that already exist on main with a null `startAt` or `endAt` — backfill to a computed date, or block edits until an admin supplies one?
3. Is a genuinely open-ended program (an ongoing drop-in) still allowed, or does policy really mean every program ends?
4. Does making `startAt` required break drafts saved before dates are known?

---
_Posted by Claude._

thpr:
**Register work this owes when implemented** — recording it here so it does not live only in a side document.

[...]

PR #1645, opened 2026-08-12
ci(discussions): read-only staleness audit for release-planning discussions (no-model comparison for #1602)

jee7s:
This is the version @thpr asked for on #1602: *"A script that reads the discussion body's checked boxes and reports what has gone stale gets all of it, with no injection surface, no API key, and no secret to add."* It sits next to #1602 rather than replacing it, so the build-at-all decision has both shapes to weigh. Neither this nor #1602 should merge on autopilot.

It also happens to be what a local, human-previewed run would consume — the report below is the input a human would read before deciding to touch anything.

**127 lines of bash, 27 of YAML. No new dependencies (`gh` + `jq` + `awk`).**

## What it does

Given a discussion number, it reads the body, pulls candidate issue numbers out of the task-list boxes, and then reads each one's live state and milestone. It reports:

- candidates that are **closed**
- candidates already on a **different** milestone
- candidates that **are not issues** — a PR number or a discussion number in a box
- issues **already on the target milestone** that the discussion does not list, split into "mentioned in prose" and "not mentioned anywhere"
- whether the milestone exists yet, and whether it is open or closed

## What it does not do

- **No writes.** Every call is a GET. It cannot create a milestone, assign an issue, or post a comment.
- **No model.** No `ANTHROPIC_API_KEY`, no secret to add to this repo, nothing to rotate.
- **No injection surface.** The untrusted body text does exactly one thing: it selects which **public issue numbers** get read. There is no other channel out of the parser — a number is a number. If a box holds a number that is not an issue, that is *reported as a finding*, never followed. Both branches are confirmed against the live API: `#1602` (a PR) comes back with `pull_request != null`; a discussion number 404s.
- **No `discussion` triggers.** The workflow is `workflow_dispatch` only, so nobody outside the repo can start it, and the input reaches the shell through `env:` rather than `${{ }}` interpolation.

## The discussion format it assumes

Derived from the two real planning discussions, #1598 and #1599 — stated plainly because the parser is the whole trust boundary:

1. **Milestone name = the first `vMAJOR.MINOR` in the discussion title.** Same rule `apply.sh` uses in #1602. `"v1.2 — Program leaders can trust their screen"` → `v1.2`. Nothing in the *body* can name a milestone.
2. **A candidate is a markdown task-list line** — `- [ ]` or `- [x]`, any indent, `*` bullets and uppercase `[X]` accepted.
3. **On a task line, bold `**#N**` refs win.** Both real discussions bold their candidates and leave prose refs plain, which matters: `- [ ] **#1441** — program dates required *(design merged in PR #1515)*` has to yield `1441` and not `1515`. If a task line has *no* bold ref, it falls back to every plain `#N` on that line — an off-convention line gets flagged rather than silently dropped.
4. **One box may hold several issues.** `- [ ] **#1435** + **#1520**` yields both.
5. **Checked and unchecked are tracked separately** and both are audited. In practice this matters more than the reviewer's phrasing suggests: right now *every* box in both live discussions is unchecked, and the closed-candidate finding below comes off an unchecked one. Auditing only checked boxes would report nothing at all today.

Assumptions 3 and 5 are the ones most likely to be wrong later. Both fail loudly (a wrong number gets looked up and reported), never silently.

## Sample report — real run against #1598

Verbatim stdout, nothing edited:

```markdown
## Staleness audit — [#1598](https://github.com/innovationtreehouse/checkin/discussions/1598)

v1.2 — Program leaders can trust their screen (candidates, not committed)

Target milestone from the title: **v1.2** — exists (#2, **open**).

Candidates parsed: **0** checked, **8** unchecked.

### Candidates that are closed

- **#1435** (unchecked) — is **closed**

### Candidates already on a different milestone

_none_

### Candidates that are not issues

_none_

### On **v1.2** but not a candidate

- **#1508** (open) — mentioned in the body's prose, but not a candidate box
- **#1154** (open) — **not mentioned anywhere** in the discussion

_Read-only: this run made no writes and called no model._
```

That is both findings #1602's PR body claims for the model version — the closed candidate (#1435, closed by PR #1555 two days before the discussion was written) and the milestone already holding issues the list does not name — recovered with no model in the loop. The prose/never-mentioned split is a little extra: #1508 is at least discussed in #1598's conditionals table, #1154 is not in that discussion at all.

#1599 was also run: milestone `v1.3` does not exist yet, 10 unchecked candidates, nothing stale. It correctly excludes the `PR #1515` and `PR #1457` refs that sit in prose on candidate lines.

## How to run it

Locally, needing nothing but an authenticated `gh`:

```sh
scripts/discussion-milestone-audit.sh 1598
```

In CI, Actions → **Discussion milestone audit** → Run workflow, with the discussion number. Output goes to the job summary. The job's token is `contents: read`, `issues: read`, `discussions: read`.

## The check

```sh
scripts/discussion-milestone-audit.sh --self-test
# OK: 3 checks
```

Parsing is factored into one function with no network in it, so the check feeds a fixture body straight through it and `diff`s the extracted `state<TAB>number` pairs. The fixture covers the cases the real discussions actually contain: checked and unchecked, an indented uppercase `[X]` with a plain ref, two issues in one box, a bold candidate on a line that also names a PR, a box with no number, and prose and table refs that must *not* be picked up. No framework — same shape as the existing `scripts/test-*.sh`, which are also run by hand.

## Not built

Any write path, a diff-since-last-run, caching, and a scheduled trigger — a report nobody asked for on a cadence is just a notification to mute. Dispatch it when a decision is near.

### DIFF
--- a/.github/workflows/discussion-milestone-audit.yml
+++ b/.github/workflows/discussion-milestone-audit.yml
@@ -23,5 +23,9 @@ jobs:
           GH_TOKEN: ${{ github.token }}
           REPO: ${{ github.repository }}
           NUM: ${{ inputs.discussion }}
+        # Without pipefail the job's status is `tee`'s, so a failed audit reports
+        # green; `2>&1` puts the error in the summary instead of leaving it blank.
+        shell: bash
         run: |
-          scripts/discussion-milestone-audit.sh "$NUM" | tee -a "$GITHUB_STEP_SUMMARY"
+          set -euo pipefail
+          scripts/discussion-milestone-audit.sh "$NUM" 2>&1 | tee -a "$GITHUB_STEP_SUMMARY"
--- a/.github/workflows/discussion-milestone-audit.yml
+++ b/.github/workflows/discussion-milestone-audit.yml
@@ -1,0 +1,27 @@
+name: Discussion milestone audit
+
+# Read-only. No secrets, no discussion triggers, no writes — the job is a
+# report, so the token is read-only and there is nothing here to hijack.
+on:
+  workflow_dispatch:
+    inputs:
+      discussion:
+        description: Discussion number to audit
+        required: true
+
+permissions:
+  contents: read
+  issues: read
+  discussions: read
+
+jobs:
+  audit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - env:
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+          NUM: ${{ inputs.discussion }}
+        run: |
+          scripts/discussion-milestone-audit.sh "$NUM" | tee -a "$GITHUB_STEP_SUMMARY"
--- a/scripts/discussion-milestone-audit.sh
+++ b/scripts/discussion-milestone-audit.sh
@@ -38,8 +38,10 @@ parse_candidates() {
 milestone_from_title() { grep -oE 'v[0-9]+\.[0-9]+' <<<"$1" | head -1 || true; }
 
 self_test() {
-  local got want
-  got=$(parse_candidates <<'FIXTURE'
+  # The fixture is read into a variable first: bash 3.2's `$( )` scanner cannot
+  # skip heredoc bodies, and this one contains both `#` and `'`.
+  local fixture got want
+  IFS= read -r -d '' fixture <<'FIXTURE' || true
 ## Candidates
 
 - [ ] **#1409** — roster "Pending Since" always reads *Unknown*
@@ -52,7 +54,7 @@ self_test() {
 Prose mentioning #9999 is not a candidate.
 | #1508 already carries v1.2 | table row, not a box |
 FIXTURE
-)
+  got=$(printf '%s' "$fixture" | parse_candidates)
   want=$'unchecked\t1409\nchecked\t1487\nunchecked\t1435\nunchecked\t1520\nchecked\t1441\nchecked\t1134'
   diff <(echo "$got") <(echo "$want") || { echo "FAIL: parse_candidates"; exit 1; }
   test "$(milestone_from_title 'v1.2 — leaders can trust their screen')" = v1.2 || { echo "FAIL: milestone"; exit 1; }
@@ -76,19 +78,28 @@ MS=$(milestone_from_title "$TITLE")
 # Dedupe by number, preferring "checked" (sorts before "unchecked").
 CANDS=$(parse_candidates <<<"$BODY" | sort -t$'\t' -k2,2n -k1,1 | awk -F'\t' '!seen[$2]++')
 
+ERR=$(mktemp "${TMPDIR:-/tmp}/dma.XXXXXX"); trap 'rm -f "$ERR"' EXIT
+
 closed=(); elsewhere=(); notissue=(); nchk=0; nunchk=0
 while IFS=$'\t' read -r st n; do
   [ -n "${n:-}" ] || continue
   if [ "$st" = checked ]; then nchk=$((nchk + 1)); else nunchk=$((nunchk + 1)); fi
-  meta=$(gh api "repos/$REPO/issues/$n" --jq '[.state,(.pull_request!=null),(.milestone.title//"")]|@tsv' 2>/dev/null) \
-    || { notissue+=("- **#$n** ($st) — not an issue in this repo (a discussion, or deleted)"); continue; }
+  # Only a real 404 is a finding. A rate limit, a 5xx or a broken token is a
+  # broken run, and must not be reported as something wrong with the discussion.
+  if ! meta=$(gh api "repos/$REPO/issues/$n" --jq '[.state,(.pull_request!=null),(.milestone.title//"")]|@tsv' 2>"$ERR"); then
+    grep -q 'HTTP 404' "$ERR" || { echo "gh api failed reading issue #$n:" >&2; cat "$ERR" >&2; exit 1; }
+    notissue+=("- **#$n** ($st) — not an issue in this repo (a discussion, or deleted)")
+    continue
+  fi
   IFS=$'\t' read -r state ispr msname <<<"$meta"
   if [ "$ispr" = true ]; then
     notissue+=("- **#$n** ($st) — is a pull request, not an issue")
     continue
   fi
   [ "$state" = open ] || closed+=("- **#$n** ($st) — is **$state**")
-  [ -z "$msname" ] || [ "$msname" = "$MS" ] || elsewhere+=("- **#$n** ($st) — already on milestone **$msname**")
+  if [ -n "$MS" ] && [ -n "$msname" ] && [ "$msname" != "$MS" ]; then
+    elsewhere+=("- **#$n** ($st) — already on milestone **$msname**")
+  fi
 done <<<"$CANDS"
 
 MSJSON=$(gh api "repos/$REPO/milestones?state=all&per_page=100" --paginate \
@@ -97,6 +108,10 @@ MSNUM=$(jq -r '.number // empty' <<<"$MSJSON"); MSSTATE=$(jq -r '.state // empty
 
 unlisted=()
 if [ -n "$MSNUM" ]; then
+  # Assigned, not `< <(…)`: a failed listing must abort the run, not silently
+  # report an empty milestone.
+  ONMS=$(gh api "repos/$REPO/issues?milestone=$MSNUM&state=all&per_page=100" --paginate \
+    --jq '.[] | select(.pull_request == null) | [.number, .state] | @tsv')
   while IFS=$'\t' read -r n state; do
     [ -n "${n:-}" ] || continue
     cut -f2 <<<"$CANDS" | grep -qx "$n" && continue
@@ -105,8 +120,7 @@ if [ -n "$MSNUM" ]; then
     else
       unlisted+=("- **#$n** ($state) — **not mentioned anywhere** in the discussion")
     fi
-  done < <(gh api "repos/$REPO/issues?milestone=$MSNUM&state=all&per_page=100" --paginate \
-    --jq '.[] | select(.pull_request == null) | [.number, .state] | @tsv')
+  done <<<"$ONMS"
 fi
 
 section() { local h=$1; shift; printf '### %s\n\n' "$h"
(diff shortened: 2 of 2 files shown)


scripts/discussion-milestone-audit.sh
```diff
@@ -0,0 +1,127 @@
...
+    }'
+}
+
+# Same convention apply.sh uses in #1602: the milestone is the first vMAJOR.MINOR
+# in the title, so no text in the body can name one.
+milestone_from_title() { grep -oE 'v[0-9]+\.[0-9]+' <<<"$1" | head -1 || true; }
+
+self_test() {
+  local got want
+  got=$(parse_candidates <<'FIXTURE'
```
thpr: **The script does not parse on stock macOS bash (3.2).** Not just `--self-test` — this is a parse error, so it kills every invocation, including the `scripts/discussion-milestone-audit.sh 1598` run the PR body documents:

```
scripts/discussion-milestone-audit.sh: line 42: unexpected EOF while looking for matching `)'
```

Bash 3.2's `$( … )` scanner does not skip heredoc bodies. Line 48 of the fixture — `*(design merged in PR #1515)*` — hits it: inside the rescan, `#` opens a comment that swallows the rest of the line including the `)`, so the paren count never closes. Minimal repro on 3.2.57:

```
FAIL : [- [x] **#1441** — dates *(design merged in PR #1515)*]
FAIL : [it's an apostrophe]
OK   : [| table | row |]
```

CI is unaffected (ubuntu runners are bash 5), which is why this is invisible from the checks. But the repo ships `docs/setup/SETUP_MACOS.md`, so macOS is a supported dev environment, and the whole point of this script over #1602 is that a human runs it locally first.

Fix is to get the heredoc out of the command substitution:

```sh
local fixture got want
IFS= read -r -d '' fixture <<'FIXTURE' || true
…
FIXTURE
got=$(printf '%s' "$fixture" | parse_candidates)
```

thpr (commented):
Review of the script and the workflow. The design holds up — read-only, no secrets, and the parser really does emit nothing but digits, so there is no channel out of the untrusted body. Four findings, all in the mechanics rather than the shape.

One of them is a blocker for the local run the PR body documents: the script does not parse on stock macOS bash. The other three are failure modes that report success or report a finding that isn't one.

Also verified and clean: `NUM` reaches the shell through `env:` rather than `${{ }}`; `${arr[@]+"${arr[@]}"}` is the correct empty-array idiom under `nounset`; the dedupe-preferring-`checked` sort works; `gh api` does not consume the `while read` loop's stdin; and `--paginate | jq -s 'add // []'` is the right shape for the milestone list.

.github/workflows/discussion-milestone-audit.yml
```diff
@@ -0,0 +1,27 @@
...
+  audit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - env:
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+          NUM: ${{ inputs.discussion }}
+        run: |
+          scripts/discussion-milestone-audit.sh "$NUM" | tee -a "$GITHUB_STEP_SUMMARY"
```
thpr: **The job reports green when the script fails.** GitHub's default shell for `run:` on Linux is `bash -e {0}` — `-e` but no `pipefail`. So the pipeline's status is `tee`'s, which is 0 no matter what the script did. Every failure path here exits green with an empty job summary: `no discussion #$NUM in $REPO`, an unauthenticated `gh`, and the bash-3.2 parse error above. The error text goes to stderr, not to `$GITHUB_STEP_SUMMARY`, so the summary a reader opens is simply blank.

```yaml
        shell: bash
        run: |
          set -o pipefail
          scripts/discussion-milestone-audit.sh "$NUM" | tee -a "$GITHUB_STEP_SUMMARY"
```

scripts/discussion-milestone-audit.sh
```diff
@@ -0,0 +1,127 @@
...
+
+# Dedupe by number, preferring "checked" (sorts before "unchecked").
+CANDS=$(parse_candidates <<<"$BODY" | sort -t$'\t' -k2,2n -k1,1 | awk -F'\t' '!seen[$2]++')
+
+closed=(); elsewhere=(); notissue=(); nchk=0; nunchk=0
+while IFS=$'\t' read -r st n; do
+  [ -n "${n:-}" ] || continue
+  if [ "$st" = checked ]; then nchk=$((nchk + 1)); else nunchk=$((nunchk + 1)); fi
+  meta=$(gh api "repos/$REPO/issues/$n" --jq '[.state,(.pull_request!=null),(.milestone.title//"")]|@tsv' 2>/dev/null) \
+    || { notissue+=("- **#$n** ($st) — not an issue in this repo (a discussion, or deleted)"); continue; }
```
thpr: **An API failure is reported as a finding.** `2>/dev/null` plus a blanket `||` collapses every non-zero exit into one confident claim. A 403 rate-limit, a 5xx, or a network blip all come out as "not an issue in this repo (a discussion, or deleted)" — indistinguishable from the real 404 the branch is written for.

Nothing is damaged, since the run is read-only. The cost lands on the human who acts on the report and deletes a candidate box for an issue that exists. Worse, a rate-limit tends to hit many numbers at once, so the section fills up and reads as a systematic problem with the discussion rather than with the run.

Keep the 404 branch, fail loud on anything else — capture stderr instead of discarding it, or check the status via `--include`.

scripts/discussion-milestone-audit.sh
```diff
@@ -0,0 +1,127 @@
...
+  IFS=$'\t' read -r state ispr msname <<<"$meta"
+  if [ "$ispr" = true ]; then
+    notissue+=("- **#$n** ($st) — is a pull request, not an issue")
+    continue
+  fi
+  [ "$state" = open ] || closed+=("- **#$n** ($st) — is **$state**")
+  [ -z "$msname" ] || [ "$msname" = "$MS" ] || elsewhere+=("- **#$n** ($st) — already on milestone **$msname**")
```
thpr: **Contradictory output when the title names no version.** With `MS` empty, the header prints "the title names no `vMAJOR.MINOR`, so nothing below is milestone-scoped" — and then this line flags every candidate that carries *any* milestone as "already on a different milestone", because `$msname` is non-empty and never equals `""`. The section directly contradicts the header two lines above it. (The heading also renders as "On **?** but not a candidate".)

Skip the whole comparison when `MS` is empty.

scripts/discussion-milestone-audit.sh
jee7s: Fixed in b92a26f9 — the fixture heredoc is now read into a variable (`IFS= read -r -d '"' fixture <<'"'FIXTURE'"'`) and piped in: `got=$(printf '%s' "$fixture" | parse_candidates)`. No heredoc sits inside a command substitution anywhere in the script now — `grep -n '<<[^<]'` finds exactly one, at top level of `self_test`; every other `<<` is a here-string, which has no body for the 3.2 scanner to skip. So the offending construct class is gone by construction rather than by test: 3.2 is not installed here (no docker either), so I have not run it on 3.2 and am not claiming I did. `bash -n` and `--self-test` pass on 5.2.

Also made `mktemp` BSD-safe while in there (`mktemp "${TMPDIR:-/tmp}/dma.XXXXXX"` — bare `mktemp` is a usage error on macOS), since the new 404-vs-error branch needs a temp file and macOS is the platform this finding is about.

.github/workflows/discussion-milestone-audit.yml
jee7s: Fixed in b92a26f9, your snippet plus one character: `shell: bash` and `set -euo pipefail`, and the script's stderr is folded into the pipe with `2>&1`. That way the failure is red *and* the summary a reader opens carries the error text instead of being blank — `no discussion #$NUM in $REPO`, an unauthenticated `gh`, and the API errors from the third finding all land in the summary now.

scripts/discussion-milestone-audit.sh
jee7s: Fixed in b92a26f9 — stderr goes to a temp file instead of `/dev/null`, and only `HTTP 404` keeps the finding. Anything else prints `gh api failed reading issue #N:` with the API error and exits 1, so a rate limit stops the run instead of filling the section.

Same hole existed one block down: the milestone-issue listing ran in `< <(gh api …)`, where a failure is invisible and the report just says the milestone holds nothing. That is now an assignment, so `set -e` aborts.

Verified both branches end-to-end with a throwaway `gh` shim on `PATH`: a 404 still yields the "not an issue" finding and exit 0; a `(HTTP 403)` rate limit prints the error and exits 1 with no report.

scripts/discussion-milestone-audit.sh
jee7s: Fixed in b92a26f9 — the comparison is guarded (`[ -n "$MS" ] && [ -n "$msname" ] && [ "$msname" != "$MS" ]`) and both milestone-scoped sections are skipped entirely when `MS` is empty, which also removes the `On **?**` heading. The header now says so out loud: "the title names no `vMAJOR.MINOR`, so the two milestone-scoped sections are skipped."

Checked with the shim: a title with no version and a candidate carrying `v9.9` now reports the candidate sections only, and flags nothing.

jee7s:
All four addressed in b92a26f9. Nothing about the shape changed — still one script, one workflow, no new dependencies; the diff is +30/−12.

| # | Finding | Fix |
|---|---|---|
| 1 | bash 3.2 parse error | Fixture read into a variable, then `printf … \| parse_candidates`. No heredoc inside a `$( )` anywhere in the file now. |
| 2 | Job green when the script fails | `shell: bash` + `set -euo pipefail`, and `2>&1` so the summary shows the error rather than nothing. |
| 3 | API failure reported as a finding | Only `HTTP 404` is a finding; every other `gh api` exit prints the error and exits 1. Milestone listing moved off `< <(…)` for the same reason. |
| 4 | Contradictory output with no version in the title | Both milestone-scoped sections skipped when `MS` is empty; header states it. |

On #1 specifically: 3.2 is not installed on this machine and there is no container runtime available, so I have **not** run it under 3.2 and am not claiming a passing test there. What I can claim is stronger than a spot test — the construct class the bug needs is gone: `grep -n '<<[^<]'` finds a single heredoc in the file, at top level of `self_test`, and every other `<<` is a here-string with no body for the scanner to skip. Worth a 30-second check on your Mac before you take my word for it.

Also fixed in passing, since it lands on the same platform: bare `mktemp` is a usage error on BSD, so the new temp file uses an explicit template.

## Re-runs

`--self-test` → `OK: 3 checks`.

The real read-only run against #1598 is byte-identical to the sample in the PR body — same two findings, unchanged:

```markdown
### Candidates that are closed

- **#1435** (unchecked) — is **closed**

### On **v1.2** but not a candidate

- **#1508** (open) — mentioned in the body's prose, but not a candidate box
- **#1154** (open) — **not mentioned anywhere** in the discussion
```

#1599 also unchanged: `v1.3` does not exist yet, 10 unchecked candidates, nothing stale.

The new branches have no live case to exercise — #1598 has no 404 candidate and its title names a version — so I checked them with a throwaway `gh` shim on `PATH` rather than shipping fixtures for them: 404 → finding, exit 0; `(HTTP 403)` → error on stderr, exit 1, no report; no version in the title + a candidate on `v9.9` → not flagged, milestone sections absent. The shim stayed in the scratchpad; the script keeps its one `--self-test`.

thpr (approved):
All four findings fixed in b92a26f9, each verified rather than taken on the commit message.

- **bash 3.2 parse** — fixture read into a variable, no heredoc inside `$( )`. On stock macOS bash 3.2.57 the script now parses and `--self-test` prints `OK: 3 checks`.
- **Green-on-failure** — `shell: bash` + `pipefail` + `2>&1` into the summary. (The `bash` keyword already implies `-eo pipefail`, so the explicit `set` is belt-and-braces, not a problem.)
- **API error vs finding** — only `HTTP 404` becomes a `notissue` entry now. Confirmed `gh` discriminates on stderr: `gh: Not Found (HTTP 404)` against `gh: Bad credentials (HTTP 401)`.
- **Empty `MS`** — both milestone-scoped sections gated, header reworded. The contradiction is gone.

The milestone-issue listing moving off `< <(…)` onto an assigned `$( )` is a real fix I had missed: the process substitution's exit status was invisible to `set -e`, so a failed listing reported an empty milestone as a clean result.

A live read-only run against #1598 is byte-identical to the sample in the PR body, so none of this shifted the output.

Approving the code. That is not a vote on the build-at-all question this sits next to — #1602 versus this shape is still the maintainer's call, as the PR body says.

PR #1648, opened 2026-08-15
fix(auth): kiosk attendance polls rejected as replays (double signature verification)

jee7s:
## Problem

The physical kiosk cannot load the attendance roster: every signed `GET /api/attendance` poll returns `401 {"error":"Replay detected"}`, even with a fresh nonce. Verified live against prod from the kiosk Pi.

`getOptionalSessionUser` — added to the route for the denied-household gate — resolves through `resolveAuthResult`, which tries kiosk-signature verification first. That verification **consumes the single-use nonce**, then the kiosk result is discarded (the helper returns only session users). The route then runs its own `verifyKioskSignature` on the same request and the seen-nonce check fires: every kiosk poll is a "replay" of itself.

## Fix

Session-only resolution for `getOptionalSessionUser`: it now skips kiosk verification entirely. Verifying there served no purpose (kiosk results were discarded), and burning the nonce broke the one route that legitimately re-verifies. The `withAuth`/`withKiosk` chokepoint path is unchanged.

## Why CI never caught it

`attendance.integration.test.ts` mocks `@/lib/verify-kiosk`, so the mock had no nonce state. The new regression test (`attendanceKioskAuth.test.ts`) uses the real verify module with real Ed25519 signatures, mirroring `scanAuth.integration.test.ts`: a single signed poll must authenticate in one pass, an actual same-nonce replay must still 401, and a bad signature must still 401. Confirmed the first two fail without the fix.

Ran the kiosk/auth/attendance/scan unit suites (27 suites, 689 tests, green) and `tsc --noEmit`.

## Context

Found while recovering the facility kiosk: the Pi was still on the pre-monorepo `checkin-client` pointed at the decommissioned nip.io backend. After migrating it to the monorepo client and current prod URL, this server-side bug was the remaining blocker — the display stays down until this deploys.

### DIFF
--- a/checkin-app/src/app/__tests__/attendanceKioskAuth.test.ts
+++ b/checkin-app/src/app/__tests__/attendanceKioskAuth.test.ts
@@ -1,0 +1,99 @@
+/**
+ * GET /api/attendance — kiosk auth with the REAL verify-kiosk module.
+ *
+ * The route resolves an optional session (denied-household gate) and then runs
+ * its own kiosk signature verification. Session resolution must not verify the
+ * kiosk signature itself: verification consumes the single-use nonce, so a
+ * request verified twice fails its second check as a replay. These tests pin
+ * that a kiosk poll authenticates in one pass — attendance.integration.test.ts
+ * can't see this because it mocks @/lib/verify-kiosk, which has no nonce state.
+ */
+import crypto from 'crypto';
+import { GET } from '@/app/api/attendance/route';
+import { getServerSession } from 'next-auth/next';
+
+jest.mock('next-auth/next', () => ({
+    getServerSession: jest.fn(),
+}));
+
+jest.mock('@/lib/getFullAttendance', () => ({
+    getFullAttendance: jest.fn().mockResolvedValue({
+        attendance: [],
+        counts: { total: 0 },
+        safety: {},
+    }),
+}));
+
+function makeKeypair() {
+    const { publicKey, privateKey } = crypto.generateKeyPairSync('ed25519');
+    const rawPublicHex = (publicKey.export({ type: 'spki', format: 'der' }) as Buffer)
+        .subarray(-32)
+        .toString('hex');
+    // Mirrors verify-kiosk.ts — nonce is bound INTO the signed payload.
+    const sign = (ts: string, nonce: string, method: string, path: string, body: string) =>
+        crypto.sign(null, Buffer.from(`${ts}:${nonce}:${method}:${path}:${body}`), privateKey).toString('hex');
+    return { rawPublicHex, sign };
+}
+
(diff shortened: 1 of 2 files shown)


thpr (approved):
Approving. Traced the diff against `verify-kiosk.ts`, both `getOptionalSessionUser` callers, the attendance route, and `client/client.py`.

**The diagnosis holds.** `verifyKioskSignature` records the nonce *after* the signature check (verify-kiosk.ts:127-134), so pre-fix the sequence was: `getOptionalSessionUser` → `resolveAuthResult` verifies and burns the nonce → discards the kiosk result → the route's own `verifyKioskSignature` sees the nonce already in the cache → `401 Replay detected`. Every poll, deterministically. Fixing it in the shared helper rather than in the route is the right level: the hazard belongs to "any caller that resolves auth and then re-verifies", not to this one route.

**Behaviour deltas I checked, all safe:**

- Kiosk-signed with no cookie (the actual Pi — its `requests.Session` never acquires a NextAuth cookie): unchanged, still resolves `isKiosk`.
- Kiosk-signed **with** a valid session cookie: now resolves as a session instead of kiosk, so the route takes the role-based branch. That's stricter, not looser — a captured signed request replayed inside the 60s window with a member's own cookie now gets the member view rather than the full roster.
- Local keyless path (`pubKeys` empty + `isLocal()`): `getOptionalSessionUser` returned `undefined` before and returns `undefined` now; the route's own keyless branch still sets `isKiosk`. No change.
- Staging gate: a kiosk result carried no org/flag claims either way, so it was already downgraded to `unauthenticated`. Both `getOptionalSessionUser` callers keep their `isStaging() && !user` early return, so routeAuthDrift rule 4 stays satisfied.
- `sessionOnly` on the exported chokepoint is fail-closed by construction — skipping kiosk verification can only downgrade a caller to anonymous, never grant anything.

`/api/attendance` is the only route calling `verifyKioskSignature` outside `lib/auth.ts`, so no sibling route has this shape.

**One adjacent observation, non-blocking and not a defect today:** `src/app/api/programs/[id]/route.ts:33` calls `authenticateRequest(req)` a second time after `handler()` has already resolved auth — the same double-verify shape. It doesn't misbehave, because a failed second verification just falls through to anonymous and that code only cares about `type === 'session'`. But it does burn a nonce it has no use for, and now that `sessionOnly` exists, passing it there would make the call say what it means. Fine as a follow-up.

**On the test:** genuine coverage, not a rubber stamp — real Ed25519, real verify module, fresh nonce per case, and cases 1 and 2 both fail without the fix. Only nit: the bad-signature case asserts status `401` without asserting the error string, so it would pass on a 401 arising for some other reason. Worth an `expect(json.error).toBe('Invalid signature')` if you touch the file again; not worth a push on its own.

PR #1649, opened 2026-08-15
test: close member-pricing coverage gaps (discount amount, fail-open horizon, purchase flow)

jee7s:
Test-only PR closing three gaps found by a member-pricing coverage audit. No product code changes; one additive seed persona.

## What was uncovered

1. **The program discount amount was never asserted.** `programDiscountCodeAPI.integration.test.ts` minted codes for a 4000/5000-cent program but never checked that `nonOrgMemberPriceCents − orgMemberPriceCents` actually reaches `mintMemberDiscountCode` — a sign flip or nulled price would have shipped silently (failed discounts deliberately degrade to undiscounted). Now asserted via a call-through `jest.mock` (SWC compiles exports non-configurable, so `jest.spyOn` cannot attach), including not-called assertions on both `{code:null}` paths.

2. **`isDuesSettledThrough`'s fail-open was unpinned on the pricing path.** `coversThrough` deliberately returns true when no `orgMembershipYearBoundary` is configured (a settings gap must not strip every member of member pricing). That was tested for `isActiveOrgMemberThrough` but not for the dues path the discount-code route actually uses. Now characterized.

3. **No flow test bought a program at member price.** New `flow-tests/program-member-pricing.flow.test.ts` drives the whole journey over HTTP: board creates a priced program → opens enrollment (programs default `enrollmentStatus: CLOSED` — a rough edge worth its own backlog note: every board-created program starts unenrollable until a separate settings PATCH) → seeded member household gets a `PRG{id}-` code, non-member gets `null` → member self-enrolls PENDING → dev orders-paid webhook activates. Requires one additive seed persona (`member.pricing@example.com`, ACTIVE OrgMembership, adult lead); Family/Family2/Family3 untouched. The program's `endAt` is deliberately near-term so the member-code assertion doesn't silently depend on the seed lacking a configured membership-year boundary.

## Verification

- Unit: orgMembership suite 19/19.
- Integration: programDiscountCodeAPI 4/4 against real Postgres.
- Flow: full suite (5 suites / 6 tests) green on a fresh stack — confirms the seed addition doesn't disturb sibling flow tests; the new test re-verified green after the endAt hardening.

### DIFF
--- a/checkin-app/flow-tests/program-member-pricing.flow.test.ts
+++ b/checkin-app/flow-tests/program-member-pricing.flow.test.ts
@@ -1,0 +1,101 @@
+/**
+ * @jest-environment node
+ */
+/**
+ * Flow test for the single-pool member-pricing journey: board creates a priced
+ * program, a member and a non-member each probe the discount-code route, the
+ * member self-enrolls, and paying (via the dev Shopify orders/paid stand-in)
+ * activates their enrollment. Exercises program creation → discount-code →
+ * enroll → pay as a whole HTTP journey, wired through real auth, routes, and DB.
+ *
+ * Personas: boardmember@example.com (board, creates the program),
+ * member.pricing@example.com (dues-settled household lead — ACTIVE OrgMembership
+ * from the seed), and certified.adult@example.com (solo household, no
+ * OrgMembership row — a genuine non-member, used read-only here).
+ */
+import { loginAs, api } from "./helpers";
+
+type PersonaList = { personas: { id: number; email: string }[] };
+type ProgramCreate = { success: boolean; program: { id: number; shopifyVariantId: string | null } };
+type DiscountCode = { code: string | null; reason?: string };
+type Enrollment = { success: boolean; enrollment: { status: string } };
+type OrdersPaid = { ok: boolean; participants: { personId: number; status: string }[] };
+
+async function personaId(list: PersonaList["personas"], email: string): Promise<number> {
+    const persona = list.find((p) => p.email === email);
+    if (!persona) throw new Error(`dev persona not found in /api/auth/dev-personas: ${email}`);
+    return persona.id;
+}
+
+describe("flow: program purchase with member pricing", () => {
+    it("board creates a priced program; member gets a discount code and settles by payment; non-member gets no code", async () => {
+        const board = await loginAs("boardmember@example.com");
+        const member = await loginAs("member.pricing@example.com");
+        const nonMember = await loginAs("certified.adult@example.com");
+
+        const personas = await api<PersonaList>(board, "/api/auth/dev-personas");
+        expect(personas.status).toBe(200);
+        const boardId = await personaId(personas.json.personas, "boardmember@example.com");
+        const memberId = await personaId(personas.json.personas, "member.pricing@example.com");
+
+        const tag = Date.now().toString(36);
+        // Keep the program short: member pricing requires the membership to cover
+        // the program's END date, and the seeded member holds no renewal. A
+        // near-term endAt keeps the code assertion valid even if the seed ever
+        // gains a configured orgMembershipYearBoundary just ahead.
+        const startAt = new Date();
+        const endAt = new Date(startAt.getTime() + 3 * 24 * 60 * 60 * 1000);
+
+        const created = await api<ProgramCreate>(board, "/api/programs", {
+            method: "POST",
+            body: JSON.stringify({
+                name: `Member Pricing Program ${tag}`,
+                leadMentorId: boardId,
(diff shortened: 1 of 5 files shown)


checkin-app/flow-tests/program-member-pricing.flow.test.ts
```diff
@@ -0,0 +1,92 @@
...
+        const nonMember = await loginAs("certified.adult@example.com");
+
+        const boardId = board.personaId;
+        const memberId = member.personaId;
+
+        const tag = Date.now().toString(36);
+        // Keep the program short: member pricing requires the membership to cover
+        // the program's END date, and the seeded member holds no renewal. A
+        // near-term endAt keeps the code assertion valid even if the seed ever
+        // gains a configured orgMembershipYearBoundary just ahead.
```
thpr: The premise here is inverted, and the property it claims doesn't hold today in either direction.

**The coverage rule is never exercised in the flow environment.** `deploy/docker-compose.flow.yml` brings the DB up with `npx prisma db push`, which does not run migrations — so `20260811130000_board_settings_year_boundary_default` never fires there. And `seedBaseline` writes no `BoardSettings` row at all. `membershipValidThrough` therefore returns `null` at its `if (!settings?.orgMembershipYearBoundary) return null`, `coversThrough` takes its deliberate fail-open path, and the member-vs-non-member split in this test rests entirely on `isDuesSettled` (status). The `endAt` value is doing nothing.

**And if a boundary ever does reach that DB, three days is the wrong margin.** The failure case is exactly the one the comment names — a boundary sitting *just ahead*. `endAt = now + 3d` crosses it whenever the run date is within three days of the boundary, `coversThrough` returns false, the route answers `{ code: null, reason: "membership_ends_before_program" }`, and the `stringContaining('PRG…')` assertion fails. With the documented 1 September boundary now on `main`, that is roughly 29–31 August each year, and the seeded `member.pricing` household holds no renewal `OrgMembershipProcess` to push its horizon a boundary further.

Either direction fixes it, and both are one line:

- Make the program **dateless** (omit `startAt`/`endAt`). `programCoverageDate` returns `null`, `coversThrough` short-circuits on `through === null`, and the test is then honest that it covers the status half of member pricing, not the duration half.
- Or seed `orgMembershipYearBoundary` plus a settled renewal for the member, so the duration half is genuinely asserted rather than fail-open.

Not blocking — the enroll-and-pay journey this adds is real coverage regardless. It's the comment asserting a safety margin that doesn't exist that's worth correcting, since the next reader will trust it.

---

thpr (commented):
Checked the three gaps against the routes and the seed. The mock strategy is right (SWC really does compile these exports non-configurable, so `jest.spyOn` can't attach), the asserted amount matches `nonOrgMemberPriceCents - orgMemberPriceCents` in the route, and the seed addition is safe — `OrgMembership.householdId` is `@unique` so the upsert is idempotent, and the baseline persona-count assertion is a `toBeGreaterThanOrEqual`, so it doesn't trip.

One finding, inline: the near-term `endAt` doesn't harden what the comment says it hardens, and the coverage rule the flow test appears to exercise is switched off in the flow environment.

checkin-app/flow-tests/program-member-pricing.flow.test.ts
jee7s: Confirmed all three points and took the dateless option (c86dc29) — `db push` runs no migrations and `seedBaseline` writes no `BoardSettings` row, so `coversThrough` was fail-open here and the `endAt` was inert; the margin also pointed the wrong way into the 1 Sep boundary.

Dropping `startAt`/`endAt` makes `programCoverageDate` return `null` and the short-circuit explicit, so the test now says plainly that it covers the status half; the duration half stays asserted in `programDiscountCodeAPI.integration.test.ts`. Flow suite is CI-only here (no Docker on this box) — typecheck, lint, and `orgMembership.test.ts` (19 passed) are green locally.

---

thpr (approved):
Addressed in c86dc296 — the dateless option, which is the honest one.

Re-checked the downstream: `parseDateOnly(undefined)` returns null and the create route has no required-date guard (only name / leadMentorId / maxParticipants), so the program lands with both dates null. `programCoverageDate` then returns null, `coversThrough` short-circuits on `through === null`, and the member/non-member split rests on `isDuesSettled` — which is what the new comment now says it rests on. Nothing else in the journey reads the dates: enrolment gates on `enrollmentStatus`, and `checkProgramAge` already handles the dateless case.

The pointer to the integration test holds up too — its `beforeAll` upserts `orgMembershipYearBoundary` and its `afterAll` restores it, so the duration half is genuinely exercised where a boundary exists, rather than fail-open.

---

Issue #270, opened 2026-06-13
shopify member segment


dkaygithub:
make a member segment to be allowlist of folks that can use membervariants for program payment

dkaygithub:
and prevent them from using non-member

dkaygithub:
@jee7s did you already do this with the re-work of variants for programs?

jee7s:
It's recorded in the docs but is not feature complete

jee7s:
**HOLD — duplicate.** This is a one-line restatement of a hole that is specified in full elsewhere.

1. Should this close as a duplicate of #278? `docs/backlog/INDEX.md` CO1 lists #270 and #278 on a single row, and the merged design `checkin-app/docs/designs/SHOPIFY_MEMBER_SEGMENT_PRICING.md` (#929) opens by naming both as one problem: "Both honor-system holes (#270 program tier, #278 volunteer discount) share one" root cause.
2. If it stays open, what does it cover that #278 does not — specifically the program-tier member variant rather than the volunteer discount code?

The comment thread already reaches this conclusion informally ("It's recorded in the docs but is not feature complete"). Closing it and tracking the work under #278 + #929 loses nothing.

---
_Posted by Claude._

PR #1650, opened 2026-08-15
feat(finance): reconcile unentitled program member-discount (PRG) code use

jee7s:
## What this is, in plain terms

Programs can sell a **member price** through a Shopify discount code (`PRG<program>-…`), minted when a dues-paid member heads to checkout. The code is **not locked to the person** — anyone holding the checkout link can redeem it for 48 hours. Misuse of the *membership* discount code already gets caught after the fact and flagged to the board; misuse of the *program* member code was never caught at all: a household that never paid dues could take the member price silently.

**This PR closes that gap.** At the moment an order is paid (the Shopify webhook, or the nightly reconciler recovering a missed webhook), we ask one question: *does this order use this program's member code, and is any person on the order actually enrolled in this program from a dues-settled household?* If no — the enrollment is **not activated** and a `DISCOUNT_UNAUTHORIZED` exception lands on the board's payment-exceptions list, exactly like the existing misuse flags.

Three judgment calls, all shaped by review:
- **We judge once, at payment time — never retroactively.** An earlier draft re-checked old orders nightly; that meant an ordinary household deactivation would spray flags across every past purchase. Removed.
- **We don't trust the cart's claim of who's on the order.** Those ids are customer-editable, so the check only counts people genuinely enrolled in this program (thpr's review caught this).
- **The 48-hour code lifetime is an accepted gap, not a bug**: a code minted while entitled can be redeemed up to 48h after the household lapses. We name it rather than build machinery for it.

Why now: interim enforcement for the discount honor-system gap (#270/#278) until segment pricing lands.

---

Interim enforcement for the program half of the discount honor-system gap (#270, #278 — plain references, both stay open for the segment-pricing end-state).

## Problem

A minted `PRG{programId}-XXXXXXXX` member discount code is single-use but not identity-bound (`customerSelection: {all: true}`): anyone holding the checkout URL can redeem it within its 48h window. The volunteer *membership* code has post-hoc enforcement (`usesVolunteerCodeUnentitled` → `DISCOUNT_UNAUTHORIZED`); the *program* member code had none — an unentitled redemption was never caught, before or after activation.

## Approach

Same policy the reconciler already applies to the volunteer code, per `docs/rules/finance-payments.md` ("the store owns what a discount comes to; the app owns who may use one and until when") — but **judged at the money event, once, and never re-judged**. Entitlement is a fact about the moment money moved; re-deriving it nightly from live household state would let the answer change under the household long after the sale (see the review thread).

- **The judgement** lives in `lib/programs/memberDiscountCode.ts`: the code-format prefix (shared with the minter so encoder and decoder can't drift), `usesProgramMemberCode` (prefix-matched on the program's OWN id; the trailing hyphen prevents `PRG4-`/`PRG42-` collisions — unit-tested both directions), and `unentitledMemberCodeUse(programId, personIds, discountCodes)` — true iff the order carries this program's code and NO enrollee's household is dues-settled through the program's coverage date (`isDuesSettledThrough` + `programCoverageDate`). `personIds` can arrive straight off the customer-controlled `CheckMeIn_Account_ID` attribute, so the helper **intersects them with this program's real `ProgramParticipant` rows first** — otherwise stuffing any dues-settled person's id into the cart permalink would buy the member price for an unentitled enrollee. An empty intersection flags rather than clears.
- **orders/paid webhook** (`api/webhooks/shopify/route.ts`): asks it seconds after checkout, right after the existing variant-id item gate, and fails the same way as `NO_ITEM` — raise `DISCOUNT_UNAUTHORIZED` for the board, do **not** activate. The route's local `ShopifyOrder` interface now declares `discount_codes`, the same one-line move `line_items` already got. This is the path that matters: webhook-activated orders never reach the reconciler's forward pass.
- **Forward pass** (`lib/finance/reconcile.ts`): the same helper, asked when recovering a MISSED webhook — that recovery *is* the money event for those orders. A recovered PENDING order carrying the code with no entitled enrollee raises and does not activate, mirroring the membership forward branch.
- **The reversal pass does NOT re-ask it.** It classifies money-out only (refund / chargeback / cancel / edited-down). It iterates every `ACTIVE` enrollment with no bound on time, so re-deriving entitlement there meant an ordinary membership-ops deactivation raised a fresh flag on every past member-priced order that household held, with a reopen loop against any resolution the board attempted.
- **Deliberately inherits the fail-open** when no membership-year boundary is configured — an entitled member with a settings gap is never falsely flagged.
- **No schema migration and no snapshot column**: the `PaymentException` row already IS the durable judgement — `(kind, shopifyOrderId)` is unique, so a row means "this order was judged unentitled" and no row means "judged clean". Idempotency rides that constraint + the `raisePaymentException` pre-check.

This deliberately sidesteps #278's still-deferred reject-at-checkout-vs-flag decision by following the already-shipped flag-for-the-board pattern. `docs/backlog/INDEX.md` CO1 row notes the interim enforcement.

### Accepted window (named, not engineered away)

Minted codes live 48h and are not identity-bound, so entitlement can move between mint and checkout. We judge at checkout: a code minted while entitled but redeemed after the household lapses is flagged at redemption. That is the intended answer — payment is when the money moved — not a gap to close with machinery.

## Verification

- Unit: full default suite green (280 suites / 2754 tests) on the reworked branch. `memberDiscountCode` 13/13 covers the prefix-collision guards and the entitlement judgement itself (flag when nobody is settled, clear when someone is, never judge an order without the code, coverage date rather than today, status-only when the program has no dates).
- Integration (real Postgres, not in the default jest run): reconcile scope pins that a **deactivated household produces NO `DISCOUNT_UNAUTHORIZED` on any number of runs**, that a real REFUND on a member-priced enrollment still classifies, that the entitled forward pass activates, and that the unentitled forward pass raises without activating; `webhookShopify` pins both halves at the money event (unentitled → flagged + still PENDING + idempotent across redelivery; dues-settled → activates, no exception).
- `tsc --noEmit` clean; eslint clean.
- `origin/main` merged in (#1641's `MAX_DATE` consolidation): `orgMembership.ts` is now byte-identical to main again — the batched helper this PR had added there went away with the sweep, so the overlap is zero.

Known pre-existing limitation (unchanged): the `(kind, shopifyOrderId)` unique key means one flagged order yields one exception row even if it covers multiple enrollees; `personId` carries the first.


thpr (commented):
The forward-pass check reads correctly and the `PRG{id}-` prefix guard is right. The reversal audit is the part I'd change, and not by tightening it: entitlement is a fact about the moment money moved, but the audit re-derives it from live household state on every daily run, so its answer changes under the household long after the sale.

The concrete failure. A board member deactivates a household through the ordinary membership-ops toggle (`src/app/api/membership-ops/households/route.ts:326`), which sets `ACTIVE` → `REVOKED`. At 09:15 the next morning, every past member-priced program order that household holds raises `DISCOUNT_UNAUTHORIZED`. Those purchases were honest when they were made; the board now sees a queue of fraud-shaped flags against a family whose only act was leaving.

Three properties combine to produce that. The sweep at `reconcile.ts:558` iterates every `ProgramParticipant` with `status: ACTIVE` and an order id, with no bound on time. `ProgramParticipantStatus` is `PENDING | ACTIVE` only, so nothing ever retires a finished enrollment; a 2026 summer program is still in that loop in 2028. And `isDuesSettledThroughMany` answers from `DUES_SETTLED_PERSON_WHERE`, which reads the household's current `OrgMembership.status` and knows nothing about the order. The coverage half is safe, because `membershipValidThrough` computes from `now` and so only moves forward; it's the status half that has no fail-open.

Nothing clears the flag either. `raisePaymentException` reopens a RESOLVED row whenever the condition re-detects (`reconcile.ts:100`), and the only condition that stops re-detection is the household becoming dues-settled again. A board member resolves the exception, and the next run reopens it and puts it back in the red-dot count (`src/app/api/nav/todo-counts/route.ts:422`). The volunteer analogue has an escape, since someone can flip `isVolunteer`; this one has none, and the PR's own policy is that the enrollment stays ACTIVE, so there's no state the board can move the enrollment to instead.

The same live derivation is arbitrary in the other direction, though more narrowly: a household that redeems the code while unentitled and then settles its dues before the next morning's run is never flagged at all. Once a row exists it stays, so this is a one-day window rather than a disappearing flag, but the window exists only because the judgment happens on cron time instead of at payment.

A snapshot column would fix the semantics, and it isn't needed, because the `PaymentException` row already is the durable record: `(kind, shopifyOrderId)` is unique, so a row means "this order was judged unentitled" and no row means "judged clean." Reconstructing the history instead of stamping it isn't available here. `membershipValidThrough(householdId, now)` is already time-parameterized, so the coverage half could be asked as of the order date, but `OrgMembership.status` keeps no history, and the AuditLog can't stand in for one: both branches of the households route write their audit row only `if (auth.type === 'session')`, so a token-actor change leaves no trace at all.

What I'd suggest instead is moving the judgment to the money event. `activateProgramEnrollment` is already the choke point both callers share (`src/lib/programs/activateEnrollment.ts:37`). The stated reason for putting the check in the audit is that webhook-activated orders never reach the forward pass, which is true; the answer to it is to check in the webhook, where entitlement is evaluated seconds after checkout, rather than to re-judge the whole org nightly. The webhook has what it needs: an `orders/paid` payload carries `discount_codes`, and the route's local `ShopifyOrder` interface simply doesn't declare the field yet (`src/app/api/webhooks/shopify/route.ts:11`), so extending it is the same one-line move `line_items` already got, and computing the predicate in the caller matches the convention `hasProgramItem` sets.

That shrinks the diff rather than growing it. Keep `usesProgramMemberCode` and the forward-pass branch as written, add the equivalent at the webhook, and delete the reversal-audit branch along with the `programById` / `codeCandidates` / `entitledByEnroll` batching that exists to serve it. `isDuesSettledThroughMany` and the `coverageReaches` extraction can probably go with it, since one order enrolls one household and the per-pair form in a short loop covers that; if it turns out worth keeping, it stands on its own merits rather than on the sweep's. No migration either way, and the retro-flagging and the reopen loop stop existing instead of being bounded.

One residual gap is worth naming rather than engineering away: codes live 48 hours, so a household that mints a code, is revoked, and then checks out is flagged at activation. That's the right answer, since payment is when the money moved.

---
🤖 Posted by [Claude Code](https://claude.com/claude-code) on @thpr's behalf

jee7s:
Taken as written — the judgement moved to the money event and the sweep is gone, not bounded.

**What moved where.** `lib/programs/memberDiscountCode.ts` now owns the whole judgement: the format prefix, `usesProgramMemberCode` (over a plain `string[]`, so the mirror's `discountCodes` and the webhook's `discount_codes` both feed it), and `unentitledMemberCodeUse(programId, personIds, discountCodes)` — carries this program's code AND no enrollee's household dues-settled through the coverage date. The `orders/paid` webhook asks it right after the variant-id item gate and fails the way `NO_ITEM` does: raise for the board, do not activate. `ShopifyOrder` declaring `discount_codes` was indeed the same one-line move `line_items` already got. The reconciler's forward pass calls the same helper — that recovery is the money event for a missed webhook.

**Deleted, per your read.** The reversal-audit branch, the `programById` / `codeCandidates` / `entitledByEnroll` batching, and `isDuesSettledThroughMany` + `coverageReaches` with them. `orgMembership.ts` is byte-identical to main again. The enrollment loop in `reconcileReversals` classifies money-out only.

**Your scenario now.** ACTIVE enrollment, member-priced order carrying the PRG code, household flipped `ACTIVE` → `REVOKED` through membership ops: zero `DISCOUNT_UNAUTHORIZED`, on the first run and on every run after — pinned as an integration test, with a sibling asserting a real REFUND on the same shape still classifies, so the delete didn't take the reversal path with it. The reopen loop and the red-dot churn stop existing rather than getting a time bound. The one-day hole in the other direction closes for the same reason: judgement happens at payment, not on cron time.

You were right that no snapshot column is needed — the `PaymentException` row is the record. Row means judged unentitled, no row means judged clean.

**The 48h window, named not engineered.** Codes live 48 hours and aren't identity-bound, so entitlement can move between mint and checkout. We judge at checkout, so a code minted while entitled and redeemed after the household lapses is flagged at redemption. That's the intended answer, not a gap — it's four lines in `unentitledMemberCodeUse`'s docblock and a line in the PR body, no machinery.

**Main merged in.** #1641 landed on `orgMembership.ts` too. No textual conflict, and the semantic overlap checks out: it consolidated `new Date(8.64e15)` into `MAX_DATE` and pointed `membershipValidThrough` at it — same value, no behavior change — and after this rework the file has no diff against main at all. Verified at the merge commit before touching anything (tsc clean, 51 tests in the orgMembership/renewal/reconcile scope), and again after: full default suite 280 suites / 2754 tests green, tsc and eslint clean. The integration suites (reconcile, webhookShopify) are updated but not runnable in this environment — no Docker for Postgres — so those are unexecuted here.

---

thpr (commented):
The rework does what it says. Entitlement is asked once at the money event, the reversal sweep classifies money-out only, `orgMembership.ts` is byte-identical to `main` again, and nothing in `src/` still references the removed exports. One problem came in with the new webhook branch.

The webhook judges entitlement over ids the customer supplies. `participantIds` is parsed from the `CheckMeIn_Account_ID` cart attribute at `src/app/api/webhooks/shopify/route.ts:153`, and the guard comment a few lines below already says those attributes are entirely customer-controlled. That list is then handed to `unentitledMemberCodeUse` at line 182, which returns `false` as soon as any one id in it is dues-settled, without checking that the id belongs to an enrollee of this program.

So the check comes apart with one extra number in a URL. A non-member enrolls through the ordinary flow, which creates a PENDING participant and takes no payment; they obtain a shared `PRG{id}-` code, then build the cart permalink with `attributes[CheckMeIn_Account_ID]=<their own id>,<any dues-settled person's id>`. Person ids are autoincrement, so a usable one is easy to guess. The entitlement check sees the settled household and passes. `activateProgramEnrollment` then ignores the stuffed id, since it has no PENDING row in this program, and activates the attacker's own enrollment at the member price with no flag raised. That is the exact redemption this PR exists to catch, on the path that carries live checkouts.

The reconciler's forward pass doesn't have the problem, which is what makes it easy to miss in the diff: it resolves the attribute ids to real PENDING rows before judging (`src/lib/finance/reconcile.ts:375`) and passes only those to the helper. The two callers look symmetric and aren't.

One query fixes it, and it belongs inside `unentitledMemberCodeUse` so both callers inherit it: intersect `personIds` with the program's actual `ProgramParticipant` rows before asking whether any household is dues-settled. That matches what `hasProgramItem` already establishes on this route, which is to believe the database about what the order is for and never the payload.

---
🤖 Posted by [Claude Code](https://claude.com/claude-code) on @thpr's behalf

jee7s:
Real hole, and the asymmetry is exactly why I missed it — the forward pass resolves attribute ids to PENDING rows before calling, the webhook didn't, and the shared helper trusted whatever it was handed.

Fixed where you said: one `ProgramParticipant` query inside `unentitledMemberCodeUse` intersects `personIds` with this program's real enrollees before any household is asked about dues, so both callers inherit it and the webhook stops believing the payload. An empty intersection now flags rather than clears — an id that isn't enrolled can't be activated by the order either, so nobody paying was allowed the price.

Pinned in both places: a unit test for the permalink-stuffing case (own PENDING id + a foreign settled id → still flagged, and entitlement asked only about the enrolled id) plus the empty-intersection case, and an integration test driving the same attack through the webhook route. Both new unit tests fail against the pre-fix helper — I reverted the intersect to confirm they actually catch it. Entitled path stays green; `tsc` and the default suite clean.

---

jee7s:
Correction to the note above, since it matters for a security path: the suite I ran clean locally was missing one — the `LIVE_PERSON` drift guard failed CI on that commit, because the new `ProgramParticipant` lookup is a class-3 site (join delegate carrying a `person` relation).

Fixed in `29e1fba0` by filtering rather than allowlisting — `where: { ..., person: LIVE_PERSON }`. A merge leaves the tombstone's enrollment row behind, so without it a merged-away person could lend their old household's dues to someone else's checkout, which is the same rule this fix is built on. All checks green on the final head.

---

dkaygithub:
Feedback: I cannot understand from any PR description or comment what is going on here or why. It has something to do with program discounts sure.

jee7s:
@dkaygithub fair — the body assumed you'd followed the review thread. Rewrote it with a plain-language section up top: what a PRG code is, the gap (program member-price codes had no misuse enforcement at all), what happens now (judged once at payment; unentitled → no activation + a DISCOUNT_UNAUTHORIZED flag for the board), and the three judgment calls from review. The dense mechanics are still below it. Happy to expand any part.

thpr (commented):
Both fixes land. The intersect closes the payload-stuffing hole for any id that isn't enrolled, and putting it inside the helper is the right home, since the reconciler inherits it without changing. `LIVE_PERSON` on that lookup earns its place independently of the drift guard: a merge leaves the tombstone's enrollment row behind, and that row would otherwise lend a former household's dues to someone else's checkout.

A weaker version of the attack survives, and it comes out of my own imprecision. I asked for an intersect with the program's real `ProgramParticipant` rows, and that is exactly what shipped; the question the check needs to answer is narrower than the one I posed, which is not "who is enrolled" but "whose enrollment is this order paying for". The lookup at `src/lib/programs/memberDiscountCode.ts:59` matches on enrollment existence at any status, so an id already ACTIVE in the program still carries entitlement into the loop below it.

That leaves this path. The attacker now needs a dues-settled person who is enrolled in the same program rather than any settled person at all. Their id clears the check, `activateProgramEnrollment` finds that row already ACTIVE and its CAS guard no-ops on it, and the attacker's own PENDING row flips to ACTIVE at the member price with nothing flagged. It is a much harder attack than the one I reported, since the attacker has to guess an id that is both enrolled in that specific program and settled, gets one attempt per purchase, and learns nothing from a failure. I'd treat it as low severity rather than a blocker.

Filtering the intersect to `status: PENDING`, which is what the reconciler's forward pass does, would close it and break something else. The webhook has no already-recorded early return the way `reconcileForwardProgram` does, so a redelivered webhook for an honest order would find every row ACTIVE, take the empty intersection, and flag a member who did nothing wrong. I assume that's why the intersect is status-blind, and that reasoning is sound.

Matching rows that are `PENDING` **or** already carry this same `shopifyOrderId` gets both properties. That set is exactly the rows this order is paying for: a co-enrollee who paid on some other order stops carrying entitlement, and a redelivery still recognizes its own rows and stays quiet.

---
🤖 Posted by [Claude Code](https://claude.com/claude-code) on @thpr's behalf

jee7s:
Fixed in cd7bb4c3, exactly your set: the intersect's where now carries `OR: [{ status: "PENDING" }, { shopifyOrderId }]` (order id threaded in from both callers — `order.id` at the webhook, `order.legacyId` in the reconciler's forward pass). A settled co-enrollee ACTIVE via another order clears neither arm; a redelivery matches its own stamped rows and stays quiet. Three new unit tests pin the OR shape, the redelivery path, and the null-order-id case (PENDING arm only). 17/17 on the helper suite, drift guard and webhook/reconcile suites green, tsc clean.

thpr (approved):
Approving. The intersect now scopes to the rows this order pays for, both callers pass their order id, and the redelivery arm keeps an honest re-sent webhook quiet. Every finding I raised is closed, and the main merge left the PR's files untouched.

One residual stays open by design rather than by oversight, and it's worth a line in the PR body next to the existing `(kind, shopifyOrderId)` note rather than more code. A dues-settled person who is PENDING in this same program still matches the first arm, so naming their id in the cart attribute clears the order and activates the attacker's own row at the member price, handing the victim a free activation stamped with the attacker's order. It needs an id guessed blind, enrolled and unpaid in that exact program, with one attempt per purchase. It is also the irreducible form of "any entitled row clears the whole order", which is this PR's stated interim semantics; closing it means judging per row and activating only the entitled ones, which is the segment-gated end state #929 and #278 already own.

---
🤖 Posted by [Claude Code](https://claude.com/claude-code) on @thpr's behalf

PR #1654, opened 2026-08-16
fix(membership): a family joining inside the renewal window buys the coming year

### LINKED ISSUE #4: dev signage
When running in dev mode, the main nav bar should say "CMI-dev" instead of "Checkmein" to remind the user what env is running

jee7s:
## Problem

We are inside the renewal window (opened Jul 1), and every application activated since then shows **Valid until: <current boundary>** in membership-ops — a family that just joined appears to expire in weeks. Meanwhile `membershipValidThrough` (program pricing, notifications) gives the same family coverage through the FOLLOWING boundary. Two horizon computations, disagreeing, with the board-facing one wrong.

## Root cause

`settledThisCycleWhere` (lifecycle.ts) was pinned to `kind: "RENEWAL"`, so an INITIAL application settled in-window never counted as "settled for the coming year" where that fragment is read: the households `validUntil`/`settledForComingYear`, and `runRenewalSweep`'s skip-test — which therefore also opened a `PENDING_RENEWAL` and sent a renewal reminder to a family for the very year it just paid for.

## Change

Two questions, two fragments (post-review shape):

- **`settledThisCycleWhere` — the money question**: terminal **ACTIVE** in-window, kind-agnostic. Read by the households `validUntil` and `membershipValidThrough`. ARCHIVED is deliberately absent: `archiveApplication` refuses ACTIVE processes, so every ARCHIVED row **never completed payment** — this also removes the pre-existing ops case where a board-archived *declined* renewal showed a valid-until a year the family never bought.
- **`handledThisCycleWhere` — the sweep question**: ACTIVE **or** ARCHIVED in-window. Only `runRenewalSweep`'s skip-test reads it (kind `RENEWAL` stays on the in-flight arm only): a process the board already closed this cycle is not reopened, and a family that joined in-window is not billed twice.
- `membership-ops/households`: no query change beyond the fragment; the `renewalGrantable` split is unaffected (`grantableRenewalWhere` keeps `kind: "RENEWAL"`).
- `docs/rules/membership.md`: new [Decision] bullet — settling inside the window buys the coming year, joining and renewing alike.
- Tests: both fragment shapes pinned; `membershipValidThrough` probe asserted kind-agnostic + ACTIVE-only; the sweep's OR shape pinned by a unit test (re-wrapping it in `kind: "RENEWAL"` now fails a test, not just CI-only integration); `householdsListGrantableAPI` integration expectations flipped to the new rule.

## Why this deliberately reverses part of #1067 (fix #4)

#1067 added the `kind: "RENEWAL"` pin "so a stray INITIAL activation no longer flips derivedValidUntil a year forward." But:

1. **The pin never guarded the system.** `membershipValidThrough` was already kind-agnostic, so a stray in-window INITIAL already flips the *pricing* horizon today. The pin only manufactured the ops/pricing disagreement; it did not remove the stray-INITIAL effect.
2. **It excludes the legitimate majority case.** Per policy the window sells the coming year; a family joining in-window and a family renewing are the same money event. The correct discriminator is *when the process settled and completed*, not its kind.
3. **The residual stray-ACTIVE-INITIAL exposure is unchanged from today.** An erroneously *activated* INITIAL cannot be archived (`archiveApplication` throws `wrong_phase` on ACTIVE) — the board's levers are revocation or a manual correction, before and after this PR. What this PR changes is that the ops display now agrees with what pricing already did, instead of contradicting it.

## Known limitation (recorded, not fixed here)

Settlement time is `stageEnteredAt` — stage completion, not payment. A family paying pre-window whose background review clears in-window reads as having bought the coming year; board review latency can swing the horizon by a year around the window edge (the June-joiner / summer-programs overlap). This predates the PR (`membershipValidThrough`'s inline probe had the same timing) but is newly load-bearing for INITIALs. If it needs closing, the honest fix is disambiguating the *product* (a year-specific membership purchase) rather than inferring the year from timing — left as an open follow-up question.

## Verification

Targeted suites 110/110; full default unit suite 2747/2747; `tsc --noEmit` clean; eslint clean. Integration suites ride on CI (no Docker locally) — the previously failing `householdsListGrantableAPI` cases are updated to pin the new rule.


thpr (commented):
## Review — 5 findings

### 1. ARCHIVED (never-paid) process now extends pricing horizon +1 year

`membershipValidThrough` gains ARCHIVED via the shared fragment, but `archiveApplication` (archive.ts:19) *refuses* ACTIVE processes — so every ARCHIVED row is by construction an application/renewal that **never completed payment**. The PR body's justification ("board-archived completed renewal … latent drop") describes a state that can't exist.

**Scenario:** ACTIVE household declines to renew; sweep opens PENDING_RENEWAL on Jul 1; board archives it Jul 15 (archive.ts stamps `stageEnteredAt = now`, in-window; membership stays ACTIVE — no auto-revoke). Pre-PR `membershipValidThrough` probed `status: "ACTIVE"` only → boundary. Post-PR it matches the ARCHIVED row → boundary + 1 year → `isDuesSettledThrough` / `coversThrough` grants member pricing for programs running into a membership year the family explicitly did not buy.

Same leak path: abandoned-then-archived INITIAL while a second application is dues-settled-awaiting-BG.

---

### 2. Architecture question: `stageEnteredAt` vs `paidAt` — which year did this family buy?

The fragment tests `stageEnteredAt` (last stage transition), not `paidAt`. For renewals this is benign — they only open in-window, so by the time they reach ACTIVE the stage timestamp is necessarily in-window too.

For INITIALs it's a different story: a family can pay in June (pre-window) and have their BG review land in July (in-window). `activate()` stamps `stageEnteredAt` at review clearance time, so now this family reads as "bought the coming year" — decided not by when they paid but by how fast the board reviewed their background check. Five days of board latency can swing a full membership year of coverage.

**The concrete risk window I'm worried about:** families paying for summer programs (FRC, drone, etc.) in June / early July. Those payment dates overlap with the renewal window opening. If a family joins June 28, pays that day, and BG clears July 3 — the system now says they bought the *following* year, the sweep never opens a renewal, and they get 13 months of coverage instead of 1. Meanwhile an identical family whose review finished June 29 gets the boundary as their horizon.

This might not be fixable within the scope of this PR's approach (kind-agnostic fragment keyed on `stageEnteredAt`). If the underlying problem is "we can't tell which year a payment bought", the question is whether the membership *product* itself needs to distinguish — e.g., a year-specific membership product so the system can 100% disambiguate what was purchased, rather than inferring it from timing. That's a bigger architectural question, but the new INITIAL-in-window path is the smaller of our two risk windows for tagging the wrong outcome (the other being the display-only disagreement this PR fixes), and I'd rather surface it now than after it silently over-covers a summer joiner.

Not necessarily a blocker for this PR — the pre-PR code already has the same `stageEnteredAt` timing on `membershipValidThrough`'s inline probe, so this isn't a *new* mechanism, just newly exposed to INITIALs. But worth recording as a known limitation of the timing-inference approach.

---

### 3. Stray ACTIVE INITIAL can't be archived — recovery path cited in PR body doesn't exist

The PR body argues the stray-INITIAL edge (#1067's motivating case) "is better handled where it originates: an erroneous INITIAL is a board-archivable process." But `archiveApplication` (archive.ts:19) throws `wrong_phase` for ACTIVE processes. An erroneously activated INITIAL — the exact scenario the `kind: "RENEWAL"` pin guarded against — cannot be archived. The board's only levers are revoking the entire membership or a manual DB edit.

---

### 4. Docblock overclaims unification — badge + outreach probes still diverge

The new docblock (lifecycle.ts:204) claims all readers "read this ONE fragment and cannot disagree", and the new membership.md bullet asserts the same. But two near-copy probes remain with deliberately different semantics:

- **Badge roster** (people/search/route.ts:53–54): `status: 'ACTIVE'` only, no ARCHIVED
- **Outreach recipients** (outreach/recipients.ts:53,76): `status: "ACTIVE"` only, no ARCHIVED

A board-archived renewal in-window: ops shows settled-for-coming-year and the sweep skips them, but outreach keeps mailing them renewal reminders (audience exclusion misses ARCHIVED) and badges print no year. The "cannot disagree" claim covers three of five readers. The next maintainer trusts a single-source assertion that is false.

Not a behavior regression — these probes were already ACTIVE-only pre-PR — but the documentation now actively misleads about the scope of unification.

---

### 5. Sweep's restructured OR shape has no unit test

`lifecycle.test` pins the fragment alone (kind-agnostic, ACTIVE/ARCHIVED, stageEnteredAt). The sweep's *use* of it — `kind: "RENEWAL"` on the in-flight arm only, settled arm kind-agnostic — has no unit test. A future refactor re-wrapping `processes.where` in `kind: "RENEWAL"` (the pre-PR shape) would restore the double-billing-of-in-window-joiners bug this PR fixes, and all 2745 default-suite tests would stay green (integration suites that exercise `runRenewalSweep` ride CI only).

---

jee7s:
All five addressed in 22fbe4a5 (plus a merge of main picking up #1650):

1. **ARCHIVED extending the horizon** — real, fixed by splitting the fragment: `settledThisCycleWhere` is ACTIVE-only again (money), and the new `handledThisCycleWhere` (ACTIVE|ARCHIVED) is read by the sweep alone. Your declined-renewal scenario now shows the plain boundary AND stays unpestered by the sweep. This also removes the pre-existing ops case where a board-archived declined renewal displayed a year it never bought.
2. **stageEnteredAt vs paidAt** — recorded as a known limitation in the fragment docblock and the PR body, including your year-specific-product framing as the honest fix if the June-edge needs closing. Not attempted here.
3. **"board-archivable" claim** — corrected in the PR body: an erroneously activated INITIAL cannot be archived; the levers are revocation or manual correction, unchanged by this PR. The justification now rests on the pin never having guarded pricing in the first place.
4. **Overclaimed unification** — the "cannot disagree" wording is gone; docblocks now name the exact readers of each fragment. Badge/outreach probes were ACTIVE-only all along and now match the money fragment's semantics, but they are not claimed as single-sourced.
5. **Sweep OR shape** — pinned by a new unit test (mocked findMany arg asserted against the exact two-arm shape); re-wrapping the probe in kind:RENEWAL fails in the default suite now.

The failing `householdsListGrantableAPI` integration cases were the old #1067 pins — flipped to the new rule (INITIAL-in-window settled TRUE / archived-renewal settled FALSE). Full unit suite 2747/2747, tsc/lint clean.

jee7s:
@thpr ptal

thpr (approved):
All five findings from the first pass addressed:

- **ARCHIVED/money split** — `settledThisCycleWhere` (ACTIVE-only) vs `handledThisCycleWhere` (ACTIVE+ARCHIVED, sweep-only) cleanly separates the money question from the don't-reopen question. Integration test assertions flipped correctly.
- **stageEnteredAt known limit** — documented in the docblock; architecture question recorded in the review thread.
- **Stray-ACTIVE-INITIAL** — PR body now explicitly acknowledges the exposure is unchanged from today.
- **Docblock overclaim** — removed; each fragment's scope correctly stated.
- **Sweep OR shape** — new unit test pins it.

PR #1657, opened 2026-08-17
ci(staging): remove the staging deploy smoke-test timeout

jee7s:
## Problem

The staging smoke test bounds its retry loop at 60×10s = 10 minutes. That number was measured against one observed wake-up, and the workflow's own comment records the failure mode: on run 29925037763 the previous 2-minute window expired 68 seconds before the waker promote finished, reporting a **succeeded** deploy as failed. Staging cold-start time varies (scale-to-zero, image pull, waker promote), so any fixed window eventually repeats that incident with a slower wake-up.

## Change

The retry loop runs until the probe passes; GitHub's job-level limit (6h default) is the only backstop. The 10s cadence is kept — each probe drives the ops-stg-waker's health poll. The now-unreachable failure tail is removed.

Genuinely-broken deploys still fail: `aws ecs wait services-stable` earlier in the job catches a crash-looping image (with `--desired-count 1` so 0/0 can't report vacuously stable); the smoke loop only decides how long we wait for a *stable* service to answer.

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -520,13 +520,15 @@ jobs:
           echo "Service is stable."
 
       - name: Smoke test the live app
+        # 20 min: generous for staging wake-up (the old 10-min loop cap was
+        # the false-negative), but bounded so an ECS-stable-yet-unreachable
+        # app (target group / listener / SG) can't burn the 6-hour default.
+        timeout-minutes: 20
         run: |
           set -euo pipefail
           # Real probe of the deployed service (mirrors deploy-prod.yml). The
           # app answers 200 or an auth redirect; retries cover target
-          # registration + first boot after scale-up, with no step-level
-          # timeout — staging wake-up time varies too much for a fixed window,
-          # so the job's GitHub-level limit is the only backstop. The 503s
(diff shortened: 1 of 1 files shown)


thpr (commented):
The `while true` with no `timeout-minutes` on the job defers to GitHub's 6-hour default. If `ecs wait services-stable` passes but the ALB never routes traffic (target group misconfigured, security group, listener rule), this loop silently burns 6 hours of runner minutes before GitHub kills it.

The old 10-minute cap was too short — but 6 hours is too long. An explicit `timeout-minutes: 20` on the job (or just this step) catches the "service stable but unreachable" class without repeating the original false-negative.

jee7s:
Good catch — took the step-level variant: `timeout-minutes: 20` on the smoke-test step in 13bb5874. The wide retry window that fixed the old 10-minute false negative stays, but the "services-stable yet unreachable" class (target group, listener rule, SG) now fails in 20 minutes instead of silently burning the 6-hour job default. Step-level rather than job-level so the deploy/wait phases keep their own natural bounds.

jee7s:
Closing this as I don't think it's a limiting factor. The problem was the holdoff, and I've set that to 0 in settings.

PR #1664, opened 2026-08-18
fix(corrections): score open-visit deletes as strong (#1630)

### LINKED ISSUE #1630: deleteSignificance scores an open visit 0, so deleting the live roster scores lowest
`checkin-app/src/lib/visit/significance.ts:88` scores a delete off the recorded duration, and an open visit has none:

```ts
const duration = visit.departedAt
    ? minutesBetween(visit.arrivedAt, visit.departedAt)
    : 0;
```

So deleting a visit that is still open yields `{ score: 0, flagged: true }` — rendered on the corrections screen as `Flagged · 0`.

## Why that is wrong

§2 of `1256_ATTENDANCE_CORRECTION_SURFACE.md` makes the score the *strength* of a change; the delete floor supplies only *whether*. Its table reads `| delete a LEAD_MARKED or SCANNER visit | high | **yes** — strong |`. The delete table never contemplates an open visit, and the code resolves that unspecified case as "nothing was lost." For an open visit the strength comes out zero, so a deleted live scanned check-in is indistinguishable from a trivial deletion on the one axis the review screen sorts and scans by.

The asymmetry is sharp. A person badges in and is on site six hours:

- **Delete it now** (still open): score `0`.
- **Wait for the kiosk checkout, delete the identical row:** `360 × 3 × 2 = 2160`.

And an open visit *is* the live roster of who is in the building — `events/[id]` `manualEditAttendance` refuses an Absent mark on one for exactly that reason. So the path that scores lowest is the one destroying the most safety-critical record.

## Proposed fix

Score an open visit against now — time on site is the magnitude — with `now` injected so the fixed-offset assertions stay exact:

```ts
const duration = minutesBetween(visit.arrivedAt, visit.departedAt ?? opts.now ?? new Date());
```

## Reachable sites

- `checkin-app/src/app/api/facility/visits/route.ts` — `DELETE`
- `checkin-app/src/app/api/my-programs/conflicts/resolve/route.ts` — an open visit is a legitimate half of a duplicate-attendance conflict
- `checkin-app/src/app/api/attendance/manual/[id]/route.ts` — **pre-existing**; its board email body reads `(significance 0)`

## Why it was not fixed in #1558

It changes `attendance/manual/[id]` scores too — including which ones cross the threshold and email the board — and it is a semantics change to weights the helper's own docstring calls "v1 defaults awaiting board calibration" (design §6.2). That is a scoring-model decision, not a bug fix, so it wants its own board conversation.

The assertion pinning the current value is in `checkin-app/src/app/__tests__/adminVisitsAPI.integration.test.ts` (the staff `DELETE` case). #1558 re-recorded it as unresolved rather than as intended, and left a `TODO` pointing here — swap this issue's number in.

## Lower priority, same shape

A visit with both ends machine-closed (`FACILITY_CLOSE` / `AUTO_CLOSE`, both weight 0) gets `weight = max(0, 0) = 0`, so it also renders `Flagged · 0` however long the visit was.

## Origin

Raised reviewing #1558, which made three more paths score and so brought the zero into the board's default lens. Text is the author's own from that PR's description, filed here so it has a home. Related: #1624, which is why the weights cannot cleanly express what these writers mean in the first place.

---
_Filed by Claude._

jee7s:
Fixes #1630.

## Bug
`deleteSignificance` scored an open visit's delete by recorded duration, which
falls back to `0` when `departedAt` is `null`. Every delete always flags
(`flagged: true`), but the *score* — the axis the corrections screen sorts and
scans by — came out lowest for exactly the delete that erases the live
in-building roster.

## Spec reading
§2 of `1256_ATTENDANCE_CORRECTION_SURFACE.md` makes the score the *strength* of
a change; the delete floor only supplies *whether* it flags. Its table says
`delete a LEAD_MARKED or SCANNER visit` → weight "high" → "yes — strong". The
table (and the delete-floor prose) never contemplates an open visit — the old
code resolved that gap as "nothing was lost," which is the one reading the
spec's own logic rules out: a visit still open is unfinished, not zero.

## Fix
Score an open visit's delete against **time on site so far** — `arrivedAt` to
now — using the same `weightOf`/max-weight logic already used for closed
deletes. `now` is injectable via `SignificanceOpts` so unit tests stay exact;
every real call site (`facility/visits` DELETE, `attendance/manual/[id]`
DELETE, `my-programs/conflicts/resolve`, `events/[id]` `manualEditAttendance`)
needed no changes — they already call with only `{ byProxy }`, so they pick up
real-clock scoring for free. Closed-visit scoring is untouched: `departedAt`
still wins the `??` chain, so nothing changes for the visit's normal (closed)
delete path.

## Before / after on the corrections screen
- **Before:** deleting a `SCANNER` visit that's still open scored `0`
  (`Flagged · 0`) — indistinguishable from a trivial delete, and *lower* than
  deleting the identical visit after checkout (`360 × 3 = 1080`, or `×2 = 2160`
  by proxy).
- **After:** deleting that same open visit now scores by elapsed minutes ×
  source weight, same as a closed delete — a 6h-old open `SCANNER` check-in
  scores `1080`, matching the "delete a SCANNER visit — high/strong" row in
  §2's table instead of falling into the unspecified gap.

Out of scope (per the issue): the "both ends machine-closed" case, which is a
separate, lower-priority instance of the same floor-collapsing-to-0 shape and
not part of this fix.

## Tests
- `checkin-app/src/lib/visit/__tests__/significance.test.ts` — added:
  - "an open visit's delete scores elapsed time, not the floor" (SCANNER, 6h
    open → `{ score: 1080, flagged: true }`)
  - "a closed visit's delete score is unaffected by `now`" (confirms closed
    delete scoring is unchanged)
- `checkin-app/src/app/__tests__/adminVisitsAPI.integration.test.ts` — updated
  the pinned staff-DELETE-of-an-open-visit assertion that issue #1630 flagged
  as `UNRESOLVED`; now asserts `flagged: true` and a score in the elapsed-time
  range instead of the old hardcoded `{ score: 0, flagged: true }`.

Local results:
- `npx tsc --noEmit` — clean
- `npx jest --testPathPatterns significance.test.ts` — 15/15 passed
- `npx eslint` (via lint-staged on commit) — clean, no fixes needed beyond auto-format

### DIFF
--- a/checkin-app/src/app/__tests__/adminVisitsAPI.integration.test.ts
+++ b/checkin-app/src/app/__tests__/adminVisitsAPI.integration.test.ts
@@ -416,10 +416,13 @@ describe('Admin Visits API Integration Tests', () => {
             // Resolved (#1630): the visit is still open, so its score is elapsed
             // time-on-site (~1h, no arrivedVia so weight 1) rather than the old
             // hardcoded 0 — deleting a live check-in no longer scores lowest.
+            // The admin (testAdminId) isn't the visit's own person (testUserId),
+            // so this delete is byProxy too — the route doubles the weight
+            // (significance.ts PROXY_MULTIPLIER), giving ~120, not ~60.
             const significance = (auditLog?.newData as { significance?: { score: number; flagged: boolean } })?.significance;
             expect(significance?.flagged).toBe(true);
-            expect(significance?.score).toBeGreaterThanOrEqual(60);
-            expect(significance?.score).toBeLessThan(65);
+            expect(significance?.score).toBeGreaterThanOrEqual(120);
+            expect(significance?.score).toBeLessThan(130);
         });
     });
 
--- a/checkin-app/src/app/__tests__/adminVisitsAPI.integration.test.ts
+++ b/checkin-app/src/app/__tests__/adminVisitsAPI.integration.test.ts
@@ -413,12 +413,13 @@ describe('Admin Visits API Integration Tests', () => {
             expect(auditLog).not.toBeNull();
             expect((auditLog?.oldData as { id?: number })?.id).toBe(doomed.id);
 
-            // UNRESOLVED (#1630) — 0 is what ships, not what is right. The visit is
-            // still open, so deleteSignificance weighs no duration: deleting a live 6h
-            // scanned visit scores 0, while deleting that same row after checkout
-            // scores 2160. The lowest score destroys the live in-building roster.
-            expect((auditLog?.newData as { significance?: { score: number; flagged: boolean } })?.significance)
-                .toEqual({ score: 0, flagged: true });
+            // Resolved (#1630): the visit is still open, so its score is elapsed
+            // time-on-site (~1h, no arrivedVia so weight 1) rather than the old
+            // hardcoded 0 — deleting a live check-in no longer scores lowest.
+            const significance = (auditLog?.newData as { significance?: { score: number; flagged: boolean } })?.significance;
+            expect(significance?.flagged).toBe(true);
+            expect(significance?.score).toBeGreaterThanOrEqual(60);
+            expect(significance?.score).toBeLessThan(65);
         });
     });
 
--- a/checkin-app/src/lib/visit/__tests__/significance.test.ts
+++ b/checkin-app/src/lib/visit/__tests__/significance.test.ts
@@ -106,6 +106,25 @@ describe("deleteSignificance", () => {
         );
         expect(r.score).toBe(360);
     });
+
+    // #1630: an open visit has no recorded duration to weigh, so it must not
+    // fall back to 0 — that made deleting the live roster the *lowest*-scoring
+    // delete. Score against time on site so far instead.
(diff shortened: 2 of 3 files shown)


thpr (commented):
Clean fix, no blocking findings.

One minor observation: the existing "always flags" test (`significance.test.ts:98`) calls `deleteSignificance` with `departedAt: null` and no `now` — score is now wall-clock-dependent (~70k against `at(14)` from July). Doesn't break anything since the assertion only checks `.flagged`, but passing `{ now: at(15) }` would make it fully deterministic.

---
*Posted by Claude Code on behalf of @thpr*

jee7s:
Applied — `{ now: at(15) }` on the open-visit call in the "always flags" test, so the score is deterministic instead of drifting with the wall clock. 15/15 in `significance.test.ts` pass. `00996387`

thpr (approved):
Observation addressed in 00996387bb. Clean fix, correct at all call sites.

---
*Posted by Claude Code on behalf of @thpr*

Issue #1347, opened 2026-07-25
Kiosk resilience epic (health machine, recovery ladder, reconciliation, DLQ) — AT8 is a slice


thpr:
**Backlog item AT14** — surfaced by the design review of PR #1216 (`docs/designs/KIOSK_RESILIENCE.md`). Folded from `backlog/sources/design_pr_open_items.md`.

Backlog **AT8 = only the offline store-and-replay + banner slice**. The doc is a full resilience program (1098 lines): §0 invariants, §2 offline queue, §3 health state-machine + recovery ladder (Chromium bounce / wifi bounce / nightly + escalation reboot), §6 reconciliation substrate + projection, server-side DLQ, a `system-status/unsynced-scans` review panel, Phase-0 dead-path fixes.
- Includes a net-new **bug (Q16)**: the 3s debounce swallows the second badge, so a keyholder double-badging faster than 3s **cannot close the facility** and gets no feedback (working confirm window is really [3s,12s]); ties AT6/#300, AT9/#254.
- Several open sequencing DECISIONS (closed-facility authority [CONTESTED], §0 invariant ratification, Stage-2 substrate timing, vs #254 race) — teed up in the backlog DECISIONS file.

AT8 becomes a slice of this epic.

thpr:
The v2 design doc is now **merged to `main`** via #1216 (merge commit `2d8853c1`).

**Location:** [`docs/designs/KIOSK_RESILIENCE.md`](https://github.com/innovationtreehouse/checkin/blob/main/docs/designs/KIOSK_RESILIENCE.md)

v2 added §0 invariants and the §6 reconciliation model (unified Stage-2 substrate + re-runnable projection, partition states, M0–M3 machines, conflict catalog) on top of the v1 queue/health sections. Still a **proposal — nothing built**; the open decisions live in the doc's §5 (notably the §2 `[CONTESTED]` advisory-vs-authoritative facility state, and §5.22 on when the server-side substrate lands). Per #1359, those open questions should be resolved before implementation of the affected slices.

jee7s:
HOLD — a merged design with `[CONTESTED]` decisions still open is exactly the case #1359 describes: "Ready" here would mean a doc exists, not that someone can build this. It also bundles an epic with a standalone defect.

Questions:

1. Is a facility's closed state advisory or authoritative — the §2 `[CONTESTED]` decision?
2. Are the §0 invariants ratified as written, or do any change before implementation?
[...]

PR #1665, opened 2026-08-18
fix(kiosk): guard self-update against non-advancing restarts (#1616)

### LINKED ISSUE #1616: Kiosk self-update can restart-loop when the Pi's git pull cannot fast-forward
`version_poller`'s self-update exit has never fired, because it compared `HEAD` against `origin/master` on a repo whose default branch is `main`. #1607 fixes the branch name, which makes the exit live for the first time. Nothing in that PR is wrong; this is the behavior it switches on.

The exit has no guard against repeating itself:

```python
if local_head and remote_head and local_head != remote_head:
    log.info(f"Client version update available ({local_head} -> {remote_head}). Restarting client.")
    os._exit(0)
```

The only thing that clears `local_head != remote_head` is the pull at the top of `kiosk.sh`'s respawn loop:

```bash
git -C "$(git rev-parse --show-toplevel)" pull origin main || true
```

`|| true` swallows every failure. A Pi with a dirty tree, a hand-edited tracked file, a local commit, or a detached HEAD can never fast-forward, so the comparison stays unequal forever: the client exits every poll interval (60s default), `kiosk.sh` kills Chromium and respawns it, the pull fails again, and the display is unusable until someone walks up to it. The failure is silent — the pull error goes nowhere and the exit logs as a normal update.

The Pis most likely to be in that state are the ones this path was written for. Because the self-update never worked, the fleet has only ever been updated by hand, which is exactly how a working tree picks up local edits.

## Fix

Make the exit idempotent against a remote head it has already tried. The client restarts, so the memory has to outlive the process — a single file next to `client.py` holding the last sha it exited for, skipped if the new `remote_head` matches, is enough. Restore the exit when the remote moves again, so a genuinely new commit still updates.

Worth pairing with making the pull failure visible: `|| true` is load-bearing under `set -e` (a failed pull must not kill the kiosk), but it should log loudly rather than pass in silence.

## Scope

Client only — `client/client.py`, possibly `client/kiosk.sh`. No server change, no schema. Related to the kiosk resilience epic #1347 but independent of every phase in it; #1607 is the PR that makes it reachable.

jee7s:
Fixes #1616

## Failure narrative

`version_poller`'s self-update exit (`os._exit(0)` when `local_head != remote_head`) was dead code until #1607 fixed the `origin/master` -> `origin/main` mismatch, making it fire for the first time. It has no guard against repeating itself. `kiosk.sh`'s respawn loop does `git -C ... pull origin main || true` before relaunching the client — on a Pi with a dirty tree, a local commit, or diverged history, that pull can never fast-forward, so `HEAD` never advances. The client sees the same mismatch on the very next poll after respawn and exits again, immediately, forever: Chromium gets killed and relaunched every cycle and the kiosk is unusable, with no error surfaced anywhere (the failed pull was swallowed by `|| true` and the repeat exit logs identically to a normal update).

## Guard design

The client now remembers, on disk, the remote head it already restarted for — `client/.self_update_last_target`, written next to `client.py`/`client.key` using the same relative-path convention (kiosk.sh `cd`s into `client/` before launching `client.py`), gitignored.

- On a `HEAD`/`origin/main` mismatch, the client only exits if `remote_head` differs from the last target it restarted for (`_should_restart_for_update`). First time seeing a given target: restart as before, and persist that target.
- Second time seeing the *same* target after a restart (pull didn't advance `HEAD`): don't exit again. Log a loud `WARNING` explaining the pull likely can't fast-forward, and keep serving scans on the current version.
- If the remote genuinely moves further (a new commit lands while the Pi is stuck), the new `remote_head` no longer matches the persisted target, so the exit fires again as normal — the guard only suppresses repeats for a target already tried, not all future updates.

This is the minimal shape the issue itself proposes ("a single file next to client.py holding the last sha it exited for, skipped if the new remote_head matches... restore the exit when the remote moves again"). It turns "restart-loop forever" into "log once, then keep working" without adding any new machinery — no health states, no retry/backoff scheduler, no server round-trip.

Paired with this (also called out in the issue): `kiosk.sh`'s `git pull || true` still can't be allowed to kill the kiosk under `set -e`, but it no longer swallows the failure silently — a failed pull now logs `WARNING: git pull failed -- kiosk will keep running the current checkout.` to stderr, so the actual root cause (bad checkout state on the Pi) is visible to whoever's watching logs instead of masquerading as a normal update cycle.

## Out of scope

The kiosk resilience epic's health-machine escalation (#1347 and its phases) is not touched — this is a standalone, minimal fix for the specific restart-loop failure mode, independent of that epic per the issue's own scope note. No server change, no schema change, no refactor of shared client scaffolding (scanner listener, proxy, SSE, attendance polling all untouched).

## Tests

Added `client/test_self_update_guard.py`:
- `_should_restart_for_update`: no prior target -> restart; same target as last restart -> don't restart; remote moved past last target -> restart again.
- State-file round trip: missing file reads as `None`; write-then-read round-trips.
- End-to-end against the real `version_poller` loop (mocked `subprocess`/`os._exit`/`time.sleep`, scripted head sequence): mismatch -> one restart; non-advancing pull with the same target -> no second exit, `WARNING` logged (asserted via `assertLogs`); remote advances to a new target -> restart allowed again.

Ran locally via the same invocation CI's "Kiosk client tests (Python)" job uses (`python -m unittest discover -v` from `client/`, `pip install -r client/requirements.txt` into a venv first): all 13 tests pass (7 pre-existing + 6 new), no regressions.

## Deferred / left alone

- The health-machine escalation ladder from the kiosk resilience epic (#1347) — explicitly out of scope per the issue.
- A separate known bug where one client version-check path polls a route that doesn't match the server's actual route shape — tracked separately, untouched here.
- Clearing `.self_update_last_target` on a successful update isn't needed: once `HEAD` catches up, the stored target is simply never matched again since git history only moves forward.

### DIFF
--- a/client/.gitignore
+++ b/client/.gitignore
@@ -1,6 +1,7 @@
 # Private key - NEVER commit
 client.key
 config.json
+.self_update_last_target
 
 # Python
 __pycache__/
--- a/client/client.py
+++ b/client/client.py
@@ -13,6 +13,7 @@ import html
 import json
 import os
 import secrets
+import subprocess
 import sys
 import time
 import threading
@@ -37,6 +38,10 @@ log = logging.getLogger("kiosk")
 # The backend page serving the kiosk display; `mode=kiosk` selects its kiosk
 # layout. Overridable per-Pi via config.json.
 DEFAULT_KIOSK_PATH = "/attendance/current?mode=kiosk"
+
+# Self-update loop guard (see version_poller): remote head we already exited
+# for. Relative path -- kiosk.sh cds into client/ before launching client.py.
+SELF_UPDATE_STATE_FILE = ".self_update_last_target"
 
 def load_config(path="config.json"):
     if not os.path.exists(path):
@@ -603,10 +608,31 @@ def attendance_poller(backend, state, interval=30):
                 log.info(f"Attendance poll: {new_counts.get('total', '?')} present")
                 state.push_event({"html": "", "counts": new_counts})
 
-def version_poller(backend, state, interval=60):
+def _read_last_restart_target(path=SELF_UPDATE_STATE_FILE):
+    # Missing/unreadable file means "never restarted" -- same as no target.
+    try:
+        with open(path) as f:
+            return f.read().strip() or None
+    except OSError:
+        return None
+
+def _write_last_restart_target(remote_head, path=SELF_UPDATE_STATE_FILE):
+    try:
+        with open(path, "w") as f:
+            f.write(remote_head)
+    except OSError as e:
+        log.error(f"Could not persist self-update state to {path}: {e}")
+
+def _should_restart_for_update(remote_head, last_restart_target):
+    # False once we've already exited for this exact remote_head and the
+    # restart didn't move HEAD past it (non-fast-forward pull); True for a
+    # never-tried or newly-advanced target.
+    return remote_head != last_restart_target
+
+def version_poller(backend, state, interval=60, state_path=SELF_UPDATE_STATE_FILE):
     """Background thread that checks for client and server version updates."""
-    import subprocess
-    
+    last_restart_target = _read_last_restart_target(state_path)
+
     # Get initial server version
     initial_server_version = None
     for _ in range(6):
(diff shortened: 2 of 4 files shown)


jee7s:
Reviewed `be5d070698` against #1616.

The guard matches the issue: persist the remote sha you already exited for, don’t restart-loop when pull can’t fast-forward, log the failed pull instead of `|| true`. Tests cover first-try / stuck / remote-moved-on.

One hole: if `_write_last_restart_target` fails you still `os._exit(0)`. Next boot the file is missing, so the same `remote_head` looks never-tried and the loop is back. Don’t exit unless that sha is on disk (or treat write-failure as “stay up,” same as a stuck pull).

jee7s:
Good catch — fixed in e91f032d. `_write_last_restart_target` now returns whether the sha actually reads back off disk, and `version_poller` only calls `os._exit(0)` on True; a failed write logs a WARNING and the client stays up, same posture as a pull that can't fast-forward. Added `test_unwritable_state_file_does_not_exit` (plus a return-value check on the round-trip test), both of which fail against the previous commit — it exited once with no state file written.

thpr (approved):
Clean fix, well-scoped to the restart-loop failure mode.

Traced the full guard logic: write→readback verification prevents exiting when state can't persist (which would reproduce the loop); same-target suppression breaks the cycle; remote-advance past a stuck target re-enables the exit. `kiosk.sh` `if !` preserves `set -e` safety while surfacing previously swallowed pull failures. Tests cover the three head-sequence scenarios from the issue plus the unwritable-state-file edge case.

No findings.

Issue #1631, opened 2026-08-10
fix(attendance): manual visit correction must not restamp arrivedVia to WEB


thpr:
## Summary

`facility/trends` decides whether a visit's hours count by reading `arrivedVia` alone:

```ts
OR: [
    { arrivedVia: null },
    { arrivedVia: { notIn: ["LEAD_MARKED", "SYSTEM"] } },
],
```

`facility/visits` PATCH stamps the corrected field and only the corrected field. So a roster-marked visit — `arrivedVia: LEAD_MARKED`, deliberately excluded from the hours because a third party asserted it happened — starts counting as real hours the moment a board member fixes its arrival time:

1. Lead marks a member Present for an event. Both fields `LEAD_MARKED`. Excluded from trends, correctly: nobody measured this, it is the event window.
2. Board member corrects the arrival. `arrivedVia` becomes `STAFF_CORRECTED`.
3. The row now passes `notIn ["LEAD_MARKED", "SYSTEM"]` and counts toward facility hours.

Correcting only the *departure* does not do this — `arrivedVia` is untouched, so the visit stays excluded. Which end of the visit staff happens to fix decides whether the whole thing counts.

## Root cause

Whether a visit *happened* is a fact about the row. It is being stored in a per-field column describing how one **time** was captured, and per-field writes overwrite it. `LEAD_MARKED` on `arrivedVia` is carrying an existence claim it has no way to defend against a later, entirely legitimate write to the same field.

The same category error one level down is #1624.

## History

Not introduced by #1558, but its model makes the gap legible. On `main` the PATCH stamped `WEB`, which also passes the filter, so the promotion has always been possible. #1558's 6249f2d1 briefly closed it by stamping `LEAD_MARKED`, which then created the opposite bug — correcting a scanned visit deleted its hours — and a2e6a586 fixed that by adding `STAFF_CORRECTED`, reopening this one.

That oscillation is the signal: one per-field column cannot hold both facts, and each fix trades one error for the other.

## Coverage

`facilityTrendsAPI.integration.test.ts` gained a case in a2e6a586 that drives the real PATCH route and pins the favourable direction — a corrected `SCANNER` visit stays counted. The `LEAD_MARKED` direction is untested and flips silently.

## Possible fixes

- Filter trends on both fields, so a visit is excluded when *either* end is `LEAD_MARKED`. Smallest change; wrong in the other direction once a roster-marked visit is genuinely corrected on both ends.
[...]

Issue #1624, opened 2026-08-10
VisitSource fuses how a time was captured with who asserted it


thpr:
## Summary

`VisitSource` names three different things across its six values, and both of its consumers read it for something it does not say. Verified on `main` and on `claude/1523-significance`:

| value | what the name states | what it silently implies |
|---|---|---|
| `SCANNER` | mechanism (the kiosk) | measured, **and** self — you badge yourself in |
| `WEB` | mechanism (a form) | *nothing about who typed it* |
| `LEAD_MARKED` | who (a lead) | *nothing about mechanism* |
| `FACILITY_CLOSE` | who (the building sweep) + mechanism | a placeholder |
| `AUTO_CLOSE` | who (the cron) + mechanism | a placeholder |
| `SYSTEM` | legacy fusion of the last three | — |

Two axes are fused into one column: **how a time was captured** (device / typed / machine placeholder) and **who asserted it** (the member / somebody else / nobody). `SCANNER` and the machine closes happen to pin both, so the fusion is invisible there. `WEB` and `LEAD_MARKED` each pin one and leave the other to the reader — and each reader guesses differently.

## What that costs today

**`facility/trends` reads it as "was this measured?"** Its filter is `arrivedVia: { notIn: ["LEAD_MARKED", "SYSTEM"] }`, so a member typing their own hours into `attendance/manual` (`arrivedVia: "WEB"`, `checkin-app/src/app/api/attendance/manual/route.ts:125`) is counted as measured hours. The route's own comment states the confusion outright — "a walk-in it adopts keeps its measured SCANNER/**WEB** and still counts." A typed time is not a measurement, whoever typed it.

**`significance` reads it as "how authoritative?"** `WEIGHTS` asserts `WEB` = "the member's own prior self-report" (`checkin-app/src/lib/visit/significance.ts`), which is a claim about *who* hung on a value named for *where*. The score then multiplies by `byProxy` — the who axis again — so who is partly counted twice.

**Writers disagree because the value cannot express what they mean.** `facility/visits/insert` stamps `WEB` on a visit the board types for someone who was never badged in, which is the most staff-asserted time in the system. `facility/visits` PATCH stamped `WEB` for the same reason until #1558 moved it to `LEAD_MARKED` — which then makes a precisely-corrected time indistinguishable from a roster window with no real clock, and drops the corrected visit out of trends. Neither value is right, because the fact being recorded needs both axes.

## Proposed split

[...]

PR #1666, opened 2026-08-18
fix(facility): backfill staff walk-in visits to LEAD_MARKED (#1632)

### LINKED ISSUE #1632: Staff walk-in visits stop counting in facility hours at deploy, with no backfill decision
## Summary

#1558 moves `facility/visits/insert` — the staff walk-in entry, where a board member types both ends of a visit for someone who was never badged in — off `arrivedVia: "WEB"` and onto `LEAD_MARKED`. That is the right value: nothing measured the time and the member never claimed it.

`facility/trends` excludes `LEAD_MARKED`, so those visits stop counting toward facility hours. Existing rows are not rewritten, so they carry `WEB` and keep counting.

The result is a step in the trend at the deploy boundary. The same operation is counted before it and not counted after it, and nothing on the chart explains the drop.

## The rows are identifiable

The migration note in #1558 says the change is "purely additive — no row is rewritten", which is accurate but reads as though a backfill were not available. It is: the insert route writes a CREATE audit row tagged with its own route.

`checkin-app/src/app/api/facility/visits/insert/route.ts:87`:

```ts
newData: { arrivedAt: ar.value, departedAt: dr.value, type: "staff_entry" },
```

So the affected visits are `AuditLog` rows with `tableName = 'Visit'`, `action = 'CREATE'` and `newData->>'type' = 'staff_entry'`, joined to `Visit` on `affectedEntityId`. That set is exactly the staff walk-in inserts and nothing else — member self-entry through `attendance/manual` writes a different `type`, so the ambiguous `WEB` rows are not caught by it.

## Decision needed before writing anything

Whether to backfill is a reporting judgement, not a technical one, and it should be made deliberately rather than by default:

- **Backfill** — the historical series becomes consistent with the new rule, and previously reported facility hours drop for past periods. Anything already reported to the board from those numbers no longer reconciles.
- **Leave it** — past periods stay as reported, and the series has a discontinuity at the deploy that has to be remembered by whoever reads it.

Either is defensible. The failure mode is picking one silently, which is where this currently stands.

If the answer is "leave it", the discontinuity belongs in `docs/rules/attendance-checkin.md` so the next person to read the trend knows why the shape changes.

## Origin

Found reviewing #1558, which is approved and not blocked on this. The reason no cleaner discriminator exists is #1624.

---
_Filed by Claude._

jee7s:
Fixes #1632.

## Decision

`jee7s` decided on the issue (2026-08-18): **backfill**. Rewrite the historical staff walk-in rows to `arrivedVia`/`departedVia: LEAD_MARKED`. Understood consequence: previously reported facility-hours for past periods drop retroactively and no longer reconcile with anything already reported from those numbers.

## Route change — and why it's landing in this PR, not #1624

The insert route did **not** already stamp `LEAD_MARKED` for every new staff walk-in. #1558 deliberately left `facility/visits/insert` on `WEB` — its own PR body says so outright: "Item A (`facility/visits/insert` stamps `WEB`) is untouched — you folded it into #1624 yourself" — deferring that route change to #1624, which is still open. (An earlier commit on the #1558 branch, `a2e6a586`, did make this exact insert→`LEAD_MARKED` change; it was removed by a reset to `6249f2d1` before merge, precisely to defer it.)

This PR's second commit moves the route to `LEAD_MARKED` anyway. That's forced by the backfill's own coherence, not by any rule #1558 already established: leave the insert route on `WEB` and the very next staff walk-in recreates the discontinuity the backfill just erased. It is a real, if narrow, re-opening of what #1558 explicitly deferred — flagged for explicit sign-off in the PR comment rather than folded in silently.

## Row-set logic

`checkin-app/src/app/api/facility/visits/insert/route.ts:82` — the only route that writes staff walk-ins — pairs every `Visit` CREATE with an `AuditLog` row tagged `newData.type: "staff_entry"`. That tag is the exact, and only, discriminator:

```sql
WHERE "id" IN (
    SELECT "affectedEntityId"
    FROM "AuditLog"
    WHERE "tableName" = 'Visit'
      AND "action" = 'CREATE'
      AND "newData"->>'type' = 'staff_entry'
)
AND ("arrivedVia" = 'WEB' OR "departedVia" = 'WEB')
```

Verified against the two other candidate writers to confirm neither leaks into this set:
- `attendance/manual/route.ts:166` (member self-entry, incl. household-lead-for-member) writes `newData.type: "manual_entry"` — a different string, untouched.
- `attendance/manual/[id]/route.ts` (self-correction) writes `type: "self_correction"` — also untouched.

`facility/trends`'s `route.ts:79-82` excludes `arrivedVia IN ('LEAD_MARKED', 'SYSTEM')` from hours — confirmed the backfill target value is exactly what the exclusion filter checks.

## Idempotency

The `WHERE ... AND ("arrivedVia" = 'WEB' OR "departedVia" = 'WEB')` guard means a re-run of this migration, or a re-apply against an environment where it already landed, touches zero rows the second time — every row in the identified set that still carries `WEB` gets rewritten once, and nothing else matches after that.

## Migration ordering

Data-only `UPDATE`, no schema change — pure backfill, so it ships expand-only per `DEPLOY_MIGRATION_ORDER_OF_OPERATIONS.md` rule 1 (nothing destructive, nothing another release depends on). Single statement, so no `BEGIN`/`COMMIT` wrap needed (matches the repo's own convention — e.g. `20260724120000_purge_adult_dob`, also a single-statement `UPDATE` backfill, ships unwrapped; multi-statement backfills like `20260803000100_visit_source_split_backfill` are the ones wrapped).

No prior data migration in this repo writes an `AuditLog` entry for its own rewrite (checked every `migration.sql` under `checkin-app/prisma/migrations/` — the only `AuditLog`-touching migrations are DDL against the table itself). Following that precedent, this migration does not synthesize an audit trail for the backfill either.

## Test coverage

Added `excludes a real staff walk-in insert from trends hours` to `checkin-app/src/app/__tests__/facilityTrendsAPI.integration.test.ts`, alongside the existing `LEAD_MARKED`/`SYSTEM` exclusion cases. It drives the actual `POST /api/facility/visits/insert` route (not a hand-built fixture) and then asserts the visit is invisible to `facility/trends` — pins the route's current `LEAD_MARKED` stamp end-to-end, which is also the exact behavior the backfill is bringing historical rows in line with.

The same test now also deletes that walk-in through the real `DELETE /api/facility/visits` route and asserts the recorded `significance` score reflects the `LEAD_MARKED` source weight (2) rather than the `WEB` weight (1) it replaced — 480 vs. the 240 a WEB-weighted read of the identical edit would give, `byProxy` doubling included. Pins that the route change this PR makes actually feeds `significance`, not just `facility/trends`.

Local: `npx tsc --noEmit` clean, `eslint --max-warnings 0` clean on the changed test file. The integration suite needs Docker/Postgres, unavailable in this sandbox — CI runs it (`migration-safety.yml` also applies the new migration against a populated DB).

## Known gap — now closed by #1668 (was: not fixed here)

**Update (2026-08-20):** #1668 landed on main and closed this gap. The section below is kept for the record; no action is left for this PR.

The gap as originally described: `attendance/manual/[id]/route.ts:117` unconditionally restamped `arrivedVia: "WEB"` whenever a member (or household lead) self-corrected their own visit's arrival — the same per-field-provenance failure #1631 already tracks for the staff PATCH route (an existence-claim source overwritten by an unconditional field write), just the mirror direction. Before this backfill, that could only misfire on roster-marked (`LEAD_MARKED`) visits; after it, every staff walk-in is `LEAD_MARKED` too, so a member nudging their own walk-in's arrival would have flipped it from trends-excluded to counted, over a much larger population.

#1668 dropped that restamp: `arrivedVia` is now left as-is on a manual correction (`departedVia` still becomes `WEB`, since trends' exclusion filter reads only `arrivedVia`). It also added two pins to `facilityTrendsAPI.integration.test.ts` — a self-corrected `LEAD_MARKED` visit stays excluded from trends hours, and a self-corrected `SCANNER` visit still counts as the control. Merged into this branch at `94fba256`.

Consequence for this PR: the backfill no longer widens any exposure. It enlarges the `LEAD_MARKED` population, and that population is exactly what #1668 now protects from member self-correction — the two changes are complementary. No test in this PR depended on the old restamp behavior (the #1632 test drives the insert and DELETE routes only, never the manual PATCH), and the docs rule added here describes the trends exclusion itself, not the restamp. #1624/#1631 remain open for the staff-PATCH mirror of the same per-field-provenance issue.

## Docs

`docs/rules/attendance-checkin.md` now carries the standing rule (added on review request, provenance trimmed per the register standard §3.8 in 9af8a9ad): staff-asserted presence is not facility hours, with the adopted-walk-in carve-out, tagged [Decision]. The backfill's retroactive-drop consequence stays recorded here and on issue #1632, not in the register.

### DIFF
--- a/checkin-app/prisma/migrations/20260818120000_staff_entry_lead_marked_backfill/migration.sql
+++ b/checkin-app/prisma/migrations/20260818120000_staff_entry_lead_marked_backfill/migration.sql
@@ -1,0 +1,15 @@
+-- #1632: backfill staff walk-in visits (insert route's CREATE audit rows
+-- tagged newData.type='staff_entry') from arrivedVia/departedVia=WEB to
+-- LEAD_MARKED, matching the post-#1558 rule. Idempotent: only touches rows
+-- still on WEB. Rationale/consequence: PR body.
+UPDATE "Visit"
+SET "arrivedVia" = CASE WHEN "arrivedVia" = 'WEB' THEN 'LEAD_MARKED' ELSE "arrivedVia" END::"VisitSource",
+    "departedVia" = CASE WHEN "departedVia" = 'WEB' THEN 'LEAD_MARKED' ELSE "departedVia" END::"VisitSource"
+WHERE "id" IN (
+    SELECT "affectedEntityId"
+    FROM "AuditLog"
+    WHERE "tableName" = 'Visit'
+      AND "action" = 'CREATE'
+      AND "newData"->>'type' = 'staff_entry'
+)
+AND ("arrivedVia" = 'WEB' OR "departedVia" = 'WEB');
--- a/checkin-app/src/app/__tests__/adminVisitsAPI.integration.test.ts
+++ b/checkin-app/src/app/__tests__/adminVisitsAPI.integration.test.ts
@@ -439,14 +439,14 @@ describe('Admin Visits API Integration Tests', () => {
             expect(res.status).toBe(403);
         });
 
-        it('creates a closed WEB visit for another person and audits actor + subject', async () => {
+        it('creates a closed LEAD_MARKED visit for another person and audits actor + subject', async () => {
             (getServerSession as jest.Mock).mockResolvedValue({ user: { id: testAdminId, isSysadmin: true } });
             const res = await post({ personId: testUserId, arrivedAt: arrived.toISOString(), departedAt: departed.toISOString() });
 
             expect(res.status).toBe(201);
             const { visit } = await res.json();
             const stored = await prisma.visit.findUnique({ where: { id: visit.id } });
-            expect(stored).toMatchObject({ personId: testUserId, arrivedVia: 'WEB', departedVia: 'WEB' });
+            expect(stored).toMatchObject({ personId: testUserId, arrivedVia: 'LEAD_MARKED', departedVia: 'LEAD_MARKED' });
             expect(stored?.departedAt).toBeInstanceOf(Date);
 
             const audit = await prisma.auditLog.findFirst({
--- a/checkin-app/src/app/__tests__/facilityTrendsAPI.integration.test.ts
+++ b/checkin-app/src/app/__tests__/facilityTrendsAPI.integration.test.ts
@@ -12,11 +12,14 @@
  * present" visits, plus its pre-split `SYSTEM` spelling) and the two cases it must
  * not swallow — an untagged arrival, and a real visit whose time staff corrected
  * (driven through the PATCH route, since a hand-built fixture cannot catch a wrong
- * stamp) — the `programId` filter, and the invalid-period 400.
+ * stamp) — the `programId` filter, a real staff walk-in insert (#1632: the
+ * route the LEAD_MARKED backfill's row-set is keyed on) staying excluded, and
+ * the invalid-period 400.
  */
 
 import { GET } from '@/app/api/facility/trends/route';
 import { PATCH } from '@/app/api/facility/visits/route';
+import { POST as INSERT_POST } from '@/app/api/facility/visits/insert/route';
 import prisma from '@/lib/prisma';
 import { getServerSession } from 'next-auth/next';
 import { getAppSettings } from '@/lib/appSettings';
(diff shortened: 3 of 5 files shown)


jee7s:
Adversarial cross-review flagged two things worth surfacing explicitly rather than folding in silently.

**Ratification needed — the route change.** The insert route still stamped WEB (#1558 deferred its Item A to #1624). Landing the decided backfill forces that deferred route change or the discontinuity reappears with the next walk-in — confirm the route hunk stays, or say the word and it's reverted pending #1624.

**FYI — enlarged #1631 exposure, not fixed here.** `attendance/manual/[id]/route.ts:117` unconditionally restamps `arrivedVia: "WEB"` when a member (or household lead) self-corrects their own visit's arrival — the same per-field-provenance failure #1631 already tracks for the staff PATCH route, just the mirror direction. Before this backfill, only roster-marked visits were exposed to that flip; after it, every staff walk-in is LEAD_MARKED too, so a member nudging their own walk-in's arrival now flips it from trends-excluded to counted, over a much larger population. Root fix belongs to #1624/#1631 — deliberately not addressed here, and no test pins the current (wrong) behavior.

jee7s:
Reviewed `b9fe2dac38` against #1632 and the 2026-08-18 backfill decision.

Backfill keyed on `staff_entry` CREATE audits is the decided row set. The insert route **must** move to `LEAD_MARKED` in this PR — leaving it on `WEB` recreates the step the next walk-in. That’s #1632 coherence, not a silent #1624 reopen. Leave the hunk.

The `manual/[id]` restamp enlargement is already #1631 (shrunk to that route). Don’t fix it here.

Missing from the PR: a Procedure line in `docs/rules/attendance-checkin.md` that staff-asserted (`LEAD_MARKED`) walk-ins are not facility hours. The trends filter is now a standing rule, not just a query. Extract it here; a later #1624 rewrite will need it.

jee7s:
Ruling taken, doc rule extracted.

**1 — insert route stays (ratified).** The `LEAD_MARKED` hunk in `facility/visits/insert` stays as-is. Understood: it's #1632 coherence, not a re-opening of #1624 — the backfill only holds if the next staff walk-in is written the way the backfilled ones now read. No further change, and nothing deferred back to #1624 on this route.

**2 — restamp enlargement stays with #1631.** Agreed, and it matches what the PR body already says: the `manual/[id]` unconditional `arrivedVia: "WEB"` restamp is #1631's (shrunk to that route), not this PR's. Left untouched, and no test here pins the current wrong behavior.

**3 — ACTION done: Procedure line added.** `a876c26c` — `docs/rules/attendance-checkin.md`, under Procedure › The visit record, immediately before the existing facility-hours split rule:

> - Staff-asserted presence is not facility hours: a visit recorded for somebody
>   else, or a roster mark, states the window it was asserted for rather than a
>   measured stay, so the trends leave it out. A walk-in that a roster mark adopts
>   keeps its badged times and still counts. The 2026-08 backfill brought
>   historical staff entries into line with this, so hours for past periods dropped
>   retroactively.  [Decision]

Written to the register's house style per `DOCUMENTATION_STANDARD.md` §3.6/§3.8 — business language, no enum spellings (`LEAD_MARKED`/`SYSTEM` are the mechanism, and the file carries no code identifiers in rule prose), `[Decision]` tag since no policy or principle governs it. Two things kept in deliberately: the adopted-walk-in carve-out the trends filter actually has (§3.8 — state the carve-outs or the register does damage), and the retroactive drop, which is the consequence `jee7s` signed off on and the reason past-period numbers no longer reconcile.

This supersedes the PR body's **Deferred** section, which said no doc note was needed — that read was wrong: the exclusion is a standing rule, not just a query.

Docs-only commit, no TS touched. Not merging.

jee7s:
Reviewed `a876c26c81` against #1632 and `docs/DOCUMENTATION_STANDARD.md` §3.8.

The standing rule and the adopted-walk-in carve-out are what the last pass asked for. The last sentence is change-history:

> The 2026-08 backfill brought historical staff entries into line with this, so hours for past periods dropped retroactively.

Rules files state constraints, not how they came to be — no dates, no PR/issue numbers. After this migration, staff-asserted visits are not hours. The retroactive drop is a one-time reporting consequence, not something later work can violate. Drop that sentence; keep the rule and the carve-out.

jee7s:
Confirmed against `DOCUMENTATION_STANDARD.md` §3.8 ("No PR numbers, issue numbers, or dates... Change history lives in git and in the PR record") — the backfill sentence was the only date anywhere in `docs/rules/`, so git plus this PR record is already the standard's designated home for it. Dropped in `9af8a9ad`; the constraint and the adopted-walk-in carve-out are untouched.

jee7s:
Merge conflict resolved — merged `origin/main` (post-#1665/#1668) at `94fba256`.

One conflicted file, `checkin-app/src/app/__tests__/facilityTrendsAPI.integration.test.ts`, keep-both on each hunk: the docblock/import block, and an adjacent-hunk collision where #1668's two self-correction tests and this branch's #1632 walk-in test insert at the same anchor and shared a cleanup tail — all three tests kept, each with its own teardown. Net diff vs `main` is unchanged from this PR's feature (migration + insert route + tests + docs rule); nothing of main reverted.

Known-gap update: the PR body's "Known gap" section is now marked closed by #1668, which dropped the `arrivedVia: "WEB"` restamp on manual corrections. No test here depended on the old behavior (the #1632 test drives the insert and DELETE routes only, never the manual PATCH), and the docs rule describes the trends exclusion, not the restamp. The backfill and #1668 are complementary — the backfill enlarges the `LEAD_MARKED` population that #1668 now protects.

`tsc --noEmit` clean, eslint clean on the resolved file, manual-attendance unit suite 23/23. Integration is CI-only.

thpr (approved):
Migration row-set correctly scoped via AuditLog `staff_entry` tag + `WEB` guard — idempotent, matches the established `CASE...END::"VisitSource"` cast pattern. Route change to `LEAD_MARKED` is the only coherent companion to the backfill. Integration test drives both the trends exclusion and the significance weight on delete through the real routes. Register entry clean, provenance trimmed per §3.8.

PR #1668, opened 2026-08-18
fix(attendance): manual correction keeps arrivedVia source (#1631)

### LINKED ISSUE #1631: fix(attendance): manual visit correction must not restamp arrivedVia to WEB

jee7s:
Fixes #1631.

## The shrunk ruling

> **v1.2 scope, shrunk.**
>
> Staff `PATCH /api/facility/visits` already leaves `arrivedVia` alone — the filed promotion does not reproduce there.
>
> This issue is only: `PATCH /api/attendance/manual/[id]` must not restamp `arrivedVia: "WEB"` on a correction. A `LEAD_MARKED` (or `SCANNER`) arrival stays that source so trends do not start counting it.
>
> Not in scope: splitting `VisitSource` (#1624).
>
> 🤖 Posted by Grok on behalf of @jee7s

## Why the fix is confined to this route

`PATCH /api/facility/visits` (staff correction) was checked and already leaves `arrivedVia` untouched — verified by reading the route, not touched here. `PATCH /api/attendance/manual/[id]` (member self-correction / household-lead correction) was the one route still doing `data: { arrivedAt: nextArrived, arrivedVia: "WEB" }` on every arrival edit, unconditionally overwriting whatever source was on the row.

## Trends-promotion mechanics

`facility/trends`'s exclusion filter (`route.ts` ~L79-82) reads **only** `arrivedVia`:

```ts
OR: [
    { arrivedVia: null },
    { arrivedVia: { notIn: ["LEAD_MARKED", "SYSTEM"] } },
],
```

A `LEAD_MARKED` arrival (a roster mark / staff-asserted presence — see the `docs/rules/attendance-checkin.md` rule added by #1666: "staff-asserted presence is not facility hours") is deliberately excluded. Before this fix, the moment a member or their household lead corrected that visit's arrival time through this route, the unconditional `arrivedVia: "WEB"` stamp promoted it past the filter and it silently started counting as measured facility hours — exactly the mechanism the parent issue (#1624's sibling) described, now confirmed to live specifically in this route.

Fix: drop the `arrivedVia` key from the update entirely when only `arrivedAt` is being corrected, so the existing source (`LEAD_MARKED`, `SCANNER`, `WEB`, etc.) survives untouched.

**`departedVia` is deliberately left restamping to `WEB`.** I checked whether the same promotion mechanism applies there: the trends filter above reads `arrivedVia` only, never `departedVia`, and no other exclusion/counting path in the codebase reads it either (`significance.ts` reads the *old* pre-update visit, `attendanceConflicts.ts` only displays it, audit logging only reads old values). So restamping `departedVia` to `WEB` on a departure correction cannot promote anything — it's inert with respect to this bug, and mirroring the fix there would be an unrequested, unjustified behavior change. Left as-is.

**Audit/significance unaffected.** `editSignificance`/`deleteSignificance` (`lib/visit/significance.ts`) are called with the visit's state *before* the update, so they already weighted corrections by the true prior source — the restamp never fed into that calculation. The audit log's `oldData` also reads pre-update values. Nothing here changes.

## Tests

- Unit (`api/attendance/manual/[id]/__tests__/route.test.ts`): updated the existing "applies a small edit" pin to assert no `arrivedVia` key is written; added a case pinning that a `LEAD_MARKED` arrival keeps that source through a correction (response body included), and a case confirming `departedVia` still restamps to `WEB` on a departure correction.
- Integration (`app/__tests__/facilityTrendsAPI.integration.test.ts`, CI-only, drives the real `PATCH` route per the file's existing pattern): added a case pinning that correcting a `LEAD_MARKED` visit's arrival through the self-correction route leaves it excluded from trends totals (0h, before and after), and a control case showing a `SCANNER` visit's self-correction is otherwise unchanged — still applies and still counts.

Ran locally: `tsc --noEmit` clean, lint clean, and the full non-integration unit suite for the touched files (27 tests) passes. Integration tests require Docker Postgres and run in CI only, per this repo's convention.

## Out of scope

Splitting `VisitSource` (#1624) — the "one column, two facts" root cause the original issue named — is explicitly not addressed here, per the shrunk ruling. This PR is the narrow stopgap: it stops the one live promotion path without touching the `VisitSource` model.

### DIFF
--- a/checkin-app/src/app/__tests__/facilityTrendsAPI.integration.test.ts
+++ b/checkin-app/src/app/__tests__/facilityTrendsAPI.integration.test.ts
@@ -9,14 +9,16 @@
  * path: bucketing visits by period, the volunteer/participant split (by program
  * enrollment — ProgramParticipant, NOT age), structured (event-associated) vs
  * unstructured hours, the `arrivedVia=LEAD_MARKED` exclusion (synthetic "marked
- * present" visits, plus its pre-split `SYSTEM` spelling) and the two cases it must
- * not swallow — an untagged arrival, and a real visit whose time staff corrected
- * (driven through the PATCH route, since a hand-built fixture cannot catch a wrong
- * stamp) — the `programId` filter, and the invalid-period 400.
+ * present" visits, plus its pre-split `SYSTEM` spelling) and the cases it must
+ * not swallow — an untagged arrival, a real visit whose time staff corrected via
+ * the staff route, and a member's self-correction (#1631) — plus the `programId`
+ * filter and the invalid-period 400. Corrections are driven through the real
+ * PATCH routes, since a hand-built fixture cannot catch a wrong stamp.
  */
 
 import { GET } from '@/app/api/facility/trends/route';
 import { PATCH } from '@/app/api/facility/visits/route';
+import { PATCH as MANUAL_PATCH } from '@/app/api/attendance/manual/[id]/route';
 import prisma from '@/lib/prisma';
 import { getServerSession } from 'next-auth/next';
 import { getAppSettings } from '@/lib/appSettings';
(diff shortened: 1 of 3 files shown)


thpr (approved):
Fix is correct and well-scoped. `arrivedVia` no longer restamped on arrival correction — prevents LEAD_MARKED promotion past the trends exclusion filter. The `departedVia` asymmetry is justified (trends never reads it). Significance/audit unaffected (both read pre-update state). Tests pin the regression and control case.

Issue #254, opened 2026-06-12
Cross-participant race: force-close updateMany not covered by per-participant advisory lock


thpr:
## Context

Flagged during review of #250 (per-participant advisory lock for concurrent same-participant scans). #250 fixes **same-participant** races; this is the remaining **cross-participant** facility-state hole. Pre-existing — not introduced by #250.

## Problem

In [scan-service.ts](checkin-app/src/lib/scan-service.ts) (the last-keyholder force-close path, ~line 118), an `updateMany` closes *everyone's* open visits when the keyholder checks out. But the scan route's advisory lock — `pg_advisory_xact_lock(participant.id)` — is keyed per participant.

A non-keyholder checking in concurrently holds a *different* lock key, so the two transactions don't contend. Sequence:

1. Keyholder scan begins force-close (`updateMany` closing all open visits).
2. Non-keyholder scan (different lock key) creates a fresh open visit.
3. Depending on interleave, the new visit can survive the force-close — the participant ends up **checked into a just-closed facility**.

## Why it's not fixed by #250

#250's lock serializes only scans for the **same** participant. The force-close touches global facility state under a per-participant key, so cross-participant invariants remain unprotected.

## Possible directions

- A facility-level advisory lock (fixed key) taken by the force-close path *and* by any check-in that depends on facility-open state, or
- A partial unique index / state check that rejects a check-in created against a facility being closed, or
- Re-validate facility-open state inside the check-in transaction after acquiring the lock.

Needs design — noting so the gap is tracked rather than assumed closed by #250.

jee7s:
**PROMOTE** — reproduces on `origin/main`; only the line reference drifted.

The force-close `updateMany` the issue cites at `scan-service.ts` ~line 118 is now `closeAllOpenVisits` at `checkin-app/src/lib/scan-service.ts:161-162` (called from :135 and :198). It still filters on `departedAt`/`deletedAt` alone with no facility-level lock. Meanwhile every advisory lock in the tree is keyed per person, not per facility: `api/scan/route.ts:111` (`participant.id`), `api/attendance/route.ts:203`, `api/attendance/manual/route.ts:83`, `api/attendance/manual/[id]/route.ts:98,168`, `api/facility/visits/route.ts:67,140`. So the cross-participant hole is untouched.

[...]

PR #1669, opened 2026-08-18
feat(kiosk): explicit force-close confirm — 15s countdown + token (#1347 slice)

jee7s:
Slice of #1347. Implements the ruling on the epic's last comment ("Phase 1 leftovers, decided.", 2026-08-18) — **not** the whole epic.

## The ruling, implemented

> - Confirm token across an outage: **honor** the pre-outage confirm. The queued second badge still closes.
> - Visible countdown: **15 seconds.**
> - Manual open visit while the server thinks closed: **match C** (hold, then project when a keyholder Visit exists). Same as a scan.
> - Explicit confirm (countdown + token) ships in **its own PR before** the outbox/replay cut.

**(a) Honor across an outage.** The confirm has **no elapsed-time gate server-side**. Validity is by issuance: the warning mints a token, and any later scan echoing it closes the facility whenever it lands — seconds later or after a multi-hour outage. This overrules §5.23's own "recommended no, expire it", so D4's "the confirm badge is its own queued event" stands as written and the §2 contract row that called a queued confirm expired is superseded (both updated in this PR).

**(b) 15-second visible countdown.** The kiosk banner carries a ticking counter; the countdown *is* the token's lifetime and it is enforced at the edge — `client.py` drops the token when the countdown lapses, and the banner clears with it. The server keeps no timer, which is exactly what makes (a) work.

**(c) Manual open while closed → match C.** Recorded on §5.26, **not built here**. On `main` the manual open backfill (`attendance/manual/route.ts`) and the scan check-in path (`processCheckin`) *already answer identically* — both return the same "Facility is closed. A Keyholder must check in first." 403. The divergence §5.26 describes only appears once the scan path becomes C, which happens in the queue/substrate cut. C's "hold, then project" half requires the §6 Stage-2 substrate + re-runnable projection, which this slice explicitly excludes; a hold with no projector would silently swallow the backfill, which is worse than today's refusal. So the decision is written down and lands with the substrate, on both paths together.

**(d)** Ships alone, before #1667.

## Mechanism

- **State: `Visit.forceCloseToken`**, beside the existing `forceCloseWarnedAt` — the §5.23 shape extended, not a parallel one. `forceCloseWarnedAt` stays as the "when we showed it" audit stamp; the token is what decides.
- **Warning** (`400 {type:"warning"}`) now carries `forceCloseToken` + `confirmSeconds`. **Confirm** = `POST /api/scan` with `forceCloseToken` matching the token on the caller's own open visit.
- **Single use.** The warning mints a *fresh* token every time, so an unconfirmed countdown can never be redeemed later; the confirm closes the visit, so the token matches nothing afterwards.
- **Debounce.** A scan carrying a **live** token is exempt from the 3s debounce — without that, the debounce swallows the confirm and Q16's dead `[0s,3s)` front of the window survives into the new mechanism (§5.16: "the next scan ends the countdown — that is the confirm"). The exemption is keyed on the token still matching an *open* visit, so the classic hazard (one badge swipe read twice, milliseconds apart) is unchanged: the first read spends the token and closes, the second matches nothing and debounces as before.
- **Web surface.** `/` self-check-in round-trips the token too (no ticking UI, the message states the window) — otherwise an explicit-confirm-only server would leave that path warning forever with no way to confirm.

## Design sections honored

- **§5.23 (B2)** — the forced choice, taken as "warning issues a short-lived confirm token"; all three residual opens (a)/(b)/(c) decided and recorded.
- **§5.16 (Q16)** — the confirm is explicit state with a visible countdown; the debounce no longer infers close from log spacing, and no longer swallows the confirm.
- **§0 invariant 1** — the debounce exception is preserved for noise, and force-close confirm is no longer inferred from that window.
- **§4 gate "Before Phase 1 — B2 / §5.23"** — this is that gate.
- Doc updated: §5 Q16, Q23, Q26, the §2 `/api/scan` contract block and its `{type:"warning"}` client row.

## Interaction with #1667 (offline store-and-replay)

**#1667 rebases on this after it merges.** No file on that branch is touched here.

#1667's F1 fix parks *every* replayed last-keyholder checkout as `force_close_review` and never reads or writes the force-close state — a blanket ban, taken while §5.23(a) was still open and recommended "expire it". The ruling went the other way, and this PR is what makes the safe version possible: a replayed second badge carrying a **valid pre-outage token** should close, because the token proves a human saw the warning and confirmed at the door. On rebase, #1667's park should narrow from "any replayed last-keyholder checkout" to "any replayed last-keyholder checkout **without** a matching `forceCloseToken`". Everything else in F1 stands.

Mechanical conflicts to expect on that rebase: `processCheckout`'s parameter list (both PRs add a 5th argument — `confirmToken` here, `replayEventId` there), the debounce block in `route.ts`, `post_scan`/`handle_scan` in `client.py`, and the same two design-doc rows.

## Tests

Local, all green:

- `checkin-app`: `npx tsc --noEmit` clean; `npm run lint` clean; **full unit suite 280 suites / 2789 tests pass**.
  - `scanServiceKeyholderWarning.test.ts` — rewritten off the timestamp window: warn mints a token and returns it with `confirmSeconds`; matching token closes; **a token "minted days ago" still closes** (the (a) pin); no token re-warns with a *different* token; wrong token re-warns instead of closing.
  - `scanRoute.test.ts` — a live token passes the debounce and is forwarded to `processCheckout`; a spent/unknown token still debounces (the double-read pin); a non-string token is treated as absent and never hits the DB.
  - `attendance/manual/[id]/route.test.ts` — the outbound-stripper pin extended to `forceCloseToken` (tier `internal`, must not reach a browser).
- `client/` (`python -m unittest discover -v`, venv + `requirements.txt`): **10/10 pass**, including `post_scan` sending the field only when a token is held, `take_confirm` being single-use and expiring with the countdown, and the warning→countdown→confirm sequence through `handle_scan`.
- **CI-only** (no local DB): `scanCausalChain.integration.test.ts` updated to the token and given a new case — warn, then badge again **inside** the 3s debounce with the token, and the facility closes.

## Migration

`20260818150000_visit_force_close_token` — one nullable column, `ADD COLUMN IF NOT EXISTS` (retry-safe), expand-only: the previous release never names it, so its reads are unaffected.

## Not in this slice

Health state machine and recovery ladder (§3), the offline queue and replay path (§2 — that is #1667), the §6 reconciliation substrate and deferred projection C, the server-side DLQ and `unsynced-scans` review panel, #254's facility-level force-close lock, and §5.24–25 (cannot-vouch display, minimize-by-surface) which stay Phase 2. Match-C for the manual open backfill is decided and recorded but lands with the substrate, per above.

### DIFF
--- a/checkin-app/prisma/migrations/20260818150000_visit_force_close_token/migration.sql
+++ b/checkin-app/prisma/migrations/20260818150000_visit_force_close_token/migration.sql
@@ -1,0 +1,4 @@
+-- Expand-only: one nullable column. The previous release never names it in a
+-- SELECT, so its reads are unaffected (rule 1). IF NOT EXISTS keeps the file
+-- retry-safe if the deploy is re-run after a partial failure.
+ALTER TABLE "Visit" ADD COLUMN IF NOT EXISTS "forceCloseToken" TEXT;
--- a/checkin-app/prisma/schema.prisma
+++ b/checkin-app/prisma/schema.prisma
@@ -1116,10 +1116,15 @@ model Visit {
   deletedById       Int?
 
   /// When the kiosk last showed this keyholder the "others are still here"
-  /// force-close warning. A force close only proceeds if the scan follows a
-  /// fresh stamp, so an ordinary double-tap cannot close the building.
+  /// force-close warning. Audit stamp only — the confirm is the token below.
   /// @sensitivity:internal
   forceCloseWarnedAt DateTime?
+
+  /// Confirm token minted with that warning and returned to the kiosk. Only a
+  /// scan echoing it force-closes, so an ordinary double-tap cannot; validity
+  /// is by issuance, not elapsed time, so a confirm queued through an outage
+  /// still closes when it lands (§5.23a). @sensitivity:internal
+  forceCloseToken   String?
 
   person Person @relation(fields: [personId], references: [id])
   event  Event?      @relation(fields: [associatedEventId], references: [id])
--- a/checkin-app/src/app/__tests__/scanCausalChain.integration.test.ts
+++ b/checkin-app/src/app/__tests__/scanCausalChain.integration.test.ts
@@ -31,10 +31,10 @@ jest.mock('@/lib/logger', () => ({
 const TAG = 'scan-causal-test';
 const flush = () => new Promise(r => setImmediate(r));
 
-function scanReq(participantId: number) {
+function scanReq(participantId: number, forceCloseToken?: string) {
     return new Request('http://localhost/api/scan', {
         method: 'POST',
-        body: JSON.stringify({ participantId }),
+        body: JSON.stringify({ participantId, ...(forceCloseToken ? { forceCloseToken } : {}) }),
     }) as unknown as import('next/server').NextRequest;
 }
 
@@ -95,14 +95,16 @@ describe('Scan causal chain — last isKeyholder closes facility', () => {
         expect(open).toBe(0);
     });
 
-    it('force-closes (departing everyone) when the scan follows a fresh warning, then sends post-event emails', async () => {
+    it('force-closes (departing everyone) when the scan echoes the confirm token, then sends post-event emails', async () => {
         const kVisit = await prisma.visit.create({
-            // The warning was shown 5s ago (within the 60s window) → confirmed.
-            data: { personId: isKeyholder.id, arrivedAt: new Date(), forceCloseWarnedAt: new Date(Date.now() - 5000) },
+            data: {
+                personId: isKeyholder.id, arrivedAt: new Date(),
+                forceCloseWarnedAt: new Date(Date.now() - 5000), forceCloseToken: 'confirm-me',
+            },
         });
         await prisma.visit.create({ data: { personId: normal.id, arrivedAt: new Date() } });
 
-        const res = await processCheckout(isKeyholder, kVisit.id, 'kiosk');
+        const res = await processCheckout(isKeyholder, kVisit.id, 'kiosk', undefined, 'confirm-me');
         const json = await res.json();
         expect(json.facilityClosed).toBe(true);
 
@@ -121,7 +123,7 @@ describe('Scan causal chain — last isKeyholder closes facility', () => {
         const beforeClose = Date.now();
 
         const kVisit = await prisma.visit.create({
-            data: { personId: isKeyholder.id, arrivedAt: new Date(), forceCloseWarnedAt: new Date(Date.now() - 5000) },
+            data: { personId: isKeyholder.id, arrivedAt: new Date(), forceCloseToken: 'confirm-me' },
         });
         const staleVisit = await prisma.visit.create({ data: { personId: normal.id, arrivedAt: stale } });
         // Open AND tombstoned — the one-open-visit index only covers live rows.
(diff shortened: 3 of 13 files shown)


thpr (commented):
**Observation (non-blocking):** The `forceCloseToken` column is not cleared on confirm — after a successful force-close, the token stays on the now-closed visit record. Safe today because all lookup paths filter `departedAt: null`, but a future query that forgets that filter could match stale tokens. Worth knowing if someone adds a Visit query that doesn't filter on open.

---

jee7s:
Applied — the confirm path now nulls `forceCloseToken` and `forceCloseWarnedAt` before the visit departs, so no redeemable state survives on the closed row (placed ahead of `closeAllOpenVisits`, which sweeps the keyholder's own visit too). The existing confirm test asserts the clearing update. `tsc --noEmit` clean; `scanServiceKeyholderWarning` + `scanRoute` 18/18 pass. `98bed365`

thpr (dismissed):
Clean mechanism, sound invariants, thorough test coverage. The new commit clears the spent token — observation addressed.

One minor note left on the table: the token comparison in `scan-service.ts:110` uses `===` rather than `crypto.timingSafeEqual`. Not practically exploitable (UUID entropy, single-use, network jitter, physical-access-only threat model), but `timingSafeEqual` would close the theoretical side-channel for free.

---

jee7s:
Rebuilt on top of `main` after #1664 and #1670 landed — conflicts resolved and the review nit applied. Head is now `28009c5e`.

**Conflicts (4), all keep-both — none of main's logic is reverted:**

| File | Resolution |
|---|---|
| `prisma/schema.prisma` | `Visit` keeps both new columns: this PR's `forceCloseToken` next to #1670's `supervisionWarnedAt`. Both migrations are present and ordered (`20260818120000_visit_supervision_warned_at`, then `20260818150000_visit_force_close_token`). |
| `src/lib/scan-service.ts` | Import block only. The supervision ladder and the force-close token confirm are separate interrupts and compose unchanged — the ladder's skip-while-closing precondition (`if (!facilityClosed)`) survives, so a force-close still bypasses it. |
| `src/security/generated/classifications.ts` | Regenerated with `prisma generate`, not hand-merged. All three `Visit` fields classify `internal`. |
| `client/test_client.py` | Both test classes kept. #1670's `FakeBackend.post_scan` / `FakeState` needed adapting to `handle_scan`'s new token argument, or its banner tests would have errored against this PR's call signature. |

`client.py` auto-merged cleanly: #1670's amber supervision banner lives in the success branch, this PR's countdown lives in the warning branch.

**thpr's nit** — `scan-service.ts:131-136`: the confirm-token comparison was `===`, now timing-safe. Both sides are hashed to a fixed length before `timingSafeEqual`, which is the house idiom (`cronAuth.ts`, the Shopify webhook verify, `outreach/unsubscribeToken.ts`). That removes the length-mismatch throw entirely rather than branching around it, and leaks no length information. The existing "token does not match" test already passes a different-length token, so it covers the throw case — I added a comment there so a future tidy-up doesn't equalise the lengths and silently drop the guard.

**Verification:** `tsc --noEmit` clean; eslint clean; jest `scanService|scanRoute|supervision` 44/44; the suites both sides touched (`significance|attendance/manual|getFullAttendance|participantPiiMinimization`) 52/52; Python `unittest discover` 21/21, covering this PR's `TestForceCloseConfirm` and #1670's `TestSupervisionWarningBanner` together. Net diff vs `origin/main` is this PR's 13 files only.

thpr (approved):
Clean mechanism, thorough tests, prior nits addressed. No findings.

Verified: timing-safe compare (SHA-256 → `timingSafeEqual`, house idiom), single-use token lifecycle (mint-fresh on re-warn, null before fetch/pop-under-lock), debounce bypass keyed on live token only, token+stamp cleared on confirm before departure, expand-only migration, `internal` sensitivity classification with test pin, monotonic countdown in Python client, `confirmSeconds` clamped both sides.

---

Issue #1442, opened 2026-08-01
Youth school-enrollment status is not captured (policy divergence; moved Programs→Membership)


thpr:
**Backlog item M25** — from the backlog register (`docs/backlog/INDEX.md`), source: policy divergence (DIVERGENCE_HANDOFF).

Youth school-enrollment status is not captured (policy divergence; moved Programs→Membership) — POLICY SHORTFALL: policy requires a youth without a high-school diploma or equivalent to be enrolled in school; the app has no field for it, so the requirement cannot be checked.

- Origin: CREATE · Readiness: NEEDS-DESIGN · Size: S

_Auto-created from the backlog INDEX; see the doc + `backlog/sources/` reconciliation for context._

jee7s:
**HOLD** — "add a field" is the solution, not the problem, and the enforcement that gives the field meaning is undecided.

To promote, answer:

1. Who supplies the answer — the youth's household lead at intake, or staff during review?
2. Is it a self-attested boolean, or does it require a document or a named school?
3. What happens when the answer is no — is membership or programme enrolment refused, flagged for board review, or only recorded?
4. Does it need re-confirming each membership year, or is it captured once?
5. The rule applies only to a youth without a diploma — how does the app learn that a youth has one?

The youth-eligibility predicates this would join already live in #167's design (`checkin-app/docs/designs/167-youth-enrollment-rules.md`); the answers should be consistent with them.

---
_Posted by Claude._

thpr:
## Register handover (from the #1528 Class-A audit)

`docs/rules/programs.md` carried this as an entry in its
`## Policy requirements not yet enforced here` appendix — the entry read
"School enrollment for youth is not captured." **That entry is being deleted**,
so this issue is now the only record of the gap.

Deleted because the appendix failed `DOCUMENTATION_STANDARD.md` §3.7's own test:
a divergence is where the app **does** model something and models it weaker than
policy. There is no field, so there is no rule to qualify — a missing feature
belongs in the tracker.

**When this is built, the register gains rules.** Note the move recorded in this
issue's title: the rules land in `docs/rules/people-households.md` (Procedure,
under `### Identity and age`), not in `programs.md` where the stale entry sat —
school enrollment is a fact about a person, not about a program. One bullet per
rule, tagged per §3.6:

- who the field is captured for, and when it is asked
- what the app does when it is missing or says "not enrolled" — gate, warn, or
[...]

Issue #300, opened 2026-06-27
isMinor null-DOB defaults to ADULT — fails open in two-deep safety check


thpr:
## Problem

`isMinor(dob)` in `checkin-app/src/lib/time.ts:56` returns `false` for null/undefined DOB → an unknown-DOB person is classified **adult**. In the two-deep youth-protection calc this fails **open**: an unknown-DOB person is (a) counted as a supervising adult and (b) dropped from the unaccompanied-minor set — both directions weaken the safety check and can hide an `isTwoDeepViolation`.

## Callers (full blast radius — only 3)

| # | Site | Context | null→adult (today) | null→minor would | Wants |
|---|------|---------|--------------------|------------------|-------|
| 1 | `getFullAttendance.ts:44` | **Safety** — two-deep calc | unknown counted as adult supervisor, dropped from students → **masks violation** (fail open) | unknown counts as unaccompanied student, never supervisor → **fail closed** | **null→minor** |
| 2 | `app/admin/participants/new/page.tsx:40` | **Display/form UX** — toggles "student" fields; `dob` starts `""` | blank form = not student (correct, quiet) | blank form shows student fields immediately → annoying | null→adult |
| 3 | `app/api/profile/onboarding-status/route.ts:28` | **Age-gate** — `needsPhone` (minors exempt, #169) | unknown asked for phone (harmless) | unknown silently exempt → mild data gap | null→adult |

Key finding: only **1 of 3** callers wants fail-closed. The other two genuinely want null→adult. A global flip is wrong.

## Options

| | Option | Diff | Risk | Verdict |
|---|--------|------|------|---------|
| a | Flip global default null→minor | small | breaks callers 2 & 3 | ✗ over-reaches |
| b | Context-scoped default: `isMinor(dob, { unknownIs: 'minor' \| 'adult' })`, default `'adult'`; safety caller passes `'minor'` | **~3 lines** | low — only caller 1 changes | ✓ |
| c | Require/backfill DOB upstream, keep isMinor strict | large + migration | high; doesn't protect existing null rows | ✗ separate hygiene task |
| d | Tri-state `'minor'\|'adult'\|'unknown'` | large — every caller rewritten | medium | ✗ over-engineered; no caller needs a 3rd state |

## Recommendation: (b)

[...]

PR #1670, opened 2026-08-18
feat(attendance): supervising-adult test + close interrupt and banner (#1436, #1550)

### LINKED ISSUE #1436: Closing/departure guard must enforce two adults with a last youth, not keyholder-count
**Backlog item SA12** — from the backlog register (`docs/backlog/INDEX.md`), source: policy divergence (DIVERGENCE_HANDOFF).

Closing/departure guard must enforce two adults with a last youth, not keyholder-count (policy divergence, youth-supervision) — POLICY SHORTFALL: the closing guard watches keyholders (last keyholder is stopped + made to confirm), so a NON-keyholder adult can leave a youth with only one remaining adult and nothing interrupts. Policy requires two adults on site whenever a youth is. Distinct from AT6 (guards the last adult →0, not the drop below two →1) and AT9 (force-close race). Reports the room compliant when it is not.

- Origin: CREATE · Readiness: NEEDS-DESIGN · Size: M

_Auto-created from the backlog INDEX; see the doc + `backlog/sources/` reconciliation for context._

### LINKED ISSUE #1550: Supervision check counts bare adults; policy requires two unrelated non-student volunteers (policy divergence)
**Policy divergence**, surfaced by the domain-register audit (#1528). Recorded in
`docs/rules/attendance-checkin.md`'s `## Policy requirements not yet enforced
here` appendix, which is being retired — this issue is now the only record.

## The shortfall

The two-deep supervision check **counts adults present**. Policy requires two
adults who are:

- **volunteers** — adult presence alone does not satisfy it
- **not students** — a student is anyone enrolled in school through twelfth
  grade, which can run to 19, so student and youth are different sets
- **unrelated and from different households**

Any of those three can be false while the app reports the room compliant. Two
parents from one household, or an 18-year-old student volunteer plus one adult,
both read as two deep today.

— *Definitions Policy, Art. III, "Two Deep"*; *Event, Location and Keyholder
Policy, Art. III and Art. VI*

## Why this one is worth naming

`docs/DOCUMENTATION_STANDARD.md` §3.7 uses this exact case as its worked example
of a genuine divergence — "a supervision check counting bare adults where policy
requires two unrelated non-student volunteers" — and it was the only entry in the
audit with no issue behind it. It is also the one where the gap is a safety
posture rather than a missing feature: the app **does** model the check, and
models it weaker than the rule it is named after.

## What the app already has, and what it lacks

The register states the surrounding rules, and they bound the work:

- Two deep is **shown as a warning**. It does not stop anyone entering or
  checking in — "surfacing the shortfall to the people in the room is the whole
  of what the app does about it." Enforcement is not what is being asked for
  here; **correctness of the count** is.
- Household membership is known (`Person.householdId`), so the same-household
  test is reachable today.
- Volunteer status and student status are the harder halves. Volunteer has a
  taxonomy already; **student enrollment is not captured at all** (#1442), so
  the not-a-student test cannot be evaluated until that lands.

## Done when

- The count distinguishes at least: same-household pairs, and adults who are not
  volunteers.
- Whatever cannot be evaluated yet (the student test, pending #1442) is stated
  as a deliberate limit **at the rule** in `docs/rules/attendance-checkin.md`,
  not left implied.
- `docs/rules/attendance-checkin.md` Procedure gains or amends the rules under
  `### Supervision` to say what the check actually tests. The current line —
  "Two deep is shown as a warning. It does not stop anyone entering or checking
  in" — describes the response, not the test, and says nothing about who counts.

## Related, deliberately separate

- #1436 — the closing/departure guard is keyed on keyholder count rather than
  two-adults-with-a-last-youth. Same policy article, different moment.
- #1438 — the second-adult-volunteer requirement as a precondition of
  *scheduling*. That gates the calendar; this one is about the room.
- #1442 — youth school-enrollment status is not captured. Prerequisite for the
  not-a-student half of this.

jee7s:
Fixes #1436
Fixes #1550

Both issues were decided together in their v1.2 rulings and share **one helper**, so they ship as one PR.

## The rulings implemented

From #1436's v1.2 decisions (and the earlier policy answers on the same issue):

> - 3rd qualifying adult badges out → warning only (two remain). No confirm.
> - Next qualifying adult badges out → warning + confirm (15s countdown; same explicit confirm as #1347).
> - Same household counts as one adult.
> - Fires on every departure. Keyholder close-guard stays separate.
> - Youth check-in while below two: warning only, not blocked.

> Who counts as a supervising adult:
> - Adult with a **still-valid background clearance**.
> - **Not** a current `ProgramParticipant` on any program that is in session now (event overlapping now, or program dates covering now). Transition-year 18+ participants with a clearance do not count while they hold that seat; the same person can count later as a volunteer on a different program.
> - Not the volunteer-family fee flag.

And the follow-up ruling on this PR's open question (2026-08-19):

> **YES — the confirm rung is youth-gated.** With no youth present, that departure downgrades to the warning-only shape. The warn rung itself stays un-gated.

From #1550's v1.2 decisions:

> The banner uses the same supervising-adult test as the close interrupt … Not in this release: school-enrollment (#1442). "Student" here means program participant, not school youth.

## The shared helper

`checkin-app/src/lib/supervision.ts` is the single expression of the test. It answers **who is supervising the room right now**, and both surfaces consume it.

```ts
supervisingAdultVisits(db?, now?): Promise<Map<number /*visitId*/, number /*householdId*/>>
supervisingAdultCount(supervising, excludeVisitId?): number
youthIsPresent(db?): Promise<boolean>
MIN_SUPERVISING_ADULTS = 2
```

A person qualifies when all of these hold:

| Prong | How it is tested | Why that check |
|---|---|---|
| adult | `isDeclaredAdult`, else `isYouth(dob, { unknownIs: 'youth' })` | matches the existing fail-closed rule (#300) and the DoB-stripped 26+ case (#1165) |
| still-valid clearance | household `OrgMembership.status === 'ACTIVE'` **and** `bgValidUntilBoundary(lastBackgroundCheck, boardSettings) >= today` | `ACTIVE` is what encodes "background check cleared" (`lib/orgMembership`); the expiry half reuses the renewal rule rather than inventing a second one. No check on file is never valid — same fail-closed posture as unknown age. With `bgRecheckMonths` unset (the default) clearance has no expiry, so this does not silently empty the room. |
| no participant seat in session | `programParticipants` where `status: ACTIVE` and the program has an event bracketing now, **or** its own dates cover today, **or** it is open-ended and `RUNNING` | the ruling's two arms, plus two fail-safe arms: `lte` never matches null, so neither an unset `endAt` nor an unset `startAt` may read as "not in session" — a RUNNING programme missing either date would otherwise fail OPEN and let its participants count as supervisors |
| household | the map is keyed visit→household and counted as **distinct households** | "same household counts as ONE" |

Deliberately **not** in the test: `OrgMembership.isVolunteer` (the volunteer-family fee flag) and school enrollment.

Returning the visit→household map rather than a bare number is what lets the interrupt answer both of its questions from **one** query — "how many are here" and "how many would remain" — and it makes the household rule fall out for free.

**Where each surface consumes it**

- `lib/getFullAttendance.ts` — `isTwoDeepViolation` now reads `supervisingAdultCount(...) < MIN_SUPERVISING_ADULTS` instead of `adultVisits.length < 2`. That one flag feeds **both** banners (`/attendance/current` and the home page) and the `TWO_DEEP_VIOLATION` board notification, so #1550 is fixed at the single site that computes it. The supervision queries are awaited only when a youth is unaccompanied — the only case the flag can be true — so an adult-only room costs the attendance poll no extra queries, the same gate `processCheckin` uses. The "is this youth accompanied by an adult of their own household" prong is unchanged — different question, not in scope.
- `lib/scan-service.ts` — `processCheckout` runs the ladder on every departure; `processCheckin` warns a youth arriving below two.

## The interrupt ladder

`supervisionInterrupt()` in `scan-service.ts`. There is **no counter and no session state** — the rung is derived from the count itself:

```
before    = supervisingAdultCount(supervising)
remaining = supervisingAdultCount(supervising, activeVisitId)
```

- `remaining >= before` → this departure removes no supervising adult (a youth leaving, a non-qualifying adult leaving, or **the second adult of a household that stays represented**). Silent.
- `remaining > 2` → silent.
- `remaining === 2` → warning only. The checkout **completes**; the text rides the success response's new `warning` field.
- `remaining < 2` **and a youth has an open visit** → warning **and** confirm. 400 with `type: "warning"`, and `Visit.supervisionWarnedAt` is stamped. Only a scan following a fresh stamp inside 15s goes through.
- `remaining < 2` with **no youth in the building** → warning only, same shape as the rung above: the checkout completes, nothing is stamped, no second badge is asked for.

**The youth gate** (ruling of 2026-08-19) is on the confirm rung *only*. Two deep is owed whenever a youth is present, so an adult-only room locking up — the ordinary evening — has nothing to confirm, and a 400 raised where nothing is at stake is the fastest way to train badge-through on the night it matters. The warn rung stays un-gated: over-warning there is deliberate and costs a line of text.

`youthIsPresent()` is the adult prong inverted over the same open-visit population (`departedAt: null, deletedAt: null`, live person): not `isDeclaredAdult`, and not 18+ by DoB — unknown age fails closed as youth (#300), the same classification `processCheckin` uses for its arrival warning. Only the `isDeclaredAdult` half is a query filter; the age half stays with `isYouth` so the rule keeps one expression. The probe runs **inside the `remaining < 2` branch only**, so every quieter departure — including the whole warn rung — pays nothing for it.

This is why "3rd adult" needs no tracking: the third qualifying adult leaving is exactly the departure whose `remaining` lands on 2, and the next one is exactly the departure whose `remaining` falls below it. The household rule is enforced by the same arithmetic rather than by a second code path.

`supervisionWarnedAt` is its own column, not a reuse of `forceCloseWarnedAt`: the keyholder close-guard is a separate interrupt and both can be pending on one visit. The ladder is skipped entirely when the keyholder guard is force-closing, since that sweep departs everyone. Migration is additive and nullable (`20260818120000_visit_supervision_warned_at`).

## #1347 convergence

**PR #1669 now owns the explicit-confirm interface.** This body previously said `feat/1347-explicit-confirm` "is empty" — true when this branch was cut, false since #1669 opened. What #1669 ships:

- `Visit.forceCloseToken`, minted per warning and consumed by the confirming scan,
- `confirmSeconds` in the 400 body, so the window is server-owned but client-rendered,
- the client countdown as `AttendanceState.arm_confirm` / `take_confirm`.

**On rebase after #1669 lands**, this PR's supervision confirm converges on that token shape — the cross-review's **S2**:

- mint a supervision token per warning, in place of the `supervisionWarnedAt` timestamp stamp,
- return `confirmSeconds` in the 400 body and let #1669's countdown render the tick, dropping the "3 to 15 seconds" copy,
- **drop the server-side 15s elapsed window** (`SUPERVISION_CONFIRM_MS`). It is the mechanism #1669 deleted for force-close, and for the same reason: an elapsed-time window can be satisfied by two queued replays arriving inside it, where a minted-and-consumed token cannot,
- **acceptance criteria from review (2026-08-19):** the debounce exemption keys on the supervision token so the confirm window truly opens at 0s (Q16 front door); convergence tests drive `POST /api/scan` end-to-end (not direct `processCheckout` calls), pinning that a second badge inside 3s confirms; `confirmSeconds` + token ride the 400 body so the kiosk countdown ticks,
- add a test that a supervision token cannot force-close and a force-close token cannot confirm a supervision departure — two interrupts can be pending on one visit, so the tokens must not be interchangeable.

## Kiosk client

The warning-only rung needs a way to say "the scan succeeded *and* the room is short". `client/client.py` gained three lines: a 2xx body carrying `warning` renders the amber `banner-warning` (12s dwell) with the CHECKED IN/OUT confirmation inside it, rather than the 5s green one. Escaped like every other backend value.

## Tests

Unit (`npx jest`, 282 suites / 2810 tests green locally):

- `src/lib/__tests__/supervision.test.ts` — the helper's edge cases: same-household counts once, participant-seat exclusion, clearance validity (none on file, expired under the recheck policy, inside the window), non-ACTIVE household, youth, unknown age vs. DoB-stripped declared adult, and the program-in-session probe's four arms.
- `src/lib/__tests__/scanServiceSupervisionInterrupt.test.ts` — the ladder end to end through the **real** helper: both rungs, the confirm window opening/expiring, "NO supervising adult" copy, the same-household non-event, a departure that removes nobody, and the youth check-in warning (plus a check that an adult check-in costs no extra query). The youth gate adds three: below two **with** no youth completes with the warning and no stamp, the last adult leaving an adult-only room still warns, and the warn-only rung never runs the youth probe.
- `src/lib/__tests__/getFullAttendance.test.ts` — retargeted at its own half (accompaniment) now that the count moved out, plus the short-circuit — no unaccompanied youth, no supervision queries.
- `client/test_client.py` — the amber warning banner, its escaping, and that an ordinary scan is untouched. `python3 -m unittest discover` green.

Integration (CI-only, no local Postgres — **not run locally**):

- `src/lib/__tests__/supervision.integration.test.ts` — the query itself against real Postgres (nested membership filter, both in-session arms), the banner for all three ways two adults fail to be two deep, and the full ladder including the stamp/second-badge round trip and its youth-free twin (same departure, no youth → warned and through, nothing stamped).

## Register

`docs/rules/attendance-checkin.md` §Supervision is rewritten: the two-deep line now states who counts, the keyholder-only interrupt is no longer a deliberate limit, the departure rule states the youth gate on the confirm (and why the warn rung keeps over-warning), and what still cannot be tested (school enrollment) is stated **at the rule** as `[Short of policy]` per §3.7.

## Out of scope

- **#1442 — school enrollment.** "Student" here is a program participant. The school-enrollment half of policy's not-a-student test is recorded as a shortfall at the rule, not built.
- **#1624.**
- **The web checkout path.** thpr's register-handover comment flagged the A10 inversion (DELETE checkout closes the last keyholder with no confirmation), but the governing v1.2 ruling scopes this to the badge interrupt — "fires on every departure" answers *which person*, not *which code path*. Stated as a deliberate limit in the register; worth its own issue.
- **`docs/DOCUMENTATION_STANDARD.md` §3.7** still uses "a supervision check counting bare adults" as its worked example of a divergence. It teaches the *shape*, so it is left alone rather than rewritten inside a feature PR.

### DIFF
--- a/checkin-app/prisma/migrations/20260818120000_visit_supervision_warned_at/migration.sql
+++ b/checkin-app/prisma/migrations/20260818120000_visit_supervision_warned_at/migration.sql
@@ -1,0 +1,5 @@
+-- Additive, nullable: old code neither reads nor writes it, so the drain window
+-- is safe. Every existing open visit starts unstamped, i.e. unconfirmed. Kept
+-- separate from forceCloseWarnedAt — the keyholder close-guard is its own
+-- interrupt and both can be pending on the same visit (#1436).
+ALTER TABLE "Visit" ADD COLUMN "supervisionWarnedAt" TIMESTAMP(3);
--- a/checkin-app/prisma/schema.prisma
+++ b/checkin-app/prisma/schema.prisma
@@ -1121,6 +1121,12 @@ model Visit {
   /// @sensitivity:internal
   forceCloseWarnedAt DateTime?
 
+  /// When the kiosk last showed this person the "the room drops below two
+  /// supervising adults" warning (#1436). Its own stamp, not forceCloseWarnedAt:
+  /// the keyholder close-guard is a separate interrupt that can be pending at the
+  /// same time. @sensitivity:internal
+  supervisionWarnedAt DateTime?
+
   person Person @relation(fields: [personId], references: [id])
   event  Event?      @relation(fields: [associatedEventId], references: [id])
 
--- a/checkin-app/prisma/schema.prisma
+++ b/checkin-app/prisma/schema.prisma
@@ -1124,7 +1124,8 @@ model Visit {
   /// When the kiosk last showed this person the "the room drops below two
   /// supervising adults" warning (#1436). Its own stamp, not forceCloseWarnedAt:
   /// the keyholder close-guard is a separate interrupt that can be pending at the
-  /// same time. @sensitivity:internal
+  /// same time.
+  /// @sensitivity:internal
   supervisionWarnedAt DateTime?
 
   person Person @relation(fields: [personId], references: [id])
--- a/checkin-app/src/app/api/household/__tests__/participantPiiMinimization.test.ts
+++ b/checkin-app/src/app/api/household/__tests__/participantPiiMinimization.test.ts
@@ -30,6 +30,9 @@ jest.mock("@/lib/prisma", () => ({
     default: {
         person: { findUnique: jest.fn() },
         visit: { findMany: jest.fn() },
+        // The supervising-adult test reads the board's background-check recheck
+        // policy (#1436); unset here, which is the default.
+        boardSettings: { findUnique: jest.fn() },
     },
 }));
 
--- a/checkin-app/src/lib/__tests__/getFullAttendance.test.ts
+++ b/checkin-app/src/lib/__tests__/getFullAttendance.test.ts
@@ -7,6 +7,16 @@
 const findMany = jest.fn();
 jest.mock("@/lib/prisma", () => ({ __esModule: true, default: { visit: { findMany: (...a: unknown[]) => findMany(...a) } } }));
 
+// Who counts as a supervising adult is lib/supervision's rule and is tested there
+// (#1436/#1550). Pinned at 1 — short of two-deep — so what this file asserts is its
+// OWN half: whether a youth is accompanied by an adult of their household.
+jest.mock("@/lib/supervision", () => ({
+    __esModule: true,
+    MIN_SUPERVISING_ADULTS: 2,
+    supervisingAdultVisits: jest.fn().mockResolvedValue(new Map()),
+    supervisingAdultCount: () => 1,
+}));
+
 import { getFullAttendance } from "@/lib/getFullAttendance";
 
 // Age fixtures are relative to now so they never age past the youth boundary the
@@ -108,8 +118,8 @@ describe("getFullAttendance() — privileged caller (unchanged)", () => {
 describe("two-deep calc fails closed on unknown DOB (#300)", () => {
     // Youth (hh 8) + real adult (hh 6) + null-DOB visitor in the youth's
     // household. Under the old null→adult default the null-DOB visitor
-    // "accompanied" the youth AND made adultVisits.length 2 — masking the
-    // violation on both prongs.
+    // "accompanied" the youth, masking the violation. The supervising-adult
+    // count is the other prong and lives in lib/supervision now.
     const nullDobRow = {
         id: 204, arrivedAt: new Date("2026-07-01T14:20:00Z"), departedAt: null, personId: 80,
         person: {
@@ -130,9 +140,10 @@ describe("two-deep calc fails closed on unknown DOB (#300)", () => {
         expect(attendance[2].participant.isYouth).toBe(true);
     });
 
-    it("a DoB-stripped declared adult (#1165) still supervises", async () => {
+    it("a DoB-stripped declared adult (#1165) still accompanies their own youth", async () => {
         // Same shape, but the null-DoB visitor is a 26+ member whose DoB was
-        // deliberately deleted: isDeclaredAdult wins over the fail-closed default.
+        // deliberately deleted: isDeclaredAdult wins over the fail-closed default,
+        // so the youth of household 8 is no longer unaccompanied.
         findMany.mockResolvedValue([...rows, {
             ...nullDobRow,
             person: { ...nullDobRow.person, isDeclaredAdult: true },
--- a/checkin-app/src/lib/__tests__/getFullAttendance.test.ts
+++ b/checkin-app/src/lib/__tests__/getFullAttendance.test.ts
@@ -10,10 +10,11 @@ jest.mock("@/lib/prisma", () => ({ __esModule: true, default: { visit: { findMan
 // Who counts as a supervising adult is lib/supervision's rule and is tested there
 // (#1436/#1550). Pinned at 1 — short of two-deep — so what this file asserts is its
 // OWN half: whether a youth is accompanied by an adult of their household.
+const supervisingAdultVisits = jest.fn().mockResolvedValue(new Map());
 jest.mock("@/lib/supervision", () => ({
     __esModule: true,
     MIN_SUPERVISING_ADULTS: 2,
-    supervisingAdultVisits: jest.fn().mockResolvedValue(new Map()),
+    supervisingAdultVisits: (...a: unknown[]) => supervisingAdultVisits(...a),
     supervisingAdultCount: () => 1,
 }));
 
@@ -52,6 +53,7 @@ const rows = [
 beforeEach(() => {
     findMany.mockReset();
     findMany.mockResolvedValue(rows);
+    supervisingAdultVisits.mockClear();
 });
 
 describe("getFullAttendance({ kiosk: true })", () => {
@@ -138,6 +140,7 @@ describe("two-deep calc fails closed on unknown DOB (#300)", () => {
         // counted as youth, not volunteer, and flagged as youth on the wire
         expect(counts).toEqual({ keyholders: 1, volunteers: 0, youth: 2, total: 3 });
         expect(attendance[2].participant.isYouth).toBe(true);
+        expect(supervisingAdultVisits).toHaveBeenCalled();
     });
 
     it("a DoB-stripped declared adult (#1165) still accompanies their own youth", async () => {
@@ -153,5 +156,8 @@ describe("two-deep calc fails closed on unknown DOB (#300)", () => {
         expect(safety.isTwoDeepViolation).toBe(false);
         expect(counts).toEqual({ keyholders: 1, volunteers: 1, youth: 1, total: 3 });
         expect(attendance[2].participant.isYouth).toBe(false);
+        // No unaccompanied youth, so the flag is false either way — the poll must
+        // not pay for the supervision queries to learn that.
+        expect(supervisingAdultVisits).not.toHaveBeenCalled();
     });
 });
--- a/checkin-app/src/lib/__tests__/scanServiceSupervisionInterrupt.test.ts
+++ b/checkin-app/src/lib/__tests__/scanServiceSupervisionInterrupt.test.ts
@@ -1,0 +1,164 @@
+/**
+ * The departure interrupt ladder (#1436). Drives the REAL supervising-adult test
+ * through processCheckout, so the two stay wired together:
+ *
+ *   third supervising adult leaves (two remain) → warning, checkout proceeds
+ *   next one leaves (below two)                 → warning AND badge-again confirm
+ *
+ * The keyholder close-guard is a separate interrupt with its own stamp; the
+ * participant here is not a keyholder, so only the supervision rungs run.
+ */
+import type { Person } from "@/generated/prisma/client";
+import type { DbClient } from "@/lib/db-client";
+import { processCheckin, processCheckout } from "@/lib/scan-service";
+
+jest.mock("@/lib/prisma", () => ({ __esModule: true, default: {} }));
+jest.mock("@/lib/notifications", () => ({
+    sendCheckinNotifications: jest.fn().mockResolvedValue(undefined),
+}));
+jest.mock("@/lib/attendanceTransitions", () => ({
+    findAssociatedEventAt: jest.fn().mockResolvedValue(null),
+    processVisitCheckout: jest.fn().mockResolvedValue([]),
+}));
+
+const adult = { id: 1, isKeyholder: false, isDeclaredAdult: true, dateOfBirth: null } as Person;
+const youth = { id: 9, isKeyholder: false, isDeclaredAdult: false, dateOfBirth: new Date("2015-01-01") } as Person;
+
+const DEPARTING_VISIT = 42;
+
+/** One open visit per household id, all fully qualifying supervising adults. */
+function openVisits(householdIds: number[]) {
+    return householdIds.map((householdId, i) => ({
+        id: i === 0 ? DEPARTING_VISIT : 100 + i,
+        person: {
+            householdId,
+            dateOfBirth: null,
+            isDeclaredAdult: true,
+            lastBackgroundCheck: new Date("2026-01-01"),
+            household: { orgMembership: { status: "ACTIVE" } },
+            programParticipants: [],
+        },
+    }));
+}
+
+function fakeDb(householdIds: number[], warnedAt: Date | null = null) {
+    const update = jest.fn().mockResolvedValue({});
+    const db = {
+        boardSettings: { findUnique: jest.fn().mockResolvedValue({ orgMembershipYearBoundary: null, bgRecheckMonths: 0 }) },
+        visit: {
+            count: jest.fn().mockResolvedValue(1),
+            findMany: jest.fn().mockResolvedValue(openVisits(householdIds)),
+            findUnique: jest.fn().mockResolvedValue({ supervisionWarnedAt: warnedAt }),
+            create: jest.fn().mockResolvedValue({ id: 500 }),
+            update,
+        },
+    };
+    return { db: db as unknown as DbClient, update };
+}
+
+async function checkout(householdIds: number[], warnedAt: Date | null = null) {
+    const { db, update } = fakeDb(householdIds, warnedAt);
+    const res = await processCheckout(adult, DEPARTING_VISIT, "kiosk", db);
+    return { res, body: await res.json(), update };
+}
+
+describe("rung 1 — the third supervising adult leaves", () => {
+    it("warns on the successful checkout and does not ask for a confirm", async () => {
+        const { res, body, update } = await checkout([10, 20, 30]);
+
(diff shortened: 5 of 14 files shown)


jee7s:
## Cross-review (#1668 / #1669 / #1670): yes, with notes

**Verdict: approve with notes.** The shared helper is the right shape — returning the visit→household map is what makes "how many are here" and "how many would remain" one query, and the household rule falls out of it instead of needing a second code path.

### Fixed at c07eb2b7

Two one-liners the review turned up, both now on the branch:

1. **`programInSessionNow` failed OPEN for a RUNNING programme with a null `startAt`.** `Program.startAt` and `endAt` are both nullable; `lte` never matches null, so such a programme matched no OR arm and its ACTIVE participants were counted as *supervising* adults while they are the ones being supervised. The null-`endAt` arm already existed — the null-`startAt` one didn't. Added `{ startAt: null, phase: "RUNNING" }` plus a probe-arm test.
2. **`getFullAttendance` paid for the two supervision queries on every poll.** `isTwoDeepViolation` is already gated on an unaccompanied youth being present, so the queries only matter when `unaccompaniedYouth.length > 0` — the same gate `processCheckin` already uses. Short-circuited; behaviour otherwise identical, and the test asserts both sides.

The PR body's "#1347 convergence" section has also been rewritten — it still claimed the 1347 branch was empty, which stopped being true when #1669 opened.

### Rebase obligation once #1669 lands (S2)

#1669 replaced the force-close confirm's elapsed-time window with a **minted-and-consumed `Visit.forceCloseToken`**, `confirmSeconds` in the 400 body, and an `arm_confirm` / `take_confirm` countdown on the client. This PR's supervision confirm must converge on that shape rather than keep its own `SUPERVISION_CONFIRM_MS` window.

That is not tidiness. A server-side 15s elapsed window can be *satisfied* by two queued replays that arrive inside it — nobody confirmed anything, the clock just agreed. A consumed token cannot: whoever spends it is the one who saw the warning. That is exactly the mechanism #1669 deleted for force-close, and this PR reintroduced it locally because the interface wasn't available yet. On rebase: mint a supervision token, return `confirmSeconds`, drop the window — and add a test that a supervision token cannot force-close and a force-close token cannot confirm a departure, since both interrupts can be pending on one visit.

### Question for @jee7s — one word answers it

**Should the confirm rung (a departure dropping the room below two supervising adults) fire only when a YOUTH has an open visit?**

The v1.2 decision text omits the qualifier and this PR implements it literally, so the ladder fires on *every* departure. But the issue title, the policy premise ("two adults whenever a youth is"), and this PR's own youth-gated banner all read as if the qualifier is load-bearing:

- as written, **every adult-only closing** — the ordinary case, two volunteers locking up with no youth in the building — hits the 400 and a forced re-badge;
- an interrupt that fires when nothing is at stake is the fastest way to train badge-through, and then it will not be noticed on the night a youth *is* present.

Minimal change if yes: gate **only the confirm rung** on any youth having an open visit. The warning-only rung and the two-deep banner stay exactly as they are.

jee7s:
Reviewed `890b3d1aa0` against #1436 / #1550.

The helper is the right shape: one function, adult + still-valid clearance, household counts as one, active `ProgramParticipant` on a program in session now excluded, fail-closed on missing DoB / missing BG. 3→2 warns without confirm, youth check-in warns only, keyholder close-guard stays a separate stamp, rules file updated.

The confirm rung is not what was decided.

#1436: next qualifying adult badges out → warning + confirm (15s countdown; **same explicit confirm as #1347**).

#1347's confirm (PR #1669) is: mint a token with the warning, 15s *visible* countdown at the edge, no elapsed-time gate server-side, token scan is debounce-exempt. This PR rebuilds the old `WarnedAt` + 15s wall-clock window:

- `supervisionWarnedAt` within `SUPERVISION_CONFIRM_MS` *is* the confirm, not a token
- Copy says "Badge again in 3 to 15 seconds" — that is Q16's dead `[0s, 3s)` front again
- Interrupt tests call `processCheckout` directly, so they never see the scan-route debounce swallow the second badge
- The kiosk change only restyles the *successful* 3→2 warning; the 400 confirm has no `confirmSeconds` / token, so even after #1669 the countdown will not tick for this interrupt

Don't ship a second adjacency window. Land #1669 first and reuse that confirm (own token column is fine — the two interrupts can be pending on one visit; don't invent a third abstraction). Then the "3 to 15 seconds" copy, and the rules sentence that encodes a server-side fifteen-second window, both go away.

jee7s:
Agreed on all three, and they resolve into the same S2 convergence this thread already records — the mechanism each one critiques is the server-side elapsed window itself, which the rebase on #1669 replaces:

1. **Q16 front door**: the token convergence brings #1669's debounce exemption (keyed on a token matching the participant's own open visit), which opens the `[0,3s)` confirm window the current copy wrongly promises.
2. **Tests bypass the route debounce**: the convergence tests will drive `POST /api/scan` end-to-end (the #1669 pattern), pinning that a second badge inside 3s confirms rather than being swallowed.
3. **400 confirm has no countdown**: `confirmSeconds` + token ride the 400 body after convergence, so the kiosk's existing `arm_confirm`/tick machinery drives it with near-zero new client code.

Deliberately not patching these pre-rebase — improving the elapsed-window mechanism would be building the thing we're deleting. All three are now enumerated in the PR body's convergence section as acceptance criteria for the rebase. The one input still needed before that work starts: the youth-qualifier question above.

jee7s:
## Ruling recorded — the confirm rung is youth-gated

> **YES — the confirm rung is youth-gated.** With no youth present, that departure downgrades to the warning-only shape. The warn rung itself stays un-gated.

Implemented at `550735be`.

### The ladder now

| `remaining` | youth in the building | outcome |
|---|---|---|
| `>= before` (removes nobody) | — | silent |
| `> 2` | — | silent |
| `== 2` | — | warning, checkout completes |
| `< 2` | **yes** | 400 `{type:"warning"}`, `supervisionWarnedAt` stamped, badge again inside 15s |
| `< 2` | **no** | warning, checkout completes, nothing stamped |

The warn rung is untouched — over-warning there is deliberate and costs a line of text. The gate is on the confirm alone, so the ordinary adult-only closing no longer eats a 400 and a forced re-badge, and the interrupt keeps its meaning for the night a youth *is* present.

### What changed

- `lib/supervision.ts` — `youthIsPresent(db?)`: the adult prong inverted over the same open-visit population (`departedAt: null, deletedAt: null`, live person). Not `isDeclaredAdult`, and not 18+ by DoB — unknown age fails closed as youth (#300). Exactly the classification `processCheckin` already uses for its arrival warning, so both halves of "two adults whenever a youth is" read the same rule. Only the `isDeclaredAdult` half is a query filter; the age half stays in `isYouth` so the age rule keeps one expression rather than being re-spelled in SQL.
- `lib/scan-service.ts` — the probe runs **inside the `remaining < 2` branch only**, before the stamp read, so the whole warn rung and every silent departure pay nothing for it. No youth → return the warning-only shape with the accurate count ("leaves only 1 supervising adult" / "NO supervising adult").
- Tests — three added to `scanServiceSupervisionInterrupt.test.ts` (below two with no youth completes with the warning and no stamp; the last adult leaving an adult-only room still warns; the warn-only rung never calls the probe), and the existing confirm pins now stand up a youth fixture. Integration gained the youth-free twin of the stamp/second-badge round trip. 282 suites / 2810 tests green; `tsc --noEmit` clean.
- `docs/rules/attendance-checkin.md` §Supervision — the departure rule states the gate: dropping below two stops the departure "but only while a youth is in the building", and says why the warn rung keeps over-warning.
- PR body — the ruling is quoted under "The rulings implemented", and the ladder bullets carry the third case.

The S2 convergence obligation is unchanged by this: the gate decides *whether* a confirm is raised, the token decides *how* one is satisfied. Rebasing onto #1669 replaces the elapsed window under the same gate.

thpr (commented):
Architecture is clean. One helper (`supervision.ts`) serves both surfaces. The visit→household map is an elegant design — `supervisingAdultCount(supervising, excludeVisitId?)` answers "how many now" and "how many after this departure" from one query, and the household-dedup falls out of the `Set<householdId>` for free.

The rung ladder is stateless — derived purely from the count, no counter, no session state. The "3rd adult" / "next adult" distinction is just `remaining === 2` vs `remaining < 2`, which is exactly right.

The youth gate (ruling of 2026-08-19) is correctly placed: only on the confirm rung (`remaining < 2`), not on the warn rung. The `youthIsPresent()` probe runs inside that branch only, so quieter departures pay nothing.

Tests are well-structured: unit tests mock supervision for `getFullAttendance` (testing its own half — accompaniment), the interrupt test drives the real helper through `processCheckout`, and integration tests hit real Postgres for the query's nested filters.

Migration is additive, nullable, no data migration — safe for zero-downtime deploy.

Client change is three lines, correctly escapes, uses the amber `banner-warning` class with 12s dwell.

Two minor observations, neither blocking:

1. `schema.prisma:1127` — `@sensitivity:internal` is on the same line as other comment text (`/// same time. @sensitivity:internal`) instead of its own `///` line like `forceCloseWarnedAt` has. Generator handled it (classifications.ts is correct), but breaks the visual pattern.

2. `supervision.integration.test.ts:120` — `prisma.visit.deleteMany({})` in `beforeEach` wipes all visits globally, not just the test's. Fine under the serial-only rule, but filtering by the test's person IDs would be more defensive.

No bugs, no missed callers, no data-loss risk. The `#1669` convergence (token-based confirm replacing the timestamp window) is clearly stated as post-rebase work, not a gap.

jee7s:
Both minor observations applied:

1. `schema.prisma` — `@sensitivity:internal` moved to its own `///` line, matching `forceCloseWarnedAt`. Confirmed with `prisma generate`: `classifications.ts` is byte-identical, so it was only the visual pattern.
2. `supervision.integration.test.ts` — the `beforeEach` wipe is now scoped to the suite's own person IDs (shared `ownIds()` helper, also reused by the `afterAll` cleanup).

`tsc --noEmit` clean; `supervision` + `scanServiceSupervisionInterrupt` 26/26 pass (the integration suite stays CI-only). `8c8d894d`

thpr (approved):
Both review observations addressed in 8c8d894. Clean PR — no bugs, no missed callers, no data-loss risk. Good to merge.

Issue #477, opened 2026-06-29
Dead route + lost validation: PATCH /api/programs/[id]/settings is never called


thpr:
## Summary

`PATCH /api/programs/[id]/settings` (`checkin-app/src/app/api/programs/[id]/settings/route.ts`) has **no client consumer**. A grep for `/api/programs/.../settings` across `checkin-app/src` finds only the route itself and its integration test. No UI page or component fetches it.

## How program settings are actually saved

The program edit form at `checkin-app/src/app/program-ops/programs/[id]/page.tsx` (`handleSaveGeneral`) sends every settings field — `name`, `begin`, `end`, `phase`, `enrollmentStatus`, `memberOnly`, `minAge`, `maxAge`, `maxParticipants`, `leadMentorId`, `memberPrice`, `nonMemberPrice` — through the generic `PATCH /api/programs/[id]` (`checkin-app/src/app/api/programs/[id]/route.ts`). The dedicated `settings` handler is a near-duplicate that is bypassed entirely.

## Why this matters (not just dead code)

The `settings` handler carries validation that the generic PATCH does **not**:

- **Age sanity:** `minAge` / `maxAge` must be non-negative, and effective `minAge <= maxAge` (uses body-overrides-current so a one-sided edit can't invert the range).
- **Capacity floor:** `maxParticipants` must be a positive integer and **not below the current enrollment count** — otherwise the capacity lock perma-rejects every future enroller.

The base `PATCH /api/programs/[id]` performs none of these checks. It writes `minAge`, `maxAge`, and `maxParticipants` straight through. So today via the UI a lead/admin can:

- Set `minAge > maxAge`, making the age gate reject everyone (`age >= minAge && age <= maxAge` is unsatisfiable). Both self/household enroll (`POST /api/programs/[id]/participants`) and public register (`POST /api/programs/[id]/public-register`) enforce that range.
- Set `maxParticipants` **below** the number already enrolled. The capacity check (`lockProgramAndCheckCapacity`) then rejects all new enrollments even though spots were intended — and existing over-cap enrollees are silently stranded.
- Set negative ages (no `< 0` guard on the base path).

## Comparison

| check | `settings` (dead) | base `[id]` PATCH (used) |
|---|---|---|
| `minAge`/`maxAge` >= 0 | yes | no |
| effective `minAge <= maxAge` | yes | no |
| `maxParticipants` positive int | yes | no |
| `maxParticipants` >= current enrollment | yes | no |
[...]

PR #1674, opened 2026-08-21
feat(programs): show capacity next to enrolled counts (3/6)

### LINKED ISSUE #1673: Make program capacity visible
from @ryannazaretian

"Would like to see Enrolled to include the total capacity, e.g. 3/6, 6/6."

jee7s:
Both program **list** views showed a bare enrolled count, while both program **detail** pages already showed capacity. `/api/programs` already returns `maxParticipants` (it's in `PUBLIC_PROGRAM_SELECT`), so this is render-only — no schema, no migration, no API change, no new files, nothing under `src/security/`.

Fixes #1673

## Surfaces

| # | Surface | Before | After |
|---|---------|--------|-------|
| 1 | Public program detail (`programs/[id]`) | already `Capacity: 3/6` + `(Full)` | unchanged (reference implementation) |
| 2 | Program-ops program detail, General tab | already `3 / 6`, but a `0` cap rendered as uncapped | `3 / 0` — truthy test → `!= null` (spacing unchanged) |
| 3 | Public programs directory card, "Enrolled" | bare `3` | `3/6` |
| 4 | Program-ops programs list, "Participants" column | bare `3` | `3/6` (keeps the `?? 0` fallback; `sortBy` still the raw count, so the column sorts numerically) |
| 5 | My Programs → Attendance | `Total Enrolled: 3` | **deliberately unchanged** — see below |

Uncapped programs (`maxParticipants: null`) still render the bare count with no slash.

## Numerator semantics

`_count.participants` on `/api/programs` counts **all** `ProgramParticipant` rows — same rule the capacity enforcer uses (`lockProgramAndCheckCapacity`, where a PENDING/scholarship-held seat consumes capacity). So on surfaces 1–4 the numerator and denominator agree with what the enroller will actually accept.

Surface 5 is different and is left alone (plan open question 1): its number comes from `/api/nav/todo-counts`, which counts `status: "ACTIVE"` only. Printing that as `3/6` would understate fullness and imply seats the enforcer would refuse. The honest fix there is a separate all-status count on `LedProgram`, not repurposing the ACTIVE one.

Also per the plan's recommended defaults: no "Full" marker on the list views (open question 2 — `6/6` already says it, and the ops list has its own Open/Closed badge), and `PATCH /api/programs/[id]` capacity validation stays with #477 (open question 3).

## Tests

Extended the three existing suites (no new files):

- `programs/__tests__/page.test.tsx` and `program-ops/programs/__tests__/page.test.tsx` — table-driven over `null → 3` (no slash), `0 → 3/0`, `6 → 3/6`, at capacity `6/6`, over capacity `7/6` (rendered as-is, unclamped).
- `program-ops/programs/[id]/__tests__/page.test.tsx` — `maxParticipants: 0` renders `2 / 0` rather than reading as uncapped.
- The public suite's existing "counts omitted (system waking)" case now carries a `maxParticipants`, so it also proves capacity alone doesn't resurrect the stat block when `_count` is absent.

Each falsy-zero test was verified to **fail** against the unfixed code by reverting the guard in place:

```
programs/__tests__/page.test.tsx                    FAIL  Unable to find an element with the text: 3/0
program-ops/programs/__tests__/page.test.tsx        FAIL  Unable to find an element with the text: 3/0
program-ops/programs/[id]/__tests__/page.test.tsx   FAIL  renders a zero max participants as a real cap, not as uncapped
```

## Local verification

```
npx tsc --noEmit                              exit 0, no output
npx eslint --max-warnings 0 <6 touched files> exit 0, no output
npx jest --testPathPatterns 'src/app/(programs|program-ops/programs)'
  Test Suites: 5 passed, 5 total
  Tests:       91 passed, 91 total
```

No integration/flow tests needed — no route or DB behaviour changed.

### DIFF
--- a/checkin-app/src/app/program-ops/programs/[id]/__tests__/page.test.tsx
+++ b/checkin-app/src/app/program-ops/programs/[id]/__tests__/page.test.tsx
@@ -97,6 +97,14 @@ describe("ProgramDetailsPage", () => {
     expect(screen.getByText("Mandy Mentor (mandy@example.com)")).toBeInTheDocument();
   });
 
+  it("renders a zero max participants as a real cap, not as uncapped", async () => {
+    setSession({ id: 1, isSysadmin: true });
+    mockFetchJson({ "/api/programs/1": { ...programData, maxParticipants: 0 } });
+    renderPage();
+
+    expect(await screen.findByText("2 / 0")).toBeInTheDocument();
+  });
+
   it("switches to the roster tab and shows volunteers/participants", async () => {
     setSession({ id: 1, isSysadmin: true });
     mockFetchJson({ "/api/programs/1": programData });
--- a/checkin-app/src/app/program-ops/programs/[id]/page.tsx
+++ b/checkin-app/src/app/program-ops/programs/[id]/page.tsx
@@ -298,7 +298,7 @@ export default function ProgramDetailsPage({ params }: { params: Promise<{ id: s
 
                 <Card withBorder radius="md" padding="md">
                   <Group justify="space-around">
-                    <Stack gap={0} align="center"><Text fz="xl" fw={700}>{program.participants?.length || 0} {program.maxParticipants ? `/ ${program.maxParticipants}` : ''}</Text><Text size="sm" c="dimmed">Participants Enrolled</Text></Stack>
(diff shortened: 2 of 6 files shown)


thpr (approved):
Clean render-only PR. Three surfaces get the same `!= null` guard consistently, falsy-zero fix on the detail page is correct (`0` is a real cap), `sortBy` stays on the raw count so column ordering isn't broken, and the `_count`-absent guard still works because `maxParticipants` alone doesn't resurrect the stat block. Tests cover the full matrix. No findings.

---

PR #1680, opened 2026-08-21
fix(schema): @ignore legacy two-variant Shopify columns (#975 Release 2a)

### LINKED ISSUE #1508: Drop the legacy two-variant Shopify columns (#975 Release 2)

thpr:
Fixes #1508

## Summary

Release 2a of the expand/contract in `975-LEGACY_VARIANT_CONTRACT.md`. Marks the four dead legacy two-variant Shopify columns `@ignore` so Prisma's generated client stops selecting them — the intermediate step before the `DROP COLUMN` migration (Release 2b, separate PR after this reaches prod).

**Why not drop directly?** (Review finding from #1680.) Prisma's generated client includes all declared schema fields in select-less queries (`findUnique`, `findMany` without `select`). `deploy-prod.yml` runs `prisma migrate deploy` *before* the ECS service swap, so the old task's client would query the dropped columns for the entire drain window — 500s on every `Program`/`BoardSettings` read.

### Changes

- `schema.prisma`: mark 4 legacy fields `@ignore` (`Program.shopifyOrgMemberVariantId`, `Program.shopifyNonOrgMemberVariantId`, `BoardSettings.shopifyNormalVariantId`, `BoardSettings.shopifyVolunteerVariantId`)
- Collapse `hasShopifyVariant` in participants DELETE route to `!!shopifyVariantId` (a Release-1 survivor that still referenced the legacy columns)
- Delete `DROPPED_SOON` exclusion from the public-tier oracle test (`@ignore` removes the fields from generated classifications, so the exclusion became an under-check)
- Regenerate `classifications.ts` (4 entries removed)
- Update design doc: Release 2 split into 2a (`@ignore`) + 2b (`DROP COLUMN`)
- Update capacity doc wording

### No migration

This is code-only — columns stay in the DB. The `DROP COLUMN` migration ships as Release 2b after this is fully rolled out in prod.

## Test plan

- [x] `tsc --noEmit` — clean
- [x] Full unit test suite — 282 suites, 2814 tests, all pass
- [x] No migration file (columns remain in DB)


jee7s:
Reviewed against `docs/designs/975-LEGACY_VARIANT_CONTRACT.md` and `DEPLOY_MIGRATION_ORDER_OF_OPERATIONS.md`. The content is faithful to the Release-2 checklist — plain `BEGIN`/`COMMIT` drops on nullable columns, `DROPPED_SOON` deleted so the oracle stops under-checking, classifications regenerated rather than hand-edited, both docs updated. Two sequencing problems, one blocking:

**1. This PR folds a missed Release-1 read into the drop release (blocking).** `api/programs/[id]/participants/route.ts`'s DELETE still reads `currentProgram.shopifyOrgMemberVariantId` / `shopifyNonOrgMemberVariantId` on `main` today — a survivor #1464 missed (the contract doc's checklist named the `hasShopifyVariant` check in `[id]/route.ts`, not this one, and `currentProgram` comes from a select-less `findUnique`). The doc's own rule: *"A dropped column cannot ship in the same release as the code that stops selecting it"* — migrations run to completion before the ECS swap, so the old task serves that route against a schema missing columns its query names, for the whole drain window. This is the #917 shape the doc exists to prevent. Fix: split the one-line `hasShopifyVariant` collapse into its own PR, and per the doc's prod note it must reach a **published prod release** (not just dev-on-merge) before the migration lands.

**2. The Release-2 plan itself has a drain-window gap this PR inherits.** Prisma clients generate explicit column lists from the schema they were built against, so the old task's client selects all four columns in **every select-less `Program`/`BoardSettings` read** — not just the route above. "After Release 1 nothing in the running fleet selects them" holds for app code but not for the generated client. Sequenced perfectly, the drop still 500s those reads between `migrate deploy` and task swap. Two honest options: (a) an intermediate release marking the four fields `@ignore` in `schema.prisma` (client stops selecting them; column stays; no migration, no drift), then this drop; or (b) the board explicitly accepts a few minutes of Program/BoardSettings errors at deploy time — a ruling to record on #1508, not a default. If (a), the contract doc's Release list should gain that step.

Also before merge: run the doc's cutover re-verify query against prod (the migration comment says "re-verified at cutover" — worth attaching the actual counts to this thread at that point).

Everything else looks right, including the parts that are easy to get wrong: the migration is a plain drop (no rebuild), the boundary-isolation exemption for `generated/` holds (removal, not a re-tier — and CI's isolation check agrees), and the `975` doc status update matches DOCUMENTATION_STANDARD.

jee7s (approved):

PR #1686, opened 2026-08-21
feat(merge): refuse decision-bearing collisions, auto-dedupe bare joins (#1456 Phase 1)

thpr:
Part of #1456 (Phase 1 — stops creating new tombstone residue; Phases 2–4 resolve stock, archive, and drop `mergedIntoId`)

## Summary

- **Refuse** merges with decision-bearing collisions: `ProgramParticipant` (seat+payment), `ToolStatus` (certification levels), `BackgroundCheckAttestation` (two-identity conflict), future `RSVP`, both-open `Visit` — returns 409 with a structured `collisions` array naming the specific entities and what to fix
- **Auto-dedupe** bare-join collisions with no distinguishing payload: `ProgramVolunteer`, `CorporationLead`, `CorporationMember`, past `RSVP` — merge-side duplicate deleted, keeper row survives
- **Analyze** route surfaces collisions before the operator commits, alongside the existing `membershipBlock`
- Defense-in-depth `left` counters handle TOCTOU races between pre-check and CAS

## What this changes

The merge was a "park everything on the tombstone and let LIVE_PERSON clean up" operation. Now it's a precondition-checked operation that either completes cleanly or refuses. No new merge creates tombstone residue after this lands.

Phases 2–4 (resolve existing stock, archive+delete tombstones, drop `mergedIntoId` + `LIVE_PERSON`) are separate.

## Design notes

**Enrollment status and collisions:** A REVOKED or DENIED `ProgramParticipant` still registers as a collision. That's the conservative default — even a dead enrollment carries payment history and downstream state that a silent merge shouldn't discard. The operator can unenroll the inactive side first if the merge is needed.

**First hard deletes in the merge path:** The auto-dedupe deletes (`ProgramVolunteer`, `CorporationLead`, `CorporationMember`, past `RSVP`) are bare join rows with no distinguishing payload — same human, same entity, nothing to decide. This is deliberate: these are the only relations where both rows are semantically identical, so keeping two is pure duplication. The audit row's `deduped` tallies plus the captured `oldData` pre-images record what was removed.

## Test plan

- [x] Matrix 1a: all decision-bearing collision types → 409 naming each entity
- [x] Matrix 1b: bare-join collisions only → 200, duplicates deleted, keeper rows survive
- [x] Matrix 1c: past RSVP collisions auto-dedupe, future refused
- [x] Matrix 13: both-open visit → 409; merge-side-only open visit migrates
- [x] Analyze collision reporting
- [x] Consumer exclusion: enrollment collision → refusal (no tombstone seat to exclude)
- [x] Match-audit: non-colliding enrollment migrates normally
- [x] Matrix 12 (zero-rows-deleted invariant) still passes — no collision scenario, all rows migrate
- [x] All 69 integration tests pass

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>


jee7s:
Reviewed. The direction is right and the implementation is careful — #1456's own text ratifies replacing collision-parking with refuse + dedupe, the refusal happens pre-transaction with a structured `collisions` array the UI can act on, `analyze` surfaces the same collisions up front, and the old in-transaction guards survive as TOCTOU defense-in-depth (the `left` tallies now only count race outcomes, which the audit row still records). The past/future RSVP split and the both-open-visit refusal both match the issue's intent, and the reworked test matrix covers refusal, dedupe, and the race fallbacks.

Two things before this can land:

1. **The unit CI job failed with no test report at all** — jest died before writing junit or coverage, which is a module-level crash rather than a red assertion (the log ends with "No test report files were found"). Worth checking suites that statically import the merge route (the drift-guard / route-inventory style tests are the usual suspects). Whatever it is, it's invisible in the diff and needs a green re-run before review sign-off means anything.

2. **`Fixes #1456` will auto-close the umbrella issue on merge, but this is Phase 1 only.** The issue's core ask — removing `LIVE_PERSON` and making tombstone exclusion structural — is unshipped. Suggest rewording to a plain reference ("Part of #1456, Phase 1"), or filing the Phase 2 issue first and letting this close #1456 with the successor linked.

Two non-blocking notes:

- A REVOKED/DENIED enrollment still registers as a `programParticipant` collision (the check reads relation rows regardless of status). That's the conservative right default — a dead enrollment still carries payment history — but a sentence in the PR body would save the next reader the archaeology.
- The auto-dedupe deletes are the first hard deletes in the merge path since the no-deletion rework; the deleted duplicates are bare joins with no payload, and the audit row's `deduped` tallies plus the captured `oldData` relations record what went. That reasoning deserves to live in the PR body too, since it's the part that looks like a policy reversal at first glance.

thpr:
All four points addressed:

1. **CI crash fixed.** Was an unused `settingsPATCH` import left over from the rewritten enrollment-cap-gate test. Pushed (b778001). Unit tests green on the re-run.

2. **`Fixes #1456` → `Part of #1456`.** Updated. Phase 1 doesn't close the umbrella.

3. **REVOKED/DENIED enrollment collisions.** Added a "Design notes" section to the PR body. Conservative default: even a dead enrollment carries payment history, so it collides. Operator can unenroll the inactive side first.

4. **First hard deletes in merge path.** Also in "Design notes." These are bare join rows with no payload — same human, same entity, nothing to decide. Audit row's `deduped` tallies + `oldData` pre-images record what was removed.

---

jee7s (dismissed):

jee7s (approved):

Issue #1432, opened 2026-08-01
Require a name on every Person; retire the kiosk email-prefix fallback (rules-fold)


thpr:
**Backlog item M24** — from the backlog register (`docs/backlog/INDEX.md`), source: rules-fold (BACKLOG_HANDOFF).

Require a name on every Person; retire the kiosk email-prefix fallback (rules-fold) — the kiosk shows the local-part of an email when a person has no name recorded; requiring a name on Person removes the need for that fallback entirely.

- Origin: ENHANCE · Readiness: NEEDS-DESIGN · Size: S

_Auto-created from the backlog INDEX; see the doc + `backlog/sources/` reconciliation for context._

jee7s:
HOLD — the ask is sound but understated: this is a data-model change with a live-column migration, and the fallback is not confined to the kiosk.

Questions:

1. What is the backfill for existing Persons with no name? `Person.name` is `String?` (`checkin-app/prisma/schema.prisma:72`) and for a member the org only has an email for, there is no source to fill it from.
2. DB-level `NOT NULL` or app-level validation? A `SET NOT NULL` on a live column has to survive the rolling-deploy drain window while old code is still writing nulls.
3. Does "retire the kiosk email-prefix fallback" mean only `checkin-app/src/lib/kiosk-names.ts:17`, or all of them — `scan-service.ts:111`, `getFullAttendance.ts:112`, `app/attendance/current/page.tsx:303/487/510`, `api/kioskdisplay/certifications/route.ts:58`?
4. What happens to `checkin-app/src/app/api/membership-ops/participants/import/route.ts:208`, which *manufactures* a name from `pr.parentEmail.split('@')[0]`? Requiring a name does not remove the email prefix from the display — it launders it into stored data, where it is harder to spot.

---
_Posted by Claude._

jee7s:
Open implementation slice: #1688 (`Part of #1432`, not `Fixes`). It enforces a name on the admin/household/intake create paths. Out of that PR on purpose: bulk-import email-prefix, `NOT NULL` on `Person.name`, and retiring the kiosk email-prefix fallback. GitHub will not show #1688 as a linked PR until a closing keyword is used.

jee7s:
## Remainder plan for #1432 — after #1688

**This plan defers entirely to #1688.** It changes nothing #1688 touches, duplicates none of its
scope, and every step below is **gated on #1688 merging first**. #1688's own "What's NOT in this
[...]

PR #1688, opened 2026-08-22
fix(membership-ops): require name on every Person creation path

thpr:
## Summary

- **Admin participants route**: reject 400 when `parentEmail` doesn't match an existing Person and no `parentName` is provided; previously created a nameless parent from just an email
- **Household add-member route**: reject 400 when `memberName` is missing or blank
- **Intake secondary parent**: reject with `name_required` when only email is provided (was silently skipped); guard the update path against writing explicit null name
- **New participant form**: add `parentName` field (appears when `parentEmail` is entered), disable submit until filled

## What's NOT in this PR

- **Bulk import route** (`import/route.ts:208`): still uses `email.split('@')[0]` as a manufactured name — different fix (CSV format change), separate PR
- **`NOT NULL` migration**: this enforces at app level first; DB constraint follows once existing nulls are backfilled
- **Kiosk email-prefix display fallback removal**: deferred until all Persons have names (already consolidated to two lib sites: `kiosk-names.ts` and `getFullAttendance.ts`)

## Test plan

- [x] Existing parent-create test updated to send `parentName`, asserts parent gets the name
- [x] New test: rejection when `parentEmail` provided without `parentName` and parent doesn't exist
- [x] Page unit tests updated for `parentName` in request body
- [ ] Integration: verify intake secondary parent email-only gets `name_required` rejection
- [ ] Integration: verify household add-member without name gets 400

Part of #1432

### DIFF
--- a/checkin-app/src/app/__tests__/adminParticipants.integration.test.ts
+++ b/checkin-app/src/app/__tests__/adminParticipants.integration.test.ts
@@ -214,10 +214,11 @@ describe('Admin Participants API Integration Tests', () => {
 
             const req = new Request('http://localhost:4000/api/membership-ops/participants', {
                 method: 'POST',
-                body: JSON.stringify({ 
-                    name: 'Child User', 
+                body: JSON.stringify({
+                    name: 'Child User',
                     email: 'new-child-participants-test@example.com',
-                    parentEmail: 'new-parent-participants-test@example.com' 
+                    parentEmail: 'new-parent-participants-test@example.com',
+                    parentName: 'New Parent'
                 })
             });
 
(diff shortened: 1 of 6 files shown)


jee7s:
Reviewed. The two paths this fixes are fixed well — the admin route's 400 with the `{lastName} Household` naming, the form's conditional required field with correct submit gating, and tests on both the accept and reject sides. Three asks before it matches its own title, plus context on the deferred half of #1432:

**1. `POST /api/household` still creates a nameless Person.** The household-lead add-member route destructures `memberName` and passes it straight into `tx.person.create` with no guard — nothing in the route validates it (its only two mentions are the destructure and the create). A body without it mints exactly the nameless row this PR exists to prevent, on a member-facing route. Same one-guard shape as the participants fix.

**2. The intake tightening trades "nameless person" for silent data loss.** With the gate now `sp.name` alone, an email-only second guardian is silently skipped — and the membership form doesn't require `secondaryName` when an email is present (`buildPayload` sends it unvalidated), so a member who submits just the email gets no error and no record. The admin path got a server-side 400; this path should throw the equivalent `IntakeError` (and ideally the form should require the name field when the email is filled), not drop the input.

**3. `Fixes #1432` closes an issue whose headline is deferred.** The NOT-NULL migration, the kiosk email-prefix fallback retirement, and the bulk-import fix are all explicitly out of scope here — same shape as the #1686 closing-link point. Suggest `Part of #1432`, or file the follow-ups first.

Minor, update-path rather than creation-path: the existing-secondary update spread `...(sp.name !== undefined && { name: sp.name })` will write an explicit `null` name onto an existing person; a `?? undefined` there closes it until the DB constraint exists.

**Sizing note for the deferred fallback retirement** (the other half of #1432): the email-prefix fallback is already consolidated — the M1 wire-privacy work moved it server-side into exactly two shared resolvers, `lib/kiosk-names.ts` (with the "falls back to email prefix when name is null" test) and `lib/getFullAttendance.ts:119`, and #1115 dropped the raw email from the kiosk warning path. So once names are required and backfilled, the retire step is deleting the fallback in those two lib sites plus the bulk-import `email.split('@')[0]` — small and well-localized, worth naming in the follow-up.

jee7s (approved):

Issue #1687, opened 2026-08-22
Auto-advance program phase when dates pass


thpr:
## Context

#1512 identified that program `phase` never advances on its own — it only changes when someone manually edits program settings. With #1441/#1512 making `startAt`/`endAt` required, the dates are now reliable enough to derive phase from.

## Future work

Once dates are required (prerequisite: #1441 + #1512), add automatic phase advancement:

- **Option A — Derived phase**: compute `PLANNING` → `UPCOMING` → `RUNNING` → `FINISHED` from `startAt`/`endAt` instead of storing it. Removes the manual step entirely but touches every `phase` read site.
- **Option B — Nightly cron**: keep the column, run a job that advances phase as dates pass. Simpler migration, but adds infra.
- **Option C — Surface staleness**: board list of programs that are `RUNNING` with `endAt` in the past. Doesn't fix, but makes the gap visible.

Deliberately deferred from the date-requirement work to keep that PR focused.

Related: #1441, #1512

PR #1675, opened 2026-08-21
fix(facility): trends counts every completed visit regardless of source

jee7s:
**Land after #1666.** Two reasons, both blocking:

1. **Conflict surfaces.** #1666 touches `facilityTrendsAPI.integration.test.ts` and `docs/rules/attendance-checkin.md`; this PR touches the first and deliberately leaves the second alone (the staff-asserted bullet arrives with #1666). Resolution on both is **invert-and-keep, never drop** — keep #1666's added rows and flip their assertions to the counting behaviour. Its added `LEAD_MARKED` stamp on `api/facility/visits/insert/route.ts` also carries a comment ending "Excluded from `facility/trends` (see that route's filter)", which the post-merge resolution rewords to the significance rationale.
2. **#1666 goes red if this merges first.** Its added test `excludes a real staff walk-in insert from trends hours` asserts exactly the behaviour this PR removes. Merging in the other order breaks it on landing.

## Summary

`facility/trends` dropped every visit whose `arrivedVia` was `LEAD_MARKED` or its legacy `SYSTEM` spelling, so a staff roster mark was not facility time. This deletes that `OR` block — one filter, one place; grepping `notIn`/`arrivedVia` across `checkin-app/src` finds no second aggregate reader.

The response shape is unchanged (`buckets`/`totals`), so `facility-ops/trends/page.tsx` needs no edit. No schema change, no migration, no new route parameter, no cache to invalidate.

Fixes #1672

## Policy

**Every completed visit counts toward facility hours regardless of source.** Where a time came from governs how a correction is weighed and reviewed — the significance weights (`SCANNER` 3 / `LEAD_MARKED` 2 / `WEB` 1) and the correction-review surface — not whether the visit is counted. Those systems are untouched, as are #1666's backfill and its insert-route stamp: both stay correct for significance and review, they simply stop affecting trends.

The exclusions that remain are structural, not editorial: an open visit has no duration to sum, and a deleted visit did not happen.

## Past-period callout

**Reported facility hours for past periods go UP.** Every staff-marked visit already in the window starts counting the moment this deploys — the mirror image of the drop #1666's body accepts. Nothing is recomputed or migrated; the numbers simply read differently on the next page load, including for periods already reported to the board.

## Test changes

`checkin-app/src/app/__tests__/facilityTrendsAPI.integration.test.ts` — exclusion assertions inverted, file docblock rewritten:

[...]

PR #1689, opened 2026-08-22
fix(programs): require start and end dates (#1441, #1512)

### LINKED ISSUE #1441: Program dates are optional; policy requires a start and a bounded end (policy divergence)

### LINKED ISSUE #1512: Nothing ends a program — endAt is optional and phase never advances on its own
## Problem

Nothing ends a program. A program created today reads as live indefinitely unless a person remembers to go and change it, and nothing surfaces the ones that need changing.

Two fields could say a program is over, and neither does so on its own:

- **`endAt` is optional.** The create form sends `endAt || null` ([program-ops/new/page.tsx](https://github.com/innovationtreehouse/checkin/blob/main/checkin-app/src/app/program-ops/new/page.tsx)), and nothing requires it later.
- **`phase` never advances by itself.** It defaults to `PLANNING` and only moves when an admin edits program settings ([programs/[id]/settings/route.ts](https://github.com/innovationtreehouse/checkin/blob/main/checkin-app/src/app/api/programs/%5Bid%5D/settings/route.ts)). Outside tests, the only writes of `RUNNING` or `FINISHED` come from that manual route. No cron, no date-driven transition.

So "is this program still going?" cannot be answered from the data unless someone maintained it by hand.

## Why it matters beyond tidiness

Downstream rules use program attachment as the proxy for "this person is still involved", and they degrade quietly when a program never ends.

- The **individual adult-child membership agreement** (#1224) bounds its population to programs running or ended within the last year. On a program with no end date that bound does nothing, and the person is asked to sign again every membership year until they age out at 26.
- The **background-check obligations** and the **board compliance lists** read attachment with no time bound at all.
- #1462's cheapest option — a sweep that clears attachment rows once `endAt` passes — **silently does nothing** on an undated program. That issue depends on this one being fixed first.

The failure mode is the bad kind: no error, no empty list, just people staying on chase lists and receiving requests after they have finished.

## Options

Not ranked, not chosen.

1. **Require `endAt` at creation.** Smallest change, and it makes the existing window predicates work. Needs an answer for genuinely open-ended programs (an ongoing shop night), and a backfill decision for programs already created without one.
2. **Derive `phase` from the dates.** `PLANNING` → `UPCOMING` → `RUNNING` → `FINISHED` computed from `startAt`/`endAt` rather than stored, so it cannot drift. Removes the manual step entirely, but the stored column is written today by the settings route and read in several places, so it is a real refactor.
3. **Nightly phase advance.** Keep the column, let a cron move it as dates pass. Cheap and mechanical; leaves undated programs untouched, so it only helps where option 1 also applies.
4. **Surface the staleness instead of fixing it.** A board list of programs that are `RUNNING` with an `endAt` in the past, or that have no `endAt` at all. Fixes nothing by itself, but makes the gap visible and is compatible with any of the above.
5. **Accept it.** Downstream windows stay best-effort and the lists keep some people on them indefinitely.

## Decision needed

Should a program's state be **derived from its dates** or **kept as an explicit thing someone sets**? Everything else follows: if derived, `endAt` has to be mandatory and `phase` stops being stored; if explicit, then something has to prompt the person who owns it, because today nothing does.

thpr:
## Summary

- **Schema**: `startAt` and `endAt` on `Program` are now `NOT NULL` (were optional). Migration backfills existing nulls (`startAt` → `CURRENT_DATE`, `endAt` → `2027-06-30`) before adding the constraint.
- **API**: POST and PATCH `/api/programs` validate that both dates are present, returning 400 if missing.
- **Form**: Create-program page marks both date fields `required` and disables submit until both are filled.
- **Dead code removal**: Null-matching WHERE arms in `supervision.ts`, `personAgreementTriggers.ts`, and the active-programs filter in `route.ts` are removed — they can never match with NOT NULL columns.
- **Simplified `programCoverageDate()`**: returns `endAt` directly (always present), signature narrows from `{ startAt: Date | null; endAt: Date | null } → Date | null` to `{ endAt: Date } → Date`.
- **Display**: Programs list and detail pages drop "TBD"/"Ongoing" fallback labels.
- **Docs**: `docs/rules/programs.md` updated to reflect both dates required.
- **Future work**: Filed #1687 for auto-phase-advance (deriving RUNNING/FINISHED from dates).

Fixes #1441
Fixes #1512

## Test plan

- [x] `programsCreateDate.test.ts` — verifies dates stored as Date objects, rejects missing startAt, rejects missing endAt
- [x] `orgMembership.test.ts` — `programCoverageDate` returns endAt
- [x] `programSignupIntegration.test.ts` — full signup flow with required dates
- [x] ~48 integration test files updated to supply both dates on program creates
- [x] tsc passes clean
- [ ] Manual: create a program without dates → form prevents submit; API returns 400
- [ ] Manual: existing programs with null dates backfilled after migration


jee7s:
Reviewed. The code side is right where it matters — the PATCH validation takes the careful form (`startAt !== undefined && !startAt`, so single-field edits keep working), POST requires both dates, the null-arm removals in `supervision.ts` / `personAgreementTriggers.ts` / the active-programs filter are correct once the constraint exists, and the fixture sweep is mechanical. Three fixes and one ruling before it can land:

**1. Wrap the migration in `BEGIN;`/`COMMIT;`.** Two UPDATEs + two ALTERs, and `prisma migrate deploy` does not wrap a file in a transaction (`DEPLOY_MIGRATION_ORDER_OF_OPERATIONS.md` rule 5, the 2026-07-02 partial-application incident). As written, the `startAt` backfill can commit while the `endAt` half fails, leaving a wedge that needs manual surgery.

**2. Rule 3 — old code can still write the nulls this constrains.** Migrations run to completion before the ECS swap, and `main`'s POST has no date validation (and its PATCH will write an explicit `startAt: null` through `parseDateOnly`). During the drain window those writes fail `23502` → 500. The #918 bar is "prove old code can't violate the new constraint" — here it can. Either split validation-first (constraint follows in the next release), or record the accepted drain-window risk explicitly on the PR — it's an admin-only write path, so acceptance is defensible, but it has to be said, not implied.

**3. Rebase over #1675** — it merged and rewrote `facilityTrendsAPI.integration.test.ts`, which flipped this branch to CONFLICTING. Resolution is keep #1675's assertions, re-add the date fixtures.

**4 (ruling, not review): the backfill fabricates product data with a pricing consequence.** `startAt → CURRENT_DATE` and `endAt → '2027-06-30'` aren't neutral placeholders once `programCoverageDate()` returns `endAt` directly: a formerly dateless program — whose member pricing deliberately rested on dues status alone (the flow-test comment this PR deletes says exactly that) — gains a coverage window to mid-2027. Two asks: run a cutover count of actually-null rows against prod (if it's zero this whole point evaporates, and the PR should say so), and if it's nonzero, the two backfill values need the owner's sign-off on the PR, not just a migration comment.

thpr:
**1. BEGIN/COMMIT** — done (commit 29b551d).

**2. Rule 3 drain-window risk** — acknowledged. The write path is admin-only (sysadmin/board creating or editing a program), so the 23502 window during ECS drain is defensible: it's a low-traffic, privileged operation, and the 500 is a clean refusal, not data corruption. Splitting validation-first adds a release cycle for the same net effect. Calling it explicitly here rather than implying it.

**3. Rebase over #1675** — done. Kept #1675's test names/comments, re-added the date fixtures.

**4. Backfill values + pricing consequence** — this needs a prod count of actually-null `startAt`/`endAt` rows. @thpr can you run that? If it's zero this point evaporates and we say so on the PR. If nonzero, the backfill values (`CURRENT_DATE` / `2027-06-30`) need owner sign-off since `programCoverageDate()` now returns `endAt` directly, giving formerly dateless programs a coverage window.

PR #1696, opened 2026-08-22
feat(person): gapless IDs from a monotonic counter (#1693)

### LINKED ISSUE #1693: Person IDs: eliminate Aurora auto-pause sequence gaps

jee7s:
## The problem

`Person.id` is `SERIAL`, and it is the only model id that reaches a physical artifact: `print-badges/page.tsx` encodes it into the QR payload, and `BadgeDocument.tsx` prints it as human-readable text beside it. Aurora Serverless v2's storage layer prefetches sequence values in batches of ~32 and discards them on every pause/resume — `CACHE 1` is honoured at the SQL layer but not underneath. With the scale-to-zero schedule that happens at least once a day, so ~250 real people already span ids up to ~2,400, and the burn is ~12k ids a year.

## The shape

One table and one helper.

```prisma
model IdCounter {
  name  String @id
  value Int
}
```

`mintPersonId(tx)` issues the next value in a single statement, and every production Person create now passes `id: await mintPersonId(tx)` inside the same transaction as the create.

**Why not `MAX(id) + 1`.** It reuses ids. `auth-options.ts`'s NextAuth adapter `deleteUser` hard-deletes a Person row today, so deleting the highest person would hand their number to the next sign-up — and every badge ever printed for the first would then scan as the second at the kiosk. In a youth check-in system that is a safety defect, not an aesthetic one. A counter only ever moves up: a deletion cannot wind it back. `mintId.integration` case 1 is the test that fails the day anyone "simplifies" it back.

**Why the statement is not a bare `UPDATE … SET value = value + 1`.** Three clauses, each for a failure that was reproduced:

- `ON CONFLICT … DO UPDATE` self-seeds. A bare `UPDATE` on a missing row affects zero rows and 500s every person create — and the row legitimately goes missing in three places: `prisma db push` provisions the flow DB and dev instances from `schema.prisma` and skips migrations entirely; the dev reset `TRUNCATE`s every table in the catalog with `RESTART IDENTITY`; and a future migration coalesce produces a schema-only baseline. The first mint puts the row back. (Same idiom as `BoardSettings`, which also has no migration `INSERT`.)
- `GREATEST(value, MAX(id))` absorbs a foreign high-water mark. 461 fixture `person.create`s across 133 test files mint from the sequence while the counter starts seeded; any fixed margin is exhausted and they collide on the PK. This is *not* MAX+1 — after a delete the counter keeps the higher of the two, so nothing is reissued.
- `setval(…, minted)` pushes the **sequence** up to the value just handed out. Without it the collision simply runs the other way: a counter-minted row raises `MAX(id)` while `Person_id_seq` stays put, so the very next sequence-minted insert — a fixture, or any site a guard missed — draws the id just issued. `setval` is non-transactional, and that is a feature in both directions: a concurrent sequence user sees it immediately, and if the minting transaction rolls back the gap it leaves is in the *sequence* (consumed only by fixtures and unconverted sites) while the counter rolls back and re-mints the value gaplessly.

**Why `@default(autoincrement())` stays.** Migrations apply *before* the rolling deploy, and old tasks then serve live traffic against the fully-migrated schema for a multi-minute drain window. Old code emits `INSERT INTO "Person"` with no id and relies on that default; dropping it here would 500 every create served by an old task for the whole window. Old-code inserts during the drain mint from the sequence, cannot collide with counter-minted ids (the counter is seeded above the sequence), and cannot reuse anything. They do leave gaps — gaplessness begins when the last old task drains.

## Site inventory

Nine production sites in five files, plus three in the dev seed.

| # | Site | Transaction context | Conversion |
|---|---|---|---|
| 1 | `lib/auth-options.ts` `createParticipantWithHousehold` | inside `prisma.$transaction` | `id: await mintPersonId(tx)`. **Choke point for three entry points** — the adapter's `createUser`, the dev-persona mint, and ops creating an emergency-contact person. The issue's inventory listed the contacts route separately; it has no `person.create` of its own. |
| 2–3 | `lib/membership/intake.ts` (secondary parent, children loop) | **none** — `saveIntake` is a bare sequence of autocommit statements | wrapped in `withTx(prisma, …)` |
| 4 | `api/household/route.ts` | inside `prisma.$transaction` | `id:` added to `data` |
| 5–7 | `api/membership-ops/participants/route.ts` | **none** — no `$transaction` in the file | wrapped in `withTx` |
| 8–9 | `api/membership-ops/participants/import/route.ts` | local helper takes `db: DbClient`; child create is inside `$transaction` | helper body became `withTx(db, …)` |
| — | `lib/dev/seed-helpers.ts` — the three `+ Family` macro creates | none | wrapped in `withTx`; the macro is click-repeatable, so it is the unbounded one |

The twelve debug-persona `upsert`s in the same file deliberately stay on the sequence and are allowlisted in the guard: an eagerly-minted id in an upsert's `create` branch is burned on every re-run where the row already exists, so seeding would put a gap in the counter each time. They are a fixed run-once set, dev databases print no badges, and `GREATEST` absorbs whatever the sequence does there anyway.

**No nested Person creates exist** (`git grep` for a Person relation under another model's `create` returns zero non-test hits), and there is no raw `INSERT INTO "Person"` anywhere in `src/` or in the migrations.

## Deviation from the plan worth flagging

The plan said the three sites carrying `household: { create: {…} }` needed no restructuring — "the minted `id` goes on the outer `data` and the nesting is irrelevant". That is not true of Prisma's generated types: `id` exists only in `PersonUncheckedCreateInput`, which is mutually exclusive with the nested relation create in `PersonCreateInput`. Those three sites now create the household as its own statement inside the same transaction and pass `householdId` — the same shape `createParticipantWithHousehold` already used. Atomicity is unchanged; `tsc` catches it either way.

## Seed math and the margin

```sql
GREATEST(
    COALESCE((SELECT MAX("id") FROM "Person"), 0),
    COALESCE(pg_sequence_last_value(pg_get_serial_sequence('"Person"', 'id')::regclass), 0)
) + 64
```

It clears the **sequence**, not merely `MAX(id)`, because old code still mints from the sequence during the drain. Margin 64 = two Aurora prefetch batches, covering a pause landing inside the deploy window; if `pg_sequence_last_value` under Aurora lags the storage-layer prefetch ceiling, the margin covers the difference.

**If the margin is ever too small the failure is a single retryable 500, not sustained breakage** — one unique violation, after which `GREATEST` absorbs the new high-water mark and the next mint is correct. There is deliberately no P2002 retry around the mint: the window is minutes wide, once, and the escape hatch is to bump the margin.

## Verification

The amended statement was re-verified against Postgres 15 in a scratch database before being wired in — all seven cases plus concurrency and reconnect:

| case | expected | got |
|---|---|---|
| mint on empty counter table (self-heal) | 1 | 1 |
| second mint | 2 | 2 |
| counter=100, persons 1..100, **delete person 100**, mint | 101 | 101 |
| foreign row at id 500, mint | 501 | 501 |
| **all** persons deleted, mint (no rewind) | 502 | 502 |
| rolled-back mint leaves no counter gap | 2 | 2 |
| **sequence `nextval` right after a mint** | minted + 1 | minted + 1 |
| fixture insert (no id) after a minted insert | no unique violation | no unique violation |
| two concurrent transactions | 1, 2 | 1, 2 |
| counter continues across a reconnect | 4 | 4 |

The migration's seed was verified separately on an empty DB (`64`) and on a populated one with the sequence prefetched past `MAX(id)` (max 2400, sequence 2432 → `2496`), and re-running it is a no-op. `prisma migrate deploy` onto a fresh DB followed by `prisma migrate diff --from-config-datasource --to-schema prisma/schema.prisma --exit-code` reports **no difference**.

Tests: 26 unit suites (248 tests) green, including the guard, the new unit suite, and every suite touching a converted file. 13 integration suites (84 tests) green — `adminParticipants.integration.test.ts` among them, which is the suite that interleaves fixture `person.create`s with route calls and is exactly the regression the `setval` exists for. `mintId.integration` covers all seven cases.

Both halves were also proved by breaking them in place:

- Deleting the `setval` from the helper and re-running case 7 fails with `Unique constraint failed on the fields: (id)` — the P2002 the amendment predicted. Restored, green.
- Deleting `mintPersonId` from `api/household/route.ts` fails the guard with `["app/api/household/route.ts"]`. Restored, green.

## The ratchet, and the recommendation against it

The follow-up would be to drop the DEFAULT, making `id` required in `PersonCreateInput` so `tsc` enforces every site instead of a guard test. **Recommendation: do not do it.** Dropping the default makes *every* `person.create` in the repo a type error — 461 call sites across 133 test files — and hand-written ids in fixtures are precisely the thing that collides with a counter. The guard test buys the same enforcement for a hundredth of the diff, and its blind spots (a delegate held in a variable, raw SQL, mint-and-create in separate transactions) are all patterns that do not exist here. Keep the default and keep the guard permanently.

## Register entry

#1694 is closed, not merged: the owner ruled against the framing it was written for. Its register entry returns here corrected — the "sparse, never dense" and "gaps are expected rather than a fault to close" clauses are removed, because they are now false, and the last sentence states the invariant the counter provides. The reuse prohibition, which is the load-bearing half, survives in shape.

Fixes #1693


thpr (dismissed):
Reviewed the full diff — SQL semantics, call-site conversions, guard test, integration tests, migration, and rolling-deploy safety.

The three-clause single statement (`ON CONFLICT` self-seed, `GREATEST` absorbs foreign high-water marks without rewinding, `setval` syncs the sequence) is correct for all edge cases I traced: concurrent inserts, rollback, missing counter row, sequence ahead of MAX(id), and the load-bearing case — delete-highest-then-mint never reissues.

The guard test is the right enforcement shape for a codebase with 461 fixture creates: catches unguarded production sites statically, lets test fixtures use the sequence freely, states its blind spots honestly and verifies none exist today.

No issues found. The Prisma unchecked-input workaround (separate household create in same tx) preserves atomicity. Rolling-deploy margin (+64, single-retryable-500 degradation) is practical.

---
*Claude Code with Claude Opus 4.6*

jee7s:
**Unit CI failure fixed** (run 32546321390, job 96965327167) — `__tests__/programSignupIntegration.test.ts`.

That suite drives the household PATCH route end-to-end against a mocked prisma. The route now mints the new member's id, so it reaches `mintPersonId(tx)` → `tx.$queryRaw`, the mock returned nothing, `const [row] = await undefined` threw "(intermediate value) is not iterable", the route 500'd where the test expects 200, and the `console.error` guard tripped on the way past. My earlier verification ran a targeted pattern rather than the full unit tier, so it never saw this; that was the miss.

Fixed in one place per mock, narrowly:

- **`jest.setup.js`'s default `@/lib/prisma` mock** now answers exactly one raw statement — matched on `IdCounter` — from an in-memory counter seeded at `100000` and incremented per call, so a multi-create flow gets distinct ids well clear of any fixture. **Every other raw query still rejects.** That reject-real-calls behaviour is the whole reason the mock exists (AGENTS.md, unit tier: *"These must never hit a real DB — the prisma mock rejects real calls"*), and the match is deliberately narrow rather than a blanket `$queryRaw` stub.
- **`programSignupIntegration.test.ts` declares its own `jest.mock('@/lib/prisma', …)`**, which overrides the default, so the setup-file fix alone could not reach it — verified by re-running with only that change. Its existing `$queryRaw: jest.fn()` now resolves a row. Same narrow shape, in the mock that is actually in play.
- **`src/__tests__/prismaUnitMockRawQuery.test.ts`** (new) pins both halves of that branch: the mint resolves and increments, and any other raw query still rejects with the original error. The invariant had no test before this, and an exception carved into it is precisely the thing that rots — widening the match to all raw queries now fails a test instead of passing silently.

The three per-suite mock updates from the first commit (`auth-options-adapter`, `auth-options-jwt`, `intake`) are unchanged and still correct.

Full unit tier now green end to end — `284 → 285 suites, 2865 → 2867 tests, 0 failures`, run both in-band and at `--maxWorkers=12`. `tsc --noEmit` and `eslint --max-warnings 0` clean on all touched files. Integration tier unchanged from the PR body.

jee7s:
**Merged `origin/main` — conflicts resolved, keep-both.** Source: #1688 (*require name on every Person creation path*), which landed on the same two blocks the counter conversion rewrote.

Conflicted files (2): `src/app/api/membership-ops/participants/route.ts`, `src/lib/membership/intake.ts`. Everything else auto-merged.

Nothing from either side was dropped — `git diff origin/main` on both files now shows **only** the counter changes, with #1688's guards intact. The two ordering calls:

- **participants route** — #1688's `parentName` guard stays *outside and before* `withTx`. A missing name is a 400 that must fail fast, and running it inside the transaction would take the `IdCounter` row lock only to roll it back. The derived household name (`parentLastName ? \`${parentLastName} Household\` : "Household"`) and `name: trimmedParentName` both moved onto the split-out `household.create` / `person.create` inside the transaction, since an explicit `id` can't ride along with a nested relation create.
- **intake.ts** — #1688 narrowed the secondary-parent branch from `sp.name || sp.email` to `sp.name` and added the `name_required` rejection as the sibling `else if (sp.email)`. Kept as-is: the rejection branch is reached *instead of* the create, so a nameless second parent never opens a transaction and never mints. `name: sp.name` (non-null) kept over my earlier `sp.name ?? null`.

Verification, all green post-merge: `prisma generate` re-run (schema merged cleanly, `IdCounter` + its classifications entry intact), `tsc --noEmit` clean, `eslint --max-warnings 0` clean on the touched files, full unit tier **286 suites / 2896 tests** (up from 285/2867 — #1688's own suite), and the integration set for every converted file — `mintId.integration`, `adminParticipants`, `membershipIntakeAPI`, `intakeConcurrency`, `householdAPI`, `householdMemberAPI`, `adminParticipantHousehold`, the four `adminBulkImport` suites, `contacts`, `profileAPI`, `seed-helpers` — **14 suites / 92 tests**.

Merge commit `8a8b369`, no force-push.

checkin-app/src/__tests__/personIdMintGuard.test.ts
```diff
@@ -0,0 +1,197 @@
...
+    }
+    return null;
+}
+
+const CREATE_RE = /\b(?:prisma|tx|db)\.person\.(create|upsert)\s*\(/g;
+// The one nesting through which a Person could be created without person.create.
+const NESTED_RE = /\bhouseholdMembers\s*:\s*\{\s*create\b/g;
+const MINTED_RE = /mintPersonId/;
+
+function scan(): Set<string> {
```
thpr, line 176: Guard regex misses `createMany` / `createManyAndReturn` — both are Prisma operations that insert Person rows without matching the current pattern. No production call site exists today (`grep` confirms one hit in a test file only), but the guard's whole purpose is catching the day someone adds one.

```typescript
const CREATE_RE = /\b(?:prisma|tx|db)\.person\.(create|createMany|createManyAndReturn|upsert)\s*\(/g;
```

---

jee7s: Done in 350927fb — good catch, and the semantics are worth stating in the guard itself.

Regex now:
```
/\b(?:prisma|tx|db)\.person\.(create|createMany|createManyAndReturn|upsert)\s*\(/g
```

The two bulk shapes are flagged **unconditionally, payload unexamined**. One statement over an array of rows has no per-row hook, so a `createMany` site can't be made to satisfy the guard in place — the call shape *is* the finding, and the fix is converting it to a mint-per-row loop inside one transaction. The assertion's failure message now spells that out rather than leaving "add `mintPersonId`" as advice that wouldn't work here. Flagging the shape rather than the payload also closes the obvious dodge: a per-row `id:` inside the array still isn't a per-row mint, and that case is pinned.

Since every shape the regex recognises has zero production sites, the alternation could rot without anything noticing — so the scan body moved into an exported `scanSource(src)` and there are now nine fixture cases against it: both bulk shapes, a `createMany` that names `mintPersonId` (still a violation), unminted create/upsert, the nested `householdMembers` route, the two clean minted forms, and a `createMany` in a comment (not a site, proving `maskComments` still applies). Also confirmed end to end by dropping a real `createMany` into `api/household/route.ts` — the guard named the file, then restored.

Full unit tier green: 286 suites / 2905 tests. `tsc` and `eslint --max-warnings 0` clean.

PR #1697, opened 2026-08-22
fix(merge): repoint every Person relation the merge previously stranded (#1456 2b-0)

jee7s:
Part of #1456 (slice 2b-0). The minimum first PR of the Phase-2 plan: no schema
change, no migration, no boundary file, no dependency on the rest of Phase 2.

## The six relations

Six `Person` relations were never repointed by the participant merge. The rows
stay pinned to a record the rest of the app treats as nonexistent — and two of
them are why the tombstone can't simply be deleted later.

| Relation | FK | Why the repoint matters |
|---|---|---|
| `RawBadgeLog.personId` | **RESTRICT** | This is the one that gates future deletion. A scan left on the tombstone makes `DELETE FROM "Person"` fail, on essentially every real tombstone. Plain `updateMany` — no unique constraint. |
| `OrgMembershipProcess.noteAckById` | SET NULL | Who read a family's intake note. Left behind, it becomes "nobody read it" the moment the row goes. |
| `Event.attendanceConfirmedById` | SET NULL | Who confirmed attendance. Same erasure. |
| `TrustedAdultReview.decidedById` | SET NULL | Which board member decided a trusted-adult review. Same erasure. |
| `PersonRole.personId` | **CASCADE** | A role left on a tombstone is a security grant that a later delete destroys **silently, with no audit row**. Repointed with dedupe (Decision 4's default). |
| `PersonRole.grantedById` | SET NULL | Who granted the role. |
| `BackgroundCheckAttestation.subjectPersonId` | SET NULL | The adult an attestation names — and the second half of the collision problem below. |

All of it lands inside the existing transaction, with tallies in the audit
`moved` object alongside the existing ones.

**`PersonRole` dedupe.** `PersonRole`'s PK is `@@id([personId, role])` — a
composite PK, no surrogate id and no separate unique. So the holder repoint uses
the same loop-or-skip idiom already established for `bgAttestations`, keyed on
`personId_role`: move each role the survivor doesn't hold, delete the duplicate
when it does (the survivor already carries that grant). The audit records
`roles: { migrated, deduped }`, so nothing goes quietly. The `grantedById`
sweep runs **after** the holder pass, so a granter stamp on a row that was just
deduped isn't counted as moved.

## Collision-key widening (Decision 3, detection half)

`BackgroundCheckAttestation` carries
`@@unique([processId, reviewerId, subjectPersonId])`, but #1686's refusal — in
both the pre-transaction guard and the analyze route — keyed on `processId`
alone. That covers the reviewer side and misses the subject side entirely:
today it cannot even *detect* the case where both identities are the SUBJECT of
the same process.

Both surfaces now refuse two further collision classes, alongside the existing
process-level reviewer check (left as-is: "the same human attested under two
identities" is an integrity signal in its own right, not merely a constraint to
route around).

**`bgAttestationSubject`** — the exact `(processId, reviewerId)` pair, the part
of the triple a `subjectPersonId` repoint does not change:

- **Same reviewer named both identities under one process** → refuse. One
  reviewer reading the same human off a PDF twice under two identities: the
  P2002 the repoint would otherwise hit, and a fact worth a human's attention.
- **Two different reviewers each named one identity** → merges cleanly. That is
  an ordinary 2-of-N review, and after the merge both attestations correctly
  name one person. Refusing it would block the very dedupe this is for.

**`bgAttestationCrossRole`** — one record REVIEWED an attestation naming the
other as its SUBJECT. This one is not a constraint problem at all: only one row
exists, so **no unique fires**, and the `subjectPersonId` repoint silently
collapses it into `reviewerId === subjectPersonId`. `approvalsForSubject`
(`src/lib/membership/review.ts`) has no self-exclusion, so that self-approval
counts toward `REQUIRED_APPROVALS = 2`, and `attest()`'s same-household gate is
create-time only and never sees a merge. Refused in both directions.

The process-level reviewer check does **not** cover this case — it fires only
when both identities hold REVIEWER rows, and here only one does. Found by a
Fable cross-check of this PR and verified live against `deploy-db-1`; see the
comment below.

This slice is detection only — the sysadmin removal endpoint Decision 3 also
calls for is a separate PR, before R1.

## Tests

`merge/__tests__/route.integration.test.ts`, following the suite's existing
fixture idiom and per-test id-tracked cleanup (the new `RawBadgeLog` rows are
RESTRICT children, so they're torn down ahead of their `Person`; ditto
`TrustedAdultReview` ahead of `TrustedAdult`):

- One case per newly-repointed relation, each asserting the row lands on the
  survivor and the audit tally counts it.
- `PersonRole` dedupe: tombstone holds BOARD + KEYHOLDER, survivor holds BOARD →
  exactly two rows survive, both on the survivor, and the dropped duplicate is
  named in the audit.
- Ordering pin: a `grantedById` stamp on the row the dedupe deletes is **not**
  counted as moved, which fails if the granter sweep is moved ahead of the
  holder pass.
- `subjectPersonId` repoint under a second reviewer already naming the survivor —
  proving the widened key doesn't over-refuse.
- Subject-side collision → 409 naming it, nothing moved; and analyze reporting
  the same collision so the picker warns before the operator commits.
- Cross-role collision, both directions → 409 with the attestation untouched
  (`reviewerId !== subjectPersonId` still holds), plus analyze reporting it
  direction-independently.

## Verification

- `tsc --noEmit` clean; `eslint --max-warnings 0` clean on all three touched files.
- Full unit tier (`npx jest --ci --maxWorkers=12`): **283 suites / 2882 tests, all passing**.
- Merge integration suite: **82/82 passing** (69 before, 13 new).
- Revert-in-place proof on a repoint: dropping the `RawBadgeLog` `updateMany`
  fails its new test (`Expected: 0, Received: 1` — the scan stays on the
  tombstone) and nothing else; restored.
- Revert-in-place proof on the new guard: disabling the cross-role clause
  reproduces the reported defect exactly (both cross-role cases return
  `200` instead of `409`, and the row becomes a self-review); restored.

## Note on #1696

**#1696 (open) also touches `merge/__tests__/route.integration.test.ts`.** The
two sets of cases are independent — whichever lands second resolves the conflict
keep-both.

### DIFF
--- a/checkin-app/src/app/api/membership-ops/participants/merge/__tests__/route.integration.test.ts
+++ b/checkin-app/src/app/api/membership-ops/participants/merge/__tests__/route.integration.test.ts
@@ -1067,6 +1067,19 @@ describe("Merge Participants API", () => {
             expect(await movedTallies()).toMatchObject({ roles: { migrated: 1, deduped: 1 } });
         });
 
+        // Ordering pin: the granter sweep runs AFTER the holder pass, so the stamp on
+        // the row the dedupe just deleted is not counted as moved. Sweep-first would
+        // report roleGrants: 1 for a row that no longer exists.
+        it("does not count a granter stamp on a role row the dedupe deleted", async () => {
+            await prisma.personRole.create({ data: { personId: pKeepId, role: "BOARD" } });
+            await prisma.personRole.create({ data: { personId: pMergeId, role: "BOARD", grantedById: pMergeId } });
+
+            const res = await POST(mergeReq(pKeepId, pMergeId));
+            expect(res.status).toBe(200);
+
+            expect(await movedTallies()).toMatchObject({ roles: { migrated: 0, deduped: 1 }, roleGrants: 0 });
+        });
+
         it("moves the PersonRole granter stamp (grantedById)", async () => {
             await prisma.personRole.create({ data: { personId: actorId, role: "KEYHOLDER", grantedById: pMergeId } });
 
@@ -1132,6 +1145,56 @@ describe("Merge Participants API", () => {
             const { collisions } = await res.json();
             expect((collisions as { type: string }[]).map(c => c.type)).toEqual(['bgAttestationSubject']);
             expect(collisions[0].message).toContain("subject of the same background check");
+        });
+    });
+
+    // One record REVIEWED an attestation naming the other as SUBJECT. Only one row
+    // exists, so no unique constraint fires: the subjectPersonId repoint silently
+    // makes reviewerId === subjectPersonId, and approvalsForSubject counts that
+    // self-approval toward REQUIRED_APPROVALS.
+    describe("cross-role attestation collision (reviewer of the other's own check)", () => {
+        const seedCrossRole = async (reviewerId: number, subjectPersonId: number) => {
+            const process = await prisma.orgMembershipProcess.create({ data: { kind: "PERSON_BG", status: "PENDING_BG_REVIEW" } });
+            createdProcessIds.push(process.id);
+            return prisma.backgroundCheckAttestation.create({ data: { processId: process.id, reviewerId, subjectPersonId, result: "APPROVE" } });
+        };
+
+        it("409s when the KEEPER reviewed an attestation naming the merged-away record", async () => {
+            const att = await seedCrossRole(pKeepId, pMergeId);
+
+            const res = await POST(mergeReq(pKeepId, pMergeId));
+            expect(res.status).toBe(409);
+            const data = await res.json();
+            expect((data.collisions as { type: string }[]).map(c => c.type)).toEqual(['bgAttestationCrossRole']);
+            expect(data.error).toContain("reviewer of their own background check");
+
+            // The row is untouched — no self-review was created.
+            const after = await prisma.backgroundCheckAttestation.findUnique({ where: { id: att.id } });
+            expect(after?.subjectPersonId).toBe(pMergeId);
+            expect(after?.reviewerId).not.toBe(after?.subjectPersonId);
+            expect((await prisma.person.findUnique({ where: { id: pMergeId } }))?.mergedIntoId).toBeNull();
+        });
+
+        it("409s in the other direction too — merged-away record as reviewer, keeper as subject", async () => {
+            await seedCrossRole(pMergeId, pKeepId);
+
+            const res = await POST(mergeReq(pKeepId, pMergeId));
+            expect(res.status).toBe(409);
+            expect(((await res.json()).collisions as { type: string }[]).map(c => c.type)).toEqual(['bgAttestationCrossRole']);
+        });
+
+        it("analyze reports it, both directions", async () => {
+            await seedCrossRole(pKeepId, pMergeId);
+
+            const res = await analyzeGET(analyzeReq(pKeepId, pMergeId));
+            expect(res.status).toBe(200);
+            const { collisions } = await res.json();
+            expect((collisions as { type: string }[]).map(c => c.type)).toEqual(['bgAttestationCrossRole']);
+            expect(collisions[0].message).toContain("reviewer of their own background check");
+
+            // Direction-independent: the same pair swapped reports the same collision.
+            const swapped = await analyzeGET(analyzeReq(pMergeId, pKeepId));
+            expect(((await swapped.json()).collisions as { type: string }[]).map(c => c.type)).toEqual(['bgAttestationCrossRole']);
         });
     });
 
--- a/checkin-app/src/app/api/membership-ops/participants/merge/__tests__/route.integration.test.ts
+++ b/checkin-app/src/app/api/membership-ops/participants/merge/__tests__/route.integration.test.ts
@@ -116,9 +116,11 @@ describe("Merge Participants API", () => {
         const personIds = [pKeepId, pMergeId, actorId, ...extraPersonIds];
 
         // RESTRICT-FK children must go before their Person rows.
-        await prisma.backgroundCheckAttestation.deleteMany({ where: { reviewerId: { in: personIds } } });
+        await prisma.backgroundCheckAttestation.deleteMany({ where: { OR: [{ reviewerId: { in: personIds } }, { subjectPersonId: { in: personIds } }] } });
+        await prisma.rawBadgeLog.deleteMany({ where: { personId: { in: personIds } } });
         await prisma.corporationLead.deleteMany({ where: { personId: { in: personIds } } });
         await prisma.corporationMember.deleteMany({ where: { personId: { in: personIds } } });
+        await prisma.trustedAdultReview.deleteMany({ where: { householdId } });
         await prisma.trustedAdult.deleteMany({ where: { householdId } });
         await prisma.toolStatus.deleteMany({ where: { personId: { in: personIds } } });
         await prisma.rSVP.deleteMany({ where: { personId: { in: personIds } } });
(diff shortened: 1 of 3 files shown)


jee7s:
**Fable cross-check: fix-first, one confirmed defect — fixed in `195c24f0`.**

Everything else held (ordering, dedupe, reviewer/subject parity, tallies, the
`LIVE_PERSON` sweep, and the `RawBadgeLog` mutation proof all reproduced). The
one real finding, verified live against `deploy-db-1` with a since-reverted
probe:

**The merge could create a self-review of a background check.** Seeding a single
`(process P, reviewerId = keep, subjectPersonId = merge, APPROVE)` row and
POSTing the merge returned **200**, and the row came out with
`reviewerId === subjectPersonId` — the survivor now APPROVE-reviews their own
background check. `approvalsForSubject` (`src/lib/membership/review.ts`) has no
self-exclusion, so that approval counts toward `REQUIRED_APPROVALS = 2`, and
`attest()`'s same-household gate is create-time only, so nothing downstream
catches it either.

Neither guard saw it. The existing reviewer-side check keys on `processId` but
requires REVIEWER rows on **both** sides — here only one side has one. My new
subject-side check keys on `(processId, reviewerId)` and needs a matching row on
each side — here there is only one row, so no unique constraint fires at all.
**My original PR body claimed the process-level reviewer check covered cross-role
cases. That was wrong**, and the body has been corrected.

**Fix** — a third clause on both surfaces (`bgAttestationCrossRole`), refusing
when an attestation crosses the pair in either direction, plus the ordering pin
the check suggested (a `grantedById` stamp on the row the dedupe deletes, which
fails if the granter sweep is moved ahead of the holder pass). Four new tests:
POST 409 both directions with the probe's exact fixture shape and the row
asserted untouched, analyze reporting it direction-independently, and the
ordering pin. Disabling the new clause reproduces the probe exactly (200, not
409), which is the mutation proof.

Re-verified: `tsc --noEmit` clean, `eslint --max-warnings 0` clean, full unit
tier 283 suites / 2882 tests passing, merge integration 82/82.

**One thing I did not fix, flagged rather than assumed:** this slice newly
repoints `TrustedAdultReview.decidedById`, which has the same cross-role shape —
a merge can make the survivor the decider of a trusted-adult review about
themselves. It is materially weaker than the attestation case (a single recorded
board decision, not a numeric clearance threshold), and the merge already
repoints `TrustedAdult.trustedAdultPersonId` and `disclosedById` on `main`
today. Worth a ruling; not folded in unasked.

jee7s:
All four addressed in `7fba9e4b`.

**1. PersonRole repoint skipped the Person mirror sync — confirmed, fixed.**
Exactly as described: the loop moved rows with `tx.personRole.update`/`delete`,
so `applyRoleFlag`'s dual-write never ran and a migrated BG_REVIEWER kept
`isBackgroundCheckReviewer = false`.

Took the batch shape you offered rather than per-role `applyRoleFlag` calls: one
`person.update` after the loop, writing the survivor's four mirrors from their
**final** role set (`keepRoles ∪ mergeRoles`), derived through `KIND_TO_MIRROR`
so a newly-mirrored role is covered the day it exists. Unconditional, so it also
repairs drift an earlier merge already left behind — there is a test for that.

**On the tombstone's mirrors — you didn't ask, but the answer is that `main`
leaves them stale, and this PR makes that worse.** Before 2b-0 the tombstone kept
both its `PersonRole` rows and its mirror columns, so the two at least agreed.
Nothing anywhere clears them: the step-1 CAS writes `mergedIntoId`, the identity
fields and `isHouseholdLead`, and stops. Now that the role rows move off, a
tombstone would be left reading `isSysadmin = true` with no row behind it —
precisely the stale-security-state class this slice exists to close. So the four
mirrors are zeroed in the CAS, next to `isHouseholdLead`, on the same reasoning:
a merged-away record holds no authority. In the CAS rather than derived after the
loop deliberately — every role row is moved or deleted, so `false` is the correct
end state even if a row raced in; the row would be the bug, not the flag.

**2. `bgSubjectKey` drift — fixed.** Moved to `membershipGuard.ts` (the shared
module analyze already imports) and imported by both routes. Route files can't
carry extra exports, which is why it was inlined; that was the wrong trade, since
a collision key drifting between the POST and its analyze GET is the #1686 defect
this PR set out to close.

**3. PersonRole via `mergeIncludes` — fixed.** `roles: true` added, both in-tx
`findMany` calls gone. The pre-tx snapshot is safe for the same reason every
sibling relation's is: nothing between the read and the transaction mutates
`PersonRole` — `setRoleFlag` is the only writer, it runs in its own request, and
its `FOR UPDATE` lock covers the BOARD row set, not this pair. A grant landing in
that window is the same pre-existing race the join relations already carry, and
it fails safe: the row stays on the tombstone rather than being lost.

**4. Exhaustiveness guard — added, and it's the part of this push I'd keep.**
`merge/__tests__/mergeRepointDriftGuard.test.ts`, modelled on
`livePersonDriftGuard` down to the two-assertion shape (nothing unlisted, nothing
stale). It parses `schema.prisma` for every to-one FK targeting `Person` — 22
today — and fails until each has a written disposition: repointed, deduped,
pre-refused, or deliberately-left, each entry naming the `onDelete` because that
is what decides how bad "do nothing" is. Verified non-vacuous: adding a probe FK
to the schema fails it with the new key named, and the parse itself is asserted.

It is declarative on purpose — it does not grep `route.ts` to prove a repoint
still exists. Deleting one is caught by the per-relation integration cases (each
fails if its `updateMany` goes away); a grep here would just match another line
using the same field name. This closes the other half: a relation nobody
considered at all, which is what all six of this PR's additions were.

**Verification** — `tsc` clean, `eslint --max-warnings 0` clean, full unit tier
**284 suites / 2885 tests**, merge integration **85/85**. Revert-in-place on the
mirror sync: dropping the `person.update` fails the migrated-role and self-heal
cases (`Expected: true, Received: false`) and leaves the deduped case passing,
which is the right blast radius.

jee7s:
**Round 2 (Fable delta pass) — addressed, but this branch is queued, so the
commit is parked rather than pushed here.**

`d551b87a` is on **`fix/merge-mirror-intx-1456`** (branched off this PR's head,
`7fba9e4b`). This branch was added to the merge queue while I was working and
GitHub refuses updates to a queued branch. I did **not** dequeue — `7fba9e4b` is
what was reviewed and shipped-off, and it is strictly better than `main`; the
parked change is a narrower fix that deserves its own review pass anyway.
**Recommendation: let this land, then I open the follow-up.** Say the word if
you would rather dequeue and fold it in here instead.

What is in it:

**1. Keeper-mirror lost-update race — real, fixed.** The mirror write asserts all
four flags at once, so deriving it from the pre-tx `keepRoles ∪ mergeRoles`
reintroduced finding 1's own shape: a role revoked from the keeper between the
read and the transaction came back `true` with no row behind it — retained
privilege — and one granted in the window was zeroed with a row present. Now a
`tx.personRole.findMany` after the dedupe loop. The snapshot still decides what to
MOVE, which is safe in a way the overwrite is not: a row it misses stays put.

**Testable, and tested.** The race does not need a scheduler hook — patching the
pre-tx snapshot to report a role no row backs is exactly what a concurrent revoke
leaves behind. `does not re-assert a role revoked between the pre-tx read and the
transaction` spies on `person.findUnique`, pushes a phantom SYSADMIN into the
keeper's snapshot, and asserts the merge ends with `isSysadmin = false` and no row.
Reverting to the pre-tx union fails exactly that one case with
`Expected: false, Received: true`.

**2. `personPreImage`** gains the four mirrors. The docstring's claim was already
false once this PR started rewriting them on both sides, and since a merge deletes
nothing, that audit row is the only record of the authority the merged-away record
held. The matrix-15 case now asserts the recovery directly: `oldData.isKeyholder`
true while the row itself reads false.

**3. Nits.** Demote-direction heal test added (mirror `true`, no row → `false`) —
the security-relevant direction, and only promote was covered. Drift-guard header
now states its parser ceiling: multi-line `@relation` and implicit m2m are
invisible to a line-oriented regex, neither shape occurs in this schema, and the
DMMF is the upgrade if one ever does. PersonRole's disposition carries the
raced-grant caveat.

**4. Filed for the epic, deliberately not here:** `PATCH /api/roles` does not check
`mergedIntoId` on its target, so a role can be granted to a tombstone. One line, in
a file outside this PR's scope, and it is the other half of the caveat above —
queued for #1456's next slice.

Verification on `d551b87a`: `tsc` clean, `eslint --max-warnings 0` clean, full unit
tier **284 suites / 2885 tests**, merge suites **107/107**.

PR #1699, opened 2026-08-22
fix(kiosk): curfew-gate the attendance poller; exempt the supervision confirm from the debounce (#1347 PR-0)

jee7s:
## Summary

Two independent, one-line-class defects from the kiosk-resilience epic's PR-0 slice (see #1347's plan comment). They share no files; landed as one PR per the plan's note that they're splittable but fine together.

### 1. `client/client.py` — the poller threads defeat the overnight curfew

The outbox drain already sleeps through the 23:00–06:00 closed window (`in_closed_window()` in `outbox.py`), but two other background threads had no such gate:

- `attendance_poller` — an unconditional signed GET every 30s.
- `version_poller` — an unconditional signed GET to `/api/system-status/kiosk-version` every 60s.

A kiosk pointed at prod 24/7 kept the service warm all night through either one — this is the keep-alive cost shape from the 2026-07-17 overnight incident, possibly leaking money today. Fix: both reuse `in_closed_window()` (imported, not reimplemented) and skip their fetch while closed. `version_poller`'s self-update half (git fetch/rev-parse against `origin/main`, no request to the server) is deliberately left ungated — overnight is the safe restart window, nobody's badging.

### 2. `checkin-app/src/app/api/scan/route.ts` — the supervision confirm inherits the force-close debounce bug

`route.ts`'s debounce exemption (`isConfirm`) keyed only on `forceCloseToken`. The supervision-interrupt confirm (#1436/#1670) mints no token — it accepts a re-scan within `SUPERVISION_CONFIRM_MS` (15s) of `Visit.supervisionWarnedAt` — so a second badge inside the 3s debounce window was silently swallowed before `processCheckout` ever saw it. The working confirm window was really `[3s, 15s]`, and the code said so itself:

```ts
/** Countdown on the supervision confirm (#1436). Same 3s debounce eats the front,
 *  so the kiosk copy says "3 to 15 seconds". */
```

A youth-present room's last-but-one supervising adult double-badging to leave could not check out and got no feedback.

**Decision 7** (ratified in the issue #1347 review): extend the existing debounce exemption to a scan against the person's open Visit whose `supervisionWarnedAt` is fresh, rather than minting a second token. The stamp is only ever written when a warning was actually shown, tying the confirm to that fact the same way the token ties it to force-close. The token clause is unchanged and stays primary; the token machinery is still there to drop in if the stamp-based exemption ever proves too loose.

**Fable cross-check finding (fixed in this PR):** the naive stamp-fresh exemption is not single-use. A USB hardware double-read of the WARNING scan itself (duplicate evdev keystrokes ~300–800ms apart, no client dedup) lands with the stamp already fresh, so the exemption skipped the debounce and `processCheckout` auto-departed the last-but-one adult with zero human acknowledgment — a regression in the very interlock this PR set out to strengthen (pre-PR, that second read debounced and the warning stood). Fixed by flooring the exemption at a new `SUPERVISION_CONFIRM_DEADFRONT_MS` (1s, exported from `scan-service.ts` beside `SUPERVISION_CONFIRM_MS`): the stamp must be `>= 1s` old **and** `<= SUPERVISION_CONFIRM_MS` old. A dead front invisible to a human keyholder; decision 7's semantics are otherwise unchanged, and the "within 15 seconds" copy stays honest since only the window's first second is unusable.

Also: `SUPERVISION_CONFIRM_MS` is now exported from `scan-service.ts` as the single source of truth (route.ts imports it rather than re-deriving the window), and the kiosk copy drops the dead front — "3 to 15 seconds" → "within 15 seconds".

## Tests

- `client/test_client.py`: `attendance_poller` swallows the fetch at 23:30, polls normally at 12:00 (fake-clock idiom mirroring `replay_drain`'s `sleep_fn`/`in_closed_window_fn` injection).
- `client/test_self_update_guard.py`: `version_poller`'s server-version GET is swallowed at 23:30 and runs at 12:00; the self-update `subprocess.run` call still fires during the closed window either way.
- `scanRoute.test.ts`: a stamp 2s old is exempt from the debounce (with the `visit.count` where-clause pinned — `personId`/`departedAt`/`deletedAt`/`gte`/`lte` — so a dropped or loosened filter fails the test instead of passing silently); a stamp 500ms old (the hardware dead-front case) still debounces; still debounces with neither a token nor a fresh stamp. Also fixed a pre-existing assertion (`visit.count` no longer stays uncalled for a tokenless scan, since the supervision check always runs).
- `scanServiceSupervisionInterrupt.test.ts`: a second badge just 1s after the warning confirms and departs (scan-service's own ladder has no floor — only the route's debounce exemption does); updated the "3 to 15 seconds" copy assertion.

**Revert-in-place proofs:**
- Restored the token-only `isConfirm` condition → the new supervision-exemption test failed (`ignored_debounce` instead of `checkout`), all other tests unaffected. Restored.
- Dropped the `lte` dead-front floor → the 500ms test failed (`checkout` instead of `ignored_debounce`, reproducing the auto-departure) and the pinned where-clause assertion failed on the missing `lte`. Restored; all green again.

## Verification

- Python: `python3 -m unittest discover` from `client/` — 55/55 passed.
- `npx tsc --noEmit` — clean.
- `npx eslint` on touched files — clean.
- Full unit tier: `npx jest --ci --maxWorkers=12` — 283 suites / 2886 tests passed.
- Scan integration tier: `INTEGRATION_DB=1 npx jest --ci --maxWorkers=12 --testPathIgnorePatterns=/node_modules/ --testPathPatterns='scan.*\.integration\.test\.ts'` — 5 suites / 28 tests passed.

## Follow-ups (not this PR)

- `Visit.supervisionWarnedAt` is never cleared on departure. Inert today — every lookup filters `departedAt: null` — but it contradicts `processCheckout`'s no-live-state-on-closed-rows principle (the same reasoning that already clears `forceCloseToken`/`forceCloseWarnedAt` on close). Worth folding into the substrate slice.
- A non-departing confirm outcome (e.g. a keyholder warning fires again inside the window) keeps the stamp-based exemption alive for the rest of the window rather than resetting it. Bounded by `SUPERVISION_CONFIRM_MS` either way; noted, not fixed here.

Part of #1347

### DIFF
--- a/checkin-app/src/app/api/scan/__tests__/scanRoute.test.ts
+++ b/checkin-app/src/app/api/scan/__tests__/scanRoute.test.ts
@@ -49,6 +49,7 @@ jest.mock('@/lib/scan-service', () => ({
     processCheckout: jest.fn(),
     finalizeFacilityClose: jest.fn().mockResolvedValue(undefined),
     SUPERVISION_CONFIRM_MS: 15_000,
+    SUPERVISION_CONFIRM_DEADFRONT_MS: 1_000,
 }));
 
 describe('POST /api/scan', () => {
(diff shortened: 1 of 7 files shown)


jee7s:
**Fable cross-check (fix-first) — 2 CONFIRMED findings, both fixed.** Everything else held, including the PR's own revert-in-place proof, which Fable independently reproduced.

**C1 — the supervision exemption wasn't single-use.** A USB hardware double-read of the WARNING scan (duplicate evdev keystrokes ~300–800ms apart, no client dedup) landed with `supervisionWarnedAt` already fresh, so the new debounce exemption skipped straight through and `processCheckout` auto-departed the last-but-one adult with zero human acknowledgment — a regression in the very interlock this PR set out to strengthen (pre-PR, that second read debounced and the warning stood). Fixed by flooring the exemption at a new `SUPERVISION_CONFIRM_DEADFRONT_MS` (1s, exported beside `SUPERVISION_CONFIRM_MS`): the stamp must be `>= 1s` old **and** `<= SUPERVISION_CONFIRM_MS` old. Decision 7's semantics are unchanged; the "within 15 seconds" copy stays honest since only the window's first second is unusable. New tests pin the exact `visit.count` where-clause and cover both sides of the floor (500ms still debounces, 2s confirms).

**C2 — the poller-curfew fix wasn't complete.** `version_poller`'s server-version GET (`/api/system-status/kiosk-version`, every 60s) was still unconditional, so the PR's money-leak claim didn't fully hold — the warm-service effect was binary regardless of the `attendance_poller` fix. Fixed by gating only that GET on `in_closed_window_fn` (same seam as `attendance_poller`); the self-update half (git fetch/rev-parse, no request to the server) stays ungated since overnight is its safe restart window.

Re-verified after both fixes: Python 55/55, full unit tier 283 suites / 2886 tests, scan integration tier 5 suites / 28 tests, all green. Revert-in-place on the dead-front floor reproduced the auto-departure regression (dropping `lte` failed the 500ms test); restored.

PR body and follow-ups updated accordingly.

PR #1698, opened 2026-08-22
sec(registry): entry for DELETE bg-attestation (#1456 prerequisite)

jee7s:
## Summary
- Registers `DELETE /api/membership-ops/bg-attestations/[id]` in `src/security/registry.ts`, sysadmin-only (`{ anyRole: ['isSysadmin'] }`), no bag (`envelope: null`).
- Registry-first per the `AGENTS.md` boundary-isolation rule: an unused `defineRoute` is inert, so this widening is reviewable in isolation before anything serves it. Precedent: #1459 (registry-only) unblocking #1387.
- Boundary-only diff — nothing else in this PR. `check-route-coverage --strict` reports the expected `[orphan-registry]` **warning** (not an error — "the register-first state while its route PR is pending") and 0 errors.

## Why
Part of #1456. Decision 3 (owner-accepted default, see the issue's ruling comment): a sysadmin-only endpoint to remove a `BackgroundCheckAttestation` row with a mandatory reason and an audit row, giving the participant-merge route's collision refusal ("Both attested the same background check") an actionable cleanup path. This PR is only the registry entry; the route handler follows in a stacked PR based on this branch.

Part of #1456

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01URSrfChi89qKn1hLbjaX3K

PR #1700, opened 2026-08-22
feat(membership-ops): remove a background-check attestation with reason + audit (#1456 prerequisite)

jee7s:
## Base branch note
This PR is stacked on `sec/bg-attestation-remove-registry` (#1698, the registry-only PR) so its diff is route-only — the registered `defineRoute()` entry it relies on is inert without this handler. **Retarget this PR to `main` once #1698 merges.**

## Purpose
Part of #1456, Decision 3 (owner accepted the default — see the issue's ruling comment). The participant-merge route refuses to carry forward a `BackgroundCheckAttestation` collision with the message "Both attested the same background check" (`merge/route.ts`, `bgAttestation` collision type) — same human reviewed under two `Person` identities. Until now that refusal was a dead end: attestations are deleted only by `overrideBlocked` in `reset`, and only for a `BLOCKED` process. This PR gives it a cleanup path: `DELETE /api/membership-ops/bg-attestations/[id]`, sysadmin-only, requiring a non-empty `reason`, deleting the row and writing an `AuditLog` entry (`oldData` = the deleted row's fields, `newData` = `{ reason }`).

Sysadmin-only rather than board — attestations are review artifacts, not a membership decision, per the plan's Decision 3.

## Process-status coupling — investigated, deliberately left alone
Checked whether removing an attestation needs to touch `OrgMembershipProcess.status` (`src/lib/membership/review.ts`):
- A `REJECT` attestation moves the process to `BLOCKED` (`attest()`). Deleting that attestation via this endpoint does **not** revert the status — the process stays `BLOCKED` with one fewer attestation behind it.
- The 2nd `APPROVE` triggers `clearBackgroundCheck` (stamps `lastBackgroundCheck`, sets `bgClearedAt`, may activate a membership). Deleting an attestation that contributed to an already-cleared process does **not** unwind any of that — the membership stays activated/cleared.

Per the plan: **do not invent a transition here.** Removing the row and leaving status alone, with the audit trail as the record, is the conservative default — a status reversal has its own side effects (re-notifying reviewers, reopening payment, etc.) that a bare row-delete should not trigger silently. If a removed attestation should ever force a status re-evaluation, that's a separate, deliberate design question — flagging it here rather than answering it in this PR.

## Tests
`src/app/api/membership-ops/bg-attestations/[id]/__tests__/route.integration.test.ts` (real Postgres):
- 400 without a `reason` (row untouched)
- 400 on a non-numeric id
- 404 for an id with no matching row
- success: row gone + `AuditLog` row with `actorId`, `oldData` (the deleted row's fields), `newData.reason`
- end-to-end: reproduces the merge's `bgAttestation` collision (409), deletes the duplicate via this endpoint, confirms the merge then succeeds (200) — the actual scenario this endpoint exists for

401/403 (including a board-member-alone case, since this is sysadmin-only NOT board-or-sysadmin) added to the shared `authzRoleRejection.integration.test.ts` table, plus the route added to `authzRegistry.test.ts`'s `AUTHZ_TESTED` drift guard.

Full unit tier (`npx jest --ci --maxWorkers=12`) and the security/authz integration suites pass; `check-route-coverage --strict` is clean (only the expected pre-existing "unmigrated" warning class, matching the sibling `DELETE /api/attendance/manual/[id]` which also uses `withAuth` + a registry entry rather than `handler()`).

Part of #1456

### DIFF
--- a/checkin-app/src/__tests__/authzRegistry.test.ts
+++ b/checkin-app/src/__tests__/authzRegistry.test.ts
@@ -97,6 +97,9 @@ const AUTHZ_TESTED = new Set<string>([
     'membership-ops/applications/certify-payment',
     'membership-ops/applications/external',
     'membership-ops/applications/review-override',
+    // 401/403 deny-path in authzRoleRejection.integration.test.ts, incl. a
+    // dedicated board-alone-is-denied case (sysadmin-only, not board).
+    'membership-ops/bg-attestations/[id]',
     // POST deny-path (401 anon / 403 plain / 403 sysadmin-excluded) in
     // membership-ops/contacts/__tests__/route.integration.test.ts.
     'membership-ops/contacts',
--- a/checkin-app/src/app/__tests__/authzRoleRejection.integration.test.ts
+++ b/checkin-app/src/app/__tests__/authzRoleRejection.integration.test.ts
@@ -54,6 +54,7 @@ import { POST as APP_EXTERNAL_POST } from '@/app/api/membership-ops/applications
 import { POST as REVIEW_OVERRIDE_POST } from '@/app/api/membership-ops/applications/review-override/route';
 import { POST as APP_ARCHIVE_POST } from '@/app/api/membership-ops/applications/archive/route';
 import { POST as APP_UNARCHIVE_POST } from '@/app/api/membership-ops/applications/unarchive/route';
+import { DELETE as BG_ATTEST_DELETE } from '@/app/api/membership-ops/bg-attestations/[id]/route';
 import { GET as MOPS_HH_GET, POST as MOPS_HH_POST } from '@/app/api/membership-ops/households/route';
 import { GET as VOL_ROSTER_GET } from '@/app/api/membership-ops/volunteer-memberships/route';
 import { POST as REVIEWS_POST } from '@/app/api/membership/reviews/route';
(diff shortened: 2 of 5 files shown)


jee7s:
Adversarial cross-check complete — **ship, no confirmed defects**. Highlights: the delete/audit pair is genuinely transactional (a crash can't lose the record); the deliberate non-decision on process status creates odd-but-never-incoherent states (probed: re-attest after mid-review removal works, a REJECT-less BLOCKED process refuses cleanly and recovers via reset, no double-clear); and the pair is compatible with #1697's cross-role refusal in both merge orders — refuse → delete the duplicate here → retried merge succeeds.

Two non-blocking minors, queued for the post-#1698 cleanup push rather than a separate cycle:
1. The DELETE transaction doesn't take the process-row `FOR UPDATE` that `review.ts:513`'s contract documents for attestation-set mutations — a concurrent second-APPROVE can clear a check off the row mid-delete (end state is one Decision 3 accepts, but one line honors the stated contract).
2. Concurrent double-delete of the same id: the loser's P2025 falls into the generic catch as a 500 instead of a 404.

Reviewer note: the Files tab currently renders the registry hunk inside this PR — that's GitHub's stale recorded merge-base after the retarget, not real duplication; it recomputes once #1698 merges and the branch cleanup lands.

jee7s:
Post-#1698 cleanup pushed (`87ce85e7`).

**Merge**: `git merge origin/main` after #1698's squash-merge — `registry.ts` came through byte-identical to main's, no conflict. `git diff origin/main...HEAD --name-only` now shows exactly the route PR's files (registry.ts is gone from the diff): `route.ts`, `route.integration.test.ts`, the new `route.test.ts`, plus the two shared authz test-registry files. Base already retargeted to `main`.

**Both Fable minors applied** (`route.ts`):
1. The delete transaction now takes `review.ts:513`'s `FOR UPDATE` lock on the attestation's `OrgMembershipProcess` row before deleting it — same `SELECT id FROM "OrgMembershipProcess" WHERE id = ${processId} FOR UPDATE` idiom as `attest()`/`overrideBlocked()`.
2. A concurrent double-delete's Prisma `P2025` now maps to 404 instead of the generic 500 (duck-typed `isPrismaError`, mirroring `programs/[id]/participants/route.ts`).

New unit suite `route.test.ts` (mocked prisma) pins both: asserts the `FOR UPDATE` query runs with the right `processId` before the delete call, and that a mocked P2025 on delete yields 404 with no audit row written. Real-DB behavior is unchanged and re-verified: `route.integration.test.ts` (400/404/success/audit + the merge-unblock scenario) and `authzRoleRejection.integration.test.ts` (incl. board-alone-denied) still pass.

**Verification**: `tsc --noEmit` clean; `eslint --max-warnings 0` clean on touched files; full unit tier `npx jest --ci --maxWorkers=12` — 284 suites / 2896 tests pass; `INTEGRATION_DB=1` run of this route's suites + `authzRoleRejection`/`registryAuthz`/`policy.contract`/`routeAuthDrift` + the merge route's own tests — 867 tests pass.

thpr's two notes (dead `auth.type !== "session"` guard, no attestation-id surfaced in the merge 409) are left as-is — out of scope for this cleanup push, not part of the queued Fable minors.

jee7s:
@thpr — both notes addressed on the thread; the cleanup push (`87ce85e7`) also landed since your review (diff is route-only now, plus the two cross-check minors: the process-row `FOR UPDATE` per `review.ts`'s documented contract, and P2025 → 404 on a concurrent double-delete).

**Dead guard**: unreachable at runtime, agreed — but it's the type narrow, not a belt-and-brace. `withAuth`'s handler receives the auth union, and `auth.type !== "session"` is what lets `auth.user` typecheck below; the registered DELETE siblings carry the same line for the same reason. Removing it trades one dead runtime branch for a cast, so it stays.

**Merge 409 without attestation IDs**: real discovery gap, and the right home is the merge route's structured `collisions` array (it already carries `type` + `message`; adding the colliding ids is additive). Queued for the #1456 epic's next slice rather than churning the in-review #1697.

